PR #15, opened 2016-08-05
Fix Grappling Hooks duplicating

BlackBeltPanda:
Since ProjectileHitEvent is delayed, we check to make sure the arrow is still a valid entity before calling handleGrapplingHook. Otherwise, it gets called twice and the Grappling Hook gets duplicated.
Also fix some instances of the Grappling Hook's arrow and leash not being removed.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/BowListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/BowListener.java
@@ -3,15 +3,12 @@ package me.mrCookieSlime.Slimefun.listeners;
 import me.mrCookieSlime.Slimefun.SlimefunStartup;
 import me.mrCookieSlime.Slimefun.Variables;
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
-import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.handlers.BowShootHandler;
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
You basically removed Slimefun's entire Custom Bow API...

src/me/mrCookieSlime/Slimefun/listeners/BowListener.java
```diff
@@ -95,23 +107,8 @@ public void run() {
...
-			}
-			Variables.arrows.remove(e.getDamager().getUniqueId());
-		}
-		
-		if (Variables.arrows.containsKey(e.getDamager().getUniqueId()) && e.getEntity() instanceof LivingEntity) {
-			for (ItemHandler handler: SlimefunItem.getHandlers("BowShootHandler")) {
-				if (((BowShootHandler) handler).onHit(e, (LivingEntity) e.getEntity())) break;
-			}
-			Variables.arrows.remove(e.getDamager().getUniqueId());
-		}
```
TheBusyBiscuit: The above part with the BowShootHandler has to stay there or you are basically removing all Bows from Slimefun and its Addons.

Why its actually twice there is unclear to me though :?

BlackBeltPanda:
Ah, my bad, forgot I removed that on my end.

BlackBeltPanda:
Ah, now I know why I removed it; it still duplicates the grappling hook with the BowShootHandler code. Very strange.

PR #16, opened 2016-08-19
Fixed full hazmat suit necessity

Poslovitch:
I fixed a bug that make players don't need to wear full hazmat suit when having a radioactive item in their inventory.


TheBusyBiscuit:
You didnt fix anything...
it worked before, you literally changed nothing...

Poslovitch:
No: players can just wear the Hazmat Chestplate or anything other (leggings, boots) and they don't get the radioactive effects. Same thing on my private server.

PR #17, opened 2016-08-19
Operational safety.

alekso56:
Fix errors :v:

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/holograms/ArmorStandFactory.java
+++ b/src/me/mrCookieSlime/Slimefun/holograms/ArmorStandFactory.java
@@ -25,14 +25,26 @@ public class ArmorStandFactory {
             	try {
             		ReflectionUtils.setFieldValue(nmsEntity, "bz", 2039583);
             	} catch(IllegalArgumentException x) {
-            		ReflectionUtils.setFieldValue(nmsEntity, "bA", 2039583);
+            		try {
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
You dont fix an Error by simply surrounding it with an empty try/catch ...

alekso56:
You could put an System.out.println there if you so wished, the field for itemslots still changed in 1.10.2
it is now  "bB"

TheBusyBiscuit:
Then change ONLY that, but adding empty tra/catch is the EXACT OPPOSITE of bug fixing...

alekso56:
So eh, this?
https://github.com/alekso56/Slimefun4/commit/97b58842a514e6979ef9ce46401770ee6f8334f2

TheBusyBiscuit:
(facepalm) No...

alekso56:
Afaik my try/catch blocks are not empty tho? .-.

TheBusyBiscuit:
-_- Still unncessary as they surpress the actual stacktrace...

alekso56:
I don't think the user should be exposed to a stacktrace here tho.

TheBusyBiscuit:
You are really making me worry about the future of programming...
"If we pretend there is no bug, there is no bug" or what?
Seriously? If that comment you just posted was honest, then you may want to rethink whether coding is something for you...

PR #18, opened 2016-08-19
Change reflection field to work on 1.10 - 1.10.2

alekso56:
There, only changed one thing.

If you misunderstood me, 
i meant that the user does not have to be exposed to a stacktrace, 
when you only call this code from ONE place. .-.


TheBusyBiscuit:
That has nothing to do with that, a stacktrace remains a stacktrace...
Literally EVERY developer in the world will tell you that errors can have more than one source... -_-

alekso56:
Yes, if you actually print it.  here you are just calling 
ReflectionUtils.setFieldValue(nmsEntity, "bA", 2039583);
on all the potential errors. .-.

TheBusyBiscuit:
exactly. You did the opposite before -_-
Seriously, stop with this, I wont answer you ever again.
Learn Java and leave me alone.

PR #26, opened 2016-11-20
Mavenize the project

MrKreeg:
This commit adds Maven to the plugin, because not every developer who wants to contribute feels like installing local dependencies.


ntfsboo:
Closed without reasoning?

TheBusyBiscuit:
Well, I think it is pretty self-explaining, your PR was rejected since I would prefer it to remain a non-Maven project.
That doesn't prevent you from continuing to mavenize your fork of course.

PR #57, opened 2016-12-19
Carbon Press

MippyTheDev:
I kinda copy and pasted the 2 lines, basically this (should) allow you to use charcoal as a material to make carbon as well.


TheBusyBiscuit:
Rejected.

PR #59, opened 2016-12-20
Coal Ore is more reasonable to give 4 coal, rather than two in Advanced Digi-Miner

NathanAdhitya:
Mostly, For some reason I think it's better for it to give out 4 Coal, rather than just two from the else statement, because other ores, like Coal is really needed, and if so, People would never opt-in for an advanced one, if instead you could use the normal one, and a fortune pick.


TEMDATEMMIE:
instead of adding the number of coal when using advanced digi-miner,why not add something like project E?

NathanAdhitya:
Uhhh.......... wot.?

PR #67, opened 2016-12-25
Bug Fixes

John000708:


NathanAdhitya:
For the Offhand Juice, Looks like you forgot an else, because if not, If you had glassbottles in your inventory, and drank a juice in offhand, you would lose one glassbottle in your inventory.

TheBusyBiscuit:
Only fixes the issues for certain cases, these bugs are anchored more deeply than this

Issue #74, opened 2016-12-28
Soulbound Hoe & Shovel Glitch!


ALYHFYAN:
## Description
So one of my players found this glitch while playing on the server. Not sure if it has been known but just to put it out there anyways. So if you use a soulbound hoe or shovel and right click a dirt, it will change to either farmland or a grass path in 1.10.2. But when you break this blocks, it will give you a soulbound hoe or shovel (depending on what you use to right click the dirt). So players can do this and get infinite hoes or shovel. 

We already tested this on our public server and our test server. We tested it on our test server because we thought that maybe another plugin on our public server might be causing it but apparently not as we tested it on our test server with only the required plugins to do this glitch.

## Server Log
There isn't any errors on console at all. So nothing to show here.

## Evidence / Proof
![slimefun glitch](https://cloud.githubusercontent.com/assets/23190446/21533008/4d205352-cd91-11e6-917f-0f306f982a44.gif)

## Environment
Spigot 1.10.2
CS-CoreLib v1.5.9
Slimefun 4.0.15

NathanAdhitya:
May happen with any SF Tools that somehow interact with the world without canceling the event, ex: Crooks too. So, one thing you can do are now adding ExcludedTools....... this "some sort"(Doesn't dupe) of bugs happens with androids too, ill post one later on.

PR #76, opened 2016-12-29
Fix Soulbound Shovel and Soulbound Hoe Duplication Bug.

AlexLander123:
This is a fix for #74 
There is probably a better way to implement this such as creating a ExcludedTools class.
So if you think this is not a good way to implement feel free to close this Pull Request. xD


Slothiful:
Unsubscribed but still receiving emails. Please unsubscribe/mute the whole
entire Slimefun page for me!

TheBusyBiscuit:
Not only Diamond Tools.

PR #81, opened 2017-01-01
Tier 2 Electric Grinder should give a 4x speed instead of 1x

NathanAdhitya:


NathanAdhitya:
Woops, my bad i derped the commit title.

NathanAdhitya:
Looks lke i forgot something once again.. ;-;

PR #92, opened 2017-01-08
Off-Hand Monster Jerky Fix

NathanAdhitya:
There may be a better way of implementing this, but.... ehh.. this is my
concept :P.

The other lines, are just naughty github not accepting the Color code
symbols.

Fix Attempt for: https://github.com/TheBusyBiscuit/Slimefun4/issues/89


TheBusyBiscuit:
Please fix the color symbols by using ChatColor.translateAlternateColorCodes('&', etc...

and youre missing
PlayInventory.update(p) there...

NathanAdhitya:
Ah, Right, my bad, and I'll fix that soon

NathanAdhitya:
@TheBusyBiscuit Is there a way that you would implement to CS-CoreLib an easier way of Filtering Chat Codes? rather than having to have ChatColor.translateAlternateColorCodes(...) for every sendMessage? (Save some file space, tbh.)

PR #108, opened 2017-01-20
Fix XP Dupe Bug

### LINKED ISSUE #106: XP Collector dupe bug
## Description
If you place Multiple XP Collectors around the same spot, they will collect the same amount of XP. So you can dupe XP.

## Evidence / Proof
https://youtu.be/M8a_rnIBk7k

## Environment
minecraft 1.11.2
Slimefun v4.1.1
Spigot

AlexLander123:
Fix the XP duplication bug with xp collectors. Fix #106


TheBusyBiscuit:
Tested?

AlexLander123:
Yeah tested. It is caused by n.remove() marking the entity for removal instead of immediately removing it.

AlexLander123:
n.isValid() checks if entity is marked for removal.

TheBusyBiscuit:
return; is the wrong statement for this scenario though.

Do

if (n.isValid()) {

}

for the following code instead of doing return;

AlexLander123:
Thanks for the advice!

PR #133, opened 2017-02-06
Altar Item fixes

XjCyan1de:
Fixed: Item puckup after server restart
Fixed: Item despawn
Fixed: Item texture when using Optifine CIT texture in Resource Pack

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
@@ -146,9 +146,9 @@ public class AncientAltarListener implements Listener {
 		if (stack != null && !stack.getType().equals(Material.AIR)) {
 			PlayerInventory.consumeItemInHand(p);
 			String nametag = StringUtils.formatItemName(stack, false);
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
Do not remove the no_pickup metadata part. Your changes are fine but this needs to stay in...

XjCyan1de:
Metadata only saves if server on, after restart its clear. If use vanilla ScoreBoard tag its works good. I just tested it

TheBusyBiscuit:
... I never said metadata is better -_-

As I said, you can add your code, but this metadata tag is widely used by multiple Addons and needs to stay...

XjCyan1de:
So its also can replace `(n.hasMetadata("no_pickup"))` (24 line) in ClearLaggIntegration.oNEntityRemove to `n.getScoreboardTags().contains("AltarItem")`

XjCyan1de:
Okay i will change it without remove

TheBusyBiscuit:
Its also 1.11 only. And therefore removes Slimefuns cross compatibility. Not gonna accept this now.

XjCyan1de:
Scoreboard tag added in 1.9 and it worked on it too

I just suggested how you can do better. But if you think that it is better to leave then I will not insist. Have a good day!

TheBusyBiscuit:
I had to leave for work, so I only took a glimpse at CraftBukkit 1.10 and it didnt seem to contain that method. If you can show me that the method exists in 1.9 I will not stand in your way.

PR #249, opened 2017-04-14
Fix Pickaxe of Vein Mining Dupe

DeathlyPanda:
Fixed pickaxe to prevent duplication of ores.


TheBusyBiscuit:
It would be wiser to use
´´drop.getType().isBlock()´´ instead.

DeathlyPanda:
That change i did to vein pick is your own code from explosive pick and it works. Tested

TheBusyBiscuit:
My point stands.

DeathlyPanda:
Changed if you want to check it out

TheBusyBiscuit:
Instead of the IRON/GOLD check not the ore check...
The point is to check if the block is an ore and the drop is a block...

DeathlyPanda:
if (SlimefunManager.isItemSimiliar(e.getPlayer().getItemInHand(), SlimefunItems.PICKAXE_OF_VEIN_MINING, true) && e.getBlock().getType().toString().endsWith("_ORE")) {

ooh look it already checks if it is an Ore.

PR #267, opened 2017-04-24
Backpack & crucible fixes

VoidAngel:
Fixes:
- Backpacks getting stuck in backpacks with number keys.
- Putting shulker boxes inside backpacks for infinite storage.
- A dupe where a player would drop the backpack at the same time as opening it.
- Crucible not making lava in return for hardened_clay

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -2297,6 +2297,7 @@ public class SlimefunSetup {
 						for (ItemStack convert: RecipeType.getRecipeInputs(machine)) {
 							if (input != null) {
 								if (SlimefunManager.isItemSimiliar(input, convert, true)) {
+									e.setCancelled(true);
 									ItemStack removing = input.clone();
 									removing.setAmount(convert.getAmount());
 									p.getInventory().removeItem(removing);
@@ -2310,7 +2311,7 @@ public class SlimefunSetup {
 												block.setData((byte) 7);
 												block.getWorld().playSound(block.getLocation(), Sound.BLOCK_LAVA_POP, 1F, 1F);
 											}
-											else if (input.getType() == Material.LEAVES) {
+											else if (input.getType() == Material.LEAVES || input.getType() == Material.HARD_CLAY) {
 												block.setType(Material.WATER);
 												block.setData((byte) 7);
 												block.getWorld().playSound(block.getLocation(), Sound.ENTITY_PLAYER_SPLASH, 1F, 1F);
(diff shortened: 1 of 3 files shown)


src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
```diff
@@ -41,27 +45,41 @@ public void onClose(InventoryCloseEvent e) {
 	}
 	
 	@EventHandler
+	public void onItemDrop(PlayerDropItemEvent e) {
+		ItemStack item = e.getItemDrop().getItemStack();
+		SlimefunItem sfItem = SlimefunItem.getByItem(item);
```
TheBusyBiscuit: Heavy method right there, definitely advised to be moved inside the if-statement

src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -2375,21 +2376,21 @@ else if (input.getType() == Material.LEAVES) {
...
 																				public void run() {
-																					if (input.getType() == Material.COBBLESTONE) {
+																					if (input.getType() == Material.COBBLESTONE || input.getType() == Material.HARD_CLAY) {
 																						block.setType(Material.LAVA);
 																						block.setData((byte) 2);
 																						block.getWorld().playSound(block.getLocation(), Sound.BLOCK_LAVA_POP, 1F, 1F);
 																					}
 																					else if (input.getType() == Material.LEAVES) {
-																						block.setType(Material.WATER);
+																						block.setType(Material.WATER || input.getType() == Material.HARD_CLAY);
```
TheBusyBiscuit: Little typo right there.

src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
```diff
@@ -41,27 +45,41 @@ public void onClose(InventoryCloseEvent e) {
 	}
 	
 	@EventHandler
+	public void onItemDrop(PlayerDropItemEvent e) {
+		if (Variables.backpack.containsKey(e.getPlayer().getUniqueId())){
+			ItemStack item = e.getItemDrop().getItemStack();
+			SlimefunItem sfItem = SlimefunItem.getByItem(item);
+			if (sfItem instanceof SlimefunBackpack || sfItem instanceof SoulboundBackpack) e.setCancelled(true);
```
TheBusyBiscuit: SoulboundBackpack extends SlimefunBackpack, so it is not needed to check for this.

TheBusyBiscuit: That actually makes me question: Does the Bound Backpack actually work like a Soulbound Item !? It would be nice if someone could investigate this.

TheBusyBiscuit: Okay yes, it does.
https://github.com/TheBusyBiscuit/Slimefun4/blob/6f61176d0c9cb944976e95813160a8df0759effc/src/me/mrCookieSlime/Slimefun/listeners/DamageListener.java#L55

VoidAngel:
:')

PR #268, opened 2017-04-25
Update TalismanListener.java

DeathlyPanda:
Fix:
*Prevents players saved item from being lost if inventory is full by sending it to the players enderchest.
Note: Need to make a better fix which allows player to acquire on the spot if possible.


TheBusyBiscuit:
Should not be send to a Player's Enderchest...

PR #272, opened 2017-04-29
Cooler fixes

VoidAngel:
Fixes:
- Prevents using number keys to insert any item into a cooler
- Prevents using the open-drop dupe with coolers


TheBusyBiscuit:
It would be a better and easier approach to just make the Cooler a SlimefunBackpack as well.

PR #331, opened 2017-06-19
Added option to toggle "enchantability" of an item

Poslovitch:
Related to https://github.com/TheBusyBiscuit/Slimefun4/issues/280

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/AutoEnchanter.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/AutoEnchanter.java
@@ -93,8 +93,13 @@ public class AutoEnchanter extends AContainer {
 			slots:
 			for (int slot: getInputSlots()) {
 				ItemStack target = BlockStorage.getInventory(b).getItemInSlot(slot == getInputSlots()[0] ? getInputSlots()[1]: getInputSlots()[0]);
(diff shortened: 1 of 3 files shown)


TheBusyBiscuit:
Doesn't this still need to handle this boolean in the machine?

Poslovitch:
I don't know. What is the problem exactly?

TheBusyBiscuit:
Well, you added a boolean.

But that boolean is not used anywhere...

Poslovitch:
Well, it seems I forgot to push the Enchanter/disenchanter files with the SlimefunItem x)
I push them in about 10 minutes

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/AutoEnchanter.java
```diff
@@ -92,6 +93,7 @@ protected void tick(Block b) {
 			slots:
 			for (int slot: getInputSlots()) {
 				ItemStack target = BlockStorage.getInventory(b).getItemInSlot(slot == getInputSlots()[0] ? getInputSlots()[1]: getInputSlots()[0]);
+				if(SlimefunItem.getByItem(target) != null && !SlimefunItem.getByItem(target).isEnchantable()) return;
```
TheBusyBiscuit: Heavy call right there.

Should be split into a variable for performance reasons.

PR #337, opened 2017-06-22
Added option to fully unregister a VanillaItem

Poslovitch:
Related to https://github.com/TheBusyBiscuit/Slimefun4/issues/231

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -60,7 +60,6 @@ public class SlimefunItem {
 	URID urid;
 	boolean ticking = false;
 	boolean addon = false;
-	boolean vanilla = false;
 	BlockTicker ticker;
 	EnergyTicker energy;
 	public String hash;
@@ -183,7 +182,7 @@ public class SlimefunItem {
 		addon = !slimefun;
 		try {
 			// Check if this item must be handled by Slimefun or return to its vanilla behavior
-			if(vanilla) SlimefunStartup.getItemCfg().setDefaultValue(this.name + ".ignore-slimefun-modifications", false);
+			if(this instanceof VanillaItem) SlimefunStartup.getItemCfg().setDefaultValue(this.name + ".ignore-slimefun-modifications", false);
 			if(SlimefunStartup.getItemCfg().getBoolean(this.name + ".ignore-slimefun-modifications")){
 			    all.remove(this);
 			    System.out.println("[Slimefun] Item has been put back to its vanilla behavior: " + this.name);
@@ -357,16 +356,8 @@ public class SlimefunItem {
 	    return ghost;
 	}
 	
-	public boolean isVanilla() {
-	    return vanilla;
-	}
-	
 	public void setReplacing(boolean replacing) {
 		this.replacing = replacing;
-	}
-	
-	public void setVanilla(boolean vanilla) {
-	    this.vanilla = vanilla;
 	}
 	
 	public void addItemHandler(ItemHandler... handler) {
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -39,7 +39,7 @@ public class SlimefunItem {
 	
 	public static Map<String, URID> map_name = new HashMap<String, URID>();
 	public static List<ItemStack> radioactive = new ArrayList<ItemStack>();
-	public static int vanilla = 0;
+	public static int slimefun = 0;
 	public static Set<String> tickers = new HashSet<String>();
 	
 	public static List<SlimefunItem> all = new ArrayList<SlimefunItem>();
@@ -60,6 +60,7 @@ public class SlimefunItem {
 	URID urid;
 	boolean ticking = false;
 	boolean addon = false;
+	boolean vanilla = false;
 	BlockTicker ticker;
 	EnergyTicker energy;
 	public String hash;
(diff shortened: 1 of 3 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -182,20 +182,32 @@ public void register() {
 	public void register(boolean slimefun) {
 		addon = !slimefun;
 		try {
+			// Check if this item must be handled by Slimefun or return to its vanilla behavior
+			if(vanilla) SlimefunStartup.getItemCfg().setDefaultValue(this.name + ".ignore-slimefun-modifications", false);
+			if(SlimefunStartup.getItemCfg().getBoolean(this.name + ".ignore-slimefun-modifications")){
```
TheBusyBiscuit: The last two lines make pretty much no sense...
There is no point for this vanilla boolean if it falls back to false at the end...

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/VanillaItem.java
```diff
@@ -9,5 +9,6 @@
 	
 	public VanillaItem(Category category, ItemStack item, String name, RecipeType recipeType, ItemStack[] recipe) {
 		super(category, item, name, recipeType, recipe);
+		setVanilla(true);
```
TheBusyBiscuit: Aaaaand doing that pretty much removes all Chain Armor Recipes, you could have just deleted all the code if that is what you wanted to do.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
Poslovitch: The "vanilla" boolean is here to add the line to the VanillaItem(s).
Well, I can do it differently using `this instanceof VanillaItem`. Let me change that.

TheBusyBiscuit: You are missing the point.
The vanilla boolean does absolutely nothing...

if (vanilla) -> false
else -> false

...

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/VanillaItem.java
Poslovitch: Why simply delete something interesting instead of providing an option to toggle that?
I personally like the ability to change vanilla items recipes : it make chain armor obtainable and elytras mor difficult to get ^^. But there is some people that don't like that. So I make it optional.

It seems you didn't understand how the code works. Well, I know this is a kind of a mess there, that's why I'll do what I said in the review above. However, the armor was removed only when I set `CHAIN_HELMET.ignore-slimefun-modifications` to true. Otherwise there was still the armor in Slimefun ^^

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
Poslovitch: The vanilla boolean acts like the "addon" boolean. If it is false, it will stay to false unless an addon does `item.setVanilla(true);`.
But I'll change it to what I said earlier. This will be more understandable

TheBusyBiscuit:
This regards all of the code above:

Think about the logic behind your code, 
some parts make logically absolutely no sense, meaning it would not work the way you probably think it would.

And even if it would work, there is actually no difference to just disabling the item...

Poslovitch:
Before submitting it, I tested my code in about "all" the possible cases.

And, there are differences between what I suggest and just disabling it.
When you disable the CHAIN_LEGGINGS, for example, you've got that in your chat when you try to use it :
spammy message

But I've got a problem. On my server, I have the possibility to get chainmail armors. But Slimefun doesn't let me use them.

Now, if I use what I just added.
Slimefun doesn't even register the item. This means that it doesn't disallow me to use the item, and so it doesn't spams me. And the recipes in Slimefun that use the chainmail armor (I have some because of a custom addon) are still working, because they now use the default chainmail armor.

TheBusyBiscuit:
You should not get this message...

https://github.com/TheBusyBiscuit/Slimefun4/blob/f1374d09239a780659521bf584ceffc664908f69/src/me/mrCookieSlime/Slimefun/api/Slimefun.java#L65*

Poslovitch:
Well, that's strange, but I get it.

However, I feel you aren't interested in the ability to toggle slimefun modifications / keep as vanilla for those VanillaItems. I really insist on the plus this ability can give to us, server owners.

TheBusyBiscuit:
https://github.com/TheBusyBiscuit/Slimefun4/blob/f1374d09239a780659521bf584ceffc664908f69/src/me/mrCookieSlime/Slimefun/api/Slimefun.java#L63

Wooops, okay whoever implemented that, forgot about the == null check prior.

The only reason I am against your changes is the fact that this should already work like that if you just disable the Item. Instead of applying some hacky new system to override the current one, fixing the current system (which should handle your problem already) may be wiser.

Poslovitch:
Hmmm. Okay I see. Thanks :)

TheBusyBiscuit:
No problem, by the way, here is an idea to fix it:

https://github.com/TheBusyBiscuit/Slimefun4/blob/c2f877b47cabcee30824a9b246244e03c4cdc0d2/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java#L308

Make that return an Enum UNKNOWN, DISABLED or VANILLA and go from there, in the method I sent earlier, the second loop in this method can also be replaced by a simple getByItem() call, you already have an instance to do the vanilla check on then as well.

Poslovitch:
I'll have a look ;)

PR #339, opened 2017-06-22
Added a State enum for SlimefunItem

Poslovitch:
Related to https://github.com/TheBusyBiscuit/Slimefun4/pull/337,
https://github.com/TheBusyBiscuit/Slimefun4/issues/231

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -230,7 +230,7 @@ public class SlimefunItem {
 				}
 				
 				if (SlimefunStartup.getCfg().getBoolean("options.print-out-loading")) System.out.println("[Slimefun] Loaded Item \"" + this.getName() + "\"");
-			} else if (!SlimefunStartup.getItemCfg().getBoolean(this.name + ".enabled")) {
+			} else {
 			    if (this instanceof VanillaItem) this.state = State.VANILLA;
 			    else this.state = State.DISABLED;
 			}
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -64,6 +64,14 @@ public class SlimefunItem {
 	EnergyTicker energy;
 	public String hash;
 	
+	State state;
+	
+	public enum State {
+	    ENABLED,
+	    DISABLED,
+	    VANILLA;
+	}
+	
 	int month = -1;
 	
 	public int getMonth() {
@@ -97,7 +105,6 @@ public class SlimefunItem {
 		this.replacing = false;
 		this.enchantable = true;
 		this.disenchantable = true;
-
 		itemhandlers = new HashSet<ItemHandler>();
 		
 		urid = URID.nextURID(this, false);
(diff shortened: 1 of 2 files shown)


src/me/mrCookieSlime/Slimefun/api/Slimefun.java
```diff
@@ -71,7 +71,7 @@ else if (isEnabled(p, item, message) && hasPermission(p, sfItem, message)) {
 			if (sfItem.getResearch() == null) return true;
 			else if (sfItem.getResearch().hasUnlocked(p)) return true;
 			else {
-				if (message && !(sfItem instanceof VanillaItem)) Messages.local.sendTranslation(p, "messages.not-researched", true);
+				if (message && sfItem.getState().equals(State.DISABLED)) Messages.local.sendTranslation(p, "messages.not-researched", true);
```
TheBusyBiscuit: Revert this.

src/me/mrCookieSlime/Slimefun/api/Slimefun.java
```diff
@@ -83,7 +83,7 @@ public static boolean hasUnlocked(Player p, SlimefunItem sfItem, boolean message
...
```
TheBusyBiscuit: Revert this.

src/me/mrCookieSlime/Slimefun/api/Slimefun.java
```diff
@@ -62,7 +62,7 @@ public static boolean hasUnlocked(Player p, ItemStack item, boolean message) {
 		SlimefunItem sfItem = SlimefunItem.getByItem(item);
 		if (sfItem == null) {
 			if (SlimefunItem.isDisabled(item)) {
-				if (message && !(sfItem instanceof VanillaItem)) Messages.local.sendTranslation(p, "messages.disabled-item", true);
+				if (message && SlimefunItem.getState(item).equals(State.DISABLED)) Messages.local.sendTranslation(p, "messages.disabled-item", true);
```
TheBusyBiscuit: You can save a lot of memory by storing the State in a variable and then doing state != ENABLED instead of isDisabled and then doing state != VANILLA for the message thing.

src/me/mrCookieSlime/Slimefun/api/Slimefun.java
```diff
@@ -60,9 +61,10 @@ public static void registerResearch(Research research, ItemStack... items) {
 
 	public static boolean hasUnlocked(Player p, ItemStack item, boolean message) {
 		SlimefunItem sfItem = SlimefunItem.getByItem(item);
+		State state = SlimefunItem.getState(item);
 		if (sfItem == null) {
 			if (SlimefunItem.isDisabled(item)) {
```
TheBusyBiscuit: You can replace isDisabled() with state != ENABLED

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -220,6 +230,9 @@ else if (SlimefunStartup.getItemCfg().getBoolean(this.name + ".enabled")) {
 				}
 				
 				if (SlimefunStartup.getCfg().getBoolean("options.print-out-loading")) System.out.println("[Slimefun] Loaded Item \"" + this.getName() + "\"");
+			} else if (!SlimefunStartup.getItemCfg().getBoolean(this.name + ".enabled")) {
```
TheBusyBiscuit: It comes right after the opposite check, so you just have to write

else, no need to specify an else if

Poslovitch: Okay, I was not sure because of the first check (#isTicking() ...)

TheBusyBiscuit:
Youre right with the ticking check, to cover that, you should actually do an if, not an else, nor an else if.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -202,10 +209,14 @@ public void register(boolean slimefun) {
 			}
 			
 			if (this.isTicking() && !SlimefunStartup.getCfg().getBoolean("URID.enable-tickers")) {
-				
+				// What is needed here?
```
TheBusyBiscuit: return; because you removed the else.

TheBusyBiscuit:
Good spot! The DISABLED State is needed as well.

PR #70, opened 2016-12-27
Research requirements & better display

Poslovitch:
Added research requirements (needed researches before being able to unlock another)

I also improved their display in the Slimefun Guide GUI (I show research name and id and what items it will unlock)

I did unit testing. Everything worked fine.
Feel free to try out and to make changes requests if needed.

PR #340, opened 2017-06-23
Documented and improved Researches

Poslovitch:
I documented the whole Research.java class (this is the beginning of the huge task of documenting about the whole Slimefun API ^^).

I also edited, renamed, and created some methods for Research. 

I'd also like to know if I can repush https://github.com/TheBusyBiscuit/Slimefun4/pull/70. It seems MySQL implementation has not begun yet.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/Research.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/Research.java
@@ -32,15 +32,15 @@ import org.bukkit.entity.Player;
  * @since 4.0
  */
 public class Research {
-    /**
-     * Whether researching is enabled or not
-     * @since 4.0
-     */
-    public static boolean enabled;
-    /**
-     * Contains all the registered researches
-     * @since 4.0
-     */
+	/**
+	 * Whether researching is enabled or not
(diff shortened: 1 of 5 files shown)


src/me/mrCookieSlime/Slimefun/Objects/Research.java
```diff
@@ -13,74 +13,233 @@
...
-	
+
+	/**
+	 * Sets the cost in XP levels to unlock the research
+	 * 
+	 * @param cost Cost in XP levels
+	 * 
+	 * @since 4.1.10
+	 */
+	public void setCost(int cost) {
```
TheBusyBiscuit: Not really necessary since the cost is overridden by the file anyway.

TheBusyBiscuit:
You should not repush #70 

I still think that the entire Research System just needs an overall rewrite.
The numeric IDs need to be replaced.
SQL support needs to be added, etc...
When there is a different ID system, your requirement tree stuff can gladly be added.
But Researches should really swap to a different ID system.

Poslovitch:
Okay, gotcha! thanks !

PR #372, opened 2017-07-27
"mavenize"

RoboMWM:
Similar to PR in CS-Corelib, though this one is a bit more interesting, since you depend on slimefun addons in this plugin...


TheBusyBiscuit:
These Addons are on GitHub, couldn't these be served by mavenizing them as well?

RoboMWM:
I can't mavenize the others without sticking adding the slimefun jar in it like I've done here (since slimefun is not yet mavenized).

I could fork these and add the jars to push to a maven repo - then your plugins would temporarily depend on my maven repo/jitpack variant. They then could be switched over, one by one I suppose; but I have not dealt with circular dependencies in maven before, so I'm not sure if that will be an issue here.

TheBusyBiscuit:
Right, so mavenize this first, let's hope that this dependency loop will work once Slimefun is on Maven.

RoboMWM:
Just fyi, this won't compile via jitpack since we are missing the coreprotect dependency, so I can't mavenize the other addons without adding the slimefun jar at the moment. I'll look into setting up a maven repo to test (where I can deploy to it from a computer that does have the coreprotect jar), but it will be a while before I get to that.

RoboMWM:
Found that bintray provides free maven repo hosting (and also realized that's what aSkyBlock uses). Although it does look for itself after I mavenized ExoticGarden, it doesn't seem to recur back after that (probably because there's no specified repo for ExoticGarden, as per this PR). So it _seems to be okay_ in that standpoint; no idea if future versions will recur all the way back to this first version without a maven repo for ExoticGarden. Maybe someone better-versed in maven can let us know.

(Thinking about this situation in general though, in order to do this recursive-depending, you'd first need a copy of an older version of the plugin; i.e., probably not great design to depend on things that depend on you...)

TheBusyBiscuit:
CoreProtect is licensed under All Rights Reserved, therefore we cannot host or serve it at all.
So, add it as a system path but don't push the lib directory, also just delte the lib directory alltogether and add it to .gitignore
EmeraldEnchants and ExoticGarden will be servable anyway, right?

PR #373, opened 2017-07-28
Use jitpack for exoticgarden dependency, add maven repo for coreprotect, emeraldenchants

RoboMWM:
Intelli's fine with distributing the CoreProtect jar, just asked him in his IRC channel. But from my understanding you want to limit jars in the git repo anyways.

I don't really feel like mavenizing Emerald Enchants as it seems to be inactive and requires NMS and PrisonUtils, both of which can't be mavenized (unless PrisonUtils isn't your plugin - I'm assuming it is since I remember seeing something about it in the description of one of your plugins). If you still don't want the jar in the repo, I can serve the jar in my maven repo if you'd like.

Also, if you want me to remove the deployment settings from the pom just let me know - doesn't hurt to keep it there and makes it easier for me to deploy new updates.


TheBusyBiscuit:
EmeraldEnchants doesn't use NMS...
But good that you asked him, still as you said Yes, I do very much like not having any distributables in the source.

Serve it or just remove it from the lib folder, whatever floats your boat.

And I don't really care about the deployment xpath.

RoboMWM:
Oh, I assumed that Craftbukkit meant CB code, not bukkit API code.

RoboMWM:
Should also be usable from jitpack as well after this.

TheBusyBiscuit:
"And I don't really care about the deployment xpath."
I actually meant that as in you can keep it if you want to but ok.

"Oh, I assumed that Craftbukkit meant CB code, not bukkit API code."
Yeah okay, it could be misleading but I just used to always compile against CB so I have NMS disposable at my hand for when I need it but all my projects use plain Bukkit API calls or Reflection from CS-CoreLib to access NMS but I had CB in my classpath so I know the method/class names to reflect.

RoboMWM:
> "And I don't really care about the deployment xpath."

Yea, I decided to remove it though since I realized Jitpack can serve this now.

TheBusyBiscuit:
Oh ok, also a heads up that the next commit will be No. 400.

PR #383, opened 2017-08-10
Fixes IllegalArgumentException with negative randomize bound with Magician Talisman

Poslovitch:
Fixes https://github.com/TheBusyBiscuit/Slimefun4/issues/366


TheBusyBiscuit:
ArrayLists cannot have negative bounds. The error lays inside the randomizer method of CS-CoreLib...
Just return 0 if n < 1 in there.

PR #384, opened 2017-08-10
Fixes a NumberFormatException when trying to read nonexistent frequency of a CargoManager

Poslovitch:
Fixes https://github.com/TheBusyBiscuit/Slimefun4/issues/363
Fixes https://github.com/TheBusyBiscuit/Slimefun4/issues/161


andris155:
Regular output node doesn't read info about cargo channels, no matter what channel it's set to, it'll draw items from every channel even if certain items have an output given by advanced output nodes

PR #387, opened 2017-08-12
Fixed radioactive items giving effects even if they were disabled

Poslovitch:
Fixes https://github.com/TheBusyBiscuit/Slimefun4/issues/297


TheBusyBiscuit:
Better use isEnabled() since the item may be disabled in the Items.yml

PR #396, opened 2017-08-14
Deleted Charge - unused

Poslovitch:
Should we delete this class?
I've never seen it.


TheBusyBiscuit:
Could be used somewhere, not deleting it wont do any harm.

PR #398, opened 2017-08-17
Crafting a Water Ancient Rune now gives 4 instead of 2

Poslovitch:
Water Ancient Rune is more expensive than the 3 other runes, and we obtain less than them... This is inconsistent.


TheBusyBiscuit:
It is not expensive, besides it is only used for Essense Of Afterlife, which should not be too easy to get, hence why I limited it through this back then.

This would make sense if the Water Rune had any practical use, but sure I can merge it anyway for your OCD

PR #457, opened 2017-10-14
Fix ConcurrentModificationException that occurred when reactor explodes.

dpkirchner:
CraftBukket's HumanEntity.closeInventory() apparently removes the human
from the inventory's viewer list, resulting in an exception when we were
calling closeInventory while iterating over the list of said viewers.

Fixes TheBusyBiscuit/Slimefun4#364.


dpkirchner:
Thanks, fixed this in my local repo. I'll include it in the next PR I'm working on.

PR #460, opened 2017-10-17
Fix EnhancedFurnace not showing speed

VanishTime:


andris155:
What is true, false?

Poslovitch:
True is force update, false is something else I forgot. Check Spigot's javadoc

Diinl:
False is no update, no block state changes will occur if its not the same type as its initial state type.

VanishTime:
God... Am I wrong?

Issue #411, opened 2017-08-27
Carbonado Edged Furnace either behaves vanilla or duplicates items


FatherWh0:
## Description
Most of the time the carbonado edged furnace behaves as a vanilla furnace. On rare occasion, when one is loaded with ores and fuel, the furnace gui closes. Then the furnace begins to generate drops of the ore and fuel it has been loaded with. Attempting to reopen the gui to remove items results in the gui closing itself again. The furnace must be broken (mined) to stop the production of drops. In the images provided the drops are being generated by the furnace.

## Server Log
There are no unusual log entries or errors that occur.
https://gist.github.com/FatherWh0/c7300ff46ab1d45ded589cbdc65a8af3

## Evidence / Proof
https://prnt.sc/gds8pg
http://imgur.com/fI1JBNo

## Environment
Server: paper 1.12.1 R0.1 # 1195
Slimefun v4.1.10
CS-CoreLib v.1.5.16
SlimefunGuidPlus v1.0
ExoticGarden v1.6.3
SlimefunLuckyBlocks v1.1

andris155:
same problem

100petr:
It also makes me...

galacticwarrior9:
Also having this. Very serious issue.

100petr:
almost 2 weeks and still unpatched.

mrcoffee1026:
Seems to preserve the fuel efficiency but runs same speed as vanilla oven.

Goldenphelix:
Issue present in Slimefun 4.1.10, serious issue for my server, please fix this asap!!
Plugins:
Slimefun v4.1.10
CS-CoreLib v.1.5.16
ExoticGarden v1.6.3
SlimefunLuckyBlocks v1.1

mrcoffee1026:
Confirmed item duplication.. although it also appears to be duplicating whatever is present in the fuel slot also (went to smelt pair of gold boots... oven began exploding out drops of boots and wood planks... it only took me a moment to mine the thing, but by then I had 10 golden boots and 10 stacks of wood planks to collect along with the oven itself).  Same versions listed less the Luckyblocks.

fonkfader:
same issue...

Stoniasawsome:
same issue

FatherWh0:
The problem has now spread to the Enhanced Furnace XI.

100petr:
Author is probably ignoring the problem -_-

[...]

PR #461, opened 2017-10-19
Compatibility with 1.12

VanishTime:
Repair enhanced furnace speed is the same as the original.
(Sorry of my bad english...)

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/EnhancedFurnace.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/EnhancedFurnace.java
@@ -29,10 +29,11 @@ public class EnhancedFurnace extends SlimefunItem {
 					if (!(b.getState() instanceof Furnace)) {
 						return;
 					}
-					if (((Furnace) b.getState()).getCookTime() > 0) {
-						((Furnace) b.getState()).setCookTime((short) (((Furnace) b.getState()).getCookTime() + getSpeed() * 10));
+                    Furnace state = (Furnace) b.getState();
+					if (state.getCookTime() > 0) {
+						state.setCookTime((short) (((Furnace) b.getState()).getCookTime() + getSpeed() * 10));
 					}
-					b.getState().update(true, false);
+					state.update(true, false);
 				} catch(NullPointerException x) {
 				}
 			}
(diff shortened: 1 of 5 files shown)


TheBusyBiscuit:
You literally haven't changed anything!?

VanishTime:
Sorry...but the previous way didn't work on PaperSpigot...

VanishTime:
Maybe it's because of 1.12 update, but this will solve the problem.

andris155:
kepkivagas

VanishTime:
Looks...I forget somethings..?

andris155:
Thanks!

andris155:
I using this pr and Slimefun furnace bugged.. https://www.youtube.com/watch?v=vveIdyuV4uA

VanishTime:
Sorry...this bug isn't made by me.  #411 
But can you tell me your environment? I didn't happen this.

mrcoffee1026:
The oven has had that bug since 4.10. (vid linked by andris155).  If your pull request repairs the speed issue that has been present since 4.9... it doesn't fix this particular issue is what he's saying.  It doesn't happen ALL the time.  The oven occasionally drops loads of fuel and items it's to be smelting non stop... independently of what was originally put into the oven.  Too bad this pull request wasn't dealing with that issue, that makes me sad.

akashaggarwal7:
@mrcoffee1026 Is this why the EnhancedFurnace has been removed from 1.12 menu? At least it doesn't show up for me in the latest build..

mrcoffee1026:
@akashaggarwal7 no this has nothing to do with that?  Since they're still showing in the menu for me I think it must be for some other reason they're not appearing.  I've not attempted *this* pull request, I rolled back to release 1.9 since 1.10 broke cargo.  Dev stated 1.11 fixes cargo again, but since there still isn't a 1.11 available to download... anyway, 1.9 does not have the bizarre item/fuel duplication/drop issue that 1.10 has either, however both 1.9 and 1.10 have the speed issue.  Both of these releases (1.9 and 1.10) appear to obey the "efficiency", however, using fuel more effectively than vanilla ovens even though they cook as slow as vanilla ovens.  If This pull fixes the speed, it seems 1.10 needs an additional revert to undo whatever else was done to the oven, but having it mysteriously vanish wasn't one (this may have been covered by the other bug report with the lava bucket earlier).

akashaggarwal7:
@mrcoffee1026 I figured it, its because the contents of the config file doesn't get copied. So the config file only has comments and no configuration.

Poslovitch:
This PR doesn't correspond to how the code actually works : it adds a "placeable" variable to SlimefunItem, which is technically already done by the interface NotPlaceable, and is an option that shouldn't be given to the users of the plugin.

TheBusyBiscuit:
Not mergeable due to file conflicts.
Not properly done as @Poslovitch stated.

PR #468, opened 2017-10-24
Fix cargo input/output with furnace

Dinnerwolph:
fix https://github.com/TheBusyBiscuit/Slimefun4/issues/114


TheBusyBiscuit:
1. Making the array static, could result in memory leaks, not a huge impact but we should prevent those nonetheless. And since it isn't referenced outside of that class (except for your use of it) it isn't needed anyway.
2. Instead of Class.forName you should refer to CS-CoreLib's ReflectionUtils.class it does all of that automatically for you, also don't iterate over the supported Versions and append R1 etc... just get the current version...
3. Getting the slot based on the face does not make too much sense because they are global and not per block, making them really really unintuitive.

PR #469, opened 2017-10-24
fix cargo input/output slot with furnace

Dinnerwolph:
fix https://github.com/TheBusyBiscuit/Slimefun4/issues/114 with changement


TheBusyBiscuit:
This is not what I meant.
Cargo Nodes need to be placed on the side of a block, please do not change that.

Dinnerwolph:
can be from the node inventory ?

TheBusyBiscuit:
What?

Dinnerwolph:
a button that selects the slot

TheBusyBiscuit:
That would be cool, you could add an Advanced Input Node which can do that.

PR #472, opened 2017-10-26
Updated deprecated methods

kiritss:


src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
```diff
@@ -158,7 +157,7 @@ private void insertItem(Player p, Block b) {
 	}
 
 	@EventHandler
-	public void onPickup(PlayerPickupItemEvent e) {
+	public void onPickup(InventoryPickupItemEvent e) {
```
TheBusyBiscuit: This Event is now assigned twice...

TheBusyBiscuit:
Also, please specify for which Versions of Minecraft this is compatible with.

src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
andris155: PlayerPickupItemEvent replaced EntityPickupItemEvent.

PR #482, opened 2017-11-01
Minor fix and enhancement to tab completion

akashaggarwal7:
Fixed a typo and added the ability to complete words on tab completion smartly.


John000708:
This might cause problems with the Colored Enderchests, since there's thousands of them

akashaggarwal7:
@John000708 Item name(s)?

John000708:
@akashaggarwal7 yeah, if I can recall correctly, it adds 8192 new items which means 8192 new ids your loop has to go through.

akashaggarwal7:
@John000708 I'm unaware of them being a part of slimefun. Is it a planned released or another plugin?

John000708:
Here you go: https://dev.bukkit.org/projects/colored-ender-chests

akashaggarwal7:
@John000708 Well that sucks.. I could remove it for the give and research command for now and perhaps we can use a better algorithm for search in the future(I can't think of one right now)?

PR #483, opened 2017-11-01
Removed smart tab completion support from research and give command t…

akashaggarwal7:
…o avoid problems with large lists

Colored Enderchest adds a lot of items. My connection on my local server was reset when trying to work with that plugin.


TheBusyBiscuit:
Just use a custom list that drives from the SlimefunItem list but remove all Items like Colored Ender Chests.

akashaggarwal7:
@TheBusyBiscuit People won't be able to give the colored enderchests if we do that.

TheBusyBiscuit:
They can just type it in manually...

PR #484, opened 2017-11-01
Better implementation for tab completion, adding it back for give and research command

akashaggarwal7:
Makes a copy of the list initially when the tab completer is loaded and works on the list asynchronously if needed.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Commands/SlimefunTabCompleter.java
+++ b/src/me/mrCookieSlime/Slimefun/Commands/SlimefunTabCompleter.java
@@ -19,7 +19,7 @@ public class SlimefunTabCompleter implements TabCompleter {
 		}
 		else if (args.length == 3) {
 			if (args[0].equalsIgnoreCase("give")) {
-				return createReturnList(Slimefun.listIDs(), args[2]);
+				return Slimefun.listIDs();
 			}
 			else if (args[0].equalsIgnoreCase("research")) {
 				List<String> researches = new ArrayList<String>();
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
0_0 No no no, do not do that.

Slimefun Addons won't be loaded midtime for reasons...
Just wait till all Items have been loaded, then truncate a copied list.

I think there is a Slimefun.registerPostLoadingHandler method for this.

TheBusyBiscuit:
I am aware why you implemented it. But all Slimefun Addons cannot be en/disabled during runtime, the current system doesn't allow plugins like PlugMan to do this, hence why this check is unnecessary.

I recommended a PostLoadingHandler since I am unsure when the class constructor will be called, there could still be Items added after the class instance was created.

Either do a PostSlimefunHandler instance and pass it to SlimefunSetup.registerPostLoadingHandler, or create the list in the MiscSetup class, otherwise you may not pick up any Addon Items.

And this will affect the performance, as you are calling .contains() on a huge list everytime tab is pressed, a check that isn't necessary.

Lastly, checking whether the list contains COLORED_ENDER_CHEST_SMALL_0_0_0 or COLORED_ENDER_CHEST_BIG_0_0_0 may work in most cases, but what if a Server Owner disabled white-white-white combinations because of some unforeseen reason, may be a better idea to check if the plugin is installed instead.

akashaggarwal7:
I guess we can work with adding an easier fix which compares the size of the list with a number say 3000? I'm not sure about slimefun addons so idk if this approach would be suitable and if so, what should be the number.

PR #497, opened 2017-11-17
added some blocks in android blacklist

yhdesai:
Added some more blocks to sf black list
1) Android Interface : android miner being to close to interface sometimes mined the interface when a wrong program was coded.
2) Gold Blocks : miner could be used to convert those sf gold blocks to vanilla gold blocks


TheBusyBiscuit:
That's not how this works.
That's not how any of this works.

Issue #491, opened 2017-11-12
Opening the SF guide with commands


oddare:
## Description
Running a Skyblock I am attempting to let my players merely use a GUI for all their commands as this is faster and easier for the majority.

The SF guide, however, is currently excluded from this as it so far requires the book to get into. Is there a way I can get into this using commands, instead of routing through to the book?

TheBusyBiscuit:
Someone should just make an Addon for this, it is a very small request.

ScuroK:
i dont agree. i think things like this are quite basic and should be included in a plugin like slimefun.

Poslovitch:
Gonna do it. This will be quite easy.
I will take this opportunity to add some features into the guide. :)

Grinthy:
On a server I play on, there's a command called **/sf guide** that lets you get a book with the default chest GUI.  Is that an add on @TheBusyBiscuit ?

Poslovitch:
Nope, it isn't. It gives you the book

Poslovitch:
Should be closed now ^^

PR #498, opened 2017-11-18
Opening the SF guide with commands #491

Redemption198:
From the suggestion of @oddare (https://github.com/TheBusyBiscuit/Slimefun4/issues/491), I've added the command to open the Slimefun Guide just typing "/sf open_guide". I also added the tab completion "open_guide" and a description of it.

### DIFF
--- a/src/plugin.yml
+++ b/src/plugin.yml
@@ -36,7 +36,7 @@ permissions:
     default: op
   slimefun.command.guide:
     description: Allows you to obtain the Slimefun guide book
-    default: op
+    default: true
   slimefun.command.open_guide:
     description: Allows you to open the SF guide without the book
     default: op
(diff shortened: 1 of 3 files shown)


TheBusyBiscuit:
You could add a permission as well.

Redemption198:
If there is something that you don't like just let me know.

oddare:
Dang, this is awesome! :D

Redemption198:
I was looking at SlimefunCommand.java and I saw that the command that gives the books doesn't check for any permission, I was wondering why.

src/plugin.yml
```diff
@@ -34,6 +34,9 @@ permissions:
   slimefun.command.versions:
     description: Allows you to do /sf versions
     default: op
+  slimefun.command.guide:
+    description: Allows you to obtain the Slimefun guide book
+    default: op
```
TheBusyBiscuit: default: true

please

Redemption198: Done. Shall I change open_guide default too?

TheBusyBiscuit: No.

Issue #499, opened 2017-11-20
Ancient Pedistal glitch


xH3LLRAIZ3Rx:
## Description
When you have a item on a pedistal and you place a fence on top of pedistal, it makes the item to where you cant pick it up

## Server Log
Theres no errors in console, I triple checked

## Evidence / Proof
No proof besides what a player told me

## Environment
Spigot v1_12_R1
CS-CoreLib v1.5.16
Slimefun v4.1.11
ExoticGarden v1.6.3
PrivateStorage v3.0
SlimeXpansion v1.1.9
ChestTerminal v1.0
EmeraldEnchants v2.0
SlimefunLuckyBlocks v1.1
ElectricSpawners v1.0
MagicLoot v3.0
ExtraHeads v1.0
Dolly v2.06
SoulJars v1.0

PR #500, opened 2017-11-21
Fixed ancient pedestal bug #499

Redemption198:
I've added a check in the Pedestal listener, now if the block above the pedestal is not air, the item will not be added and the player will be notified about that. 

P.S. The item can still be pushed by a piston, I don't know how this can be fixed.


Redemption198:
@xH3LLRAIZ3Rx reported this issue on (https://github.com/TheBusyBiscuit/Slimefun4/issues/499)

src/me/mrCookieSlime/Slimefun/Setup/Messages.java
```diff
@@ -86,6 +86,7 @@ public static void setup() {
 		local.setDefault("machines.ANCIENT_ALTAR.unknown-catalyst", "&4Unknown Catalyst! &cUse the correct Recipe instead!");
 		local.setDefault("machines.ANCIENT_ALTAR.unknown-recipe", "&4Unknown Recipe! &cUse the correct Recipe instead!");
 		local.setDefault("machines.HOLOGRAM_PROJECTOR.enter-text", "&7Please enter your desired Hologram Text in your Chat. &r(Color Codes are supported!)");
+		local.setDefault("PEDESTAL.obstructed", "&4Pedestal is obstructed! &cRemove the block above the pedestal!");
```
Poslovitch: Please use machines.ANCIENT_PEDESTAL.obstructed instead ;)

Redemption198: Done :)

TheBusyBiscuit:
This is an attempt to 'avoid' the Issue, but doesn't fix it.
What if you place an Item, then a Block?

That is also what this Issue was actually about.
This is halfly done, you could prevent the block placement as well.

Redemption198:
I've added the event that prevents the block placement, but water and pistons can still push away the item.

Redemption198:
@Poslovitch without line 188 the server throws org.bukkit.event.EventException: null

Poslovitch:
Okay gotcha

src/me/mrCookieSlime/Slimefun/Setup/Messages.java
```diff
@@ -72,6 +72,7 @@ public static void setup() {
 		local.setDefault("commands.research.reset-target", "&cYour Knowledge has been reset");
 		local.setDefault("machines.in-use", "&cThis Block's Inventory is currently opened by a different Player.");
 		local.setDefault("machines.ignition-chamber-no-flint", "&cIgnition Chamber is missing Flint and Steel.");
+    		local.setDefault("messages.cant-place" ,"&cYou can't place that block there!");
```
Poslovitch: Change "messages.cant-place" to "messages.cannot-place". This is a detail, but we have to keep a "neutral" english for convenience reasons.

Poslovitch:
Then, this will be good to go for me :)

Redemption198:
@Poslovitch thanks for your advice :)

src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
```diff
@@ -174,5 +179,17 @@ else if (!e.getItem().hasMetadata("no_pickup") && e.getItem().getItemStack().has
 			e.getItem().remove();
 		}
 	}
+	
+	@EventHandler
+	public void onBlockPlace(BlockPlaceEvent e) {
+		if(e.isCancelled()) return;
```
TheBusyBiscuit: Use @EventHandler(ignoreCancelled=true) instead.

TheBusyBiscuit: Also you should give it a priority.

PR #501, opened 2017-11-22
Explosive pickaxe could break BARRIER and COMMAND blocks

Poslovitch:


TheBusyBiscuit:
This however only sets this as a default value, existing Servers won't get this fix.

Poslovitch:
Yeah, I know. Should we make it like the ProgrammableAndroid then? It would remove any configurability, but will make sure that the servers will all have the fix ^^

PR #503, opened 2017-11-22
Ore washer dupe fix

AtomicScience:
Ore washer has a bug, which lets players dupe certain kind of dust. So, I created a tiny fix, which'll remove this dupe.

Let's, for example, dupe Copper Dust. For this, we have to fill the Dispencer of Ore Washer in this way:
image
Where:
image - Sifted Ore
image - Copper Dust.
image - Any other item. For example i used a Gravel

While clicking Ore Washer, we will get messages about overflow of Dispenser's inventory, but, with some probablity, our machine'll generate us a Copper Dust.
It work with any kind of Dusts.

**Why it works:**
CSCoreLib's method InvUtils.fits() returns true even when there're no free slots in the inventory.
So, I replaced it with a check for a free slot. (inv.firstEmpty() returns -1 only if inventory have no free slots)

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -1337,8 +1337,8 @@ public class SlimefunSetup {
 										removing.setAmount(1);
 										inv.removeItem(removing);
 										inv.addItem(adding);
-										p.getWorld().playEffect(b.getLocation(), Effect.STEP_SOUND, Material.WATER);
 										p.getWorld().playSound(b.getLocation(), Sound.ENTITY_PLAYER_SPLASH, 1, 1);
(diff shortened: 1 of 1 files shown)


andris155:
What is texture link?

Poslovitch:
There isn't any "texture link" reference in this PR

Poslovitch:
I confirm this fix is working, however, it would be better to fix the CSCoreLib's #fits() method, instead of having to apply this workaround everywhere.

AtomicScience:
No, .fits() works fine, but it's logic lets players dupe items.
I think, in CSCoreLib must be added new method, let's name it .hasEmptySlots(inv).
It must look like this:
```java
public static void hasEmptySlots(Inventory inv) {
    return inv.firstEmpty() != -1;
}

Poslovitch:
Okay, this suggested method might be added later in CSCoreLib, but for now, let's use the current fix.

This PR can be merged ;)

PR #504, opened 2017-11-23
Fixed Android being able to break command blocks

Poslovitch:


TheBusyBiscuit:
I'm holding back on this for now, when were these blocks added?

Poslovitch:
1.9. I doubled check.

TheBusyBiscuit:
Then I haven't opened Minecraft for longer than I thought...

PR #506, opened 2017-11-25
Error reports : prioritized Slimefun Environment (addons, etc)

Poslovitch:
Done to allow an easier reading of plugins in error reports in order to find the addons.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/api/TickerTask.java
+++ b/src/me/mrCookieSlime/Slimefun/api/TickerTask.java
@@ -3,11 +3,9 @@ package me.mrCookieSlime.Slimefun.api;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.PrintStream;
-import java.util.HashMap;
-import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -136,15 +134,31 @@ public void run() {
...
+																if (p.getDescription().getDepend().contains("Slimefun") || p.getDescription().getSoftDepend().contains("Slimefun"))
+																	addons.add("  - " + p.getName() + " " + p.getDescription().getVersion());
 															}
 														}
+
+														stream.println(" Installed Addons (" + addons.size() + ")");
+														for (String addon : addons) stream.println(addon);
+														stream.println();
+														stream.println("Installed Plugins (" + plugins.size() + "):");
+														for (String plugin : plugins) stream.println(plugin);
```
TheBusyBiscuit: I would rather always use brackets for loops, makes it more readable in my opinion.

PR #502, opened 2017-11-22
SlimefunItem : renames of some methods, made ghost configurable, changed "name" to "id" and privatized all fields

Poslovitch:
This PR adds or modifies to the SlimefunItem and its derivative classes : 
- all fields are now private, and can only be accessed using a method
- "name" is now "ID"
- some documentation has been added (really succinct, eg. only the version a method has been added, etc)
- "ghost" is now "hidden", as well as being configurable in Items.yml with "hide-in-guide"
- some fields are initialized inline instead of constructors (to make reading a bit easier)

PR #507, opened 2017-11-25
4.1.12

Poslovitch:
Fixes a few bug, and update Slimefun's version to 4.1.12.

However, #502 must be merged before this one.


TheBusyBiscuit:
#502 is in no relation to this, it is even on an entirely different branch, therefore it is not required to be merged first...

Issue #333, opened 2017-06-22
Slimefun Juices with Saturation do nothing. (MC1.11.2 and MC1.12)


RightyBird:
## Description
While using Minecraft 1.11 or higher (confirmed in 1.12 as well), drinking any potion that gives saturation is consumed with no effect given.

## Server Log
https://pastebin.com/PcNLe28Y
Note: No server log messages at time of consumption. 

## Evidence / Proof
Simple test.  Using any 'juice' from either slimefun or exotic garden that is supposed to satisfy hunger will show it has no saturation effect when consumed.  Other potions work.  Tested on MC1.11.2 and MC1.12

## Environment
CraftBukkit v1_12_R1
CS-CoreLib v1.5.16
Slimefun v4.1.9
Installed Addons (8)
HotbarPets v1.1
ExoticGarden v1.6.3
ZeldaHearts v2.0.2
SoulShards v1.9.5
SlimeXpansion v1.1.9
SlimefunOres v1.4.6
Shelter v1.0.2
SlimefunGuidePlus v1.0

## Research
https://bugs.mojang.com/browse/MC-112389 - seems to be related to Minecraft 1.11.2 and up.

## Suggested solution
Ability to turn-off juices selectively, or the migration of all recipes into an external text file that could be red and modified.  I understand I can comment out the juices in the code but this is not a very plugin friendly solution.  It would be better if controlled by a .config or a .yml file

Poslovitch:
Your suggested solution looks similar to #325. I will work on that.

TheBusyBiscuit:
@Poslovitch 
I would suggest you to write an Interface for SlimefunItems such as IStaticRecipe or so that prevents Recipe Editing. Backpacks for example or ExoticGarden Bushes or Machine-internal Recipes like the Crucible rely heavily on static Recipes.
Also, this should rather be an Addon, side note: Addons can call SlimefunItem#setRecipe()
You could export/import JSON files for this but keep in mind that fancy JSONs can take massively longer to load.

Goldentoenail:
Concerning the Juice consumption issue.
I have looked at the class me.mrCookieSlime.Slimefun.listeners.ItemListener
At the Event 'public void onEat(PlayerItemConsumeEvent e)', line 276 which handles the Juices. There is no player consume event dealt with.

Sorry, if I have missed this happening elsewhere.

TheBusyBiscuit:
@Goldentoenail 
There is no Event needed.
[...]

Issue #87, opened 2017-01-05
Juices do not restore hunger/saturation


coldcode69:
## Description
None of the juices (Carrot Juice, Melon Juice, etc) restore any hunger or saturation when using spigot 1.11.2.

## Evidence / Proof
Checked using a clean server.
![hunger](https://cloud.githubusercontent.com/assets/24900603/21663799/5de1378a-d2da-11e6-92cc-7147c5ed43c2.gif)

I also checked the NBT data in the player file to see if the hidden saturation value changed at all.

Before:
![nbt1](https://cloud.githubusercontent.com/assets/24900603/21663808/6ea94544-d2da-11e6-8a20-7ca2ed4e2278.png)
After consuming all juices:
![nbt2](https://cloud.githubusercontent.com/assets/24900603/21663810/700c3a0e-d2da-11e6-9f95-28cdadabeb3d.png)

## Environment
spigot 1.11.2
cscorelib version 1.5.12
slimefun version 4.1.1

NathanAdhitya:
Anything on console?

SoSeDiK:
Same problem (also with ExoticGarden). I just added the command /effect [player] 23 [time] [level] (/effect [player] 23 3 1 for juice, for example), when the potion is consumed. Maybe there is a better way.

Momshroom:
Sorry, how did you do that? :3

CyberiumShadow:
`	@EventHandler
	public void onStarve(FoodLevelChangeEvent e) {
		if (e.getFoodLevel() < ((Player) e.getEntity()).getFoodLevel()) {
			Player p = (Player) e.getEntity();
			for (ItemStack item: p.getInventory().getContents()) {
				if (SlimefunManager.isItemSimiliar(item, SlimefunItem.getItem("COOLER"), false)) {
					Inventory inv = Backpacks.getInventory(p, item);
					if (inv != null) {
						ItemStack drink = null;
						for (ItemStack i: inv.getContents()) {
							if (i != null && i.getType() == Material.POTION && i.hasItemMeta()) {
								drink = i;
								break;
							}
						}
						if (drink != null) {
							PotionMeta im = (PotionMeta) drink.getItemMeta();
							for (PotionEffect effect: im.getCustomEffects()) {
								p.addPotionEffect(effect);
							}
							p.setSaturation(6F);
							p.playSound(p.getLocation(), Sound.ENTITY_GENERIC_DRINK, 1F, 1F);
[...]

PR #513, opened 2017-11-29
Fixes juices not restoring hunger - #333 & #87

Poslovitch:


src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
```diff
@@ -283,8 +285,18 @@ else if (SlimefunManager.isItemSimiliar(item, SlimefunItems.MEDICINE, true)) {
 				else if (item.getType() == Material.POTION) {
 					SlimefunItem sfItem = SlimefunItem.getByItem(item);
 					if (sfItem != null && sfItem instanceof Juice) {
+						// Fix for 1.11 and 1.12 where Saturation potions are no longer working
+						if (ReflectionUtils.getVersion().startsWith("v1_11_") || ReflectionUtils.getVersion().startsWith("v1_12_")) {
```
TheBusyBiscuit: Heavy call.

Should be saved in a class-level variable as it is an expensive operation, since it performs code reflection.
Also, to make future support easier, I'd instead check for 1_10 and 1_9.

Poslovitch: For heavy calls, here is a better fix : https://github.com/TheBusyBiscuit/CS-CoreLib/pull/32/files

I prefered to check for 1.11 and 1.12 because I've seen on MC's bug tracker that a fix is planned for 1.13

Poslovitch: My mistake, I just doubled check https://bugs.mojang.com/browse/MC-112389.
No fix is currently planned.

Let's switch to 1.9 and 1.10 to make 1.13 support easier to do.

TheBusyBiscuit:
Now I am confused, I meant a negative check.

Just find out what Version this needs to be applied to, and assign them appropriately.

Poslovitch:
Oh my bad xD
Sorry, I'll send the fix asap

TheBusyBiscuit:
"NOT 1.9 or 1.10" 
*cough* *cough*
"NOT 1.9 AND NOT 1.10"...

Poslovitch:
Oh my bad. I'll definitely stop coding when I caught a cold. X)

TheBusyBiscuit:
Get well soon.

PR #549, opened 2017-12-25
Fixed Pedestal not dropping item when is broken or explodes

Redemption198:
Now the pedestal drops the item which is currently above of it when it's broken.
Edit: Changed obstructed message + minor changes.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
@@ -30,6 +30,7 @@ import org.bukkit.event.Listener;
 import org.bukkit.event.block.Action;
 import org.bukkit.event.block.BlockBreakEvent;
 import org.bukkit.event.block.BlockPlaceEvent;
+import org.bukkit.event.entity.EntityExplodeEvent;
 import org.bukkit.event.player.PlayerInteractEvent;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.meta.ItemMeta;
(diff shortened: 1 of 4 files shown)


Poslovitch:
Don't agree with splitting the obstructed message. This is a useless information to know what is above the pedestal, I think.

Redemption198:
@Poslovitch what about this: "&4Pedestal is obstructed! &cRemove anything above the pedestal!"

Poslovitch:
I agree :)

Redemption198:
@Poslovitch I found another "bug" when a TNT explodes the item which is above the pedestal drops on the floor.



I can try to fix it, but I don't know if the item should be dropped or deleted.

Poslovitch:
add a config option for it

TheBusyBiscuit:
Don't register additional events for Slimefun Blocks, use the BlockBreakHandler instead when calling .register()

It also takes care of explosions.
I also wouldn't add an option for it, just drop it everytime.

src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
```diff
@@ -171,5 +171,32 @@ public void onBlockPlace(BlockPlaceEvent e) {
...
+			if(item.getID().equalsIgnoreCase("ANCIENT_PEDESTAL")) {
+				Item stack = findItem(b);
+				if(stack == null) return;
+				if(SlimefunStartup.getCfg().getBoolean("options.drop-item-when-ancient-pedestal-explodes"))
+					b.getWorld().dropItem(b.getLocation(), fixItemStack(stack.getItemStack(), stack.getCustomName()));
+				stack.remove();
+			}
+		}
+	}
+}
```
TheBusyBiscuit: for all of the above:

**https://github.com/TheBusyBiscuit/Slimefun4/blob/6f61176d0c9cb944976e95813160a8df0759effc/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/handlers/BlockBreakHandler.java**

TheBusyBiscuit: Nevermind, it's this class:

https://github.com/TheBusyBiscuit/Slimefun4/blob/6f61176d0c9cb944976e95813160a8df0759effc/src/me/mrCookieSlime/Slimefun/Objects/SlimefunBlockHandler.java

TheBusyBiscuit: Could also be this class:

https://github.com/TheBusyBiscuit/Slimefun4/blob/6f61176d0c9cb944976e95813160a8df0759effc/src/me/mrCookieSlime/Slimefun/Objects/SlimefunBlockHandler.java

Redemption198:
@TheBusyBiscuit I don't know if it is what you meant.

TheBusyBiscuit:
Yes, this is exactly where the fix needed to happen.
Thanks, one minor correction:

When comparing non-primitives, use .equals() over ==

Issue #512, opened 2017-11-29
Carbonado Oven Becomes Item Duplication Machine Instead of Oven...


mrcoffee1026:
## Description
Oven from Enhanced+ run same speed as Vanilla/unmodified oven, although efficiency is increased as levels increase.  At some random interval, oven will stop producing cooked output and instead begin dropping an (infinite?) amount of whatever was put in to be cooked along with what was put in to be used as fuel.  (Use of hoppers to duplicate this issue is optional but does seem to speed up the appearance of the effect).

Issue originally appeared in SF v4.1.10, the v4.1.11's release notes elluded to the possibility this issue had been dealt with, but it clearly has not been.

## Server Log
Nothing relevant, there are no errors, things are doing what they were programmed to do.

## Evidence / Proof
https://imgur.com/a/VuQ3h

## Environment
CraftBukkit version
git-Spigot-3d850ec-19ed80c (MC: 1.12.2) (Implementing version 1.12.2-R0.1-SNAPSHOT)
CS-CoreLib v1.5.16
Slimefun v4.1.11
ExoticGarden v1.6.3

Poslovitch:
Duplicate of https://github.com/TheBusyBiscuit/Slimefun4/issues/411.

I'm starting to think it may be caused by the 10x speed bonus.

Nevermind, disable them on your server to avoid duplication, we didn't find any fix for it for 2 months.

mrcoffee1026:
Caused by.. speed that doesn't get applied anymore?  Also all of the "faster" ovens are having the same issue, it would seem that it would be necessary to disable ALL slimefun ovens if that is the "solution" to this.

Poslovitch:
This is currently the only solution we can provide to you. As far as we know, this is a Minecraft or Spigot bug. And up to now we havent found a better workaround than disabling the Enhanced furnaces.

mrcoffee1026:
A Spigot/Minecraft bug... that only effects slimefun v4.1.10 and up but didn't effect 4.1.9?  That seems rather unlikely... I think the changes made in 4.1.10 should be more closely examined.

mrcoffee1026:
Disabling them isn't really a solution either since already-made ovens... still have the issue and can't be 'unmade'.

Poslovitch:
Are you sure it comes from 4.1.10 exactly? Not due to a MC update? I might investigate it more closely

mrcoffee1026:
[...]

Issue #453, opened 2017-10-12
SlimeFun furnaces not showing speed/effeciency and taking lava bucket on first use


NerdyBirder:
## Description
We lined up all the Slimefun Furnaces (vanilla to carbonado-edged furnace) to test out what is working and what is not. All the furnaces cooked the sand at the same speed as a vanilla furnace and all but two of the slimefun furnaces used the fuel at the same rate. Enhanced Furnaces X and XI ran out of fuel 1/2 way through cooking the sand when all the others lasted much longer than the time one stack of sand took to cook.
Also, I have been having a problem with higher level furnaces taking the first bucket of lava and not lighting up the furnace at all, requiring a second bucket of lava to start the furnace. 

## Server Log
there are no odd server logs or errors for this issue.

## Evidence / Proof
https://imgur.com/a/875rr    (these images are in order showing how even the vanilla furnace was        still running but the two Slimefun Furnaces had stopped early)
https://youtu.be/nXcm7tU4pZ8

## Environment
CrafkBukkit v1_12_R1
CS-CoreLib v1.5.16
Slimefun v4.1.10
Installed Addons (2)
HotbarPets v1.1
ExoticGarden v1.6.3

TheBusyBiscuit:
The Video is unavailable.

NerdyBirder:
The video should be available now. I apologize for the video being unavailable.

andris155:
SlimeFun furnaces not work on 1.12.1, and 1.12.2

akashaggarwal7:
I wanted to work on this bug and some other bugs as well, but the items are not showing in the menu in the latest commit on 1.12.2(Is it only me?). If someone could link me to the commit where they were removed that'd be very helpful.

NerdyBirder:
The higher level furnaces are not in my menu as well. I only see Electric Furnaces I, II, and III.

akashaggarwal7:
@NerdyBirder I neither have any of the reinforced furnaces, nor other inventory based stuff like auto-enchanter etc. Auto-enchanter is bugged as well. If only I could figure out how to enable them for testing.. I had another friend looking into it but he couldn't find it either.

NerdyBirder:
Could you use an old version of Slimefun? or use the /sf give command?

akashaggarwal7:
@NerdyBirder What for?

NerdyBirder:
[...]

Issue #566, opened 2018-01-04
Items are missing!


LecraM2003:
## Description
Describe your Issue AS DETAILED AS POSSIBLE.
In the cathegory "Energy and Electricity" many Items are missing! (see Screenshot)

## Server Log
Upload your FULL SERVER LOG (/logs/latest.log) to http://pastebin.com/ and paste your link in here.
Without this File, we may not be able to help you at all.
https://pastebin.com/VggvrbKR
## Evidence / Proof
Attach as many Screenshots or Videos if possible.
Anything that helps us understand the Issue better, is gonna get your Issue fixed even faster.
![2018-01-04_22 59 19](https://user-images.githubusercontent.com/22198605/34586167-8b9b59a8-f1ab-11e7-8411-b18b6b5d77c8.png)

## Environment
We need the EXACT VERSIONS of the Software you are using.
This includes your Minecraft Version, your CS-CoreLib version and your Slimefun Version.
If you cannot figure out what versions you are running, then please run /sf versions and
show us the output of that command.
CS-CoreLib: 1.5.16
Slimefun: 4.1.12
Minecraft: Spigot 1.12.2
Java: Java 8

Poslovitch:
Same problem on my dev server

They are loaded, but not displayed.

Bredosen:
Same here

Poslovitch:
@Bredosen Could you please also provide your logs and your environment, so we can do some comparison?

RyseEmpire:
I am also having this issue. If I drop it down a version and turn off auto-update, would it fix it?

Poslovitch:
@RyseEmpire Probably not.
But please, as I asked above, provide logs and your environment, so we can compare different setups.

RyseEmpire:
Of course, @Poslovitch 

**LOGS**
https://pastebin.com/jdtY9ZJJ
**ENVIRONMENT**
CS-Corelib 1.5.16
Slimefun 4.1.12
Spigot 1.12.2
Java 8

phuchptty:
i have the same problem too 
CS-Corelib 1.5.16
Slimefun 4.112
Spigot 1.10.2
Java 8

TheBusyBiscuit:
The missing items is connected to the empty config as there are booleans for enabling ticker items, etc...
#564

TheBusyBiscuit:
Apparently, noone, not even I noticed this:
But Maven apparently doesnt put the config.yml into the jar :?
[...]

PR #567, opened 2018-01-05
Fixed speed modifier not getting applied to Enhanced Furnaces

John000708:
Fixed an issue mentioned in #512  #453 #411

Happy new year.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/EnhancedFurnace.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/EnhancedFurnace.java
@@ -22,20 +22,14 @@ public class EnhancedFurnace extends SlimefunItem {
 		this.fortune = fortune - 1;
 		
 		addItemHandler(new BlockTicker() {
-			
 			@Override
 			public void tick(Block b, SlimefunItem item, Config data) {
-				try {
-					if (!(b.getState() instanceof Furnace)) {
-						return;
-					}
+				if (b.getState() instanceof Furnace)
 					if (((Furnace) b.getState()).getCookTime() > 0) {
 						Furnace furnace = (Furnace) b.getState();
 						furnace.setCookTime((short) (furnace.getCookTime() + getSpeed() * 10));
 						furnace.update(true, false);
 					}
-				} catch(NullPointerException x) {
-				}
 			}
 
 			@Override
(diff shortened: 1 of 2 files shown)


Poslovitch:
Happy New Year first.

But it looks like you didn't change anything : you just store the state in a variable instead of casting it each time.
Did you run some tests?

John000708:
@Poslovitch I did and I was extremely amazed that it worked. You can go ahead and test it as well.

Poslovitch:
Wow. Sounds kinda magical. Is the Christmas magic still in the air today? haha
Let me run some tests, and send you the results ;)

Poslovitch:
Do you think it also fixes the duplication bug?

John000708:
I don't think so. Have you been able to reproduce the duplication issue? I haven't messed with Slimefun for a very long time so I'm a bit rusty.

Poslovitch:
No, I've never been able to reproduce the duplication issue.

Tried to test your fix, however due to #566, I can't give myself an Enhanced Furnace. I'll have to work on that thing too. So much bugs have popped from nowhere these times...

Poslovitch:
By the way, here is a tiny code-optimization for the whole #tick() method : 

```java
@Override
public void tick(Block b, SlimefunItem item, Config data) {
	if (b.getState() instanceof Furnace) {
		Furnace furnace = (Furnace) b.getState();
		if (furnace.getCookTime() > 0) {
			furnace.setCookTime((short) (furnace.getCookTime() + getSpeed() * 10));
			furnace.update(true, false);
		}
	}
}
```

John000708:
> No, I've never been able to reproduce the duplication issue.
> 
> Tried to test your fix, however due to #566, I can't give myself an Enhanced Furnace. I'll have to work on that thing too. So much bugs have popped from nowhere these times...

Hm.... That's weird. The items appear just fine for me. 

> By the way, here is a tiny code-optimization for the whole #tick() method :
> 
> @Override
> public void tick(Block b, SlimefunItem item, Config data) {
> 	if (b.getState() instanceof Furnace) {
> 		Furnace furnace = (Furnace) b.getState();
> 		if (furnace.getCookTime() > 0) {
> 			furnace.setCookTime((short) (furnace.getCookTime() + getSpeed() * 10));
> 			furnace.update(true, false);
> 		}
> 	}
> }

I purposely put the variable declaration inside the if-statement. No need to declare it every tick, even when the furnace isn't running.

Poslovitch:
Then this?

```java
@Override
public void tick(Block b, SlimefunItem item, Config data) {
	if (b.getState() instanceof Furnace)
		if (((Furnace) b.getState()).getCookTime() > 0) {
			Furnace furnace = (Furnace) b.getState();
			furnace.setCookTime((short) (furnace.getCookTime() + getSpeed() * 10));
			furnace.update(true, false);
		}
	}
}
```

So basically the same code as yours, without the NPE catch.

John000708:
Yeah I assume so. I don't see anything that could cause an NPE there.

Poslovitch:
Okey, then can you push a commit with this change

John000708:
Could somebody else also test this fix, just to be sure I'm not insane?

PR #570, opened 2018-01-06
Fixed Android Miner not breaking Slimefun blocks correctly

### LINKED ISSUE #553: Android mining interface glitch
## Description
When a mining android breaks an android interface, instead of breaking the block and storing it as an interface, it reset's it back to a standard dispenser.

For example, place an android interface below a miner and use the android miner to break the interface, it will drop a dispenser rather than an interface.

## Server Log
N/A

## Environment
Slimefun v4.1.11
Minecraft V1.12.1 (We are running ViaVersion on a bungeecord network)
CS-CoreLib V 1.5.15

Redemption198:
Fixed #553

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
+++ b/src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
@@ -732,11 +732,9 @@ public abstract class ProgrammableAndroid extends SlimefunItem {
 		Collection<ItemStack> drops = block.getDrops();
 		if (!blockblacklist.contains(block.getType()) && !drops.isEmpty() && CSCoreLib.getLib().getProtectionManager().canBuild(UUID.fromString(BlockStorage.getBlockInfo(b, "owner")), block)) {
 			try {
-				System.out.println("DEBUG: " + item);
 				if(item != null) {
 					if(fits(b, BlockStorage.retrieve(block))) {
 						pushItems(b, BlockStorage.retrieve(block));
-						System.out.println("DEBUG: IF I'm HERE");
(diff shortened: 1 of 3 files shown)


src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
```diff
@@ -700,36 +706,64 @@ private void move(Block b, BlockFace face, Block block) throws Exception {
...
 		if (!blockblacklist.contains(block.getType()) && !drops.isEmpty() && CSCoreLib.getLib().getProtectionManager().canBuild(UUID.fromString(BlockStorage.getBlockInfo(b, "owner")), block)) {
-			ItemStack[] items = drops.toArray(new ItemStack[drops.size()]);
-			if (fits(b, items)) {
-				pushItems(b, items);
-				block.getWorld().playEffect(block.getLocation(), Effect.STEP_SOUND, block.getType());
-				block.setType(Material.AIR);
+			if(item != null) {
+				if(fits(b, item.getItem())) {
+					pushItems(b, BlockStorage.retrieve(block));
+					if(SlimefunItem.blockhandler.containsKey(item.getID())) SlimefunItem.blockhandler.get(item.getID()).onBreak(null, block, item, UnregisterReason.ANDROID_DIG);
```
TheBusyBiscuit: Block Handlers return booleans.
Use them, because it might be that a Slimefun block doesn't want to be broken by an Android.

src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
```diff
@@ -700,36 +706,72 @@ private void move(Block b, BlockFace face, Block block) throws Exception {
 		}
 	}
 
-	private void mine(Block b, Block block) {
+	private void mine(Block b, Block block, SlimefunItem item) {
```
TheBusyBiscuit: You could also just declare

SlimefunItem item = BlockStorage.check(block);

inside both methods.

PR #573, opened 2018-01-07
Fixed Cooler rejecting Juices

AtomicScience:
Way to reproduce bug:
1. Open the Cooler and try to move any Juice inside of it (Do not press Shift!)
2. When you'll click empty slot with Juice at cursor, Cooler will cancel click

This bug exist because clicking at empty slot isn't handled properly - when we click at empty slot, .getCurrentItem() returns us not _null_ but an ItemStack with type "Air". Probably, this behavour was added in new versions of Bukkit API

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
@@ -15,6 +15,7 @@ import me.mrCookieSlime.Slimefun.api.Slimefun;
 
 import org.bukkit.ChatColor;
 import org.bukkit.Sound;
+import org.bukkit.Material;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
```diff
@@ -65,7 +66,7 @@ public void onClick(InventoryClickEvent e) {
 			else {
 				SlimefunItem sfItem = SlimefunItem.getByItem(e.getCurrentItem());
 				if (SlimefunManager.isItemSimiliar(item, SlimefunItem.getItem("COOLER"), false)) {
-					if (e.getCurrentItem() == null);
+					if (e.getCurrentItem().getType() == Material.AIR);
 					else if (sfItem == null) e.setCancelled(true);
```
TheBusyBiscuit: sfItem == null 
can be removed, instanceof takes care of that.
I probably didn't know about that, back when I wrote this code.

src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
```diff
@@ -65,7 +66,7 @@ public void onClick(InventoryClickEvent e) {
 			else {
 				SlimefunItem sfItem = SlimefunItem.getByItem(e.getCurrentItem());
 				if (SlimefunManager.isItemSimiliar(item, SlimefunItem.getItem("COOLER"), false)) {
-					if (e.getCurrentItem() == null);
+					if (e.getCurrentItem().getType() == Material.AIR);
```
TheBusyBiscuit: I would probably do (just to be sure)

== null || getType().equals(Material.AIR)

Also, it is good practice to use .equals instead of == when comparing non-primitives, especially enums.

PR #600, opened 2018-02-18
use diamond operator

MiniDigger:
you said we should pr stuff, so here I start

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/AncientAltar/AltarRecipe.java
+++ b/src/me/mrCookieSlime/Slimefun/AncientAltar/AltarRecipe.java
@@ -13,7 +13,7 @@ public class AltarRecipe {
 	
 	public AltarRecipe(List<ItemStack> input, ItemStack output) {
 		this.catalyst = input.get(4);
-		this.input = new ArrayList<ItemStack>();
+		this.input = new ArrayList<>();
 		for (int i = 0; i < input.size(); i++) {
 			if (i != 4) this.input.add(input.get(i));
 		}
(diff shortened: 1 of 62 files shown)


TheBusyBiscuit:
The diamond operator does not add any benefit to the source code, it is purely aesthetic.

stefvanschie:
That's not true. The diamond operators indicate the usage of generics which should be used in order to avoid having classes which can take in any kind of object. Here is a link to the Java documentation about generics: https://docs.oracle.com/javase/tutorial/java/generics/index.html.

MiniDigger:
I give up

TheBusyBiscuit:
@stefvanschie 
Generic Types need to be inferred, but it has no benefit to fill them in automatically or manually.

stefvanschie:
It is unnecessary to explicitly specify the generic type again when you just specify it. In the edited files it is clearly visible what each type is for every variable/field. By not specifying the type your code becomes more readable, especially with nested generics. If you specifically want to keep them for whatever reason, you should probably make a style guide file in the repository in which you mention this.

Also your argument as to why you should use them is a little strange. There are a lot of things you don't need to specify which will be added later, which are specifically there to make your life easier and make your code more readable. Not making use of those features is basically like writing your code in Assembly because in the end your code will end up as machine code anyway.

TheBusyBiscuit:
@stefvanschie I never said you should do it my way, nor did I reject this PR.

And of course you can leave the explicit type out, this code as well as I just come from a time where this was still necessary and I continued doing it for consistency.

I was merely pointing out that it has no benefit to remove them. (Disregarding that it may seem saner to read for some people)

I could create guidelines for my style of code but I dont really want to restrict people.

stefvanschie:
I'd highly recommend creating a style guide even if that means forcing people to do something in a particular way. If you don't create such a style guide I wouldn't be surprised if more issues/pull requests will be created which are similar to this one (e.g. mainly streamlining code without adding anything to it), which would be a waste of everyone's time and effort. By forcing people a certain style you avoid such issues/pull requests and any pull requests that do get accepted will look similar to the code already in place, making it more consistent across the project.

PR #601, opened 2018-02-18
enforce encapsulation

MiniDigger:


TheBusyBiscuit:
While, I have nothing against setting default fields to private.
Making public methods private is a mistake, as these methods are essential to the expandability of Slimefun via addons.

You can still resubmit your PR, if you fix that.

PR #603, opened 2018-02-18
Fix everything

seasonguy:
This will fix everything wrong with this code.


JanTuck:
Why did you close this? I found it very amusing, and it WOULD actually have solved all your problems though.

NickAcPT:
+1

Redemption198:
@seasonguy I can't understand why do you have to disturb people who mind their own business.

PR #619, opened 2018-03-02
Fix #617: Electric Dust Washer - Dust Choice.

### LINKED ISSUE #617: Electric Dust Washer permits chosing the output Dusts
Creating a Dust Washer that is not input into a cargo node (Although it is possible to do this too), you can leave at least one dust in each output slot to make it produce only that one or those two dusts.

I will be happy to provide a PR to fix this issue in a short while.

meiamsome:
See the issue for description, similar to the fix in #503 except it has to work on output slots only.


meiamsome:
@HawkxX1337 That filtering behaviour was a bug not a feature and this fixings it by severely limiting the machine's output. You can use a cargo network with it to take the dusts out to create them in their intended distributions only.

meiamsome:
@HawkxX1337 You have to run an older version of the plugin or recompile it with the changes listed on the following link reverted:
https://github.com/TheBusyBiscuit/Slimefun4/pull/619/files

I doubt this will become a configuration option as it's a pretty game breaking bug for the plugin's intended design. Please see #503 for the manual equivalent which was fixed near the end of last year for the same reason

KyleEstein:
@HawkxX1337 the configuration options legacy-ore-washer and legacy-dust-washer have been added in v4.1.15. If set to true the respective device will revert to the old behavior.

PR #620, opened 2018-03-02
Simplify error file name generation

meiamsome:
Cleans up the source code a bit (although not much)


TheBusyBiscuit:
Did I really do it with a ton of ifs...
Wow... Just wow for me being such an idiot back then.

PR #629, opened 2018-03-09
Fix AReactor Issue

meiamsome:
The issue is as follows:
When certain AReactors (Nuclear) produces energy it stores it in its own chargeable state.
The current charge is returned from this function. ([Line 303](https://github.com/TheBusyBiscuit/Slimefun4/blob/master/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java#L303))

However, in the energy net code, it then takes this value and adds on any storage that the block has:
https://github.com/TheBusyBiscuit/Slimefun4/blob/master/src/me/mrCookieSlime/Slimefun/api/energy/EnergyNet.java#L116-L119

Which means the the Nuclear Reactors currently produce twice their described power and when the energy network is full, they produce the equivalent of their storage every tick (So 16kJ) which makes them incredibly OP and broken. This just returns zero here so that just the storage is used, which means it does not break any other generators.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
@@ -299,10 +299,11 @@ public abstract class AReactor extends SlimefunItem {
 					extraTick(l);
 					int timeleft = progress.get(l);
 					if (timeleft > 0) {
-						if (ChargableBlock.getMaxCharge(l) - ChargableBlock.getCharge(l) >= getEnergyProduction()) {
+						boolean should_charge = ChargableBlock.getMaxCharge(l) - ChargableBlock.getCharge(l) >= getEnergyProduction();
+						if (should_charge) {
 							ChargableBlock.addCharge(l, getEnergyProduction());
 						}
-						if (ChargableBlock.getMaxCharge(l) - ChargableBlock.getCharge(l) >= getEnergyProduction() || !BlockStorage.getBlockInfo(l, "reactor-mode").equals("generator")) {
+						if (should_charge || !BlockStorage.getBlockInfo(l, "reactor-mode").equals("generator")) {
 							progress.put(l, timeleft - 1);
 
 							Bukkit.getScheduler().scheduleSyncDelayedTask(SlimefunStartup.instance, new Runnable() {
--- a/src/me/mrCookieSlime/Slimefun/api/energy/EnergyNet.java
+++ b/src/me/mrCookieSlime/Slimefun/api/energy/EnergyNet.java
@@ -114,9 +114,6 @@ public class EnergyNet {
 				}
 				else {
 					supply = supply + energy;
-					if (ChargableBlock.isChargable(source)) {
-						supply = DoubleHandler.fixDouble(supply + ChargableBlock.getCharge(source));
-					}
 				}
 				TickerTask.block_timings.put(source, System.currentTimeMillis() - timestamp);
 			}


TheBusyBiscuit:
Took me some time to understand how this happens.
Note that there is also the AGenerator.class which does exactly the same.

But I think this issue should rather be approached on this end:
https://github.com/TheBusyBiscuit/Slimefun4/blob/master/src/me/mrCookieSlime/Slimefun/api/energy/EnergyNet.java#L102

The method .generateEnergy() needs to be called, either not store this as a variable.
However with this variable, you get the block's charge for granted.

Or we could address it here:
https://github.com/TheBusyBiscuit/Slimefun4/blob/master/src/me/mrCookieSlime/Slimefun/api/energy/EnergyNet.java#L117-L119

By removing the ChargableBlock calls, as .generateEnergy does that for us.

meiamsome:
Yep, I think that makes more sense, I hadn't fully checked that all the other machines functioned like that and was thinking about it from a view of all blocks in an energy network being capacitors but this isn't true. I've implemented the method you suggested now and also fixed a bug I just noticed which would make AReactors not use up a timing step when they could only fit a single charge unit into their capacitance. (Which also means they wouldn't use coolant.) Hopefully that is ok to include with this other fix because they are thematically similar!

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
```diff
@@ -299,10 +299,11 @@ public double generateEnergy(final Location l, SlimefunItem sf, Config data) {
 					extraTick(l);
 					int timeleft = progress.get(l);
 					if (timeleft > 0) {
-						if (ChargableBlock.getMaxCharge(l) - ChargableBlock.getCharge(l) >= getEnergyProduction()) {
+						boolean should_charge = ChargableBlock.getMaxCharge(l) - ChargableBlock.getCharge(l) >= getEnergyProduction();
+						if (should_charge) {
 							ChargableBlock.addCharge(l, getEnergyProduction());
 						}
-						if (ChargableBlock.getMaxCharge(l) - ChargableBlock.getCharge(l) >= getEnergyProduction() || !BlockStorage.getBlockInfo(l, "reactor-mode").equals("generator")) {
+						if (should_charge || !BlockStorage.getBlockInfo(l, "reactor-mode").equals("generator")) {
```
TheBusyBiscuit: Apparently this caused some problems.
This shouldn't have changed. Because if should_charge is true, then charge is added, making the charge maybe go completely full capacity.

And this statement should check for that, but since it uses the same boolean as before, this is not taken into account and Reactors wrap around to start at 0 when they are full or burn items that are not needed.

Just letting you know that.

meiamsome: This change was to fix Reactors not burning fuel when they should have - if the first if statement was executed then it meant that one unit of energy had been created. By changing the second if to be based on the same condition it makes sure that if there is a charge then the fuel is taken away. For example a reactor that has a storage capacity of 1000J, and a charge rate of 100J/step the following could happen if they are not the same condition:

Step 1
- charge is at 1000J
- nothing happens
Step 2
- Something in the energy network uses 50J of power
- The remaining energy is 950J -> neither if statement executes.
Step 3
- Something in the energy network uses 50J of power
- The remaining energy is 900J -> 
- The first if statement executes, the remainig energy is now 950J.
- The second if does not execute so no time is taken from the fuel
Step 4
- Something in the energy network uses 50J of power
- The remaining energy is 900J -> 
- The first if statement executes, the remainig energy is now 950J.
- The second if does not execute so no time is taken from the fuel

etc., where step 4 is the same as step 3 which means that it can loop infinitely using an infinite amount of energy without using any time remaining on a fuel. You can see this in that in an energy system that is full with a reactor where it will keep the fuel time remaining at the highest value possible.

TheBusyBiscuit: I see, you are right.

Still, how is #666 caused then?
Also ignore my rushed approach then.

meiamsome: I think this is more complicated, relating to the algorithm used for energy generation. I'll post in issue #666, can you re-open it?

TheBusyBiscuit: Did so.

PR #630, opened 2018-03-09
Create a Network class to manage networks in a stateful way

meiamsome:
The diff is huge because a lot of whitespace got deleted from the end of lines, sorry.

I'll consider adding a linter to the repo to enforce style rules to stop the inconsistent whitespace behaviours. 

I am hoping to do some Cargo Network efficiency updates very soon, but thought I'd post this so you can review.

Shouldn't change functionality, but the cargo net visualizer currently does not show in the same way it did before, but that would be fixable with an upgrade to the base Network class in the near future.


TheBusyBiscuit:
Just a heads up that I am not ignoring these Pull Requests.
It's just convenient that I am ill during very stressful and important weeks.

I will review these large sets of changes when I got time and feel better.

Issue #487, opened 2017-11-06
Error executing task 106


viocu:
## Description
Console throwing this error:

`11/06/2017 01:25:23

01:25:23

[Slimefun] Plugin Slimefun v4.1.11 generated an exception while executing task 106

Console
 
11/06/2017 01:25:23

01:25:23

java.lang.IllegalStateException: Asynchronous entity track!
at org.spigotmc.AsyncCatcher.catchOp(AsyncCatcher.java:14) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at net.minecraft.server.v1_12_R1.EntityTracker.addEntity(EntityTracker.java:108) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at net.minecraft.server.v1_12_R1.EntityTracker.track(EntityTracker.java:82) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at net.minecraft.server.v1_12_R1.WorldManager.a(WorldManager.java:21) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at net.minecraft.server.v1_12_R1.World.b(World.java:1095) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at net.minecraft.server.v1_12_R1.WorldServer.b(WorldServer.java:1160) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at net.minecraft.server.v1_12_R1.WorldServer.a(WorldServer.java:1125) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at net.minecraft.server.v1_12_R1.Chunk.addEntities(Chunk.java:753) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at org.bukkit.craftbukkit.v1_12_R1.chunkio.ChunkIOProvider.callStage2(ChunkIOProvider.java:45) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
at org.bukkit.craftbukkit.v1_12_R1.chunkio.ChunkIOProvider.callStage2(ChunkIOProvider.java:1) ~[spigot-1.12.2.jar:git-Spigot-5695bca-53fccdf]
[...]

PR #632, opened 2018-03-10
Check if the chunk is loaded instead of relying on exceptions

### LINKED ISSUE #509: SlimeFun spitting errors in Console
## Description
Describe your Issue AS DETAILED AS POSSIBLE.
Slimefun Spitting Errors in Console
## Server Log
https://pastebin.com/NG0GMnXx
## Evidence / Proof
Attach as many Screenshots or Videos if possible.
Anything that helps us understand the Issue better, is gonna get your Issue fixed even faster.
-
## Environment
We need the EXACT VERSIONS of the Software you are using.
This includes your Minecraft Version, your CS-CoreLib version and your Slimefun Version.
If you cannot figure out what versions you are running, then please run /sf versions and
show us the output of that command.
PaperSpigot 1.12.2 v1_12_R1
CS-CoreLib v1.5.16
Slimefun v4.1.10
Barrels v1.11
SlimeXpansion v1.1.9

meiamsome:
(Maybe) Fixes #509

It would be ideal for @itslukemango or @andris155 or anyone running Paper to check if this fixes it. It does not throw any exceptions on my Spigot testing, so it shouldn't cause issues on Paper.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -3196,11 +3196,7 @@ public class SlimefunSetup {
 
 			@Override
 			public double generateEnergy(Location l, SlimefunItem item, Config data) {
-				try {
-					if (l.getBlock().getLightFromSky() != 15) return 0D;
-				} catch(IllegalStateException x) {
-					return 0D;
-				}
+				if (!l.getWorld().isChunkLoaded(l.getBlockX() >> 4, l.getBlockZ() >> 4) || l.getBlock().getLightFromSky() != 15) return 0D;
 				if (l.getWorld().getTime() < 12300 || l.getWorld().getTime() > 23850) return 2D;
 				return 0D;
 			}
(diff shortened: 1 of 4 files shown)


lukemango:
We changed to Spigot rather than Paper, so I can't test it right now.

meiamsome:
@lukemango Thanks for getting back to me regarding this.

In other news, this fix doesn't work (`.getChunk()` loads the chunk these days) - I'll push an update in a bit.

meiamsome:
The second commit includes an attempted fix for #487 - by using Location instead of Block the ticker task won't cause an Async Chunk Load. this adds some Deprecations to the API, but I haven't resolved these in the other classes yet.

TheBusyBiscuit:
Migrating everything to Locations is great. But I wouldnt advise renaming the method to getLocationInfo(), addLocationInfo etc...
just keep the name, and deprecate all methods that use a Block argument rather than a Location one.

meiamsome:
I originally kept the names of them with Block in, and then I got to this method:
```java
public static Set<Block> getTickingBlocks(Chunk chunk)
```
Which conflicts with the form of 
```java
public static Set<Location> getTickingBlocks(Chunk chunk)
```
And so I had to change the name of that one to Location and therefore I decided I should rename the others for consistency.

TheBusyBiscuit:
@meiamsome I see, okay.

Issue #290, opened 2017-05-19
Changing a Broker Spawner in a Pig Spawner


comonier:
_## Description_
If you Take any Mob-Spawner, and broke it with the Pickaxe of Containment, it will be renamed to: Broken Spawner Type "Mob". So now, put the Mob-Spawner on a anvil and rename it. Now put the Broken Spawner on the ground. Done you have a Pig Spawner

_## Server Log_
cant open pastebin.com > ERR_CONNECTION_TIMED_OUT using codeshare.io
https://codeshare.io/5ePj04

_## Evidence / Proof_
https://www.youtube.com/watch?v=LAPTkSEMIXg

_## Environment_
Server Spigot 1.11.2 Build: 3fb9445eca2
Client Minecraft Original 1.11.2
Plugin: SlimeFun 4.1.8
Plugin: CS-CoreLib v1.5.14

andris155:
Found a solution:
https://www.spigotmc.org/threads/disable-item-renaming-in-anvil.131780/

comonier:
solved ty

PR #645, opened 2018-03-31
Removes ability to rename a Broken Spawner in the Anvil, resulting in…

mrcoffee1026:
… a fixed "pig spawner" regardless of original spawner type, players Slimefun knowledge, and bypasses any need for using ingredients or the Ancient Altar to create a Reinforced Spawner.  Addresses #425, #290 (these tickets should not have been closed, this problem persists).

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
@@ -378,12 +378,16 @@ public class ItemListener implements Listener {
 	@EventHandler
     public void onAnvil(InventoryClickEvent e) {
         if (e.getRawSlot() == 2 && e.getWhoClicked() instanceof Player && e.getInventory().getType() == InventoryType.ANVIL) {
-		if (SlimefunManager.isItemSimiliar(e.getInventory().getContents()[0], SlimefunItems.ELYTRA, true)) return;
-
-        	if (SlimefunItem.getByItem(e.getInventory().getContents()[0]) != null && !SlimefunItem.isDisabled(e.getInventory().getContents()[0])) {
-            	e.setCancelled(true);
-                Messages.local.sendTranslation((Player) e.getWhoClicked(), "anvil.not-working", true);
-            }
+			if (SlimefunManager.isItemSimiliar(e.getInventory().getContents()[0], SlimefunItems.ELYTRA, true)) return;
+			
+			/* SlimefunItems.BROKEN_SPAWNER 
+			 * does SlimefunItem.isDisabled not apply to this item for some reason?
+			 * Let us hope that's the only item with this issue.. Adding a check for the name to this IF statement to workaround. */
+	
+	    	if (SlimefunItem.getByItem(e.getInventory().getContents()[0]) != null && (!SlimefunItem.isDisabled(e.getInventory().getContents()[0])) || (SlimefunManager.isItemSimiliar(e.getInventory().getContents()[0], SlimefunItems.BROKEN_SPAWNER, false)) ) {
+	        	e.setCancelled(true);
+	            Messages.local.sendTranslation((Player) e.getWhoClicked(), "anvil.not-working", true);
+	        }
         }
     }
 }


src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -231,7 +238,7 @@ public boolean onClick(Player p, int arg1, ItemStack arg2, ClickAction arg3) {
 				p.closeInventory();
 				p.sendMessage("");
 				p.sendMessage(ChatColor.translateAlternateColorCodes('&', "&7&ohttps://github.com/TheBusyBiscuit/Slimefun4/issues"));
-				p.sendMessage("");
+				p.sendMessage("");
```
sjx233: These whitespace characters should be removed

src/me/mrCookieSlime/Slimefun/api/Soul.java
```diff
@@ -21,9 +21,14 @@ public static void storeItem(UUID uuid, ItemStack drop) {
...
-				if(!p.getInventory().contains(item)) {
+				if (((!p.getInventory().getHelmet().isSimilar(item)) &&
+				(!p.getInventory().getChestplate().isSimilar(item)) &&
+				(!p.getInventory().getLeggings().isSimilar(item)) &&
+				(!p.getInventory().getBoots().isSimilar(item)) &&
+				(!p.getInventory().getItemInOffHand().isSimilar(item))) &&
+				(!p.getInventory().contains(item))) {
 					p.getInventory().addItem(item);
-				}
+				}
```
sjx233: As well as these

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -80,6 +83,10 @@ public static ItemStack getItem() {
 		return getItem(BookDesign.CHEST);
 	}
 
+	public void inventoryClickEvent(InventoryClickEvent e) {
```
TheBusyBiscuit: Please follow our naming conventions.

'onInventoryClick' for example.

src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
```diff
@@ -378,12 +378,16 @@ else if (e.getEntity() instanceof Wither) {
...
-		if (SlimefunManager.isItemSimiliar(e.getInventory().getContents()[0], SlimefunItems.ELYTRA, true)) return;
-
-        	if (SlimefunItem.getByItem(e.getInventory().getContents()[0]) != null && !SlimefunItem.isDisabled(e.getInventory().getContents()[0])) {
-            	e.setCancelled(true);
-                Messages.local.sendTranslation((Player) e.getWhoClicked(), "anvil.not-working", true);
-            }
+			if (SlimefunManager.isItemSimiliar(e.getInventory().getContents()[0], SlimefunItems.ELYTRA, true)) return;
+			
+			/* SlimefunItems.BROKEN_SPAWNER 
+			 * does SlimefunItem.isDisabled not apply to this item for some reason?
```
TheBusyBiscuit: Probably because getByItem has no special case for Broken Spawners and tries to compare the lore?

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
mrcoffee1026: I can't change the function name.  At that point this fix ceases to work.  It works because it's using the bukkit name for the function, I believe.  While I understand you have your own naming convention, trying to rewrite how bukkit deals with inventory results in exactly these sorts of glitches.  The aforementioned duplication issue exists because individual items were left out of whatever loop that assigned override actions for clicks on them... but even if those were put back in, the underlying problem that "the book chest display is being handled as a bukkit inventory" still exists so... blank slots in the book's "inventory" can still have items put into them... making the book gui an unintentional trash can, since items can't always be taken back once placed into those empty slots.    I would take a further step of removing all events that aren't hyperlink actions from all the menu items from the code to see if that further speeds up the loading of the guide, but I thought I'd just do a small thing for now.  Well, use this commit or don't, but I don't know how else to make it work.   But changing the name for me broke the fix so... I'm thinking that can't be done in this case.  Let me know if I'm wrong.

TheBusyBiscuit: The method name plays no role, you are extending. Not overwriting.
Also, the event should not work because you are missing an EventHandler annotation and a registration of this Listener.
Speaking of which, this being a static utility class, should not contain any Event.
Also cancelling it based on no statements, would result in no Inventory on the Server working at all.

Did you really test this?
Because normally your code should not even work, and if it worked, it would break everything.

Issue #435, opened 2017-09-12
Broken Spawner Exploit/Bug


mrcoffee1026:
## Description
1. Use Pickaxe of Containment to harvest spawner (sheep spawner in picture). 
2. Rename spawner to anything using vanilla anvil.
3. Place revised spawner anywhere.. is now a fixed and fully functional Pig Spawner.

## Server Log
Nothin relevent

## Evidence / Proof
[http://caffeinatedweb.com/mc/spawner1.jpg](url)
[http://caffeinatedweb.com/mc/spawner2.jpg](url)
[http://caffeinatedweb.com/mc/spawner3.jpg](url)
[http://caffeinatedweb.com/mc/spawner4.jpg](url)

## Environment
CraftBukkit v1_12_R1
CS-CoreLib v1.5.16
Slimefun v4.1.10
ExoticGarden v1.6.3

TheBusyBiscuit:
This has also been posted before...

Issue #282, opened 2017-05-08
Soulbound  Backpack Recipe Bug


kljigs500:
## Description
IDs are already generated once the small backpack is made. But to make a soulbound backpack, you must have 'no' ID from a glided backpack,, thus making it impossible to craft since upgrading the backpack until glided backpack maintains the ID given. There were already reports by other users of the same issue. Although I failed to make screenshots because I was too focused on making slime fun machines.

## Server Log
No log (I'm a player of the server)

## Evidence / Proof
Recipe for small backpack
![2017-05-08_16 50 27](https://cloud.githubusercontent.com/assets/26787840/25797411/b0032c80-340f-11e7-9e38-72b1c32ea440.png)

Small backpack made automatically makes the ID.
![2017-05-08_16 50 31](https://cloud.githubusercontent.com/assets/26787840/25797413/b005ab4a-340f-11e7-9831-ee2bb08d6412.png)

Attempting to make soulbound backback via magical workbench.
![2017-05-08_16 54 20](https://cloud.githubusercontent.com/assets/26787840/25797412/b0053e9e-340f-11e7-894f-0c9ec30c950b.png)

Attempt failed as shown in the chatbox on the lower left
![2017-05-08_16 54 29](https://cloud.githubusercontent.com/assets/26787840/25797416/b01dfeac-340f-11e7-91ca-e657cb0c8cd7.png)

Attempting again to make a soulbound backpack, but this time with no ID given.
![2017-05-08_16 55 49](https://cloud.githubusercontent.com/assets/26787840/25797414/b0106594-340f-11e7-9054-9c9d184e4fee.png)

Soulbound backpack was a success.
![2017-05-08_16 56 04](https://cloud.githubusercontent.com/assets/26787840/25797415/b013fc90-340f-11e7-86d9-ac0d8c3b5253.png)

Therefore there must be no id for the crafting to work.

## Environment
Minecraft Version: 1.11.2
Server version: CraftBukkit v1_11_R1m (provided by admin)
CS-CoreLib v1.5.14 (provided by admin)
Slimefun: v4.1.8 (provided by admin)

Addons:
Exotic Garden: v1.6.3 (provided by admin)

kljigs500:
it would be clearer if i recorded a video, but my upload is very slow and it takes a lot of time... I'll try to do better in the future. i hope the descriptions help at least so you can take a look. :) 

P.S. I was assisted by an admin who could /sf cheat

[...]

PR #649, opened 2018-04-09
Anvil / Broken Spawner patch, Further Fix for keep-inv flags, Soulbound Backpack Recipe fix

### LINKED ISSUE #633: Soulbound ARMOR/ELYTRA in KeepInventory region duplicate on player death
## Description
same as #579. the fix didn't help for items being worn by player.

mrcoffee1026:
Redoing the last pull request here, cleaned up a bit, thing with book removed since I couldn't make it work. Copied logic from Enhanced Crafting table to the Magic Workbench so the Soulbound Backpack recipe works (did not include the auto-id-ing stuff... still happens when opening one of these but they get crafted so very rarely... and I wasn't up to rewriting all of that for the Soulbound flavor)... in any case I don't really care if someone has the ingredients... if they want to craft a stack of them to sell in a chest shop, more power to them.

Closes #633 (and #529 again since it wasn't fully resolved), #282, #435 (this last shouldn't have been closed what was previously reported was not this problem, and this problem persists)

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -1157,10 +1157,19 @@ public class SlimefunSetup {
 							boolean craft = true;
 							for (int j = 0; j < inv.getContents().length; j++) {
 								if (!SlimefunManager.isItemSimiliar(inv.getContents()[j], inputs.get(i)[j], true)) {
-									craft = false;
-									break;
+									if (SlimefunItem.getByItem(inputs.get(i)[j]) instanceof SlimefunBackpack) {
+										if (!SlimefunManager.isItemSimiliar(inv.getContents()[j], inputs.get(i)[j], false)) {
+											craft = false;
+											break;
+										}
+									}
+									else {
+										craft = false;
+										break;
+									}
 								}
 							}
+
 
 							if (craft) {
 								final ItemStack adding = RecipeType.getRecipeOutputList(machine, inputs.get(i));
--- a/src/me/mrCookieSlime/Slimefun/api/Soul.java
+++ b/src/me/mrCookieSlime/Slimefun/api/Soul.java
@@ -21,7 +21,12 @@ public class Soul {
 	public static void retrieveItems(Player p) {
 		if (Variables.soulbound.containsKey(p.getUniqueId())) {
 			for (ItemStack item: Variables.soulbound.get(p.getUniqueId())) {
-				if(!p.getInventory().contains(item)) {
+				boolean armormatch = false;
+				for (ItemStack armoritem: p.getInventory().getArmorContents()) {
+					if (item.equals(armoritem)) armormatch = true;
+				}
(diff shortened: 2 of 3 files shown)


xLastt:
Works?

src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
```diff
@@ -380,7 +380,7 @@ public void onAnvil(InventoryClickEvent e) {
         if (e.getRawSlot() == 2 && e.getWhoClicked() instanceof Player && e.getInventory().getType() == InventoryType.ANVIL) {
 		if (SlimefunManager.isItemSimiliar(e.getInventory().getContents()[0], SlimefunItems.ELYTRA, true)) return;
 
-        	if (SlimefunItem.getByItem(e.getInventory().getContents()[0]) != null && !SlimefunItem.isDisabled(e.getInventory().getContents()[0])) {
+		if (SlimefunItem.getByItem(e.getInventory().getContents()[0]) != null && (!SlimefunItem.isDisabled(e.getInventory().getContents()[0])) || (SlimefunManager.isItemSimiliar(e.getInventory().getContents()[0], SlimefunItems.BROKEN_SPAWNER, false)) ) {
```
TheBusyBiscuit: From a designing point of view, an extra case should not be needed.
Why does isDisabled() need an extra case for Broken Spawners?
You should probably try to figure that out and fix it, otherwise we will be implementing work-arounds all over the place.

src/me/mrCookieSlime/Slimefun/api/Soul.java
```diff
@@ -21,7 +21,12 @@ public static void storeItem(UUID uuid, ItemStack drop) {
 	public static void retrieveItems(Player p) {
 		if (Variables.soulbound.containsKey(p.getUniqueId())) {
 			for (ItemStack item: Variables.soulbound.get(p.getUniqueId())) {
-				if(!p.getInventory().contains(item)) {
+				boolean armormatch = false;
+				for (ItemStack armoritem: p.getInventory().getArmorContents()) {
+					if (item.equals(armoritem)) armormatch = true;
```
TheBusyBiscuit: Use a break-statement, otherwise we will just iterate over unnecessary cases.

src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -1157,11 +1157,20 @@ public boolean onInteract(final Player p, MultiBlock mb, final Block b) {
...
+									else {
+										craft = false;
+										break;
+									}
 								}
 							}
 
+
 							if (craft) {
 								final ItemStack adding = RecipeType.getRecipeOutputList(machine, inputs.get(i));
```
TheBusyBiscuit: You should also implement the full fix for the Enhanced Crafting Table for the Magic Workbench, as it lacks the second and destroys the id and the backpack contents.

https://github.com/mrcoffee1026/Slimefun4/blob/7d001f2b69b5d21301717ea0375122cc97277521/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java#L232-L284

PR #657, opened 2018-04-18
Slimefun book GUI dupe fix

VoidAngel:
Adds a player to a List when they have the guide open and cancels _all_ inventory clicks until the menu is closed.


VoidAngel:
oof.
My java knowledge is weak x'D
Are you saying something similar should be done in CSCoreLib instead since the map already exists? or are you hinting at a whole new approach?

TheBusyBiscuit:
> Are you saying something similar should be done in CSCoreLib instead since the map already exists

Yes.

Issue #631, opened 2018-03-10
Operators with full perms cannot remove objects


Firebastard:
Operators with full permissions are unable to remove hologram projectors and Personal GPS Activaton Devices placed by other players on Spigot Version git-Spigot-c3093ef-3a482bd (MC: 1.12.2), CS-CoreLib v. 1.5.16 with Slimefun v. 4.1.14.  This may not seem like a large issue as they are deep in the techtree but any server operator can tell you that trolls can be quite tenacious.  I will disable them in the meantime as a workaround, but it is unfortunate that I have to do so.

Redemption198:
This happens because on SlimefunSetup.java, the #onBreak event only checks if the player who breaks the block is the owner of the block, it should be easy to fix.

mrcoffee1026:
As a workaround other than disabling them you could also use worldedit to set the blocks to air... though for the projector you also need to /minecraft:kill @e[type=armorstant,c=1]

Firebastard:
@mrcoffee1026  This is not the case. Slimefun will replace the block as soon as it is destroyed.  In the case of replacing the block with another block besides air, the original block is set back instead.  Killing the armor stand works for a time, but it is occasionally checked and respawned if not present.

mrcoffee1026:
if you replace the blocks with air, it stops being present.  right and left click each block with your //wand then //set air  .  After doing that you can do the kill command to get rid of the floating text left behind.

mrcoffee1026:
I found a way to resolve this by removing the additional protection that checks to verify the user doing the breaking is the "owner" of the block.  Initially this struck me as... oh my god what a massive security risk... and then realized that virtually none of the slimefun blocks checked for ownership prior to allowing a break.  Perfect example would be Elevator plates... which stores a user's entered settings but nothing in an inventory of any sort... can be broken by anyone if not protected by some other means.  In any case, this seems to be either the correct - or only - approach?  I'm not really certain.  I was trying to find an example of an item that allowed being broken by either the owner OR an OP user, but it doesn't appear any of them are set up that way, so I have no idea how to accomplish that.  Anyway, unless anyone screams immediately about this particular approach, I'll put that in a pull request.

PR #659, opened 2018-04-20
Anvil / Broken Spawner patch, Further Fix for keep-inv flags, Soulbound Backpack Recipe fix

### LINKED ISSUE #633: Soulbound ARMOR/ELYTRA in KeepInventory region duplicate on player death

mrcoffee1026:
I tried making changes to the old pull request but FAILED.  I couldn't seem to get my fork updated to the last changes without making a huge mess out of my fork to the point it was illegible.  Anyway, here are these files with the changes requested. 

I removed the changes from ItemListener.java.  The isDisabled function calls isItem (I believe also getByItem under other circumstances... where if lore does not match exactly, the item will not be classified as a SlimeFun item.  However, the item should still be classified as a SlimeFun item whether or not lore is identical.  Enchants, IDs and the like are virtually guaranteed to vary from item to item, that doesn't make them any less of a SlimeFun item.

As requested, brought in "the rest" of the fix for the Soulbound recipe fix into the Magic Workbench.  I really wanted this to make id-less backpacks in the end but... copying the old logic required... copying all of it.  I found it did not actually 'delete' the ID, simply recycled this ID for the new pack, changing the size in the text file to match the one just crafted.  It might be possible to have the thing look up the ID in the text file and simple delete the whole entry, but I'm not as familiar with text file manipulation with Java so decided to leave it as is.  I'm not entirely confident, but this particular approach might have the added benefit of keeping the contents of the backpack used in the recipe in the newly created backpack, but I haven't checked to be sure.

I couldn't alter the SoulBound code with "breaks" as asked as that would actually stop items in the inventory from being checked once a match had been found, and they'd still need to all be checked to keep things from duplicating... but I think I get your meaning, if so... my use of return here should alleviate any efficiency concerns.

Closes #633 (and #529 again since it wasn't fully resolved), #282, #435 (this last shouldn't have been closed what was previously reported was not this problem, and this problem persists).

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -334,7 +334,7 @@ public class SlimefunItem {
 			else if (sfi instanceof DamagableChargableItem && SlimefunManager.isItemSimiliar(item, sfi.getItem(), false)) return sfi;
 			else if (sfi instanceof ChargedItem && SlimefunManager.isItemSimiliar(item, sfi.getItem(), false)) return sfi;
 			else if (sfi instanceof SlimefunBackpack && SlimefunManager.isItemSimiliar(item, sfi.getItem(), false)) return sfi;
-			else if (SlimefunManager.isItemSimiliar(item, sfi.getItem(), true)) return sfi;
+			else if (SlimefunManager.isItemSimiliar(item, sfi.getItem(), false)) return sfi;
 		}
 		return null;
 	}
@@ -344,7 +344,7 @@ public class SlimefunItem {
 		if (this instanceof ChargableItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
 		else if (this instanceof DamagableChargableItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
 		else if (this instanceof ChargedItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
-		else return SlimefunManager.isItemSimiliar(item, this.item, true);
+		else return SlimefunManager.isItemSimiliar(item, this.item, false);
 	}
 	
 	public void load() {
--- a/src/me/mrCookieSlime/Slimefun/api/Soul.java
+++ b/src/me/mrCookieSlime/Slimefun/api/Soul.java
@@ -21,6 +21,12 @@ public class Soul {
 	public static void retrieveItems(Player p) {
 		if (Variables.soulbound.containsKey(p.getUniqueId())) {
 			for (ItemStack item: Variables.soulbound.get(p.getUniqueId())) {
+				if (item.equals(p.getInventory().getHelmet())) return;
+				if (item.equals(p.getInventory().getChestplate())) return;
+				if (item.equals(p.getInventory().getLeggings())) return;
+				if (item.equals(p.getInventory().getBoots())) return;
+				if (item.equals(p.getInventory().getItemInOffHand())) return;
+
 				if(!p.getInventory().contains(item)) {
 					p.getInventory().addItem(item);
 				}
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -1158,8 +1158,16 @@ public class SlimefunSetup {
 							boolean craft = true;
 							for (int j = 0; j < inv.getContents().length; j++) {
 								if (!SlimefunManager.isItemSimiliar(inv.getContents()[j], inputs.get(i)[j], true)) {
-									craft = false;
-									break;
+									if (SlimefunItem.getByItem(inputs.get(i)[j]) instanceof SlimefunBackpack) {
+										if (!SlimefunManager.isItemSimiliar(inv.getContents()[j], inputs.get(i)[j], false)) {
+											craft = false;
+											break;
+										}
+									}
+									else {
+										craft = false;
+										break;
+									}
 								}
 							}
 
@@ -1171,6 +1179,62 @@ public class SlimefunSetup {
 										inv2.setItem(j, inv.getContents()[j] != null ? (inv.getContents()[j].getAmount() > 1 ? new CustomItem(inv.getContents()[j], inv.getContents()[j].getAmount() - 1): null): null);
 									}
 									if (InvUtils.fits(inv2, adding)) {
+										SlimefunItem sfItem = SlimefunItem.getByItem(adding);
+
+										if (sfItem instanceof SlimefunBackpack) {
+											ItemStack backpack = null;
+
+											for (int j = 0; j < 9; j++) {
+												if (inv.getContents()[j] != null) {
+													if (inv.getContents()[j].getType() != Material.AIR) {
+														if (SlimefunItem.getByItem(inv.getContents()[j]) instanceof SlimefunBackpack) {
+															backpack = inv.getContents()[j];
+															break;
+														}
+													}
+												}
+											}
+											String id = "";
+											int size = ((SlimefunBackpack) sfItem).size;
+
+											if (backpack != null) {
+												for (String line: backpack.getItemMeta().getLore()) {
+													if (line.startsWith(ChatColor.translateAlternateColorCodes('&', "&7ID: ")) && line.contains("#")) {
+														id = line.replace(ChatColor.translateAlternateColorCodes('&', "&7ID: "), "");
+														Config cfg = new Config(new File("data-storage/Slimefun/Players/" + id.split("#")[0] + ".yml"));
+														cfg.setValue("backpacks." + id.split("#")[1] + ".size", size);
+														cfg.save();
+														break;
+													}
+												}
+											}
+
+											if (id.equals("")) {
+												for (int line = 0; line < adding.getItemMeta().getLore().size(); line++) {
+													if (adding.getItemMeta().getLore().get(line).equals(ChatColor.translateAlternateColorCodes('&', "&7ID: <ID>"))) {
+														ItemMeta im = adding.getItemMeta();
+														List<String> lore = im.getLore();
+														lore.set(line, lore.get(line).replace("<ID>", Backpacks.createBackpack(p, size)));
+														im.setLore(lore);
+														adding.setItemMeta(im);
+														break;
+													}
+												}
+											}
+											else {
+												for (int line = 0; line < adding.getItemMeta().getLore().size(); line++) {
+													if (adding.getItemMeta().getLore().get(line).equals(ChatColor.translateAlternateColorCodes('&', "&7ID: <ID>"))) {
+														ItemMeta im = adding.getItemMeta();
+														List<String> lore = im.getLore();
+														lore.set(line, lore.get(line).replace("<ID>", id));
+														im.setLore(lore);
+														adding.setItemMeta(im);
+														break;
+													}
+												}
+											}
+										}
+										
 										for (int j = 0; j < 9; j++) {
 											if (inv.getContents()[j] != null) {
 												if (inv.getContents()[j].getType() != Material.AIR) {


mrcoffee1026:
Amended commits dealing with #631 -- no longer checks ownership of block before allowing break (these are not inventory keeping items and other non-inventory keeping blocks do not check for this).

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -344,7 +344,7 @@ public boolean isItem(ItemStack item) {
 		if (this instanceof ChargableItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
 		else if (this instanceof DamagableChargableItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
 		else if (this instanceof ChargedItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
-		else return SlimefunManager.isItemSimiliar(item, this.item, true);
+		else return SlimefunManager.isItemSimiliar(item, this.item, false);
```
TheBusyBiscuit: Please leave these as "true", just implement a special case, like the above. Check for the ID or so, then you fixed the actual problem.

src/me/mrCookieSlime/Slimefun/api/Soul.java
```diff
@@ -21,6 +21,12 @@ public static void storeItem(UUID uuid, ItemStack drop) {
 	public static void retrieveItems(Player p) {
 		if (Variables.soulbound.containsKey(p.getUniqueId())) {
 			for (ItemStack item: Variables.soulbound.get(p.getUniqueId())) {
+				if (item.equals(p.getInventory().getHelmet())) return;
+				if (item.equals(p.getInventory().getChestplate())) return;
+				if (item.equals(p.getInventory().getLeggings())) return;
+				if (item.equals(p.getInventory().getBoots())) return;
+				if (item.equals(p.getInventory().getItemInOffHand())) return;
```
TheBusyBiscuit: return; terminates the entire method which basically cancels all items if just one is equal. Use continue; instead

mrcoffee1026:
Sorry about that, still a bit new at the use of git and whatnot.  Thanks!

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -334,6 +335,7 @@ public static SlimefunItem getByItem(ItemStack item) {
 			else if (sfi instanceof DamagableChargableItem && SlimefunManager.isItemSimiliar(item, sfi.getItem(), false)) return sfi;
 			else if (sfi instanceof ChargedItem && SlimefunManager.isItemSimiliar(item, sfi.getItem(), false)) return sfi;
 			else if (sfi instanceof SlimefunBackpack && SlimefunManager.isItemSimiliar(item, sfi.getItem(), false)) return sfi;
+			else if ((SlimefunManager.isItemSimiliar(item, SlimefunItems.BROKEN_SPAWNER, false)) || (SlimefunManager.isItemSimiliar(item, SlimefunItems.REPAIRED_SPAWNER, false))) return sfi;
```
TheBusyBiscuit: This method performs a for loop, you should place this special case outside of the for loop, otherwise it will perform the same check a thousand times per ItemStack, we only need it once though.

src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
```diff
@@ -37,6 +37,8 @@
 
 public class AncientAltarListener implements Listener {
 
+	public static boolean altarinuse = false;
```
TheBusyBiscuit: This restricts the maximum amount of Altars on the server to 1.
Use a HashSet with Locations instead to make this variable be per-altar.
And maybe move it to the Variables.class

src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
```diff
@@ -94,14 +105,13 @@ else if (item.getID().equals("ANCIENT_ALTAR")) {
...
 							}
-
 							ItemStack result = Pedestals.getRecipeOutput(catalyst, input);
 							if (result != null) {
+								this.altarinuse = true;
 								List<ItemStack> consumed = new ArrayList<ItemStack>();
 								consumed.add(catalyst);
 								PlayerInventory.consumeItemInHand(e.getPlayer());
-								Bukkit.getScheduler().scheduleSyncDelayedTask(SlimefunStartup.instance, new RitualAnimation(altars, b, b.getLocation().add(0.5, 1.3, 0.5), result, pedestals, consumed), 10L);
-							}
```
TheBusyBiscuit: Ummmm, where did that bracket go?

TheBusyBiscuit:
@mrcoffee1026 No problem, it is just conventional to use short names for commits and to provide a large summary of the changes in the Pull Request's comment here.

src/me/mrCookieSlime/Slimefun/listeners/AncientAltarListener.java
mrcoffee1026: You sure about that?  I just tried making a second one and it works fine too...

TheBusyBiscuit: Yes. You use the same value for all altars.
You can exploit your "fix" by just activating one after another, waiting for one to finish, that sets it to false, allowing abuse of the other one that is still in use, but the value is set to false now.

Use a Location HashSet instead.

mrcoffee1026: Or... does this mean only one can be "in use" at any given time... as I've virtually no ability to test that.  I'll have to see how this is done elsewhere, I think I saw something like the Jet Boots stored a speed somehow, maybe I can use that as a starting point.

mrcoffee1026: alright killing this so the rest can go through... i'll work on using the location hashset after. reputting these files as undone.

TheBusyBiscuit:
About the for loop, you are right, you can't access the SlimefunItem there, but you can fall back to using getByID(), seperate the two if statements and implement them this way:

PSEUDO CODE INCOMING

if (Broken Spawner) return getByID(Broken Spawner);
if (Reinforced Spawner) return getByID(Reinforced Spawner);

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -344,6 +345,7 @@ public boolean isItem(ItemStack item) {
 		if (this instanceof ChargableItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
 		else if (this instanceof DamagableChargableItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
 		else if (this instanceof ChargedItem && SlimefunManager.isItemSimiliar(item, this.item, false)) return true;
+		else if ((SlimefunManager.isItemSimiliar(item, SlimefunItems.BROKEN_SPAWNER, false)) || (SlimefunManager.isItemSimiliar(item, SlimefunItems.REPAIRED_SPAWNER, false))) return true;
```
TheBusyBiscuit: This is not a static method.
Check if this.getID() matches BROKEN_SPAWNER or REINFORCED_SPAWNER for each of the two cases.

mrcoffee1026: While "this.getID()" is sometimes a BROKEN or REINFORCED spawner, it never is when interfacing with the anvil, and there doesn't appear to be a getID() for the item being checked.  The Very next line which is " return SlimefunManager.isItemSimiliar(item, this.item, true); " suggests that this was the proper thing to put there since this is part of the current release and supposedly works for everything else.  Also, this works in this form, if I start checking "this", it stops trying to match any item that the player is using/holding/whatever... and instead checks whatever item was last referenced by whichever function was last used... providing virtually random information that has nothing to do with the item I'm putting in the anvil (which is item).  I may still be doing something wrong here, but "this is not a static method" doesn't make sense to me since the very next line does nearly the identical thing, and that was just fine for the release.

mrcoffee1026:
I'm abandoning the getbyitem "fix" since it seems to affect nothing.  The isItem is the only need the anvil has... I sorta get what you're saying with the pseudocode but... if I can't get past adding a line of code that's nearly identical to one that already existed in the method before I got to it, well then it's just not worth the trouble.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
TheBusyBiscuit: public boolean isItem(...

It isn't static. I don't really get what you are trying to explain, I think you are thinking a bit too complicated.

TheBusyBiscuit: I can understand if you want to undo this change as well, so I can at least merge the other changes for now.
Pull Requests don't have to be packed full of tweaks.

mrcoffee1026: "what" isn't static, what exactly are you talking about here, I didn't redefine anything here or change the method in any way?  And why does this work if it doesn't "work"?

TheBusyBiscuit: The method isn't static. Yes, you didn't redefine anything.

But the method belongs to an object, an instance of SlimefunItem.
And you are performing a check that does not include the current instance.

e.g. (again just pseudo)

SlimefunItem item = "Magic Eye Of Ender";
item.isItem(Broken Spawner) => true

SlimefunItem item = "Grandmas Walking Stick";
item.isItem(Broken Spawner) => true

It should only return true, if "item" was actually a Broken Spawner, the method you are working is a comparison and you just return true if it is a spawner, without checking "this".

mrcoffee1026: "this" cannot be accessed in a static context... I can't very well change the whole method to static, that would completely break it.  I really don't get what you're wanting changed here.

TheBusyBiscuit: > "this" cannot be accessed in a static context...

Again, the method is NOT static, so you not only can but must use "this"...
I want you to use "this", re-read my example above.

TheBusyBiscuit: You basically need to replace this line with this:

else if (this.getID().equals("BROKEN_SPAWNER")) return SlimefunManager.isItemSimiliar(item, SlimefunItems.BROKEN_SPAWNER, false);
else if (this.getID().equals("REINFORCED_SPAWNER")) return SlimefunManager.isItemSimiliar(item, SlimefunItems.REINFORCED_SPAWNER, false);

That's what it has to be, you need to use this in order to refer to the instance because this is object-oriented.
You can just use that snippet as it is, feel free to ask questions if you still don't understand why it needs to be this way.

TheBusyBiscuit:
Also, it looks like the Anvil uses getByItem and not isItem

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
mrcoffee1026: The result of this change is that it ceases to function and goes back to being broken again (broken spawners can be renamed on the anvil, bypassing the altar and all ingredients to become a placeable pig spawner).  I don't get why so caught up on this point.  In the two cases where this needs to return TRUE and no other thing... an item has been passed to this method which 
 -- >SlimefunManager.isItemSimiliar(item, SlimefunItems.REPAIRED_SPAWNER, false)  <-- 
The result of this statement is also TRUE.  Why are you having me check to make sure that "this...." anything??? It has nothing to do with what needs to be done here.  I'm not comparing to "this" like other methods may be doing and still need to do... in THIS CASE I need it to return "true", regardless of what "this" is, the item has been established to be a reinforced or broken spawner, isItem necessarily needs to return TRUE for those items.

mrcoffee1026: Also I've tried doing a lot of echoing and things, and from the best of my knowledge at the time of the check when the spawner is placed into the anvil and attempted to be renamed... this.getID() is either MOB_SPAWNER or ANVIL.

TheBusyBiscuit: Show me that code.

mrcoffee1026: ...
else if ((SlimefunManager.isItemSimiliar(item, SlimefunItems.BROKEN_SPAWNER, false)) || (SlimefunManager.isItemSimiliar(item, SlimefunItems.REPAIRED_SPAWNER, false))) return true;

Right here... if item is either of these, this is true.  No use of this becasue that's not what's being compared.

as taken from ItemListener.java....
public void onAnvil(InventoryClickEvent e) { ... 
if .. !SlimefunItem.isDisabled ... 

in this very same file SlimefunItem.java:
	public static boolean isDisabled(ItemStack item) {
	    for (SlimefunItem i: all) {
			if (i.isItem(item)) {
				return i.isDisabled();
			}
		}
	    return false;
	}


return i.isItem(item) --> this needs to be true for REINFORCED and BROKEN spawners, this is all that needs to happen here, this is what my one added line of code does.

mrcoffee1026:
Anvil uses isItem... see the code i pasted in the comment prior.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
TheBusyBiscuit: > return i.isItem(item) --> this needs to be true for REINFORCED and BROKEN spawners
Yes but it has to be true ONLY if they actually are Spawners...
Returning true no matter what doesn't fix anything, it just breaks everything.

And as you pointed out, the for loop does "i.isItem(item)"
In other words this compares "i" to "item".
But you only check whether "item" is a spawner but you don't check if "i" is one too.

In the method isItem() you have to check if "i" is a spawner. "i" is available as "this" in that method's context. This is basic Object Oriented Programming.

TheBusyBiscuit: And by "Show me that code" I actually meant the tests in which a SlimefunItem's id would suddenly be MOB_SPAWNER or ANVIL, which is outright impossible.
Perhaps, you confuse getID() with item.getType()

mrcoffee1026:
After a bunch more retesting discovered that the changes I recently removed from the getByItem method was what was fixing it, changing isItem was accomplishing nothing.  Hopefully, this is what you had in mind, I put that back in the way I thought you meant and just wiped everything I did out of isItem.

mrcoffee1026:
Also is there some sort of "keep-alive" event that uses getByItem?  I put a println in there just to see what items were being checked... before I even logged into my local test server it was vomiting out 3 lines a second checking a "book" item... and which book there'd be no way of knowing.  I've a mind to track that down and stop it if it's not intentional.

TheBusyBiscuit:
By the way, successfully compiled:
https://thebusybiscuit.github.io/builds/TheBusyBiscuit/Slimefun4/master/#34

mrcoffee1026:
Thanks and sorry for the misunderstanding.

VoidAngel:
@mrcoffee1026 the thing you did in the SlimefunItem class breaks the spawner recipes in the guide D:
Its kind of hard to explain/generalize it so mention me if you want more details, but its pretty self explanatory if you try clicking on the spawner recipes in game.

VoidAngel:
Yeah I'm not sure why it caused that, but I dont have time to check today, so all I can say is that I just reverted the changes to getByItem and its working fine again.

mrcoffee1026:
I tried again leaving out the check for the repaired_spawner... and while that does mean a person could rename a repaired spawner in an anvil thereby making it a pig spawner... that's not as big of a deal as bypassing the entire crafting procedure for a Reinforced Spawner.  Any other solution here seems like it would be pretty heavy handed like... checking through every possible mob type for a match.  I'm not sure why that would be necessary but apparently leaving the type out produces a "null" result and breaks it.  Anyway I'll recommit without the second check.

While checking things I had to do some echoing again and have determined that this method is being called extensively every moment to check the contents of every AutoEnchanter on the server.  Each stack of empty books for these is being checked every single tick it seems.  I believe there should be a requisite "if book" do nothing type deal to curtail some lag.  In any case this routine seems to be running every moment regardless of how far from the AutoEnchanter I can get including leaving the world entirely.  With no chunks loaded, this really shouldn't be checking this excessively so I'm going to look into this right away.  New commit coming up to deal with the recipe tho.

PR #661, opened 2018-04-23
Fix Two Player Item Duplication Bug

### LINKED ISSUE #412: Discrepancy between Block Drops and live contents
## Description
Duplicate item from machine.
If one is breaking his machine ,another jsut open that gui, get double
## Server Log

## Evidence / Proof
https://www.youtube.com/watch?v=j58I5albHig&feature=youtu.be

## Environment
Spigot 1.12
Slimefun 4.1.9
CS-Corelib 1.5.16

KyleEstein:
This is my proposed fix to prevent a player from removing an item from the inventory of a Slimefun device that expands AContainer (which is most electrical devices) after another player has broken that device and the inventory contents have dropped, thus duplicating the item.

One thing I'm not sure about is if my inv.toInventory().clear(slot) is acceptable or if I should clear the contents of the slot using inv.replaceExistingItem(slot, null), which is the method used by the TrashCan.

This should close #412

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,7 @@
         <dependency>
 	        <groupId>com.github.TheBusyBiscuit</groupId>
 	        <artifactId>CS-CoreLib</artifactId>
-	        <version>v1.5.16</version>
+	        <version>v1.5.18</version>
 	    </dependency>
         <dependency>
 	        <groupId>me.minebuilders</groupId>
--- a/src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
+++ b/src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
@@ -211,9 +211,15 @@ public abstract class ProgrammableAndroid extends SlimefunItem {
 				boolean allow =  reason.equals(UnregisterReason.PLAYER_BREAK) && (BlockStorage.getBlockInfo(b, "owner").equals(p.getUniqueId().toString()) || p.hasPermission("slimefun.android.bypass"));
 
 				if (allow) {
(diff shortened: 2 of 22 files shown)


TheBusyBiscuit:
replaceExistingItem is safer, as it initialises the Inventory Object if necessary, toInventory() just does "return this.inv;" without any checks. replaceExistingItem also calls an ItemManipulationEvent but this is of no interest for this case anyway.

Also, this interface is overwritten in a lot of machines that derive of this class. E.g. machines that have input slots but no output slots and therefore need an override.
So make sure to implement fixes for these as well.

And by the way, what if you were to simply call menu.close() at the end of this block of code, should also work, right?

KyleEstein:
> And by the way, what if you were to simply call menu.close() at the end of this block of code, should also work, right?

Sorry, but this is my first time working with java and with plugins (I'm sure you're getting tired of hearing that). I've searched and I don't see a menu class anywhere. No where in slimefun is there a menu.close(). The closest thing I see is InventoryView.close(). Is this what you were referring to? I don't think I could use that since I only know the player that broke the block. I don't know which player(s) have the inventory of the block open.

TheBusyBiscuit:
@KyleEstein My bad, I meant the variable "inv", it is an instanceof of BlockMenu.

By the way (if you are new to Java this may be helpful), in most languages there are certain conventions.
In Java Variables tend to be in lower case, hence why a class called "menu" is probably not likely.

And BlockMenu#close() closes the View of all players who have this Inventory opened, it knows that.
And I am not saying .close() is a better option than doing .replaceExistingItem()
Just threw out an idea, in the end .replaceExistingItem() may even be safer.

KyleEstein:
I synced my fork with upstream master then commited my changes to my fork and now there are many commits in this pull request that are not mine. Should I just close this pull request and create a fresh one?

PR #664, opened 2018-04-25
Guide dupe fix + Elevator trap fix

VoidAngel:
This PR proposes to fix:
1) A bug where players can click items out of the guide in specific situations
2) A bug where players can get stuck on elevators and teleporters (usually if there is no available destinations) resulting in the need for the player to be forcefully teleported off of the plate.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Variables.java
+++ b/src/me/mrCookieSlime/Slimefun/Variables.java
@@ -24,5 +24,6 @@ public class Variables {
 	public static List<UUID> cancelPlace = new ArrayList<UUID>();
 	public static Map<UUID, ItemStack> arrows = new HashMap<UUID, ItemStack>();
 	public static List<UUID> usingGuide = new ArrayList<UUID>();
+	public static Map<UUID, String> dialogueCooldown = new HashMap<UUID, String>();
 
 }
--- a/src/me/mrCookieSlime/Slimefun/Variables.java
+++ b/src/me/mrCookieSlime/Slimefun/Variables.java
@@ -23,5 +23,6 @@ public class Variables {
 	public static List<UUID> blocks = new ArrayList<UUID>();
 	public static List<UUID> cancelPlace = new ArrayList<UUID>();
 	public static Map<UUID, ItemStack> arrows = new HashMap<UUID, ItemStack>();
+	public static List<UUID> usingGuide = new ArrayList<UUID>();
 
 }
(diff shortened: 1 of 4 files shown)


src/me/mrCookieSlime/Slimefun/listeners/TeleporterListener.java
```diff
@@ -63,8 +69,18 @@ else if (item.getID().equals("GPS_ACTIVATION_DEVICE_PERSONAL")) {
...
+			else Variables.dialogueCooldown.put(e.getPlayer().getUniqueId(), e.getClickedBlock().getLocation().toString());
 			Elevator.openDialogue(e.getPlayer(), e.getClickedBlock());
 		}
 	}
-
+	
+	@EventHandler
+	public void onPlayerMove(PlayerMoveEvent e) {
+		if(Variables.dialogueCooldown.containsKey(e.getPlayer().getUniqueId())) {
+			if(!e.getPlayer().getLocation().toString().equals(Variables.dialogueCooldown.get(e.getPlayer().getUniqueId()))) {
```
TheBusyBiscuit: Be very careful with this Event, it can be called much more often than you would think, for big servers it probably should not be used at all. Maybe add this check to a new **sync** task right here: https://github.com/TheBusyBiscuit/Slimefun4/blob/master/src/me/mrCookieSlime/Slimefun/SlimefunStartup.java#L357

Also the current Player Location will never match the Pressure Plate's Location. use getBlock().getLocation() instead.

src/me/mrCookieSlime/Slimefun/Variables.java
```diff
@@ -23,5 +23,10 @@
 	public static List<UUID> blocks = new ArrayList<UUID>();
 	public static List<UUID> cancelPlace = new ArrayList<UUID>();
 	public static Map<UUID, ItemStack> arrows = new HashMap<UUID, ItemStack>();
+<<<<<<< HEAD
+	public static Map<UUID, String> dialogueCooldown = new HashMap<UUID, String>();
+=======
+	public static List<UUID> usingGuide = new ArrayList<UUID>();
+>>>>>>> parent of f595cb0... Fix for players getting stuck in elevator books/teleporter guis
```
TheBusyBiscuit: Ehhhm, this github artifact should probably not be in the code...

VoidAngel: Lol
*derp*
Sorry. This is fixed now

TheBusyBiscuit:
Uploaded the new CS-CoreLib Update to BukkitDev.
But it is always tricky to use new features, so we should expect a lot of "I updated Slimefun, this throws errors" posts.

VoidAngel:
Should I use try-catch around each one with a update message? or just let them figure it out? XD

TheBusyBiscuit:
Try/catch would be an option...
Meh, I don't know.

VoidAngel:
Its at the top of the method so if it were to break during runtime it wouldnt cause a serious issue, besides of course the guide being empty.

PR #667, opened 2018-04-26
Fix Two Player Device Inventory Item Duplication Bug

### LINKED ISSUE #412: Discrepancy between Block Drops and live contents

KyleEstein:
I updated every onBreak() handler I could find that included a call to dropItemNaturally(). The updated code will remove the item from the device's inventory immediately after it is dropped using replaceExistingItem(). The only device I didn't update was the ANCIENT_PEDESTAL. It did not appear to use the inventory setup the other devices used so I wasn't sure how to handle it. I don't think all the items I updated were actually vulnerable to the duplication bug. I have a feeling it was limited to the onBreak() handlers which used a for loop to cycle through all the inventory slots. But I also think consistency is correct. I also tested using inv.close() at the start of the onBreak() and it does appear to correct the bug as well. I chose not to use this solution as there could exist some scenario, such as a very laggy server, that would allow a player to reopen the device's inventory and grab the item during the onBreak(). I still feel that removing the item from the device's inventory as it is dropped is the correct solution.

This should close #412

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
+++ b/src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
@@ -211,14 +211,17 @@ public abstract class ProgrammableAndroid extends SlimefunItem {
 				boolean allow =  reason.equals(UnregisterReason.PLAYER_BREAK) && (BlockStorage.getBlockInfo(b, "owner").equals(p.getUniqueId().toString()) || p.hasPermission("slimefun.android.bypass"));
 
 				if (allow) {
-					if (BlockStorage.getInventory(b).getItemInSlot(43) != null) {
-						b.getWorld().dropItemNaturally(b.getLocation(), BlockStorage.getInventory(b).getItemInSlot(43));
-						BlockStorage.getInventory(b).replaceExistingItem(43, null);
-					}
-					for (int slot: getOutputSlots()) {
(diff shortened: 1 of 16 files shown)


TheBusyBiscuit:
Yep, Ancient Pedestals do not really have an Inventory and this does not apply to them.

By the way I saw a lot of these call BlockStorage.getkInventory(b); quite often, especially in a loop.
Might be worth packing it into a variable before the loop then.

KyleEstein:
> By the way I saw a lot of these call BlockStorage.getkInventory(b); quite often, especially in a loop.
Might be worth packing it into a variable before the loop then.

Sure, I can do that. If there was a variable I used it. If there wasn't, I used BlockStorage.getInventory because I thought there was a reason a variable was not being used. I also noticed some handlers had a check for an empty inventory and some did not. Should I add `if (inv != null) {` to all of them when I add the variable?

TheBusyBiscuit:
A null check can never hurt.

KyleEstein:
All modified onBreak() handlers should now use a variable `inv` for the inventory contents instead of repeatedly calling `BlockStorage.getInventory(b)`. Checks for a null inventory have also been added.

PR #673, opened 2018-04-29
Allow admins to chose ore washer and dust washer behavior

KyleEstein:
When the Slimefun plugin was updated on the server I play on, many players were less than pleased with the new behavior of the ore washer. They liked being able to force the ore washer to produce specific dust types. In addition, the server I play on has cargo networks disabled, so the new behavior of the electric dust washer makes them all but unusable. This PR adds two configuration options: legacy_ore_washer and legacy_dust_washer. They default to false which will maintain the current, updated behavior of each device. If an option is changed to true, then the associated device will revert to the old style behavior.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ElectricDustWasher.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ElectricDustWasher.java
@@ -83,19 +83,16 @@ public abstract class ElectricDustWasher extends AContainer {
 		else {
 			for (int slot: getInputSlots()) {
 				if (SlimefunManager.isItemSimiliar(BlockStorage.getInventory(b).getItemInSlot(slot), SlimefunItems.SIFTED_ORE, true)) {
-					boolean empty_slot = false;
 					if (!legacy_dust_washer) {
+						boolean empty_slot = false;
 						for (int output_slot: getOutputSlots()) {
 							if (BlockStorage.getInventory(b).getItemInSlot(output_slot) == null) {
 								empty_slot = true;
 								break;
 							}
 						}
+						if (!empty_slot) return;
 					}
-					else {
-						empty_slot = true;
-					}
-					if (!empty_slot) return;
 
 					ItemStack adding = SlimefunItems.IRON_DUST;
 					if (SlimefunStartup.chance(100, 25)) adding = SlimefunItems.GOLD_DUST;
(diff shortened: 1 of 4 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ElectricDustWasher.java
```diff
@@ -101,6 +108,7 @@ else if (ChargableBlock.isChargable(b)) {
 					else if (SlimefunStartup.chance(100, 25)) adding = SlimefunItems.SILVER_DUST;
 					
 					MachineRecipe r = new MachineRecipe(4 / getSpeed(), new ItemStack[0], new ItemStack[] {adding});
+					if (!fits(b, r.getOutput())) return;
```
TheBusyBiscuit: This should probably only be called if legacy_dust_washer is true.

Aka (legacy_dust_washer && fits...)

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ElectricDustWasher.java
```diff
@@ -83,13 +84,19 @@ else if (ChargableBlock.isChargable(b)) {
 			for (int slot: getInputSlots()) {
 				if (SlimefunManager.isItemSimiliar(BlockStorage.getInventory(b).getItemInSlot(slot), SlimefunItems.SIFTED_ORE, true)) {
 					boolean empty_slot = false;
```
TheBusyBiscuit: the empty_slot boolean should be moved inside the (!legacy_dust_washer) statement as it is not needed otherwise.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ElectricDustWasher.java
```diff
@@ -83,13 +84,19 @@ else if (ChargableBlock.isChargable(b)) {
...
+							if (BlockStorage.getInventory(b).getItemInSlot(output_slot) == null) {
+								empty_slot = true;
+								break;
+							}
 						}
 					}
+					else {
+						empty_slot = true;
+					}
 					if (!empty_slot) return;
```
TheBusyBiscuit: This is also only needed if legacy_dust_washer is false...

KyleEstein:
All requested changes have been made.

Issue #583, opened 2018-01-25
Saw Mill not handling protection plugins correctly


BlackBeltPanda:
## Description
Using the Saw Mill is very inconsistent unless the player has all permissions. It will work sometimes but mostly it won't do anything. As soon as the player has full perms, it seems to work fine. There are no errors in the console.

## Server Log
Log's too big for pastebin. =(

## Evidence / Proof
Video: https://www.youtube.com/watch?v=E1ur6vSkX1I

## Environment
MC: Paper v1_12_R1
CS-CoreLib v1.5.16
Slimefun v4.1.13
ChestTerminal v1.0

Poslovitch:
I'm seeing two reasons for this to happen : 
1. There is an exception thrown when clicking on the Saw Mill (therefore we'd need logs)
2. Another plugin is interfering with the click event. (afaik, mcMMO doesn't affect anything related to Slimefun, but another plugin might do it).

Here are a few things you should do or answer to in order to see what's really wrong : 
1. What Skyblock plugin are you using? (ASkyBlock has its own protection handling, which doesn't affect Slimefun, but uSkyblock uses WorldGuard, which might (i'm not sure))
2. Run a clean test server with Slimefun and CSCore-Lib installed. Does the bug occur?
3. Run a copy of your server and remove each "unrelated" plugin, one by one, until the bug vanishes.

AtomicScience:
I encouraged the same bug and, I think, the reason of it is AntiFastBreak, which is included in almost any anti-cheat. CSCoreLib breaks block on behalf of the player and anti-cheat cancel event.
So, related methods in CSCoreLib must be rewritten with using .breakNaturally() or .setMaterial(Material.AIR)

BlackBeltPanda:
AtomicScience is correct. After testing on a clean server and loading plugins individually, the culprit was AAC, my anti-cheat plugin.

TheBusyBiscuit:
The Saw Mill (and possibly other blocks) should switch to CS-CoreLibs protection handler rather than just calling an event, marking it as a Bug Report therefore.

AtomicScience:
Well, also I noticed some problems with Slimefun jetpacks - some anti-cheats consider what player use Fly cheat and kick or even auto-ban him. I'm not ensure this bug exist, but...
So, Slimefun or/and CSCoreLib must integrate with anti-cheat to prevent this issues

madolebr:
[...]

PR #674, opened 2018-04-29
Updated saw mill protection check

VoidAngel:
Fix for #583 and also a tiny dupe glitch where players could break barrels on the saw mill and get both planks and the barrel back.


src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -1887,10 +1887,24 @@ public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List
 			@Override
 			public boolean onInteract(Player p, MultiBlock mb, Block b) {
 				if (mb.isMultiBlock(SlimefunItem.getByID("SAW_MILL"))) {
-					if(CSCoreLib.getLib().getProtectionManager().canAccessChest(p.getUniqueId(), b, true)) {
+					if(CSCoreLib.getLib().getProtectionManager().canBuild(p.getUniqueId(), b, true)) {
```
TheBusyBiscuit: Should rather be b + (0,1,0) since the above block is being destroyed.

PR #683, opened 2018-05-03
Resolving the Ancient Altar dupe glitch.

mrcoffee1026:
Reforked and resubmitted.


SoSeDiK:
Sorry, but why did you use the 'Â' character in the "PROGRAMMABLE_ANDROID_MINER"? I think '§' should be replaced with '&' but I have no idea about what that 'Â' means.

TheBusyBiscuit:
First of all, thanks for the cleanup.
Then as @SoSeDiK pointed out, there are malformed characters in the code.

Besides that, how are tests of this new version going?

SoSeDiK:
Well, I should also note that you changed the head's texture _(the first 'e' disappeared)_. Won't it cause a problem? And still, it would be nice if you replace those '§' with '&' to avoid some issues with encoding in future, please.

TheBusyBiscuit:
The .getByItem() behaviour for Repaired Spawners seems pretty strange.
Anyway, what do your tests of the Ancient Altar show? Is it fixed?

mrcoffee1026:
Well in every test I've done I've not been able to duplicate anything, even on the live server.  If someone else is still able to make the duplication happen by breaking during craft, then I'm not sure where else it could be doing it... but this seems to be a timing thing even under that one condition, I may not be mining the blocks at the perfect moment to recreate that circumstance (if that's the case, try to be more descriptive about when the breaking needs to take place).

I agree ithe repaired spawner behavior is very strange, why it would matter for display in the guide doesn't make a lot of sense to me anyway?  I believe this is doing it:
```
		final SlimefunItem sfItem = SlimefunItem.getByItem(item);
		if (sfItem == null) {
			if (!all_recipes) return;
		}
```
But... why the change I made wouldn't result in a match here, that I don't know.  Seems to me the item for the guide would have had the lore of "Type: <Type>" instead of the actual type... but... wouldn't the change I made also have caused that to match as well?  It's basically saying.. if it is a Repaired_Spawner and the lore matches or not... it should match right?  Or does the "false" value mena that lore absolutely cannot match for a "true" value.  That seems unlikely but... commenting this out for the repaired spawners means they can be renamed in an anvil, that's... not the worst way things could be.  If you rename a reinforced spawner, it becomes a pig spawner... but it's not tragic that ranaming it changes the type, and does only that.

Replaced the color code character with ampersands, I'm not sure why they're changed in your current build.

TheBusyBiscuit:
After a quick look at the source: 
It is called REINFORCED_SPAWNER not REPAIRED_SPAWNER
That's why it returns null...

mrcoffee1026:
I only just said it wrong.  That was the correct usage in the if statement that was set up, but it breaks the repaired spawner recipe in the guide if I leave it in there.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -331,7 +331,8 @@ public static SlimefunItem getByID(String id) {
 	public static SlimefunItem getByItem(ItemStack item) {
 		if (item == null) return null;		
 		if (SlimefunManager.isItemSimiliar(item, SlimefunItems.BROKEN_SPAWNER, false)) return getByID("BROKEN_SPAWNER");
-		if (SlimefunManager.isItemSimiliar(item, SlimefunItems.REPAIRED_SPAWNER, false)) return getByID("REPAIRED_SPAWNER");
```
TheBusyBiscuit: This should be... getByID("REINFORCED_SPAWNER")

mrcoffee1026:
No, it shouldn't:
"REINFORCED_SPAWNER cannot be resolved or is not a field"

mrcoffee1026:
Keep in mind that uncommenting that line DOES work as intended... except that it also breaks the recipe in the book.  It has to be commented out for now.

PR #697, opened 2018-05-12
Update

SoSeDiK:


NathanAdhitya:
Nice code cleanup, some doesn't seem necessary for even the compiler to work properly, but eh, who cares when one has OCD.

TheBusyBiscuit:
Sorry but I will not go through 188 files that have been changed to hunt for your "update"...
Please do not convert the files to any other line endings format or whatever this does.

SoSeDiK:
Well, actually it was my goal :) In different classes, there were different code format.
I replaced many deprecated methods (getBlockInfo() > getLocationInfo(), SlimefunGuide.getItem()  etc).
Also replaced spaces with tabs.
Changed some code parts like this one: if() > if ().
Removed tabs where there was no code (this was in many classes and others keep update code with supporting this format, so yeah).
Added newlines at the end of files.
Added missing spaces in comments.
That's why there are so many changes but I really didn't change anything important in the first commit and it compiles without any problems. There is nothing that really matters, however, the code will have the same format everywhere.

TheBusyBiscuit:
Still... a pain to review...

SoSeDiK:
Haha. I can make 188 pull requests, so you can review it 188 times per file instead of a one-time review of 188 files. This will make it much easier.
Anyway, I can delete the branch, right?

PR #706, opened 2018-05-21
Fixes soulbound backpack crafting

VoidAngel:
Also a few more .clones() were added as a precaution for the future.


TheBusyBiscuit:
Nice.

Just a warning: This build fails to compile on my machine.
https://thebusybiscuit.github.io/builds/TheBusyBiscuit/Slimefun4/master/

But that is due to some SSL Certificate Exceptions which I have absolutely no clue how to fix.

PR #712, opened 2018-06-12
Research unlock event update

TSEngineer:


TheBusyBiscuit:
Why do you need a Research "undo" event?
And this change causes a lot of confusion because it has nothing do with unlocking and will probably trip up every listener out there.

PR #718, opened 2018-06-15
Parse block info at startup

creator3:
Profiler
As the screenshot shows,
```java
new Config("data-storage/Slimefun/temp.yml")
```
is particularly expensive because it has to try to load a file from disk.
This PR will avoid that call for block info and also save time spent parsing JSON.

Block info has always been limited to Map<String, String> despite the Config interface, so this won't change anything there. Mutating the Config will change the block info now, but I doubt that happens anywhere without the block info being set anyway. If it's a problem, the Map can of course be cloned.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/api/BlockInfoConfig.java
+++ b/src/me/mrCookieSlime/Slimefun/api/BlockInfoConfig.java
@@ -1,0 +1,132 @@
+package me.mrCookieSlime.Slimefun.api;
+
+import java.io.File;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.bukkit.configuration.file.FileConfiguration;
+
+import me.mrCookieSlime.CSCoreLibPlugin.Configuration.Config;
+
+public class BlockInfoConfig extends Config {
+	private Map<String, String> data;
+	public BlockInfoConfig() {
+		this(new HashMap<>());
+	}
+	public BlockInfoConfig(Map<String, String> data) {
+		super((File)null,(FileConfiguration)null);
+		this.data=data;
+	}
+	public Map<String, String> getMap(){
+		return data;
+	}
+	@Override
+	protected void store(String path, Object value) {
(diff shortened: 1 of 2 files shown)


Poslovitch:
Looks good. Have you got some timings that can show the effects of this change?

TheBusyBiscuit:
I like this design change.

However please also implement all the other methods because "it only supports String" is definitely false, all types are supported and should be made available.

creator3:
[parseJSON](https://github.com/TheBusyBiscuit/Slimefun4/blob/634c3063426c28a349c55630b15bd0e23fcc8ebf/src/me/mrCookieSlime/Slimefun/api/BlockStorage.java#L311) returns a Map<String, String>, and [the serialization code](https://github.com/TheBusyBiscuit/Slimefun4/blob/634c3063426c28a349c55630b15bd0e23fcc8ebf/src/me/mrCookieSlime/Slimefun/api/BlockStorage.java#L387) does `json.put(key, cfg.getString(key));`


I assumed it was unlikely that anything used `getBlockInfo(Location l, String key)`, but I looked into it some. Although `getLocationInfo(Location l, String key)` operates directly on a Map returned by parseJSON and supports periods, addBlockInfo uses a Config and would not.

However, it's obviously not a big deal to make getBlockInfo/addBlockInfo use the Map directly.

creator3:
@Poslovitch `matchesFilter` is down from 97% of `insert` to 2.6%.

TheBusyBiscuit:
@creator3 
Alright, somehow I thought that these things were supported in the past, okay then.

TheBusyBiscuit:
I am gonna merge this once I got my compiler working again.

VoidAngel:
I don't know who you are but I love you ;'D

TheBusyBiscuit:
Merging this anyway, but my compiler is still broken so I cannot upload any working jars right now.

If anyone has an idea on how to fix it, here's the Issue:
https://github.com/TheBusyBiscuit/builds/issues/2

Issue #715, opened 2018-06-15
Ignition Chamber Lag


realchessx:
## Description (Required)
Ignition Chamber LAG

## Evidence / Proof (Required)
https://timings.spigotmc.org/?url=yelepifihi
http://prntscr.com/jv5sp2

## Environment (Required)
slimefun 4.1.15
cs-corelib 1.5.18

madolebr:
you do have some lag due to the ignition chamber,, but u need to do more work to your server in general to get rid of your massive amounts of lag. install react and limit mobs, or something, because u are getting town up with entity movements. reduce view distance and entity activation range

TheBusyBiscuit:
Whoever added the Ignition chamber (I didn't even know such a thing existed) didn't really design it carefully.
It should be improved for sure.

https://github.com/TheBusyBiscuit/Slimefun4/blob/183a34a99c1d46b1223f7360bbb40f0aa408141c/src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java#L74

Poslovitch:
I will have a look at that thing.

andris155:
Use Paper and optimize your server!

Poslovitch:
PaperSpigot has a lot of breaking fixes for a month or two. I wouldnt recommend using it anymore with Slimefun

andris155:
I don't like Spigot because use old timings and not available hopper optimization.

madolebr:
i use paper spigot personally and have less sf bugs because of it, along with great improvement in performance

realchessx:
kk thanks for the advice, paper spigot is not applicable to my server, it corrupt or show error more than it fixed the lag, i tried to optimize more spigot setting it reduce some lag, i still get ignition chamber lag

mrcoffee1026:
[...]

PR #723, opened 2018-06-23
Fixed issue #715

AtomicScience:
Yea, I should have been more careful while creating onIgnitionChamberItemMove listener, but now I fixed it to remove lags.

Old listener checked item transfer every tick, and handling every InventoryMoveItemEvent caused huge server load. But new method handles this event only once and lock the Chamber

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
@@ -66,18 +66,23 @@ public class ItemListener implements Listener {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


VoidAngel:
Hey just  doing some testing of the latest PRs. The imports here look a little weird to me, specifically the CraftDispenser. What type of Bukkit API are you using to compile?

VoidAngel:
Heh sorry not trying to cause problems. When using spigot 1.12.2 or paper 1.12.2 the compiler didn't recognize it. So just a bit confused. Could just be on my end.

TheBusyBiscuit:
Feel free to resubmit this when you added the necessary imports.

PR #744, opened 2018-07-27
Updated to 1.13

xH3LLRAIZ3Rx:
I went ahead and updated Silmefun to 1.13.

Do note I did not fix any warnings, I mainly just replaced all materials with the updated versions. I also added TODO's to the block.setData for fixing since setData() is removed in 1.13.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
     		<dependency>
     			<groupId>org.bukkit</groupId>
     			<artifactId>bukkit</artifactId>
-    			<version>1.12-R0.1-SNAPSHOT</version>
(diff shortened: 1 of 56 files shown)


Meta-Win:
I think it's necessary to fix deprecations and update CSCoreLib too

xH3LLRAIZ3Rx:
I did update CS-CoreLib, but im struggling on a alternative to setData() and changing material data of wool and logs

thefallenone12:
is there a download for both of these we can download yet?

xH3LLRAIZ3Rx:
At its current form no because there's bugs that need fixing

thefallenone12:
oh alright thats understandable. thank you

PR #773, opened 2018-09-04
WIP: 1.13 port

MisterErwin:
Getting slimefun to run on a 1.13.1 spigot.

The new version would not be compatible with a version below spigot api version 1.13.

The following things are still to be done (see file 1.13 port todo.md):

- [ ] Test everything
- [ ]  Add recipes for stained variants (e.g. the rainbow blocks) - For now WHITE_ is used
- [ ]  Add recipes for other variants (slabs, every wood based item) - For now OAK and STONE is the default
- [ ]  Improve the MaterialHelper and move it elsewhere?
- [ ] All the `//ToDo:`s in the code - ALL OF THEM

Oh.. and a CS-CoreLib version with api version 1.13 is also required.

### DIFF
--- a/1.13 port todo.md
+++ b/1.13 port todo.md
@@ -1,0 +1,14 @@
+* Test if the crops actually work
+* Check for color-ignoring recipes (rainbow blocks)
+* Recipes using (trap)doors
+* crucible leaves & terracota
+
+* nether-brick/nether-brocks - maybe broke recipes?
+
+* check for carrot/carrots and potato/potatos
+
+* all that fancy block data :/
+
+* Materialhelper - maybe move that elsewhere/into the lib
+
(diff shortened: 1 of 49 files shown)


TheBusyBiscuit:
Nice, could you please do me a few favours though?

1) Move your MaterialHelper methods into this new CS-CoreLib class
https://github.com/TheBusyBiscuit/CS-CoreLib/blob/master/src/me/mrCookieSlime/CSCoreLibPlugin/compatibility/MaterialHook.java
It is specifically designed for this stuff.

2) Could you please split this up into several reasonably sized Pull Requests and commits.
Having to review all at once can take a long time, especially because I cannot mark areas as approved and then do the rest later.
If you could split this PR up into several packets, then I can review and merge them one by one.

MisterErwin:
Sure - (I squashed my local commits so I would not spam you)
Will do it once my human body has absorved food

PR #785, opened 2018-09-24
MySQL support for Slimefun

freethemice:
MySQL support, I added this to slimefun and thought I would share.
1: Start server it will make a mysql.yml file, then stop the server. 
2: Enter your mysql database info and set enabled to true.
3: Start server, slimefun will upload all current  saved block info to the database, will take sometime.
4: Server will rename old data files to world_old for backup.
5: Done.


xH3LLRAIZ3Rx:
This is exactly what Slimefun needs, I am so glad someone took the time to implement this. I hope this gets merged!

VoidAngel:
Im happy to see it too xD but how much has this been tested?

xH3LLRAIZ3Rx:
That is a good question. How much has been tested and does it possibly fix the old issue where machines would turn into CS-CoreLib heads?

freethemice:
It does catch the machine turning to heads, but just ignore the change. I plan on updating it so it fixes it also. 
I’ve test my self on Linux and Windows Machine with a block size of 523,000. All that really changes is the way it loads and saves.

VoidAngel:
Ill see if I can do some testing too :D

TheBusyBiscuit:
Most of your code is already part of CS-CoreLib, as it already has a MySQL API.
You could maybe take a look at it and see if you can switch to that (you can of course add missing handles into there, so that other plugins can use that too).
I'm just suggesting that to avoid unnecessary double-implementation of code/features.

freethemice:
I’ll take a look and try to use cs-corelib. Thanks for the feed back

PrzemoVi:
You should open connection to the database only when needed and keep it opened as short as possible.

xH3LLRAIZ3Rx:
Does anyone know how up to date this code is with the latest slimefun build? Also will @freethemice address that of what @PrzemoVi said, because that would be a big issue if it is a problem.

PrzemoVi:
I already created a solution:
https://github.com/TheBusyBiscuit/CS-CoreLib/pull/65

After this blocks are not turning into heads anymore.
Also I fixed particles and lastest worldedit & worldguard version support ;)

Just waiting for @TheBusyBiscuit to merge it :)

freethemice:
I currently run my server on MySQL updates to current version of slimefun but have other code in it just for my server. That said change to MySQL didn’t help the problem of machine breaking. 
Unfortunately I don’t have time to update slimefun the way it should be to MySQL using the cs-corelib.

xH3LLRAIZ3Rx:
holy sh*t @PrzemoVi, you are awesome, I am gonna merge that in my own fork.

xH3LLRAIZ3Rx:
oh wow @PrzemoVi, the fix was just a simple if statement fix? How did you determine that was the problem and how was a simple if statement causing such a big problem?

PrzemoVi:
@xH3LLRAIZ3Rx After that you should go to data-storage/Slimefun/stored-blocks/world directory and delete "null" file.

EnderSlime123:
@freethemice Does this work with the latest Slimefun build, and I think an auto setup thing would be nice since I have no clue how to use MySQL.

mrcoffee1026:
The "machine turning into heads" on restart problem is usually resolved by removing the file "null.sfb" before restarting (located in target\data-storage\Slimefun\stored-blocks\the_world_you're_having_problems_with).  The issue may or may not be resolved on the next restart if you do this first.  For some reason the location got logged as an error... so long as that location remains listed in the "null.sfb" file. it will guarantee anything placed there will become a head or worthless item upon a restart... there does not seem to be a way of altering this list or remedying the issue in game, you have to manually remove the file or the issue persists for that location on the server forever.  Hopefully the database approach will either not forget which table its using mid-operation so won't create a "null" table that recreates this issue... I'm... pretty sure it won't let you make a table called "null" in mySQL so that should actually resolve the issue more permanently.

EnderSlime123:
> I currently run my server on MySQL updates to current version of slimefun but have other code in it just for my server. That said change to MySQL didn’t help the problem of machine breaking.
> Unfortunately I don’t have time to update slimefun the way it should be to MySQL using the cs-corelib.

Really think a sepreate Slimefun addon should be made, since this isn't gonna be merged. Ever.

freethemice:
I've updated the code to current and to use my plugin [TitanSQL] (https://github.com/freethemice/TitanSQL)
Also, to effect as little as possible code changes in slimefun

EnderSlime123:
@freethemice Fix the conflicting issue.

PrzemoVi:
This is the worst idea ever. MRCs will never merge this. CS-CoreLib has integrated mysql database API and he told us to make use of that. And I can do this in free time.

TheBusyBiscuit:
I feel like it's time for me to step into this conversation again, before more people call this out or whatever.
I appreciate this Pull Request, no matter what API is used.

So, will this be merged?
Probably not, but that isn't the point anyway. I haven't merged this for multiple reasons.
But I have also let this Pull Request stay open and I definitely will continue to keep it open, so people can see it.
After all this is Open-Source, so if you prefer to use a Slimefun4 Version that supports MySQL, feel free to check out this fork and install that instead.

Thanks again @freethemice

EnderSlime123:
@TheBusyBiscuit I really think this should be merged since having MySQL support would actually reduce RAM usage and TPS drops by quite a lot if I´m correct.

TheBusyBiscuit:
@EnderSlime123 
TPS would not be affected by it.
RAM maybe but with this implementation also not affected at all.

Not everyone can afford a Database by the way, so it's good that this Pull Request added a config option.
But even if performance was affected by it. it still would not have to do anything with the fact whether this feature exists on the main repo or a fork.

freethemice:
It would only help if your on a hdd and not ssd. The way I found that’s works best is to load all data from sql when server starts and push changes ever-so often and at shut down. I’m going to release a slimefun expiation that use the my sql plugin. Thats why I updated the code for this. 
The plugin stores all data and sends it as bulk ever few minutes. Also, pause shutdown and send any threats waiting to be sent to prevent data lose. Other MySQL plugins I tried would lose  data on shutdown or freeze the server when access the database. 
I would prefer to use cs-core lib but i haven’t had time to go over or had the need to.

PrzemoVi:
If server will crash before bulk send, blocks will turn into heads... We should push changes to sql on async thread, not affecting server-tps.

EnderSlime123:
> It would only help if your on a hdd and not ssd or limited for space. The way I found that’s works best is to load all data from sql when server starts and push changes ever-so often and at shut down. I’m going to release a slimefun expiation that use the my sql plugin. Thats why I updated the code for this.
> The plugin stores all data and sends it as bulk ever few minutes. Also, pause shutdown and send any threats waiting to be sent to prevent data lose. Other MySQL plugins I tried would lose data on shutdown or freeze the server when access the database.
> I would prefer to use cs-core lib but i haven’t had time to go over or had the need to.

Do the dev builds have the MySQL support cause the latest build you have is for 1.12.2.

EnderBro123:
@freethemice Where are the jar files to make this actually work. I can't use MySQL without the compatible version.

freethemice:
https://www.spigotmc.org/resources/titansql.67830/

EnderBro123:
> https://www.spigotmc.org/resources/titansql.67830/
> 
> https://www.spigotmc.org/resources/67831

I have no idea how to open jar files. I also have no idea where to drop this Slimefun4MySQL zip file in.

TSIXS:
@freethemice Where is the Slimefun4MySQLPatch source code? I hope to make some changes. Thank you.

MinecraftAdmin:
If newer versions of Slimefun are patched using this method, does it still work?

WalshyDev:
What... what do you mean "are patched using this method"...

His branch is over a year behind in commits. You can use his but it will have a lot of issues and not be supported by us. It isn't able to be merged into here unless it's massively rebased. 

Stuff is more reliable now anyway due to the changes we've done in the last year.

PR #810, opened 2018-11-24
Fixed Sawmill Bug + Furnace Bug

Sniddunc:
This in pair with a PR submitted to the CS-CoreLib repo will fix the Sawmill bug.


PrzemoVi:
I tested it on my server. Everything works :)
Nice work!

Issue #359, opened 2017-07-13
Add Cost and Click to unlock to messages.yml


xH3LLRAIZ3Rx:
## Description
I would love it if you can add the Cost: %level% and Click to unlock to the messages.yml

Poslovitch:
Can you explain it a bit more? I don't understand what you meant

xH3LLRAIZ3Rx:
I would love it where you can modify the message over the researches that are locked where it tells you click to unlock and the cost: http://i.imgur.com/I8b1NC0.png

Poslovitch:
You mean, be able to edit/translate them?

I think I'll go through all the code to make as much texts editable as we can. 
I'll do it in about a week.

xH3LLRAIZ3Rx:
If anyone still wants to do it, go ahead, but I have no use for it now since I decided to learn a little java to modify the code myself ;)

Gusted:
This suggestion has been solved in my pull request #818

PR #818, opened 2018-12-16
Translate Locked items

Gusted:
In this pull request, the Suggestion/issue #359 Solved to translate it in config.yml you can change the messages I hope this pull request will be accepted.

### DIFF
--- a/src/config.yml
+++ b/src/config.yml
@@ -14,6 +14,8 @@ options:
   research-give-fireworks: true
   legacy-ore-washer: false
   legacy-dust-washer: false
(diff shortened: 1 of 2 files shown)


TheBusyBiscuit:
The config is not supposed to be used for localization.

Gusted:
Where then?/

Else you cant edit it :\

StarWishsama:
> Where then?
> @TheBusyBiscuit
> Else you can't edit it :\

Messages.yml I guess. BTW "open" in "Click to open" is in CS-CoreLib, go check it

PR #829, opened 2019-01-16
TO DELETE

PrzemoVi:
To delete.
Something messed up with my repositories.
I`ll create new pull request.


PrzemoVi:
It`s not perfect, machine texture turns into steve head, but machine still works and not turns to "CS-Corelib head" on destroy. I think, that it`s client bug.

PR #831, opened 2019-01-26
Fix particles display

TSEngineer:


PrzemoVi:
Already fixed.
Not compatible with latest cs-corelib pull:
https://github.com/TheBusyBiscuit/CS-CoreLib/pull/65

PR #863, opened 2019-03-27
Update FurnaceListener.java

### LINKED ISSUE #856: Using Enhanced Furnace - Console Spam
## Description (Required)
As soon as an Enhanced Furnace V is used, item put into it, an error is spammed in the console, server seems to be running fine while in game, worried the Enhanced Furnace could cause issues.

## Server Log
https://pastebin.com/eghp5zjH 

## Evidence / Proof (Required)
There is no error provided in game, and furnace operates as it should.

## Environment (Required)
Spigot 1.13.2 Server
CS-CoreLib- vDEV - 35 (git 30d7c341)
Slimefun vDEV - 68 (git 89a8b6b5)

SoSeDiK:
I was trying to fix #856 but I can't reproduce the error.
I changed the code anyway, maybe it'll help.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/FurnaceListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/FurnaceListener.java
@@ -22,15 +22,17 @@ public class FurnaceListener implements Listener {
 	
 	@EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
 	public void onBurn(FurnaceBurnEvent e) {
-		EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
-		if (furnace instanceof EnhancedFurnace && furnace.getFuelEfficiency() > 0)
-			e.setBurnTime(((int) ((1 + 0.2 * furnace.getFuelEfficiency()) * e.getBurnTime())));
+		if (BlockStorage.check(e.getBlock()) instanceof EnhancedFurnace) {
+			EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
+			if (furnace.getFuelEfficiency() > 0)
+				e.setBurnTime(((int) ((1 + 0.2 * furnace.getFuelEfficiency()) * e.getBurnTime())));
+		}
 	}
 
 	@EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
 	public void onSmelt(FurnaceSmeltEvent e) {
-		EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
-		if (furnace instanceof EnhancedFurnace) {
+		if (BlockStorage.check(e.getBlock()) instanceof EnhancedFurnace) {
+			EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
 			Furnace f = (Furnace) e.getBlock().getState();
 			int amount = f.getInventory().getSmelting().getType().toString().endsWith("_ORE") ? furnace.getOutput() : 1;
 			ItemStack result = f.getInventory().getResult() == null ? RecipeCalculator.getSmeltedOutput(f.getInventory().getSmelting().getType()) : f.getInventory().getResult().clone();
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/listeners/FurnaceListener.java
```diff
@@ -22,24 +22,19 @@ public FurnaceListener(SlimefunStartup plugin) {
 	
 	@EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
 	public void onBurn(FurnaceBurnEvent e) {
-		if (BlockStorage.check(e.getBlock()) != null && BlockStorage.check(e.getBlock()) instanceof EnhancedFurnace) {
-			EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
-			if (furnace.getFuelEfficiency() > 0) e.setBurnTime(((int) ((1.0 + 0.2 * furnace.getFuelEfficiency()) * e.getBurnTime())));
-		}
+		EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
+		if (furnace instanceof EnhancedFurnace && furnace.getFuelEfficiency() > 0)
```
TheBusyBiscuit: Never do an "instanceof"-Check **after** Casting.
This would result in a ClassCastException

src/me/mrCookieSlime/Slimefun/listeners/FurnaceListener.java
```diff
@@ -22,24 +22,19 @@ public FurnaceListener(SlimefunStartup plugin) {
...
+			e.setBurnTime(((int) ((1 + 0.2 * furnace.getFuelEfficiency()) * e.getBurnTime())));
 	}
 
 	@EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
 	public void onSmelt(FurnaceSmeltEvent e) {
-		if (BlockStorage.check(e.getBlock()) != null && BlockStorage.check(e.getBlock()) instanceof EnhancedFurnace) {
-			EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
+		EnhancedFurnace furnace = (EnhancedFurnace) BlockStorage.check(e.getBlock());
+		if (furnace instanceof EnhancedFurnace) {
```
TheBusyBiscuit: Same applies for here, use SlimefunItem before the instanceof check.

src/me/mrCookieSlime/Slimefun/listeners/FurnaceListener.java
```diff
@@ -22,24 +22,19 @@ public FurnaceListener(SlimefunStartup plugin) {
...
 			Furnace f = (Furnace) e.getBlock().getState();
-			int amount = f.getInventory().getSmelting().getType().toString().endsWith("_ORE") ? furnace.getOutput(): 1;
-			ItemStack output = RecipeCalculator.getSmeltedOutput(f.getInventory().getSmelting().getType());
-			ItemStack result = f.getInventory().getResult();
-			if (result != null) result = result.clone();
-			f.getInventory().setResult(null);
-			if (result != null) f.getInventory().setResult(new CustomItem(result, result.getAmount() + amount > result.getMaxStackSize() ? result.getMaxStackSize(): result.getAmount() + amount));
-			else f.getInventory().setResult(new CustomItem(output, output.getAmount() + amount > output.getType().getMaxStackSize() ? output.getType().getMaxStackSize(): output.getAmount() + amount));
+			int amount = f.getInventory().getSmelting().getType().toString().endsWith("_ORE") ? furnace.getOutput() : 1;
+			ItemStack result = f.getInventory().getResult() == null ? RecipeCalculator.getSmeltedOutput(f.getInventory().getSmelting().getType()) : f.getInventory().getResult().clone();
```
TheBusyBiscuit: The error would probably still persist, as the error came from "getSmeltedOutput" returning null.
A null-check after storing the output like before would be better.

Maybe getSmeltedOutput() is not working properly anymore and that is the cause of this problem.

SoSeDiK:
I have no idea how i missed that.
Huge thanks!
NPE is definitely from getSmeltedOutput().
I'll won't make a null check for now and look into CS-CoreLib instead.

TheBusyBiscuit:
@SoSeDiK Alright, also to reduce load: Perform BlockStorage.check() only once and store it, then cast it.

SoSeDiK:
I see.
Seems not related to Slimefun & CS-CoreLib.
There probably is another plugin which replaces result with null without deleting the result directly.

PR #869, opened 2019-03-27
Implement createDir() method

SoSeDiK:
& Replaced deprecated code in the same class.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/SlimefunStartup.java
+++ b/src/me/mrCookieSlime/Slimefun/SlimefunStartup.java
@@ -465,7 +465,9 @@ public class SlimefunStartup extends JavaPlugin {
 	}
 
 	private void createDir(String path) {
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/SlimefunStartup.java
```diff
@@ -480,6 +464,10 @@ public void onDisable() {
 		}
 	}
 
+	private void createDir(String path) {
+		if (!new File(path).exists()) new File(path).mkdirs();
```
TheBusyBiscuit: Not to be extra picky 🤓 , but new File(path) should be stored in a variable, rather than instantiating it twice.

SoSeDiK: Hm...
That makes sence.

PR #872, opened 2019-03-30
Rename ORGANIC_FOOD and FERTILIZER

SoSeDiK:
- Using full names instead of numbers according to 1.13 standard as those are different items and not improved ones *(like CROP_GROWTH_ACCELERATOR and CROP_GROWTH_ACCELERATOR_2)*.
  - Not really necessary but I had some problems recognizing them :P


TheBusyBiscuit:
I think it would be a better idea to reverse the names, having them all start with the prefix "ORGANIC_FOOD" sounds more intuitive.

SoSeDiK:
Yes, I agree. But in vanilla everything is like in this commit and that's why I did that it that way.
OAK_LOG, ACACIA_LOG & %COLOR%_WOOL, %COLOR%_STAINED_GLASS and on and on.
Although I can do that if you want.

PR #604, opened 2018-02-18
Switching to lambdas

Trigary:
Lambdas make the code a lot cleaner, especially in situations showcased in this video (while changes on top of lambadas should be applied to that piece of code): https://drive.google.com/file/d/1V3EVXq_aI5VnT-GdPIbyl3vY9DUSKhBb/view

PR #877, opened 2019-03-30
Use lambda expressions

SoSeDiK:
Nicer code & reduced file size.


src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -562,7 +551,6 @@ else if (!(category instanceof LockedCategory)) {
 						if (!(category instanceof SeasonCategory)) {
 							menu.addItem(index, category.getItem());
 							menu.addMenuClickHandler(index, new MenuClickHandler() {
```
TheBusyBiscuit: I suspect the file size could be reduced even further by replacing all MenuClickHandlers with
(p, slot, item, action) -> false

SoSeDiK: Alright, I'll do this.
Also, I can make a separate PR later using diamond operators everywhere if you don't mind.

SoSeDiK:
Oh!
I'm really sorry to everyone who still maintain their custom forks with updates.
I respect you all.

Trigary:
I wonder... The changes in this PR seem awfully similar. Oh, I remember now! #604

SoSeDiK:
Yeah 👀
Though I've tried my best by then to reduce the *mess* by removing some repeating code blocks (SlimefunSetup even got its own PR #875; and the Altar was rewritten somewhere else I believe).

PR #880, opened 2019-04-02
Immediately try to send outgoing terminal items

bverhoeven:
This is an attempt to fix the chest terminal duplication issue that seems to mainly be caused by the fact that the cargo network sees the item in the "out slot", registers it but doesn't move it to the network yet until the next call to the scheduled task. This allows the player to take the item before the task is executed, but will still execute the task and thus insert the item if there is enough space.

Now, before we attempt any storage operations we first check if there's an item in the "outgoing slot", if there is, we'll try to send it instead of waiting for the next call to Cargo Network's scheduled task.

This doesn't seem to impact the system in a negative way and seems to be an appropriate fix for the issue reported here: https://github.com/TheBusyBiscuit/Slimefun4/issues/879

I would appreciate any feedback though.


TheBusyBiscuit:
Sorry, but this pull request conflicted with an earlier Pull Request that I had to merge.
Can you update your fork and perform the change again?

@bverhoeven About feedback: I no longer really know how all the stuff works, it has been waaay too long and others have altered the code so much, that I barely have a clue what is going on.
But I would also like to see someone's opinion on this change and whether someone has tested it.

bverhoeven:
I will close this pull request as there appear to be more issues with Chest Terminal and I think focus should be on getting 1.14.x to work.

PR #922, opened 2019-05-30
Damage explosive pickaxe on use (configurable, off by default)

bverhoeven:
As discussed on Discord, the Explosive Pickaxe currently doesn't take any damage and therefore doesn't lose durability, which isn't a desirable behavior for some of us.

This commit adds a 'damage-on-use' item setting (in `plugins/Slimefun/items.yml`) for EXPLOSIVE_PICKAXE, which is disabled by default, but can be enabled like this:

```
EXPLOSIVE_PICKAXE:
  enabled: true
...
  damage-on-use: true
```

When enabled, the Explosive Pickaxe will get damaged for each block it breaks, in the same way a normal Diamond Pickaxe would get damage if it were to break those blocks. This seems to be similar to how Slimefun damages other tools.

Players can still enable Unbreaking (e.g. durability) on the pickaxe through enchants.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -2081,10 +2081,11 @@ public class SlimefunSetup {
 
 		@SuppressWarnings("unchecked")
 		final String[] explosiveblacklist = Slimefun.getItemValue("EXPLOSIVE_PICKAXE", "unbreakable-blocks") != null ? ((List<String>) Slimefun.getItemValue("EXPLOSIVE_PICKAXE", "unbreakable-blocks")).toArray(new String[((List<String>) Slimefun.getItemValue("EXPLOSIVE_PICKAXE", "unbreakable-blocks")).size()]): new String[] {"BEDROCK", "BARRIER", "COMMAND", "COMMAND_CHAIN", "COMMAND_REPEATING"};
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -2081,10 +2081,11 @@ public boolean onRightClick(ItemUseEvent e, final Player p, ItemStack item) {
 
 		@SuppressWarnings("unchecked")
 		final String[] explosiveblacklist = Slimefun.getItemValue("EXPLOSIVE_PICKAXE", "unbreakable-blocks") != null ? ((List<String>) Slimefun.getItemValue("EXPLOSIVE_PICKAXE", "unbreakable-blocks")).toArray(new String[((List<String>) Slimefun.getItemValue("EXPLOSIVE_PICKAXE", "unbreakable-blocks")).size()]): new String[] {"BEDROCK", "BARRIER", "COMMAND", "COMMAND_CHAIN", "COMMAND_REPEATING"};
+		final boolean damageOnUse = Boolean.TRUE.equals(((Boolean) Slimefun.getItemValue("EXPLOSIVE_PICKAXE", "damage-on-use")));
```
TheBusyBiscuit: You probably should rather just do Boolean.parseBoolean(string) to get a boolean rather than this comparison...

bverhoeven: I thought of that but the problem I faced was that `getItemValue()` for this value either returns a Boolean or `null`, and not a String. Unless I'm missing something here, there's no way to convert it to the String that Boolean.parseBoolean() wants without doing some extra checks (is it null, if not, make string).

PR #942, opened 2019-06-19
Fix hologram locations for energy regulator and cargo manager

BluGhostYT:


BluGhostYT:
*NOTE*

The hologram for the Nuclear Reactor is fine. All it is, is the coolant percentage used, which does not need to change locations. :)

Issue #899, opened 2019-04-18
Using explosive tools duplicates all Slimefun items.


VersionOmega:
## Description (Required)
When using an explosive tool (any tool that mines more than 1x1x1) near Slimefun items, the items are duplicated.

## Server Log
N/A

## Evidence / Proof (Required)
https://streamable.com/8hqd6
https://streamable.com/d007b
(Note: While this does not show Slimefun tools being used, this glitch is, in fact, present when using the Slimefun explosive pickaxe)

## Environment (Required)
Minecraft 1.13.2
Slimefun Dev - 92
CSCoreLib Dev - 36

mrcoffee1026:
Odd this isn't happening for me.  In fact I'm losing the item and getting a CSCoreLib head INSTEAD of the item, not in addition to.  Also should be noted that the explosive pick doesn't actually "work"?  Like when mining a player_head item it only mines that ONE item, it does not affect the other player_head items in range below it.  Also tried mining my Slimefun electric machines with the explosive pickaxe and was very please to see the pickaxe destroy the floor beneath them while leaving the machines completely untouched.  I'm thinking another plugin may be interacting and allowing the duplication.

PR #943, opened 2019-06-20
Change boolean allow = true to false

BluGhostYT:
The reason for this is because it solves a duplication glitch. Found from issue #899


TheBusyBiscuit:
I can see that this won't compile just by looking at it.
Always pull before you push. You reverted the last few commits.
And you changed materials to an invalid value.

PR #944, opened 2019-06-20
Switch boolean allow = true; to false

BluGhostYT:
The reason for this is because it has created a duplication glitch with the explosive pickaxe. 

*NOTE*

I checked past commits and it seems that nothing has been reverted as far as I can tell. Also apologies for the material mix-up. I was testing it on 1.13 at the time and I know for sure it will work for 1.14


BluGhostYT:
Ah nevermind, I see. The commit to my fork fixed those lines that you were talking about as well.

Issue #953, opened 2019-07-03
Infinite books with /sf guide


Harveyplays02:
## Description (Required)
Players can get infinite books using only /sf guide and a grindstone.

## Steps to reproduce the Issue (Required)
1. Get a slimefun guide with /sf guide
2. Put book inside of grindstone
3. Get book out of grindstone, unenchanted.
4. Book works as normal, such as working in crafting, enchanting tables, etc. 

## Expected behavior (Required)
Book should not be disenchantable.

## Environment (Required)
![image](https://user-images.githubusercontent.com/40532510/60628301-bbc90e00-9de8-11e9-9d0e-40823eb25171.png)

 - Minecraft Version: 1.14.3
 - Slimefun Version: vDEV 123
 - CS-CoreLib Version: vDEV 62

BluGhostYT:
Bump to this issue

PR #957, opened 2019-07-08
Fix for #953, add Slimefun books to anvil and grindstone blacklist.

NathanAdhitya:
1. Added the event for grindstone inventory click to prevent grindstoning any Slimefun Items, including books, (except legacy), #953
2. Added the Slimefun books into blacklist for Anvil, to prevent renaming the books which can be grindstoned without getting filtered.

EDIT:
Uses InventoryType.GRINDSTONE to check. Might break those under 1.14??? Needs testing.


TheBusyBiscuit:
@NathanAdhitya It's for 1.14 only anyway , so backwards compatibility is nothing to worry about.

PR #969, opened 2019-07-24
Dniym patch 1

dniym:
This patch fixes the Incorrect Constructor being called with cargo node whitelists that erroneously creates stone/golden swords in the GUI with a quantity of 20 rather than a damage value of 20.

Addresses this issue:
https://github.com/TheBusyBiscuit/Slimefun4/issues/966


SoSeDiK:
This constructor is deprecated as durability is now part of ItemMeta.
I think you should rather use Damageable meta or maybe create a new proper CustomItem constructor in CS-CoreLib for durability.

dniym:
> 
> 
> This constructor is deprecated as durability is now part of ItemMeta.
> I think you should rather use Damageable meta or maybe create a new proper CustomItem constructor in CS-CoreLib for durability.

I absolutely agree, however I assumed the deprecated constructor was being used for backwards compatibility.    All I did was add the 1 as a quantity, the (byte) 20 was the original code.

SoSeDiK:
This code was written a long time ago when this wasn't a problem. But right now Slimefun supports only 1.14 (not sure about 1.13 but it doesn't matter in this case), so there's no need for the backward compatibility anyway.

dniym:
Can be deleted, updated again using damagable interface rather than deprecated methods.

PR #971, opened 2019-07-24
Update CargoInputNode.java

dniym:
Updated cargo node whitelists to use current (Damageable) objects to set damage instead of using deprecated methods.


SoSeDiK:
You forgot about imports.
*And there's no need to specify the ItemStack amount if it equals to 1 :P*
And you may just use ItemMeta with only `((Damageable) meta).setDamage(value)` cast instead.
Also, you removed the second CustomItem constructor which was used for applying the lore. Right now that line is broken.

Issue #966, opened 2019-07-23
Cargo Input/Output Node whitelist has a non vanilla stack size item in it, due to the quantity being set rather than the durability.


dniym:
## Description (Required)
If you open a cargo output node and click the whitelist, a Stone_Sword shows up with a quantity of 20.   This is an issue because IF a player manages to get this item out due to any inventory related bug they then have a stone_sword with 20x durability.     Likewise this causes an issue with my plugin IllegalStack which monitors inventories for items with quantities over the vanilla stack size.   

In the Source code for SlimeFun in both CargoInputNode and CargoOutputNode.java line #60
new ItemStack(Material.STONE_SWORD, (byte) 20)   
FWIW it should probably utilize the new MaterialData / (Damageable) interface rather than byte data.

Should probably be new ItemStack(Material.STONE_SWORD, 1, (byte) 20)
as The damage value is not supported on the two argument constructor, only three or more.   

## Steps to reproduce the Issue (Required)
1) make cargo input / output node
2) click whitelist
3) see Stone_Sword x 20 appear in the gui.

## Expected behavior (Required)
Should be ONE stone sword with a damage value (if durability is to be considered)

## Server Log / Error Report
No Errors

## Environment (Required)
http://prntscr.com/oipwe2
Note:  /sf Version reports CraftBukkit v1_14_r1 however we are actually running Spigot 1.14.3
This server is running CraftBukkit version git-Spigot-9a643a6-a77ba00 (MC: 1.14.3) (Implementing API version 1.14.3-R0.1-SNAPSHOT)

PR #972, opened 2019-07-24
Fix issue #966

dniym:
Fixed item durability not being applied to whitelist on AdvancedCargoNode as well as CargoInputNode

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -14,7 +14,6 @@ import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.Sound;
-import org.bukkit.Tag;
 import org.bukkit.block.Block;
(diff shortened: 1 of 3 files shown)


src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -718,11 +718,11 @@ public boolean onInteract(Player p, MultiBlock mb, Block b) {
 										for (ItemStack converting: inputs.get(i)) {
 											if (converting != null) {
 												for (int j = 0; j < inv.getContents().length; j++) {
-													if (j == (inv.getContents().length - 1) && !SlimefunManager.isItemSimiliar(converting, inv.getContents()[j], true)) {
```
TheBusyBiscuit: This file has a lot of changes that should be undone immediately.

dniym:
Maybe I read it wrong by it looks like this file is being reverted? Back it’s original state before I added the tags?

TheBusyBiscuit:
@dniym You reverted SlimefunSetup.java to a state from over a month ago or so, yes, you should untouch the file.

PR #975, opened 2019-07-26
Add new fuel's to the bio reactor

dniym:
Updated the bio reactor's fuels to cover all flowers instead of just POPPY and DANDELION, as well as added support for bamboo, kelp and other new plants.


TheBusyBiscuit:
Identical to #972
You pushed all commits to the same branch and submitted the same pull request twice. Try splitting them up.

Will be merged next weekend.

PR #977, opened 2019-07-27
Bio_Reactor fuel update

dniym:
Updates The Bio Reactor to include new minecraft plants


src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -5207,7 +5208,8 @@ public int getSpeed() {
 	}
 	
 	public static void registerPostHandler(PostSlimefunLoadingHandler handler) {
-		MiscSetup.post_handlers.add(handler);
+		MiscSetup.post_handlers.add(
+handler);
```
TheBusyBiscuit: Please remove that line-break over here.

PR #980, opened 2019-08-01
Mob drop chance

dniym:
Adds an optional % chance for a drop to happen, currently only checked when a mob would drop the item.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -67,7 +67,7 @@ public class SlimefunItem {
 	private EnergyTicker energyTicker;
 	private String[] keys = null;
 	private Object[] values = null;
-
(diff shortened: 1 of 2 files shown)


TheBusyBiscuit:
Nice idea, but no Item uses it? Also unsure whether it should go into SlimefunItem.class or a subclass of it.

dniym:
Yeah, nothing uses it just yet, but I’ve got plans for it :D, if the chance is set and it’s a mob drop it’ll only drop on the chance otherwise it assumes 100% to not interfere with existing drops

TheBusyBiscuit:
I know, I still think it could be moved to a subclass though.
And sorry but I am going to hold back on merging this for now until it has a real use.

LinoxGH:
If this was added this could be implemented to Basic Circuit Boards, as I think 100% drop chance is absurd.

TheBusyBiscuit:
I still think it shouldn't be part of the SlimefunItem base class, as 99% of items don't use it.
Best would be an Interface, then you could do an instanceof check and get the chance from there.

PR #983, opened 2019-08-01
Altar item swap exploit fix

dniym:
Currently there is a bug with the dropped items used on an ancient altar, if a player uses a piston, or water to push the item off the pedestal (after starting the ritual), they can then push a less expensive item back on the pedestal and the ritual will complete as if the correct item was used.

Fix saves the items and locations when the ritual is initiated and if a change is detected the ritual is aborted.

Machine to duplicate exploit is here:  
http://prntscr.com/on2fk1

1)Load pedestals with actual materials required for ritual.
2)Load pedestal adjacent to actual pedestal with a dummy item ie dirt.
3)Start ritual
4) activate piston to push off valuable item, in this case Magical Lump
5) activate piston to push on dummy item.
6) ritual completes giving the desired output from the ritual and leaving behind the Magical Lump.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/AncientAltar/RitualAnimation.java
+++ b/src/me/mrCookieSlime/Slimefun/AncientAltar/RitualAnimation.java
@@ -1,13 +1,19 @@
 package me.mrCookieSlime.Slimefun.AncientAltar;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import me.mrCookieSlime.Slimefun.SlimefunStartup;
 import me.mrCookieSlime.Slimefun.listeners.AncientAltarListener;
 import me.mrCookieSlime.Slimefun.Variables;
 
-import org.bukkit.*;
+import org.bukkit.Effect;
+import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.Particle;
+import org.bukkit.Sound;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Item;
 import org.bukkit.inventory.ItemStack;
@@ -23,6 +29,7 @@ public class RitualAnimation implements Runnable {
 	List<ItemStack> items;
 
 	List<Location> particles;
+	Map<Item,Location> itemLock = new HashMap<>();
 
 	boolean running;
 	int stage;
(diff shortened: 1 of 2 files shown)


src/me/mrCookieSlime/Slimefun/AncientAltar/RitualAnimation.java
```diff
@@ -82,7 +109,11 @@ private void checkPedestal(Block pedestal) {
 				e.printStackTrace();
 			}
 
+			
+			System.out.println("UnLocking item: " + item.getCustomName() + " / " + item.getItemStack().getType().name());
```
TheBusyBiscuit: Pretty sure that this debug message should be removed.

src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -3611,30 +3612,30 @@ public void registerDefaultRecipes() {
...
 				registerFuel(new MachineFuel(2, new ItemStack(Material.CACTUS)));
 				registerFuel(new MachineFuel(2, new ItemStack(Material.LILY_PAD)));
 				registerFuel(new MachineFuel(8, new ItemStack(Material.CHORUS_FRUIT)));
-
+				registerFuel(new MachineFuel(1, new ItemStack(Material.BAMBOO)));
+				registerFuel(new MachineFuel(1, new ItemStack(Material.KELP)));
+				registerFuel(new MachineFuel(2, new ItemStack(Material.DRIED_KELP)));
+				registerFuel(new MachineFuel(20, new ItemStack(Material.DRIED_KELP_BLOCK)));
+				registerFuel(new MachineFuel(1, new ItemStack(Material.SEAGRASS)));
+				registerFuel(new MachineFuel(2, new ItemStack(Material.SEA_PICKLE)));
```
TheBusyBiscuit: This change belongs to a different PR, so it should be excluded from this one.

src/me/mrCookieSlime/Slimefun/AncientAltar/RitualAnimation.java
dniym: Doh yeah, I’ll fix that

src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
dniym: My bad, though for sure I added it as a new branch will fix shortly

PR #984, opened 2019-08-02
Add copper wire

NihilistBrew:
**Description**
Added copper wire, a crafting component made from copper ingots.

3 copper ingots --> 8 copper wires

This pull request changes the recipe of electric motor and heating coil to use this component instead.
Instead of using 6 and 8 ingots in their recipe, they now use 6 and 8 copper wires, respectively.

Copper wire takes 15 levels to research, and is in the category TECH_MISC (Technical Components)

**Why?**
Slimefun is notorious for having draining copper reserves, more than any other dust. This change aims to rebalance dust usage a bit. Copper will still be a highly sought-after dust, but not as extremely unbalanced.

**What's next?**
By not making copper the bottleneck to literally everything, this paves way for other dusts to be used in recipes. It doesn't make sense that copper reserves are constantly empty while abundant dusts listed below always hoard several double chests worth of dust storage, never getting used up.

- Magnesium (By definition useless unless you use SlimeXpansion, which still makes it niche)
- Lead 
- Aluminium
- Tin
- Zinc

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/ResearchSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/ResearchSetup.java
@@ -234,7 +234,7 @@ public class ResearchSetup {
 	    Slimefun.registerResearch(new Research(235, "Nether Ice Coolant", 45), SlimefunItems.NETHER_ICE, SlimefunItems.ENRICHED_NETHER_ICE, SlimefunItems.NETHER_ICE_COOLANT_CELL, SlimefunItems.NETHER_DRILL);
 	    Slimefun.registerResearch(new Research(236, "Nether Star Reactor", 60), SlimefunItems.NETHERSTAR_REACTOR);
 	    Slimefun.registerResearch(new Research(237, "Blistering Radioactivity", 38), SlimefunItems.BLISTERING_INGOT, SlimefunItems.BLISTERING_INGOT_2, SlimefunItems.BLISTERING_INGOT_3);
-		Slimefun.registerResearch(new Research(239, "Automatic Ignition Chamber", 12), SlimefunItems.IGNITION_CHAMBER);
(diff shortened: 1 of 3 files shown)


src/me/mrCookieSlime/Slimefun/Setup/ResearchSetup.java
```diff
@@ -234,6 +234,7 @@ public static void setupResearches() {
 	    Slimefun.registerResearch(new Research(235, "Nether Ice Coolant", 45), SlimefunItems.NETHER_ICE, SlimefunItems.ENRICHED_NETHER_ICE, SlimefunItems.NETHER_ICE_COOLANT_CELL, SlimefunItems.NETHER_DRILL);
 	    Slimefun.registerResearch(new Research(236, "Nether Star Reactor", 60), SlimefunItems.NETHERSTAR_REACTOR);
 	    Slimefun.registerResearch(new Research(237, "Blistering Radioactivity", 38), SlimefunItems.BLISTERING_INGOT, SlimefunItems.BLISTERING_INGOT_2, SlimefunItems.BLISTERING_INGOT_3);
-	    Slimefun.registerResearch(new Research(239, "Automatic Ignition Chamber", 12), SlimefunItems.IGNITION_CHAMBER);
+		Slimefun.registerResearch(new Research(239, "Automatic Ignition Chamber", 12), SlimefunItems.IGNITION_CHAMBER);
+		Slimefun.registerResearch(new Research(241, "Thinned-down Conductivity", 15), SlimefunItems.COPPER_WIRE); // Skips 1 to make room for eventual output_chest (see pull request)
```
TheBusyBiscuit: The indentation is kinda messed up over here.

NihilistBrew:
There! Should be all fixed!

I also took some liberty and fixed an indent at the beginning.

PR #985, opened 2019-08-02
Add sandstone to grind stone

NihilistBrew:
Added sandstone and red sandstone to grindstone.
1 (red) sandstone --> 4 (red) sand

Idea came from Jareth Maxter#4344

> Add Sandstone to Sand option for Grindstone - gives people a reason to pickaxe down a desert and increase the sand resource on servers

This suggestion were upvoted 5:0 suggesting that it's a desired feature by the community. 

NOTE: Due to the limitation of the sf guide, the red sandstone recipe doesn't fit in the /sf guide view.
It works, but just isn't visually shown. For this recipe, it doesn't really matter due to red sandstone simply being a variation of sandstone, and you could argue that they don't need 2 seperate recipe slots. The user can simply experiment and find that the red sandstone works just as well.

Maybe modifying the sf guide to be able to show several recipes would be an idea for a future pull request, but it's out of the scope for this current PR.


TheBusyBiscuit:
Yep, optimally the machine recipes should be paginated, like having an arrow on the left and right side.

PR #988, opened 2019-08-03
Fix Altar Item Swap Exploit

dniym:
Fix item swap exploit with AncientAltar Ritual

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/AncientAltar/RitualAnimation.java
+++ b/src/me/mrCookieSlime/Slimefun/AncientAltar/RitualAnimation.java
@@ -54,8 +54,10 @@ public class RitualAnimation implements Runnable {
 	@Override
 	public void run() {
 		idle();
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/AncientAltar/RitualAnimation.java
```diff
@@ -38,11 +45,17 @@ public RitualAnimation(List<Block> altars, Block altar, Location drop, ItemStack
...
+			Item itm = AncientAltarListener.findItem(ped);
+			this.itemLock.put(itm, itm.getLocation().clone());
+		}
 	}
 
 	@Override
 	public void run() {
 		idle();
+		if(!checkLockedItems())
+			abort();
```
TheBusyBiscuit: You should probably also add a return; over here, since we don't want to continue anyway.

PR #989, opened 2019-08-03
Fix spawner dupe and objectified e.getBock()

NihilistBrew:
A)
Fixed spawner dupe.

Reproducing:

A) Pickaxe of Containment
B) Try to break a reinforced spawner
C) You will get a reinforced spawner and a broken spawner.

Fix:

Add the if-statement which basically boils down to if (BlockStorage.hasBlockInfo(b)) return true;
This will nullify the event. If the if-statement resolves to true then it will is not a vanilla spawner and
the containment pick should therefore not work.

The actual implementation is if (b.getType() != Material.SPAWNER || BlockStorage.hasBlockInfo(b)) return true; and is just a way to bake in the relatively simple boolean into a similar check.

B)
Added some quick refactoring, so instead of doing,
foo(e.getBlock);
you now do,
Block b = e.getBlock();
such that the call is foo(b);

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -1840,7 +1840,6 @@ public class SlimefunSetup {
 					Block b = e.getBlock(); // Refactored it into this so we don't need to call e.getBlock() all the time.
 					if (b.getType() != Material.SPAWNER || BlockStorage.hasBlockInfo(b)) return true; 
 					// If the spawner's BlockStorage has BlockInfo, then it's not a vanilla spawner and shouldn't give a broken spawner.
-					BlockStorage.clearBlockInfo(b);
 					ItemStack spawner = SlimefunItems.BROKEN_SPAWNER.clone();
 					ItemMeta im = spawner.getItemMeta();
 					List<String> lore = im.getLore();
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -1837,17 +1837,19 @@ public class SlimefunSetup {
 			@Override
 			public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List<ItemStack> drops) {
 				if (SlimefunManager.isItemSimiliar(item, SlimefunItems.PICKAXE_OF_CONTAINMENT, true)) {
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -1837,17 +1837,19 @@ public boolean onRightClick(ItemUseEvent e, Player p, ItemStack item) {
 			@Override
 			public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List<ItemStack> drops) {
 				if (SlimefunManager.isItemSimiliar(item, SlimefunItems.PICKAXE_OF_CONTAINMENT, true)) {
-					if (e.getBlock().getType() != Material.SPAWNER) return true;
-					BlockStorage.clearBlockInfo(e.getBlock());
+					Block b = e.getBlock(); // Refactored it into this so we don't need to call e.getBlock() all the time.
+					if (b.getType() != Material.SPAWNER || BlockStorage.hasBlockInfo(b)) return true; 
+					// If the spawner's BlockStorage has BlockInfo, then it's not a vanilla spawner and shouldn't give a broken spawner.
+					BlockStorage.clearBlockInfo(b);
```
TheBusyBiscuit: Ehm, this is quite redundant if you already return when there is BlockInfo.

NihilistBrew: But wouldn't it go through all calls with all sf blocks then?

TheBusyBiscuit: What?

NihilistBrew: I assume the b.getType() != Material.SPAWNER is a check to see if a spawner is broken.
If I remove that and only use the check that I added, BlockStorage.hasBlockInfo(b), then wouldn't it go through with the event call with all non-sf items?

TheBusyBiscuit: != Material.SPAWNER   makes sure that the block is a spawner.

But this is now what I am commenting on, I made the comment on line 1843.
Clearing the block info is redundant after you checked that there is no block info anyway.

NihilistBrew: oh now I understand what you meant. Lol.

NihilistBrew: Fixed it

PR #992, opened 2019-08-03
Duplication glitch fix for ChestTerminals & Androids

BluGhostYT:
Resolves issue 18 from the ChestTerminal repository.

Issue occurred before this fix: The Miner Android, when programmed properly would mine the CT Terminal, and duplicate items within it. This is no longer possible due to the blacklisting of Player Heads & their walls.


TheBusyBiscuit:
While this fix works, it probably still works for other Slimefun machines then that are not heads.
I think it'd be a better approach to just prevent breaking Slimefun Blocks instead.

BluGhostYT:
> While this fix works, it probably still works for other Slimefun machines then that are not heads.
> I think it'd be a better approach to just prevent breaking Slimefun Blocks instead.

Understandable

src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
```diff
@@ -713,14 +714,18 @@ private void mine(Block b, Block block) {
...
+			boolean allow = false;
 			if (item != null) {
 				if (fits(b, item.getItem())) {
 					if (SlimefunItem.blockhandler.containsKey(item.getID())) {
-						if (SlimefunItem.blockhandler.get(item.getID()).onBreak(null, block, item, UnregisterReason.ANDROID_DIG)) {
-							pushItems(b, BlockStorage.retrieve(block));
-							if (SlimefunItem.blockhandler.containsKey(item.getID())) SlimefunItem.blockhandler.get(item.getID()).onBreak(null, block, item, UnregisterReason.ANDROID_DIG);
-							block.getWorld().playEffect(block.getLocation(), Effect.STEP_SOUND, block.getType());
-							block.setType(Material.AIR);
+						if (allow) {
```
TheBusyBiscuit: "allow" is never changed, this will always be false.

BluGhostYT: Though it remained false, it still prevented the items from being destroyed. The only items it would destroy are vanilla items. I understand that you wish for me to try another way of implementing this fix, but that is what worked out for me.

BluGhostYT: And there we go, that should do it.

src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
```diff
@@ -215,6 +216,8 @@ public void onPlace(Player p, Block b, SlimefunItem item) {
 			public boolean onBreak(Player p, Block b, SlimefunItem item, UnregisterReason reason) {
 				boolean allow =  reason.equals(UnregisterReason.PLAYER_BREAK) && (BlockStorage.getLocationInfo(b.getLocation(), "owner").equals(p.getUniqueId().toString()) || p.hasPermission("slimefun.android.bypass"));
 
+				if (BlockStorage.check(b) != null) return false;
```
TheBusyBiscuit: You should rather use BlockStorage.hasBlockInfo() for this.

BluGhostYT: So it would be 

if (BlockStorage.hasBlockInfo() != null) return false;

Correct? Or would it be this: if (BlockStorage.hasBlockInfo(b)) return false;?

I just woke up, so my apologies if I am incorrect

TheBusyBiscuit: It returns a boolean, so a null-check would make no sense.

PR #993, opened 2019-08-04
Some small fixes

huynhqtienvtag:
* Explosive Pickaxe will ignore liquid blocks (water, lava, etc)
* Smelter's Pickaxe will ignore Slimefun Blocks (fix duplication)

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
@@ -1536,7 +1536,7 @@ public class SlimefunSetup {
 			@Override
 			public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List<ItemStack> drops) {
 				if (SlimefunManager.isItemSimiliar(item, SlimefunItems.AUTO_SMELT_PICKAXE, true)) {
-					if (BlockStorage.check(e.getBlock()) != null) return false;
+					if (BlockStorage.hasBlockInfo(e.getBlock())) return false;
 					if (e.getBlock().getType().equals(Material.PLAYER_HEAD)) return true;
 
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -1536,6 +1536,7 @@ else if (SlimefunManager.isItemSimiliar(current, SlimefunItems.PULVERIZED_ORE, t
 			@Override
 			public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List<ItemStack> drops) {
 				if (SlimefunManager.isItemSimiliar(item, SlimefunItems.AUTO_SMELT_PICKAXE, true)) {
+					if (BlockStorage.check(e.getBlock()) != null) return false;
```
NihilistBrew: You should use `if (BlockStorage.hasBlockInfo(e.getBlock())) return false;`. It's what the method is for.

src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -1536,6 +1536,7 @@ else if (SlimefunManager.isItemSimiliar(current, SlimefunItems.PULVERIZED_ORE, t
 			@Override
 			public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List<ItemStack> drops) {
 				if (SlimefunManager.isItemSimiliar(item, SlimefunItems.AUTO_SMELT_PICKAXE, true)) {
+					if (BlockStorage.hasBlockInfo(e.getBlock())) return false;
```
TheBusyBiscuit: This should return true.

true means that this Event was handled and should not be passed onto other Items.

PR #995, opened 2019-08-05
Reactor fix

dniym:
Overhals The reactor and reactor port to be a bit more user friendly and useful.   

Currently there is no way to see the status of a reactor via the access port, this PR makes the reactor port a bit more useful as a remote tool.    

Once a reactor detects an attached Access Port clicking on the access port will show the GUI of the active reactor,giving the user the option to adjust the focus of the reactor as well as see remaining coolant.  

An information icon has been added to the Reactor's GUI which shows if an access port is connected or not with some info about how to connect one if none is found..   Provided a port is found the user can click on the icon to open the Access Port's gui if so desired,  OR they can shift click the access port to access it rather than the reactor's gui.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
@@ -46,14 +46,14 @@ public abstract class AReactor extends SlimefunItem {
 
 	private static final BlockFace[] cooling =
 		{
-			BlockFace.NORTH,
-			BlockFace.NORTH_EAST,
-			BlockFace.EAST,
-			BlockFace.SOUTH_EAST,
-			BlockFace.SOUTH,
-			BlockFace.SOUTH_WEST,
-			BlockFace.WEST,
-			BlockFace.NORTH_WEST
+				BlockFace.NORTH,
+				BlockFace.NORTH_EAST,
+				BlockFace.EAST,
+				BlockFace.SOUTH_EAST,
+				BlockFace.SOUTH,
+				BlockFace.SOUTH_WEST,
+				BlockFace.WEST,
+				BlockFace.NORTH_WEST
 		};
 
 	private Set<MachineFuel> recipes = new HashSet<MachineFuel>();
@@ -101,9 +101,9 @@ public abstract class AReactor extends SlimefunItem {
 					if(ap != null) {
 						menu.replaceExistingItem(infoSlot, new CustomItem(new ItemStack(Material.GREEN_WOOL), "&7Access Port", "", "&6Detected", "", "&7> Click to view Access Port"));
 						menu.addMenuClickHandler(infoSlot, (p, slot, item, action) -> {
-						ap.open(p);
-						newInstance(menu, b);
-							
+							ap.open(p);
+							newInstance(menu, b);
+
 							return false;
 						});
 					} else {
@@ -114,12 +114,12 @@ public abstract class AReactor extends SlimefunItem {
 							return false;
 						});
 					}
-					
+
 				} catch(Exception x) {
 				}
 			}
 
-		
+
 			@Override
 			public boolean canOpen(Block b, Player p) {
 				return p.hasPermission("slimefun.inventory.bypass") || CSCoreLib.getLib().getProtectionManager().canAccessChest(p.getUniqueId(), b, true);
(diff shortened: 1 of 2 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -1,192 +1,233 @@
...
+	
+		return null;
+	}
+	
+	public BlockMenu getReactorMenu(Location l) {
+		Location reactorL = new Location(l.getWorld(), l.getX(), l.getY() - 3, l.getZ());
+		
+		if (BlockStorage.check(reactorL, "NUCLEAR_REACTOR")) return BlockStorage.getInventory(reactorL);
+		
+		if (BlockStorage.check(reactorL, "NETHERSTAR_REACTOR")) return BlockStorage.getInventory(reactorL);
```
TheBusyBiscuit: You should rather just store the ID once and then compare it, otherwise you are making two heavy calls after another.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -1,192 +1,233 @@
...
+					BlockMenu bm = getReactorMenu(b.getLocation());
+					if(bm != null) {
+						for(int slot:reactor.getCoolantSlots())
+							if(bm.getItemInSlot(slot) != null)
+								empty = false;
+						for(int slot:reactor.getFuelSlots())
+							if(bm.getItemInSlot(slot) != null)
+								empty = false;
+					
+						if(!p.isSneaking() && !empty && (p.hasPermission("slimefun.inventory.bypass") || CSCoreLib.getLib().getProtectionManager().canAccessChest(p.getUniqueId(),b,true))) {
```
TheBusyBiscuit: So Sneaking = Open the actual Inventory, Not sneaking = Open the Reactor Menu?
I think this should be swapped around.

TheBusyBiscuit: You should probably move the sneaking and permissions check to the top, otherwise you wasted a lot of calculations to end up doing nothing.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
```diff
@@ -1,454 +1,475 @@
...
+					if(ap != null) {
+						menu.replaceExistingItem(infoSlot, new CustomItem(new ItemStack(Material.GREEN_WOOL), "&7Access Port", "", "&6Detected", "", "&7> Click to open Access Port"));
+						menu.addMenuClickHandler(infoSlot, (p, slot, item, action) -> {
+						ap.open(p);
+						newInstance(menu, b);
+							
+							return false;
+						});
+					} else {
+						menu.replaceExistingItem(infoSlot, new CustomItem(new ItemStack(Material.RED_WOOL), "&7Access Port", "", "&cNot Detected", "", "&7Access Port Must Be", "&7Placed 3 Blocks Above", "&7Reactor!"));
```
TheBusyBiscuit: Don't capitalise every letter please.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
dniym: Well the reactor port almost always is empty as the items are quickly moved from the port to the reactor.    The reactor is the part that blows up if not taken care of properly so this is the reason for that functionality,       Since its the "REACTOR" access port I feel like it makes the most sense to open the Reactor's inventory first as that is the whole reason for the port's existence.

dniym: I can move the permissions check but the isSneaking check needs to stay as it's optional and only relevant if a reactor exists, it should always open the port's inventory if there is no connected reactor.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
```diff
@@ -1,454 +1,475 @@
...
+					}
+					BlockMenu ap = getAccessPort(b.getLocation());
+					if(ap != null) {
+						menu.replaceExistingItem(infoSlot, new CustomItem(new ItemStack(Material.GREEN_WOOL), "&7Access Port", "", "&6Detected", "", "&7> Click to view Access Port"));
+						menu.addMenuClickHandler(infoSlot, (p, slot, item, action) -> {
+						ap.open(p);
+						newInstance(menu, b);
+							
+							return false;
+						});
```
TheBusyBiscuit: Please fix the formatting on the above lines.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -1,192 +1,240 @@
...
+	
+		return null;
+	}
+	
+	public BlockMenu getReactorMenu(Location l) {
+		Location reactorL = new Location(l.getWorld(), l.getX(), l.getY() - 3, l.getZ());
+		
+		SlimefunItem item = BlockStorage.check(reactorL);
+		
+		if(item != null && (item.getID().equals("NUCLEAR_REACTOR") || item.getID().equals("NETHERSTAR_REACTOR")))
```
TheBusyBiscuit: You don't need "item", you should use BlockStorage.checkID instead, parsing it as a SlimefunItem is a heavy call and not needed here, since you are only comparing the ID anyway.

dniym: Updated to use BlockStorage.checkID

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AReactor.java
dniym: Should be fixed now, reformatted to correct indentation

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -1,192 +1,239 @@
...
+		if(item instanceof AReactor)
+			return (AReactor) item;
+	
+		return null;
+	}
+	
+	public BlockMenu getReactorMenu(Location l) {
+		Location reactorL = new Location(l.getWorld(), l.getX(), l.getY() - 3, l.getZ());
+		
+		if(BlockStorage.checkID(reactorL).equals("NUCLEAR_REACTOR") || BlockStorage.checkID(reactorL).equals("NETHERSTAR_REACTOR"))
```
TheBusyBiscuit: You are still executing the checkID method twice...
Use a variable.

dniym: Updated

TheBusyBiscuit:
Approved.
But I will hold back on merging this until a fix for #1009 was made.
Since #1009 is a compilation error, merging this would just result in another failed build.

PR #996, opened 2019-08-05
Fix for energy generators suddenly losing power

Firebuggie:
Apparently, energy generators can lose their charge when either fully charged or using up their current fuel.
This PR fixes that.

Tested on CraftBukkit v1_14_R1.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AGenerator.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AGenerator.java
@@ -253,9 +253,8 @@ public abstract class AGenerator extends SlimefunItem {
 							if (ChargableBlock.getMaxCharge(l) - ChargableBlock.getCharge(l) >= getEnergyProduction()) {
 								ChargableBlock.addCharge(l, getEnergyProduction());
 								progress.put(l, timeleft - 1);
-								return ChargableBlock.getCharge(l);
 							}
-							return 0;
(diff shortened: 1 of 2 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AGenerator.java
```diff
@@ -273,7 +272,7 @@ public double generateEnergy(Location l, SlimefunItem sf, Config data) {
 						
 						progress.remove(l);
 						processing.remove(l);
-						return 0;
+						return ChargableBlock.getCharge(l);
```
TheBusyBiscuit: This should check "ChargableBlock.isChargable(l)" before-hand, otherwise we can run into issues.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AGenerator.java
```diff
@@ -297,7 +296,7 @@ public double generateEnergy(Location l, SlimefunItem sf, Config data) {
 						processing.put(l, r);
 						progress.put(l, r.getTicks());
 					}
-					return 0;
+					return ChargableBlock.getCharge(l);
```
TheBusyBiscuit: This should check "ChargableBlock.isChargable(l)" before-hand, otherwise we can run into issues.

TheBusyBiscuit:
Closed due to merge conflicts.

PR #1023, opened 2019-08-20
Adding Explosive Shovel.

ajan-12:
!!! This addition needs my PR on CS-CoreLib to be accepted as it has an addition needed for Explosive Shovel to be working. !!!
!!! If you accept this PR before the PR made to CS-CoreLib, the build won't work. !!!

- Renaming "Explosive Pickaxe" research to "Explosive Tools".
- Adding a new tool called Explosive Shovel.
  - Is able to break blocks 3x3 similar to Explosive Pickaxe.
  - Can only break diggable blocks(blocks that have shovels as their main tools).

- Adding Explosive Shovel to the "Explosive Tools" research.
- Adding Explosive Shovel to item setup.
  - Used MaterialHelper#isDiggable() in here.
  - Is only able to break diggable blocks:
    - Dirt, Coarse Dirt, Farmland, Podzol, Mycelium, Gravel, Grass Block, Grass Path, Sand, Red Sand, Soul Sand, Snow, Snow Block and Clay.


TheBusyBiscuit:
First of all, I am not going to merge something that requires a CS-CoreLib update.
Most people do not use the dev version and therefore will encounter lots of errors suddenly.
So you would have to wait until the next major CS-CoreLib release was published and a reasonable number of people have been updated.

You also need to update the pom accordingly then, which you cannot atm, making this Pull Request just not feasible.

Two options:
1. Wait a month or two to get this implemented
2. You could also do an implementation of your method for CS-CoreLib2 (which is shaded and therefore is instantly available without having to wait for any dependencies. https://github.com/TheBusyBiscuit/CS-CoreLib2
But since this is supposed to be very lightweight and well-designed, I advise you to fancy it up a bit.
I can give you a class that lays the groundwork and send you a link.

TheBusyBiscuit:
Closed due to merge conflicts and for the above mentioned reasons.

PR #1024, opened 2019-08-21
Fixes inability to open reactor access port via right click

dniym:
fixed a bug that would prevent the Reactor Access Port from being opened via directly right clicking the port.   Currently the only way to open it is via the reactor, which was un-intended.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
@@ -26,7 +26,7 @@ import me.mrCookieSlime.Slimefun.api.inventory.BlockMenuPreset;
 import me.mrCookieSlime.Slimefun.api.item_transport.ItemTransportFlow;
 
 public class ReactorAccessPort extends SlimefunItem {
-	
+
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -26,64 +26,73 @@
...
-			
+
 			@Override
 			public void init() {
 				constructMenu(this);
 			}
 
 			@Override
 			public void newInstance(BlockMenu menu, Block b) {
+				System.out.println("New menu instance?");
```
TheBusyBiscuit: Debug message that probably should have been removed...

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -26,64 +26,73 @@
...
+				try {
+					if (BlockStorage.getLocationInfo(b.getLocation(), "reactor-port") == null){
+						BlockStorage.addBlockInfo(b, "reactor-port", "true");
+					}
+				} catch(Exception x) {
+					System.out.println("exception caught: " + x.getLocalizedMessage());
```
TheBusyBiscuit: Is the try/catch really necessary there...?

PR #1026, opened 2019-08-22
Fixed Reactor Access Port being unable to be opened

dniym:


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -26,64 +26,69 @@
...
 			@Override
 			public void init() {
 				constructMenu(this);
 			}
 
 			@Override
 			public void newInstance(BlockMenu menu, Block b) {
+				
+					if (BlockStorage.getLocationInfo(b.getLocation(), "reactor-port") == null){
+						BlockStorage.addBlockInfo(b, "reactor-port", "true");
```
TheBusyBiscuit: Why exactly is this tag addded to the block btw?

dniym: Not exactly sure why I did that, must have tried as the port wouldn't open previously and forgotten to remove it.   Updated now should be good to go

PR #1028, opened 2019-08-22
Droid fix2

### LINKED ISSUE #997: Andriods cannot be broken.
## Description (Required)

Any type of andriods cannot be removed.

Place any type of Andriod (Normal, Advanced, Empowered) and try removing it.

The andriod being removed and its item being returned.

## Server Log / Error Report

No errors.

dniym:
Fixes #997


src/me/mrCookieSlime/Slimefun/Android/ProgrammableAndroid.java
```diff
@@ -1,1433 +1,1433 @@
...
+				if (!open) {
+					Messages.local.sendTranslation(p, "inventory.no-access", true);
+				}
+				return open;
+			}
+
+			@Override
+			public void newInstance(BlockMenu menu, final Block b) {
+				try {
+					menu.replaceExistingItem(15, new
```
TheBusyBiscuit: This check might as well be removed. It shouldn't even be here, must have slipped through my code review in an earlier PR.
You can remove this line I think.

dniym: Yeah I was debating removing it because i didn’t think the code would ever even get that far if it wasn’t in block storage.

PR #1030, opened 2019-08-23
Fix pickaxe of containment and repaired spawners

dniym:
Pickaxe of Containment can now mine reinforced spawners.   Previously mining them would give you a reinforced spawner without saving the entity type so that it would default to pigs.   Also discovered that a diamond pickaxe could be used to mine regular spawners, this has been corrected and a diamond pick will now destroy a spawner (vanilla and reinforced) and drop xp.   Only the pickaxe of containment can be used to pick up spawners.


TheBusyBiscuit:
This PR has merge conflicts.

TheBusyBiscuit:
Obviously not. 
You have to rebase your branch and perform the change again.

PR #1033, opened 2019-08-24
Fix busted spawners

dniym:
Friggin finally, stupid maven.


src/me/mrCookieSlime/Slimefun/Setup/SlimefunSetup.java
```diff
@@ -2702,19 +2708,6 @@ public boolean onBlockPlace(BlockPlaceEvent e, ItemStack item) {
 				}
 				else return false;
 			}
-		}, new BlockBreakHandler() {
```
TheBusyBiscuit: Why is this part of the code removed?

dniym: Because it’s handled by the Containment Pickaxe handler like the Broken_Spawner is, didn’t see any reason to have redundant code for setting the type when it could be handled in one spot.

PR #1037, opened 2019-08-25
Update lore on Refinery and Access Port

dniym:
Updated lore on access port to better explain its new use.
Updated lore on Refinery to list it's energy consumption.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
+++ b/src/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
@@ -533,7 +533,7 @@ public class SlimefunItems {
 	public static ItemStack BUCKET_OF_FUEL = null;
 	public static ItemStack OIL_PUMP = null;
 
-	public static ItemStack REFINERY = new CustomItem(Material.PISTON, "&cRefinery", "", "&rRefines Oil to create Fuel");
+	public static ItemStack REFINERY = new CustomItem(Material.PISTON, "&cRefinery", "", "&rRefines Oil to create Fuel", "&8\u21E8 &7Speed: 1x", "&8\u21E8 &e\u26A1 &720 J/s");
 	public static ItemStack COMBUSTION_REACTOR = null;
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
```diff
@@ -1,798 +1,798 @@
...
-	public static ItemStack NETHER_DRILL = new CustomItem(Material.RED_TERRACOTTA, "&4Nether Drill", "", "&rAllows you to mine Nether Ice", "", "&4End-Game Machine", "&8\u21E8 &7Speed: 1x", "&8\u21E8 &e\u26A1 &7102 J/s", "", "&c&l! &cCan only be used in the Nether!", "&c&l! &cMake sure to Geo-Scan the Chunk first");
-	
-	// Cargo
-	public static ItemStack CARGO_MANAGER = null;
-	public static ItemStack CARGO_NODE = null;
-	public static ItemStack CARGO_INPUT = null;
-	public static ItemStack CARGO_OUTPUT = null;
-	public static ItemStack CARGO_OUTPUT_ADVANCED = null;
-
-	public static ItemStack AUTO_BREEDER = new CustomItem(Material.HAY_BLOCK, "&eAuto-Breeder", "", "&rRuns on &aOrganic Food", "", "&4End-Game Machine", "&8\u21E8 &e\u26A1 &71024 J B
```
TheBusyBiscuit: "reactor" is written with a capitalized R throughout the entire lore, I think we should stay consistent and do one or the other.

BTW changing the lore will of course break any Items that Players currently own.

dniym: Ah crap, that’s mildly annoying.  So only way to change it would be to add a new version and hide the old one?

TheBusyBiscuit: Well, adding a new version for these blocks is unneccessary, I am afraid you will have to accept the prior lore.

PR #1060, opened 2019-08-29
Bigger code cleaning

Sfiguz7:
- Merged if nests where collapsible

PR #1061, opened 2019-08-29
Sfiguz7/fix code smells

Sfiguz7:


TheBusyBiscuit:
This is already part of #1060

PR #1076, opened 2019-09-01
Fixes #253

### LINKED ISSUE #253: Grappling Hook Stops Working When Fired Into Void
## Description
When using the Grappling Hook, if a player fires it into the Void, the player is then unable to use the Grappling Hook again until the server is restarted.

## Server Log
No errors

## Evidence / Proof
GIF: https://i.gyazo.com/7237c15ba97ce133227e7f3282eccae7.gif
You can't really see it in the GIF, but I was unable to fire the Grappling Hook again after that until the server was restarted.

## Environment
MC: 1.11.2 git-Paper-1072
CS-CoreLib: 1.5.12
Slimefun: 4.1.2

LinoxGH:
## Description

Cleaned up the code of grappling hooks.

## Changes

- Now after a minute(configurable) a grappling hook is used, plugin removes all instances to it.

## Related Issues

Resolves #253 

## Testability

[X] I have fully tested the proposed changes and promise that they will not break everything into chaos.
[ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/GrapplingHook.java
```diff
@@ -19,46 +19,68 @@
...
+                                if (n.isValid()) n.remove();
+                            }
 
+                            Bukkit.getScheduler().scheduleSyncDelayedTask(SlimefunPlugin.instance, () -> {
+                                utilities.damage.remove(uuid);
+                                utilities.jumpState.remove(uuid);
+                                utilities.remove.remove(uuid);
+                            }, 20L);
+                        }
+                    }, (int) Slimefun.getItemValue("GRAPPLING_HOOK", "despawn-seconds") * 20);
```
TheBusyBiscuit: You should save this setting as a field instead of grabbing it from the file everytime.

Note that you need to set the field AFTER register(boolean) was called.
Easiest approach would be to override register, call super.register and then set the field

LinoxGH: or adding a postRegister() method to SlimefunItem?

TheBusyBiscuit: There already is, I think it's install() or create()

You need to look up which one it is.
They should be renamed to pre and post register though at some point.

LinoxGH:
Did the requested changes.

PR #1078, opened 2019-09-02
Added item-permission in messages

xH3LLRAIZ3Rx:
## Description
I added a item-permission message to Messages and implemented it to the SlimefunGuide. I am not great at java so if you need to make it more efficient by all means, I don't quite know String Lists but it what I did does work within a limit.

## Changes
Added item-permission into Messages class
Added item-permission getTranslation to SlimefunGuide

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Setup/Messages.java
+++ b/src/me/mrCookieSlime/Slimefun/Setup/Messages.java
@@ -30,7 +30,6 @@ public final class Messages {
 		local.setDefault("messages.fortune-cookie", "&7Help me, I am trapped in a Fortune Cookie Factory!", "&7You will die tomorrow...     by a Creeper", "&7At some Point in your Life something bad will happen!!!", "&7Next week you will notice that this is not the real World, you are in a kind of \"Matrix\" or lets call it Computer game. Yes you are in a Computer Game", "&7This Cookie will taste good in a few seconds", "&7You will die soon and the last word you will hear is gonna be \"EXTERMINATE!!!\"", "&7Whatever you do, do not hug a Creeper... I tried it. It feels good, but it's not worth it.");
 		local.setDefault("messages.only-players", "&4This Command is only for Players");
 		local.setDefault("messages.no-permission", "&4You do not have the required Permission to do this");
-		local.setDefault("messages.item-permission", "&rYou do not have Permission", "&rto access this Item");
 		local.setDefault("messages.usage", "&4Usage: &c%usage%");
 		local.setDefault("messages.not-online", "&4%player% &cis not online!");
 		local.setDefault("messages.not-valid-item", "&4%item% &cis not a valid Item!");
@@ -67,6 +66,8 @@ public final class Messages {
 		local.setDefault("messages.fire-extinguish", "&7You have extinguished yourself");
 		local.setDefault("messages.cannot-place" ,"&cYou cannot place that block there!");
 		local.setDefault("messages.no-pvp" ,"&cYou cannot pvp in here!");
+
+		local.setDefault("tooltips.item-permission", "&rYou do not have Permission", "&rto access this Item");
 
 		local.setDefault("machines.pattern-not-found", "&eSorry, I could not recognize this Pattern. Please place the Items in the correct Pattern into the Dispenser.");
 		local.setDefault("machines.unknown-material", "&eSorry, I could not recognize the Item in my Dispenser. Please put something in that I know.");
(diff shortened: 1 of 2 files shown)


src/me/mrCookieSlime/Slimefun/Setup/Messages.java
```diff
@@ -30,6 +30,7 @@ public static void setup() {
 		local.setDefault("messages.fortune-cookie", "&7Help me, I am trapped in a Fortune Cookie Factory!", "&7You will die tomorrow...     by a Creeper", "&7At some Point in your Life something bad will happen!!!", "&7Next week you will notice that this is not the real World, you are in a kind of \"Matrix\" or lets call it Computer game. Yes you are in a Computer Game", "&7This Cookie will taste good in a few seconds", "&7You will die soon and the last word you will hear is gonna be \"EXTERMINATE!!!\"", "&7Whatever you do, do not hug a Creeper... I tried it. It feels good, but it's not worth it.");
 		local.setDefault("messages.only-players", "&4This Command is only for Players");
 		local.setDefault("messages.no-permission", "&4You do not have the required Permission to do this");
+		local.setDefault("messages.item-permission", "&rYou do not have Permission", "&rto access this Item");
```
TheBusyBiscuit: This is not a message and therefore shouldn't go into that namespace.
I'd suggest the prefix "tooltips." instead.

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -716,7 +716,11 @@ public void run(final Player p) {
 							index++;
 						}
 						else {
-							menu.addItem(index, new CustomItem(Material.BARRIER, StringUtils.formatItemName(sfitem.getItem(), false), "", "&rYou do not have Permission", "&rto access this Item"));
+							if (Messages.local.getTranslation("messages.item-permission").get(1).isEmpty()) {
+								menu.addItem(index, new CustomItem(Material.BARRIER, StringUtils.formatItemName(sfitem.getItem(), false), "", Messages.local.getTranslation("messages.item-permission").get(0)));
+							} else {
+								menu.addItem(index, new CustomItem(Material.BARRIER, StringUtils.formatItemName(sfitem.getItem(), false), "", Messages.local.getTranslation("messages.item-permission").get(0), Messages.local.getTranslation("messages.item-permission").get(1)));
```
TheBusyBiscuit: You shouldn't use List Indexing that way here, just pass the whole list, instead of accessing single elements twice.

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -716,7 +716,11 @@ public void run(final Player p) {
 							index++;
 						}
 						else {
-							menu.addItem(index, new CustomItem(Material.BARRIER, StringUtils.formatItemName(sfitem.getItem(), false), "", "&rYou do not have Permission", "&rto access this Item"));
+							if (Messages.local.getTranslation("messages.item-permission").get(1).isEmpty()) {
```
TheBusyBiscuit: That isEmptyCheck is really unneccessary, you should work with the list instead.
This is just bad design.

src/me/mrCookieSlime/Slimefun/Setup/Messages.java
xH3LLRAIZ3Rx: Done

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
TheBusyBiscuit: This message key also needs to be updated.

xH3LLRAIZ3Rx: Done, not 100% sure if this works because I am not familiar with this, I never took java courses...

LinoxGH:
I'd suggest implementing this to default item configs instead of in messages.

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
TheBusyBiscuit: That will definitely not work.
This is the totally wrong approach for this.

xH3LLRAIZ3Rx: Alright, I cannot continue on this then, I tried everything I can think of, I simply do not have enough knowledge about Lists. You can close the PR or if you can make it work that would be better but I know you are very busy so by all means just close it. Sorry for wasting your time, any other PR's I do will be of things I do know, I just need to take a couple lessons about lists.

xH3LLRAIZ3Rx: Alright, what about now?

TheBusyBiscuit: No no no no no. Don't do that.
You just have to call list.toArray(new String[list.size()]) to convert it to an Array.
But you will have to remove the empty lore line and include it in the messages.yml

xH3LLRAIZ3Rx: Alright, I figured it out I hope, I was dumb and forgot to check the various constructors CustomItem has.

PR #1090, opened 2019-09-05
MultiBlockListener overhaul

Liruxo:
## Description
I replaced the code checking for multiblocks

## Changes
Removed the vast majority of the old Multiblock checking system and replaced it with a new recursive approach. I also changed some parts of the MultiBlock to fit the new code and also added support for the now deprecated old constructor which should prevent stuff from breaking without using the new constructor.

## Related Issues
There is no open related issue, but this fixes the currently (as of 
307a439) broken Multiblocks. (Multiblock are not recognized anymore by the Code replaced here)

## Testability

Addedum:
I did not find any code style regulations, so i just used the standart eclipse ones.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
@@ -3,6 +3,7 @@ package me.mrCookieSlime.Slimefun.Objects;
 import java.util.List;
 
 import org.bukkit.Material;
+import org.bukkit.block.BlockFace;
 
 import me.mrCookieSlime.CSCoreLibPlugin.compatibility.MaterialHelper;
 import me.mrCookieSlime.Slimefun.SlimefunPlugin;
@@ -12,18 +13,56 @@ import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunMachine;
 public class MultiBlock {
 	
 	private Material[] blocks;
-	private Material trigger;
-	
+	private BlockFace trigger;
+	private boolean isSymmetric;
+
+	@Deprecated
 	public MultiBlock(Material[] build, Material trigger) {
 		this.blocks = build;
+		this.isSymmetric = isSymmetric(build);
+		//Hacky, the other constructor should be used
+		for (int i = 0; i < 9; i++)
+		{
+			if (trigger.equals(build[i]))
+			{
+				switch (i) {
+					case 1:
+						this.trigger = BlockFace.DOWN;
+						break;
+					case 4:
+						this.trigger = BlockFace.SELF;
+						break;
+					case 7:
+						this.trigger = BlockFace.UP;
+						break;
+				}
+				break;
+			}
+		}
+	}
+	
+	public MultiBlock(Material[] build, BlockFace trigger)
+	{
+		this.blocks = build;
 		this.trigger = trigger;
+		this.isSymmetric = isSymmetric(build);
+	}
+	
+	private boolean isSymmetric(Material[] blocks)
+	{
+		if (blocks[0] == blocks[2]
+			&& blocks[3] == blocks[5]
+			&& blocks[6] == blocks[8]) {
+			return true;
+		}
+		return false;
 	}
 	
 	public Material[] getBuild() {
 		return this.blocks;
 	}
 	
-	public Material getTriggerBlock() {
+	public BlockFace getTriggerBlock() {
 		return this.trigger;
 	}
 	
@@ -68,6 +107,11 @@ public class MultiBlock {
 		}
 		
 		return true;
+	}
+	
+	public boolean isSymmetric()
+	{
+		return this.isSymmetric;
 	}
 
 }
--- a/src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
@@ -21,7 +21,7 @@ public class MultiBlock {
 		this.blocks = build;
 		this.isSymmetric = isSymmetric(build);
 		//Hacky, the other constructor should be used
-		for (int i = 0; i < 9; i++)
+		for (int i = 1; i < 9; i *=3)
 		{
 			if (trigger.equals(build[i]))
 			{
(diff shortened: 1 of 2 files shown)


src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
```diff
@@ -12,18 +13,56 @@
...
+	}
+	
+	public MultiBlock(Material[] build, BlockFace trigger)
+	{
 		this.blocks = build;
 		this.trigger = trigger;
+		this.isSymmetric = isSymmetric(build);
+	}
+	
+	private boolean isSymmetric(Material[] blocks)
```
TheBusyBiscuit: This method does not access any class fields, therefore it should be static.

src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
```diff
@@ -12,18 +13,56 @@
...
+	{
+		if (blocks[0] == blocks[2]
+			&& blocks[3] == blocks[5]
+			&& blocks[6] == blocks[8]) {
+			return true;
+		}
+		return false;
```
TheBusyBiscuit: You can replace this if-statement by just returning the boolean-expression inside the brackets.

src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
```diff
@@ -12,18 +13,56 @@
...
-	
+	private BlockFace trigger;
+	private boolean isSymmetric;
+
+	@Deprecated
 	public MultiBlock(Material[] build, Material trigger) {
+		this.blocks = build;
+		this.isSymmetric = isSymmetric(build);
+		//Hacky, the other constructor should be used
+		for (int i = 0; i < 9; i++)
```
TheBusyBiscuit: Iterating over every 9 elements is unnecessary if you only need element 1, 4, 7.
You should change the loop to 
for (int i = 1; i < 9; i *=3) to make it more performant.

src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
```diff
@@ -12,18 +13,56 @@
...
+		{
+			if (trigger.equals(build[i]))
+			{
+				switch (i) {
```
TheBusyBiscuit: Do we really need a switch/case here for free options?
I mean, 3 if-statements would do the trick as well.

WalshyDev: since this is a numeric lookup I support the use of a switch here, allows for much better JVM optimisation

TheBusyBiscuit: Okay then

Liruxo: Thats up to the coding style/preferences. If you prefer three if statements i can change that, i personally prefer switch statements.

src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
Liruxo: I have implemented it, but just wanted to mention three things: first this constructor shouldn't be used in the first place, second this is only called during bootup once per multiblock and third, as far as my knowledge goes, the difference between the old and new code is somewhere in the nanoseconds.
But anyways judging from my limited java bytecode and low machine code knowledge your solution should be faster.

TheBusyBiscuit: I never said it'd be a huge impact.
Empty statements are quite a bad-practice though, so you shouldn't really waste looping over items that will be ignored anyway, hence why I pointed it out.

src/me/mrCookieSlime/Slimefun/listeners/BlockListener.java
```diff
@@ -75,73 +74,15 @@ public void onPistonRetract(BlockPistonRetractEvent e) {
...
-						blocks[3] != null && blocks[3] == blocks[5]
-						&& BlockAdjacents.hasMaterialOnBothSides(b.getRelative(0, 1, 0), blocks[5]) &&
-						blocks[6] != null && blocks[6] == blocks[8]
-						&& BlockAdjacents.hasMaterialOnBothSides(b, blocks[8])
-					)) {
-						multiblocks.add(mb);
-					}
+				Player p = e.getPlayer();
+				Block b = e.getClickedBlock();
+				List<MultiBlock> multiblocks = new ArrayList<>();
```
TheBusyBiscuit: I know you didn't design it this way, that was my past self actually.

But you could swap this for a LinkedList, that would make the calls to .add() and retrieving the last element (getLast() or so) a lot cheaper on the CPU.

src/me/mrCookieSlime/Slimefun/Objects/MultiBlock.java
Liruxo: I agree with you on that one, bad coding practices should not be merged into the master branch

src/me/mrCookieSlime/Slimefun/listeners/BlockListener.java
```diff
@@ -155,4 +96,38 @@ public void onRightClick(PlayerInteractEvent e) {
 			}
 		}
 	}
+	
+	protected boolean compareMaterials(Block b, Material[] materials, boolean onlyTwoWay)
+	{
+		return compareMaterials(b, materials[0], materials[1], materials[2], materials[3], materials[4], materials[5], materials[6], materials[7], materials[8], onlyTwoWay);
+	}
+	
+	protected boolean compareMaterials(Block b, Material mat1, Material mat2, Material mat3, Material mat4, Material mat5, Material mat6, Material mat7, Material mat8, Material mat9, boolean onlyTwoWay)
```
TheBusyBiscuit: Creating a method with this many arguments is really not a good idea.
You can also just use the method and above and read the Materials directly from the array instead.

src/me/mrCookieSlime/Slimefun/listeners/BlockListener.java
```diff
@@ -155,4 +96,38 @@ public void onRightClick(PlayerInteractEvent e) {
...
+	{
+		if (!compareMaterialsVertical(b, mat2, mat5, mat8))
+			return false;
+		
+		BlockFace[] directions = onlyTwoWay ? new BlockFace[] {BlockFace.NORTH, BlockFace.EAST} : new BlockFace[] {BlockFace.NORTH, BlockFace.EAST, BlockFace.SOUTH, BlockFace.WEST};
+		for (BlockFace direction : directions)
+		{
+			if (compareMaterialsVertical(b.getRelative(direction), mat1, mat4, mat7))
+				if (compareMaterialsVertical(b.getRelative(direction.getOppositeFace()), mat3, mat6, mat9))
```
TheBusyBiscuit: You can link these two booleans with an "and".

Liruxo:
You mentioned that i should update the constructors aswell, can/should i add these changes to this PR as well or wait until it is merged?

TheBusyBiscuit:
@Liruxo Whatever you prefer, I don't mind merging another PR, you can of course also just include these changes in here. It's up to you.

TheBusyBiscuit:
Merging this now, because bug reports have come in about it.

PR #1092, opened 2019-09-05
Micro-optimisations

WalshyDev:
## Description
Just done some minor micro-optimisations

## Changes
* Changed all single-char strings to char (unless it would be less beneficial)
* Removed useless String.valueOf's
* Used Loggers in-built formatter
* Save doing replace multiple times (especially replaceFirst which always compiles the regex)

## Related Issues
N/A

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/hooks/github/IntegerFormat.java
+++ b/src/me/mrCookieSlime/Slimefun/hooks/github/IntegerFormat.java
@@ -19,7 +19,7 @@ public final class IntegerFormat {
 	
 	public static Date parseGitHubDate(String str) {
 		try {
-			return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").parse(str.replace("T", " ").replace("Z", ""));
+			return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").parse(str.replace('T', ' ').replace("Z", ""));
 		} catch (ParseException x) {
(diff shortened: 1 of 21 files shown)


src/me/mrCookieSlime/Slimefun/hooks/github/IntegerFormat.java
```diff
@@ -19,7 +19,7 @@ public static String formatBigNumber(int i) {
 	
 	public static Date parseGitHubDate(String str) {
 		try {
-			return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").parse(str.replace("T", " ").replace("Z", ""));
+			return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss").parse(str.replace('T', ' ').replace("Z", ""));
```
TheBusyBiscuit: You missed the Z :)

WalshyDev: No I didn't, a char is by definition a single character. You can't do an empty char per say so it's best to just do an empty string here. And you can't do the "Z" alone as it would then not be matching CharSequences and fail compilation.

TheBusyBiscuit: Right, didn't see the right-side was an empty string.

PR #1094, opened 2019-09-06
Changed Slimefun Guide sound to Page Turn

xH3LLRAIZ3Rx:
## Description
I changed the sound for the Slimefun Guide to the new Page Turn sound.

## Changes
Changed ENTITY_BAT_TAKEOFF to ITEM_BOOK_PAGE_TURN

## Testability


WalshyDev:
What... why are you using Float.MAX_VALUE for volume? It's between 0-1... really not sure what the point of that is but... sure?

xH3LLRAIZ3Rx:
> What... why are you using Float.MAX_VALUE for volume? It's between 0-1... really not sure what the point of that is but... sure?

Yeah honestly I am not sure why I did that lol, tired..

TheBusyBiscuit:
You also shouldn't use a pitch of 0...
Either 1 or 0.8 or so but definitely not 0.

PR #1100, opened 2019-09-08
Changed item-permission message to be per item rather than one global list

xH3LLRAIZ3Rx:
## Description
I decided that it would be better all around to be able to change the permission message per item rather than one global message.

## Changes
Added permission-message string list
Removed tooltips.item-message in messages class

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -50,6 +50,7 @@ public class SlimefunItem {
 	private boolean replacing = false;
 	private boolean addon = false;
 	private String permission = "";
+	private List<String> noPermissionTooltip;
 	private Set<ItemHandler> itemhandlers = new HashSet<>();
 	private boolean ticking = false;
 	private BlockTicker blockTicker;
@@ -178,6 +179,7 @@ public class SlimefunItem {
 	 * @since 4.1.11
 	 */
 	public String getPermission() 			{		return permission;		}
+	public List<String> getNoPermissionTooltip()    {       return noPermissionTooltip;       }
 	public Set<ItemHandler> getHandlers() 		{		return itemhandlers;		}
 	public boolean isTicking() 			{		return ticking;			}
 	/**
@@ -244,6 +246,7 @@ public class SlimefunItem {
 				this.enchantable = SlimefunPlugin.getItemCfg().getBoolean(this.id + ".allow-enchanting");
 				this.disenchantable = SlimefunPlugin.getItemCfg().getBoolean(this.id + ".allow-disenchanting");
 				this.permission = SlimefunPlugin.getItemCfg().getString(this.id + ".required-permission");
+				this.noPermissionTooltip = SlimefunPlugin.getItemCfg().getStringList(this.id + ".permission-message");
 				SlimefunPlugin.getUtilities().enabledItems.add(this);
 				if (slimefun) SlimefunPlugin.getUtilities().vanillaItems++;
 				SlimefunPlugin.getUtilities().itemIDs.put(this.id, this);
@@ -334,6 +337,15 @@ public class SlimefunItem {
 		}
 		if (SlimefunManager.isItemSimiliar(item, SlimefunItems.BROKEN_SPAWNER, false)) return getByID("BROKEN_SPAWNER");
 		if (SlimefunManager.isItemSimiliar(item, SlimefunItems.REPAIRED_SPAWNER, false)) return getByID("REINFORCED_SPAWNER");
+		return null;
+	}
+
+	public static List<String> getNoPermissionTooltip(ItemStack item) {
+		for (SlimefunItem i: SlimefunPlugin.getUtilities().allItems) {
+			if (i.isItem(item)) {
+				return i.noPermissionTooltip;
+			}
+		}
 		return null;
 	}
 
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -218,7 +218,7 @@ public class SlimefunItem {
 			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".allow-enchanting", this.enchantable);
 			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".allow-disenchanting", this.disenchantable);
 			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".required-permission", this.permission);
-			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".permission-message", new String[] {"&4&lLOCKED", "", "&rYou do not have Permission", "&rto access this Item"});
+			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".no-permission-tooltip", new String[] {"&4&lLOCKED", "", "&rYou do not have Permission", "&rto access this Item"});
 
 			if (this.keys != null && this.values != null) {
 				for (int i = 0; i < this.keys.length; i++) {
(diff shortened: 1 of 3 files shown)


ajan-12:
Told ya lol

xH3LLRAIZ3Rx:
> Told ya lol

Yup lol, I ended up needing it.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -50,6 +50,7 @@
 	private boolean replacing = false;
 	private boolean addon = false;
 	private String permission = "";
+	private String[] permissionMessage = {"&4&lLOCKED", "", "&rYou do not have Permission", "&rto access this Item"};
```
TheBusyBiscuit: Should rather be "noPermissionTooltip" as permissionMessage is misleading...

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -713,8 +713,8 @@ public void run(final Player p) {
 							index++;
 						}
 						else {
-							List<String> tooltip = Messages.local.getTranslation("tooltips.item-permission");
-						    menu.addItem(index, new CustomItem(Material.BARRIER, StringUtils.formatItemName(sfitem.getItem(), false), tooltip.toArray(new String[tooltip.size()])));
+							List<String> message = SlimefunPlugin.getItemCfg().getStringList(sfitem.getID() + ".permission-message");
```
TheBusyBiscuit: Why would you read from the config if you implemented a getter in SlimefunItem...??

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
TheBusyBiscuit: This value is read-only atm, keeping it in RAM for all the thousands of Slimefunitems makes zero sense in your design, you should write to it in the register() block were all the other values are set as well.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -178,6 +179,7 @@ public SlimefunItem(Category category, ItemStack item, String id, RecipeType rec
 	 * @since 4.1.11
 	 */
 	public String getPermission() 			{		return permission;		}
+	public String[] getPermissionMessage()    {       return permissionMessage;       }
```
TheBusyBiscuit: This getter makes no sense if it just returns the same constant default value for each object.
You should read the value from config in .register()

xH3LLRAIZ3Rx: Honestly I am not sure why I put a getter

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
xH3LLRAIZ3Rx: Would this work?

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
xH3LLRAIZ3Rx: What about now?

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
xH3LLRAIZ3Rx: Alright

TheBusyBiscuit:
Not going to submit another review, I already said what should be changed.
You actually did the opposite XD

I said that your getter and field are pointless if you don't write to them.
And that you shouldn't read from the config hundreds of times when the guide is opened...

Instead USE the field and getter how they should be used, not remove them...

xH3LLRAIZ3Rx:
> Not going to submit another review, I already said what should be changed.
> You actually did the opposite XD
> 
> I said that your getter and field are pointless if you don't write to them.
> And that you shouldn't read from the config hundreds of times when the guide is opened...
> 
> Instead USE the field and getter how they should be used, not remove them...

Sorry dude, I have not slept yet so I am not as focused as I should be, some caffeine will fix it though.

ajan-12:
Didn't check your code fully but this is how I'd do it:
Set default value on register.
Read values on postRegister.
When item is used check if player has unlocked the item and if not send the message read in postRegister.
(Messages wouldn't be tooltip but instead sent to player via chat.)
Note: Changing tooltips may cause issues in the future.

TheBusyBiscuit:
@ajan-12 If you use postRegister than every implementation in Slimefun that overrides that needs to get a super.postRegister() call.
I agree, this just has to be considered and done too then.

xH3LLRAIZ3Rx:
> 
> 
> @ajan-12 If you use postRegister than every implementation in Slimefun that overrides that needs to get a super.postRegister() call.
> I agree, this just has to be considered and done too then.

Honestly the current code works, so I think you should merge it then let him or someone else do the rest because I don't know enough java to do the rest.

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -713,8 +713,8 @@ public void run(final Player p) {
 							index++;
 						}
 						else {
-							List<String> tooltip = Messages.local.getTranslation("tooltips.item-permission");
-						    menu.addItem(index, new CustomItem(Material.BARRIER, StringUtils.formatItemName(sfitem.getItem(), false), tooltip.toArray(new String[tooltip.size()])));
+							List<String> message = SlimefunItem.getNoPermissionTooltip(sfitem.getItem());
```
TheBusyBiscuit: Oh god no.
Why would you loop through all Slimefun items to find the Slimefun Item that you already had...???

Just do sfItem.getNoPermissionTooltip() ...

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -336,6 +340,15 @@ public static SlimefunItem getByItem(ItemStack item) {
 		return null;
 	}
 
+	public static List<String> getNoPermissionTooltip(ItemStack item) {
```
TheBusyBiscuit: Remove this method.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -243,6 +246,7 @@ public void register(boolean slimefun) {
 				this.enchantable = SlimefunPlugin.getItemCfg().getBoolean(this.id + ".allow-enchanting");
 				this.disenchantable = SlimefunPlugin.getItemCfg().getBoolean(this.id + ".allow-disenchanting");
 				this.permission = SlimefunPlugin.getItemCfg().getString(this.id + ".required-permission");
+				this.noPermissionTooltip = SlimefunPlugin.getItemCfg().getStringList(this.id + ".permission-message");
```
TheBusyBiscuit: The path should match the field name "no-permission-tooltip"

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -216,7 +218,8 @@ public void register(boolean slimefun) {
 			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".allow-enchanting", this.enchantable);
 			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".allow-disenchanting", this.disenchantable);
 			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".required-permission", this.permission);
-			
+			SlimefunPlugin.getItemCfg().setDefaultValue(this.id + ".permission-message", new String[] {"&4&lLOCKED", "", "&rYou do not have Permission", "&rto access this Item"});
```
TheBusyBiscuit: The path should match the field name "no-permission-tooltip"

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
xH3LLRAIZ3Rx: Yeah I was tired when I did that, fixed ;)

PR #1110, opened 2019-09-10
Fixed MultiBlock RecipeType material

xH3LLRAIZ3Rx:
## Description
I noticed that the RecipeType Multiblock's material was a brick and not a brick block.

## Changes
Changed Material.BRICK to Material.BRICKS

## Testability


WalshyDev:
This hasn't been changed in 3 years, reason why you want to change get it now?

xH3LLRAIZ3Rx:
> This hasn't been changed in 3 years, reason why you want to change it now?
> 
> I think both make sense, you use bricks to build a wall. A brick block could also represent a built structure. I just don't see the point in changing this. Especially since it hasn't changed for 3 years (at least).

Oh I thought it was recent, alright I will close this.

PR #1128, opened 2019-09-26
Multiblock constructor refactor

Liruxo:
## Description
This is the follow up to PR #1090
This PR changes all Multiblocks in Slimefun to use the new BlockFace-Trigger Constructor instead of the old Material-Trigger

## Changes
Reaplaced various constructors to use the different type and updated all MultiBlocks to use the new constructors

## Testability


Liruxo:
Oh i did not think about that, you are right

Liruxo:
I have added Deprecated constructors for all new ones. There is some performance overhead created by converting the old arguments to the new, but i cant think of a more efficient way without introducing new variables to the class.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunMachine.java
```diff
@@ -72,21 +88,28 @@ public void install() {
 			}
 		}
 	}
-	
-	public MultiBlock toMultiBlock() {
+
+	private static Material[] convertItemStacksToMaterial(ItemStack[] items) {
 		List<Material> mats = new ArrayList<>();
-		for (ItemStack i: this.getRecipe()) {
+		for (ItemStack i: items) {
```
TheBusyBiscuit: You should use getRecipe() instead because the method could be overridden.

TheBusyBiscuit: Other than that... looks good!

Liruxo: Uhm, i dont seem to understand your request. That is a static method, the argument 'items' is passed to it. The former method toMultiBlock() still uses "this.getRecipe()" to generate the actual mutliblock, only the Deprecated Constructors use this method directly to generate the BlockFace-Trigger Argument

TheBusyBiscuit: The static method can of course not be overridden.
The method getRecipe() can be overridden though, using this method would ensure that this case is covered, by accessing the field directly, it could cause redundancies.

Liruxo: I am confused. Just to clarify: We are talking about the static Method convertItemStacksToMaterial(ItemStack[] items). This Method is called by toMultiBlock() which uses getRecipe() to get the argument "ItemStack[] items" and passes it to convertItemStacksToMaterial(ItemStack[] items) where you now propose to use getRecipe() instead of the passed "ItemStack[] items" argument. As far as i can see, even If i would implement it there wouldnt be any change in any result, because the 'items' argument is generated by the getRecipe().

TheBusyBiscuit: Oh, sorry I actually did not see that "items" was a parameter, I thought it was a field.
My bad.

TheBusyBiscuit:
#1129 

You used the wrong BlockFace for the Smeltery.

WalshyDev:
Is there a good way of unit testing this? There's quite a few multiblock machines, having a unit test for them (hopefully in a dynamic way) would be best.

PR #1137, opened 2019-09-29
Add notice for stable

WalshyDev:
## Description
Some people disable auto-updates on dev builds because they can have issues. Let's instead advertise the stable branch

## Changes
Adds a notice to the config

## Related Issues
N/A

## Testability
N/A


TheBusyBiscuit:
Sounds good.
Slimefun still uses a very bad CS-CoreLib1 way of doing the config though...
I am unsure whether it will copy other comments...

PR #1138, opened 2019-09-30
Guide Search!

WalshyDev:
## Description
OwO what's this? OwO it's a guide search

## Changes
Add a nametag in slot 8, when clicked it sends a chat message which will then search and open up the guide again but with the filtered search result!

## Related Issues
N/A 

## Testability


src/me/mrCookieSlime/Slimefun/listeners/SearchListener.java
```diff
@@ -0,0 +1,35 @@
...
+import org.bukkit.Bukkit;
+import org.bukkit.event.EventHandler;
+import org.bukkit.event.Listener;
+import org.bukkit.event.player.AsyncPlayerChatEvent;
+
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class SearchListener implements Listener {
```
TheBusyBiscuit: This class is redundant since we already have a class like this...

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -948,7 +1017,15 @@ else if (r instanceof FurnaceRecipe) {
 		
 		menu.open(p);
 	}
-	
+
+	private static void handleHistory(Player pl, Object last, boolean book, boolean cheat) {
```
TheBusyBiscuit: Nice.
A method like that was needed!

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -726,21 +713,104 @@ public void run(final Player p) {
...
+			menu.addMenuClickHandler(i, (arg0, arg1, arg2, arg3) -> false);
+		}
+
+		// Search feature!
+		menu.addItem(7, new CustomItem(Material.NAME_TAG, "&7Search",
+			"&7Search directly for items"));
+		menu.addMenuClickHandler(7, (player, i, itemStack, clickAction) -> {
+			player.closeInventory();
+			player.sendMessage(ChatColor.AQUA + "What would you like to search for?");
+			SearchListener.addSearchingPlayer(player.getUniqueId());
```
TheBusyBiscuit: Should rather use MenuHelper.awaitChat

or something like that...

WalshyDev: Ah nice, didn't know this already existed. I don't know of any chat prompts so just made a quick 30 second class to handle it for now until such time there was Loren use cases.

TheBusyBiscuit: Elevator, Hologram Projector and a few other blocks also require chat prompts.

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -726,21 +713,107 @@ public void run(final Player p) {
...
+			menu.addItem(i, new CustomItem(new ItemStack(Material.GRAY_STAINED_GLASS_PANE), " "));
+			menu.addMenuClickHandler(i, (arg0, arg1, arg2, arg3) -> false);
+		}
+
+		// Search feature!
+		menu.addItem(7, new CustomItem(Material.NAME_TAG, "&7Search",
+			"&7Search directly for items"));
+		menu.addMenuClickHandler(7, (player, i, itemStack, clickAction) -> {
+			player.closeInventory();
+			player.sendMessage(ChatColor.AQUA + "What would you like to search for?");
```
TheBusyBiscuit: I think this String should be added to Slimefun's localization file.
And maybe add another line saying "Enter your search term in chat" or whatever?
To make it clear for everyone.

PR #1147, opened 2019-10-03
Changed distance of decaying

J3fftw1:
## Description

i changed the distance that it starts decaying from 5 to 1 this wouldnt perse reduce the sound but would make it less obvious at longer distnaces
edit

## Changes

i changed the volume 5 to 1
## Testability


WalshyDev:
Interesting fact, the client actually caps this at 1 already. There's no difference between 1 and 10. Standard for these float values

J3fftw1:
the game sound doesnt get changed but the distance it starts decaying will get changed  as seen in this video the sound only decayed after being arround 50 blocks away https://streamable.com/bz115

WalshyDev:
In terms of volume there's no change, should have added that part. 

Yes, if > 1 it will multiply the packet send distance by the float. Base distance is 16. So in these cases it's 16 * 5 of the range where people get the packet.

PR #1148, opened 2019-10-03
Update TeleportationSequence.java

J3fftw1:
## Description

i changed the distance the sound will decay
## Changes

changed a volume from 1.7 to 1
changed a volume from 2 to 1
## Testability

- [x ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.


TheBusyBiscuit:
Great, now all the rest is missing.

PR #1149, opened 2019-10-03
Changed values of sound decaying

J3fftw1:
## Description

i changed the distance the sound will decay
## Changes

i changed the following volume values
2 to 1 2x
1.7 to 1
5 to 1 3x
## Testability

- [ x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ x] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/tasks/MagnetTask.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/tasks/MagnetTask.java
@@ -18,7 +18,7 @@ public class MagnetTask extends SlimefunTask {
 		for (Entity item: Bukkit.getPlayer(uuid).getNearbyEntities(6D, 6D, 6D)) {
 			if (item instanceof Item && !item.hasMetadata("no_pickup") && ((Item) item).getPickupDelay() <= 0) {
 				item.teleport(Bukkit.getPlayer(uuid).getEyeLocation());
-				Bukkit.getPlayer(uuid).getWorld().playSound(Bukkit.getPlayer(uuid).getEyeLocation(), Sound.ENTITY_ENDERMAN_TELEPORT, 5L, 2L);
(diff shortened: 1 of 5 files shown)


WalshyDev:
Here are all occurances just in case you wanted quick access since you aren't using an editor...
Values

src/me/mrCookieSlime/Slimefun/Objects/tasks/MagnetTask.java
```diff
@@ -18,7 +18,7 @@ void executeTask() {
 		for (Entity item: Bukkit.getPlayer(uuid).getNearbyEntities(6D, 6D, 6D)) {
 			if (item instanceof Item && !item.hasMetadata("no_pickup") && ((Item) item).getPickupDelay() <= 0) {
 				item.teleport(Bukkit.getPlayer(uuid).getEyeLocation());
-				Bukkit.getPlayer(uuid).getWorld().playSound(Bukkit.getPlayer(uuid).getEyeLocation(), Sound.ENTITY_ENDERMAN_TELEPORT, 5L, 2L);
+				Bukkit.getPlayer(uuid).getWorld().playSound(Bukkit.getPlayer(uuid).getEyeLocation(), Sound.ENTITY_ENDERMAN_TELEPORT, 1L, 2L);
```
TheBusyBiscuit: Should be F, not L.

J3fftw1: it was already like that didnt know why

WalshyDev: Also change the other param to F too

TheBusyBiscuit:
Well what about the Infused Hopper and all the other stuff from before...?

J3fftw1:
what do you mean Biscuit

TheBusyBiscuit:
I mean what I said.
You originally wanted to change the Infused Hopper, but you didn't do.
This Pull Requests only contains like 10% of the stuff you wanted to change.

J3fftw1:
changed it right now i think thats all

PR #1150, opened 2019-10-03
Persistent data on items

WalshyDev:
## Description
Adds the item ID under the sf_item tag.

## Changes
In the SlimefunItem root constructors add the item ID. Some methods which previously went through the whole item to check now check for the ID first and if equal return there. This will improve performance on these lookups.

## Related Issues
N/A

## Testability

Eventually this entire plugin could not even have to rely on files. Everything could just be in NBT >:)
Would be good in terms of space and general data corruption.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
@@ -62,41 +62,40 @@ public class SlimefunItem {
 	private String wiki = null;
 
 	public SlimefunItem(Category category, ItemStack item, String id, RecipeType recipeType, ItemStack[] recipe) {
-		this.item = item;
-		this.category = category;
-		this.id = id;
-		this.recipeType = recipeType;
-		this.recipe = recipe;
+		this(category, item, id, recipeType, recipe, null);
 	}
 
 	public SlimefunItem(Category category, ItemStack item, RecipeType recipeType, ItemStack[] recipe) {
-		if (!(item instanceof SlimefunItemStack)) throw new IllegalArgumentException("item must be of Type SlimefunItemStack!");
-		
-		this.item = item;
-		this.category = category;
-		this.id = ((SlimefunItemStack) item).getItemID();
-		this.recipeType = recipeType;
-		this.recipe = recipe;
+		this(category, item, recipeType, recipe, null);
 	}
 
 	public SlimefunItem(Category category, ItemStack item, String id, RecipeType recipeType, ItemStack[] recipe, ItemStack recipeOutput) {
-		this.item = item;
-		this.category = category;
-		this.id = id;
-		this.recipeType = recipeType;
-		this.recipe = recipe;
-		this.recipeOutput = recipeOutput;
+		this(category, item, id, recipeType, recipe, recipeOutput, null, null);
 	}
 
 	public SlimefunItem(Category category, ItemStack item, RecipeType recipeType, ItemStack[] recipe, ItemStack recipeOutput) {
+		this(category, item, recipeType, recipe, recipeOutput, null, null);
+	}
+
+	public SlimefunItem(Category category, ItemStack item, RecipeType recipeType, ItemStack[] recipe, String[] keys, Object[] values) {
+		this(category, item, recipeType, recipe, null, keys, values);
+	}
+
+	public SlimefunItem(Category category, ItemStack item, String id, RecipeType recipeType, ItemStack[] recipe, String[] keys, Object[] values) {
+		this(category, item, id, recipeType, recipe, null, keys, values);
+	}
+
+	public SlimefunItem(Category category, ItemStack item, String id, RecipeType recipeType, ItemStack[] recipe, boolean hidden) {
+		this(category, item, id, recipeType, recipe);
+		this.hidden = hidden;
+	}
+
+	// Root constructors
+	public SlimefunItem(Category category, ItemStack item, RecipeType recipeType, ItemStack[] recipe, ItemStack recipeOutput, String[] keys, Object[] values) {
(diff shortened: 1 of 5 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -57,88 +59,66 @@
...
 		this.category = category;
+		this.item = item;
 		this.id = ((SlimefunItemStack) item).getItemID();
 		this.recipeType = recipeType;
 		this.recipe = recipe;
+		this.recipeOutput = recipeOutput;
 		this.keys = keys;
 		this.values = values;
+
+		PersistentDataAPI.setString(this.item.getItemMeta(), Constants.SF_ITEM, this.id);
```
TheBusyBiscuit: This should be removed.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -57,88 +59,66 @@
...
-	}
 
-	public SlimefunItem(Category category, ItemStack item, String id, RecipeType recipeType, ItemStack[] recipe, boolean hidden) {
-		this.item = item;
-		this.category = category;
-		this.id = id;
-		this.recipeType = recipeType;
-		this.recipe = recipe;
-		this.hidden = hidden;
+		PersistentDataAPI.setString(this.item.getItemMeta(), Constants.SF_ITEM, this.id);
```
TheBusyBiscuit: Definitely should be removed.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -311,7 +291,13 @@ public static SlimefunItem getByID(String id) {
 	}
 
 	public static SlimefunItem getByItem(ItemStack item) {
-		if (item == null) return null;		
+		if (item == null) return null;
+
+		if (item.getItemMeta() != null) {
```
TheBusyBiscuit: You should do hasItemMeta() instead of a null-check

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -328,6 +314,13 @@ public static SlimefunItem getByItem(ItemStack item) {
 
 	public boolean isItem(ItemStack item) {
 		if (item == null) return false;
+
+		if (item.getItemMeta() != null && this.item.getItemMeta() != null) {
```
TheBusyBiscuit: Use .hasItemmeta() instead of null-checks

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -328,6 +314,13 @@ public static SlimefunItem getByItem(ItemStack item) {
...
+			String thisId = PersistentDataAPI.getString(this.item.getItemMeta(), Constants.SF_ITEM);
```
TheBusyBiscuit: Instead of getting the id this way you can literally just call "getID()"...

src/me/mrCookieSlime/Slimefun/Setup/SlimefunManager.java
```diff
@@ -54,30 +56,36 @@ public static void registerArmorSet(ItemStack baseComponent, ItemStack[] items,
...
+
 	}
-	
+
 	public static boolean isItemSimiliar(ItemStack item, ItemStack sfitem, boolean lore) {
 		if (item == null) return sfitem == null;
 		if (sfitem == null) return false;
-		
+
+		if (item.getItemMeta() != null && sfitem.getItemMeta() != null) {
```
TheBusyBiscuit: Use .hasItemMeta() instead of doing null checks.

src/me/mrCookieSlime/Slimefun/utils/Constants.java
```diff
@@ -0,0 +1,13 @@
+package me.mrCookieSlime.Slimefun.utils;
+
+import me.mrCookieSlime.Slimefun.SlimefunPlugin;
+import org.bukkit.NamespacedKey;
+
+public final class Constants {
+
+    // Note the naming convention for Namespaces is snake_case
+
+    public static final NamespacedKey SF_ITEM = new NamespacedKey(SlimefunPlugin.instance, "sf_item");
```
TheBusyBiscuit: I think instead of creating this memory leaking constants class, just put it into SlimefunPlugin, make it not static and add a static method getNamespacedItemKey() or so.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
WalshyDev: hasItemMeta literally is just a null check and this stops IDEs from complaining that it may be null.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
WalshyDev: Me smart

src/me/mrCookieSlime/Slimefun/utils/Constants.java
WalshyDev: I mean, people shouldn't reload to begin with and then Bukkit should handle it properly... *sigh*

Either way, it's 16 bytes that can leak and for the sake of constants being in one place along with not fulling SlimefunPlugin more. I think it's fine. It's already a 450 line class.

PR #1160, opened 2019-10-05
Improved Performance for Slimefun Armor

TheBusyBiscuit:
## Description

@WalshyDev plz reviw

## Changes

+ Added HashedArmorpiece
+ Implemented caching for armorpieces to reduce SlimefunItem.getByItem() calls.
* Moved ArmorTask to its own Class

## Related Issues

None

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/tasks/ArmorTask.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/tasks/ArmorTask.java
@@ -1,5 +1,9 @@
 package me.mrCookieSlime.Slimefun.Objects.tasks;
 
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
 import org.bukkit.Bukkit;
 import org.bukkit.entity.Player;
 import org.bukkit.inventory.ItemStack;
@@ -20,6 +24,18 @@ import me.mrCookieSlime.Slimefun.utils.Utilities;
 public class ArmorTask implements Runnable {
 	
 	private final Utilities utilities = SlimefunPlugin.getUtilities();
+	private final Set<PotionEffect> radiationEffects;
+	
+	public ArmorTask() {
+		Set<PotionEffect> effects = new HashSet<>();
+		effects.add(new PotionEffect(PotionEffectType.WITHER, 400, 2));
+		effects.add(new PotionEffect(PotionEffectType.BLINDNESS, 400, 3));
+		effects.add(new PotionEffect(PotionEffectType.CONFUSION, 400, 3));
+		effects.add(new PotionEffect(PotionEffectType.WEAKNESS, 400, 2));
+		effects.add(new PotionEffect(PotionEffectType.SLOW, 400, 1));
+		effects.add(new PotionEffect(PotionEffectType.SLOW_DIGGING, 400, 1));
+		radiationEffects = Collections.unmodifiableSet(effects);
+	}
 
 	@Override
 	public void run() {
@@ -71,13 +87,9 @@ public class ArmorTask implements Runnable {
 
 					// If the item is enabled in the world, then make radioactivity do its job
 					if (Slimefun.isEnabled(p, radioactive, false)) {
+						SlimefunPlugin.getLocal().sendMessage(p, "messages.radiation");
 						Bukkit.getScheduler().runTask(SlimefunPlugin.instance, () -> {
-							p.addPotionEffect(new PotionEffect(PotionEffectType.WITHER, 400, 3));
-							p.addPotionEffect(new PotionEffect(PotionEffectType.BLINDNESS, 400, 3));
-							p.addPotionEffect(new PotionEffect(PotionEffectType.CONFUSION, 400, 3));
-							p.addPotionEffect(new PotionEffect(PotionEffectType.WEAKNESS, 400, 3));
-							p.addPotionEffect(new PotionEffect(PotionEffectType.SLOW, 400, 1));
-							p.addPotionEffect(new PotionEffect(PotionEffectType.SLOW_DIGGING, 400, 1));
+							p.addPotionEffects(radiationEffects);
 							p.setFireTicks(400);
 						});
 						
--- a/src/me/mrCookieSlime/Slimefun/Objects/tasks/ArmorTask.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/tasks/ArmorTask.java
@@ -45,14 +45,16 @@ public class ArmorTask implements Runnable {
 				}
 				
 				if (item != null && armorpiece.getItem().isPresent()) {
-					for (PotionEffect effect: armorpiece.getItem().get().getEffects()) {
-						p.removePotionEffect(effect.getType());
-						p.addPotionEffect(effect);
-					}
+					Bukkit.getScheduler().runTask(SlimefunPlugin.instance, () -> {
+						for (PotionEffect effect: armorpiece.getItem().get().getEffects()) {
+							p.removePotionEffect(effect.getType());
+							p.addPotionEffect(effect);
+						}
+					});
 				}
 			}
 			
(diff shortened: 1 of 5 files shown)


src/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -268,49 +264,7 @@ else if (getDescription().getVersion().startsWith("RC - ")) {
...
-									p.addPotionEffect(new PotionEffect(PotionEffectType.SLOW_DIGGING, 400, 1));
-									p.setFireTicks(400);
-									break; 
-									// Break the loop to save some calculations
-								}
-							}
-						}
-					}
-				}, 0L, config.getInt("options.armor-update-interval") * 20L);
+				getServer().getScheduler().runTaskTimer(this, new ArmorTask(), 0L, config.getInt("options.armor-update-interval") * 20L);
```
WalshyDev: Do we need this to be sync? The only thing I can think which MAY need to be sync is the potion effects. I haven't ever tried to add them async so I don't know if they're thread-safe or not. We should definitely make it async though even if we just offload the potion effects to a sync thread.

src/me/mrCookieSlime/Slimefun/api/HashedArmorpiece.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.bukkit.inventory.ItemStack;
+
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunArmorPiece;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
+
+public final class HashedArmorpiece {
+	
+	private int hash;
+	private Optional<SlimefunArmorPiece> item;
```
WalshyDev: What kind of reference will this be? Maybe it should be enforced as a weak reference?

src/me/mrCookieSlime/Slimefun/Objects/tasks/ArmorTask.java
```diff
@@ -0,0 +1,87 @@
...
+		for (Player p: Bukkit.getOnlinePlayers()) {
+			if (!p.isValid() || p.isDead()) {
+				continue;
+			}
+			
+			ItemStack[] armor = p.getInventory().getArmorContents();
+			HashedArmorpiece[] cachedArmor = PlayerProfile.get(p).getArmor();
+			
+			for (int slot = 0; slot < 4; slot++) {
+				ItemStack item = armor[slot];
```
WalshyDev: we should null check and continue

TheBusyBiscuit: It already does a nullcheck in .hasDiverged()
I think it should remain that way, since we otherwise would never clean up the hashed armorpieces if we keep continuing;
Would be a small waste of memory to be honest.

src/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
TheBusyBiscuit: I mean I don't think Bukkit.getOnlinePlayers() can even be sync, that would very likely cause lots of ConcurrentModificationExceptions, wouldn't it?
Or is getOnlinePlayers() actually thread-safe?

WalshyDev: There are no modifications and it's immutable so thread-safe.

src/me/mrCookieSlime/Slimefun/api/HashedArmorpiece.java
TheBusyBiscuit: Why?

src/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
TheBusyBiscuit: It's not modified when players join and leave?
Is it just an immutable copy of the current state?

src/me/mrCookieSlime/Slimefun/api/HashedArmorpiece.java
WalshyDev: If this is held elsewhere then the reference should be controlled. Not sure if it is or not though

src/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
WalshyDev: IIRC it's just an immutable list of the current net clients player

TheBusyBiscuit: Alright then, sync it shall be!

src/me/mrCookieSlime/Slimefun/api/HashedArmorpiece.java
TheBusyBiscuit: Well, only the class itself access it, it does have a getter but the entire PlayerProfile will be unloaded from memory when the player leaves anyway.

src/me/mrCookieSlime/Slimefun/Objects/tasks/ArmorTask.java
```diff
@@ -0,0 +1,103 @@
...
+					Bukkit.getScheduler().runTask(SlimefunPlugin.instance, () -> {
+						for (PotionEffect effect: armorpiece.getItem().get().getEffects()) {
+							p.removePotionEffect(effect.getType());
+							p.addPotionEffect(effect);
+						}
+					});
+				}
+			}
+			
+			if (SlimefunManager.isItemSimiliar(p.getInventory().getHelmet(), SlimefunItems.SOLAR_HELMET, true) && Slimefun.hasUnlocked(p, SlimefunItem.getByID("SOLAR_HELMET"), true) && (p.getWorld().getTime() < 12300 || p.getWorld().getTime() > 23850 && p.getEyeLocation().getBlock().getLightFromSky() == 15)) {
```
WalshyDev: Can you please split this up into multiple lines for my sanity

src/me/mrCookieSlime/Slimefun/Setup/SlimefunLocalization.java
```diff
@@ -72,6 +72,7 @@ public SlimefunLocalization(SlimefunPlugin plugin) {
 		setDefaultMessage("messages.fire-extinguish", "&7You have extinguished yourself");
 		setDefaultMessage("messages.cannot-place" ,"&cYou cannot place that block there!");
 		setDefaultMessage("messages.no-pvp" ,"&cYou cannot pvp in here!");
+		setDefaultMessage("messages.radiation" ,"&4You have been exposed to deadly Radiation! &cGet rid of the radioactive Item or equip a full Hazmat Suit!");
```
WalshyDev: *equip the hazmat suit

TheBusyBiscuit: I think we should emphasize "full" here.
Because the suit consists of four pieces, "Hazmat Suit" is one of them, so that'd cause some confusion.

WalshyDev: also radiation shouldn't be capitalised

WalshyDev: Well, then make it "equip the full..."

TheBusyBiscuit: You brits with your weird way of capitalising :p
The entirety of Slimefun is written that way, so it should also stay.

TheBusyBiscuit: Why "the" and not "a".
It's not a specific hazmat suit they need, they just need any.

WalshyDev: There aren't multiple types of hazmat suit, there's _the_ hazmat suit.

TheBusyBiscuit: Okay fine.

TheBusyBiscuit: Watch #suggestions

PR #1167, opened 2019-10-07
Added lightning rune to research and gave it its own research

J3fftw1:
## Description

Added lightning rune to the research.
## Changes

Added lightning rune to the research.
because Cookie stated `I actually think that the Lightning Rune should get a seperate research.`

## Testability

- [ x] I have fully tested the proposed changes and promise that they will not break everything into chaos.


TheBusyBiscuit:
I think 10 levels is way too low, more like 24...

J3fftw1:
updated it to 24

PR #1173, opened 2019-10-09
Allow BlockPlacer to place blocks while retaining certain data

### LINKED ISSUE #1164: Block placer put empty shulker boxes
## Description (Required)
Using a block placer with a shulker box containing items place an empty shulker box instead

- fill a shulker box with any items
- put it in a block placer
- put the shulker box in the block placer
- activate the block placer
- the shulker box is empty

the placed shulker box should be the same (filled) as the one put in the block placer

## Server Log / Error Report
No console errors to report

## Environment (Required)

 - Minecraft Version: git-paper-192
 - Slimefun Version: vDEV 276
 - CS-CoreLib Version:

AlexLander123:
## Description

 Allow BlockPlacer to place blocks while retaining certain data of the block.

## Changes

BlockPlacer will set block's name and inventory according to its itemstack after placing it.

## Related Issues

Resolves #1164 

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
@@ -57,7 +57,7 @@ public class BlockPlacer extends SimpleSlimefunItem<AutonomousMachineHandler> {
 					}
 					else {
 						block.setType(e.getItem().getType());
-						if (e.getItem().getItemMeta() instanceof BlockStateMeta) {
+						if (e.getItem().hasItemMeta() && e.getItem().getItemMeta() instanceof BlockStateMeta) {
 							BlockState itemBlockState = ((BlockStateMeta) e.getItem().getItemMeta()).getBlockState();
 							BlockState blockState = block.getState();
 							
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
@@ -61,7 +61,7 @@ public class BlockPlacer extends SimpleSlimefunItem<AutonomousMachineHandler> {
 							BlockState itemBlockState = ((BlockStateMeta) e.getItem().getItemMeta()).getBlockState();
 							BlockState blockState = block.getState();
 							
-							if(blockState instanceof Nameable) {
+							if((blockState instanceof Nameable) && e.getItem().getItemMeta().hasDisplayName()) {
 								((Nameable) blockState).setCustomName(e.getItem().getItemMeta().getDisplayName());
 							}
 							
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
@@ -57,11 +57,11 @@ public class BlockPlacer extends SimpleSlimefunItem<AutonomousMachineHandler> {
 					}
 					else {
 						block.setType(e.getItem().getType());
-						if(e.getItem().hasItemMeta() && (e.getItem().getItemMeta() instanceof BlockStateMeta)) {
+						if (e.getItem().getItemMeta() instanceof BlockStateMeta) {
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
```diff
@@ -53,6 +57,23 @@ public AutonomousMachineHandler getItemHandler() {
 					}
 					else {
 						block.setType(e.getItem().getType());
+						if(e.getItem().hasItemMeta() && (e.getItem().getItemMeta() instanceof BlockStateMeta)) {
+							BlockState itemBlockState = ((BlockStateMeta) e.getItem().getItemMeta()).getBlockState();
+							BlockState blockState = block.getState();
+							
+							if(blockState instanceof Nameable) {
```
TheBusyBiscuit: Should probably also check for hasDisplayName on the ItemMeta first

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
```diff
@@ -53,6 +57,23 @@ public AutonomousMachineHandler getItemHandler() {
...
+								((Nameable) blockState).setCustomName(e.getItem().getItemMeta().getDisplayName());
+							}
+							
+							//Update block state after changing name
+							blockState.update();
```
TheBusyBiscuit: Shouldnt .update() be called at the end? 
Even if it doesn't affect the inventory, it is more logical to place it below.

AlexLander123: It does affect the inventory as the inventory from the block will be different after .update(). Thus, I have to update() first to apply the name and then edit the latest inventory.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
AlexLander123: Will do.

TheBusyBiscuit: Ah, please add a comment to the code explaining that then.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
WalshyDev: It's redundant updating in general considering you aren't retrieving the state again after this, you're re-using the same one.

AlexLander123: It will get the Inventory based on the block instead of the blockstate but I will just change it to get a new/latest block state instead to avoid confusion.

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/BlockPlacer.java
```diff
@@ -53,6 +57,23 @@ public AutonomousMachineHandler getItemHandler() {
 					}
 					else {
 						block.setType(e.getItem().getType());
+						if(e.getItem().hasItemMeta() && (e.getItem().getItemMeta() instanceof BlockStateMeta)) {
```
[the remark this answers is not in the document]
AlexLander123: Thanks! Did not know that getItemMeta() will never return null.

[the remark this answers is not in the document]
TheBusyBiscuit: Well I think you should keep the hasItemName() for that exact reason.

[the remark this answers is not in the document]
AlexLander123: > Well I think you should keep the hasItemName() for that exact reason.

Although hasItemMeta() is redundant because getItemMeta() never returns null?

TheBusyBiscuit:
Since getItemMeta() will not return null you will need to do a null-check
via hasItemMeta() yourself.

Items can of course have no ItemMeta, but .getItemMeta will create a new
instance if no previous one exists, therefore creating a lot of unneeded
objects in memory.

So you do need to add .hasItemMeta(), it is not redundant.

AlexLander123 <notifications@github.com> schrieb am Mi., 9. Okt. 2019,
17:05:

PR #1174, opened 2019-10-11
Implement PlayerProfile async fetch

WalshyDev:
## Description
Implemented async fetch for PlayerProfile

## Changes
PlayerProfile.get(OfflinePlayer, Consumer<PlayerProfile>) - this has been implemented in SlimefunGuide (was lazy though so just changed openMainMenu to take a PlayerProfile and make a openMainMenuAsync). Also added a LockedCategory to accept PlayerProfile so there doesn't need to be extra hashtable lookups. Finally, implemented this into backpacks and moved some dupe code.

Also added opening messages, reason being if it will take 20+ seconds then the player will think nothing is happening and probably just spam or complain or whatever.

## Related Issues
Timeout when player has a large profile file, no GH issue was made.

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
+++ b/src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
@@ -51,12 +51,12 @@ import me.mrCookieSlime.Slimefun.hooks.github.Contributor;
 import me.mrCookieSlime.Slimefun.hooks.github.IntegerFormat;
 
 public final class SlimefunGuide {
-	
+
 	private SlimefunGuide() {}
-	
+
 	private static final int CATEGORY_SIZE = 36;
 	private static final int[] SLOTS = new int[] {0, 2, 3, 5, 6, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35};
-	
+
 	@Deprecated
 	public static ItemStack getItem() {
 		return getItem(SlimefunGuideLayout.CHEST);
@@ -79,15 +79,15 @@ public final class SlimefunGuide {
 	public static ItemStack getItem(boolean book) {
 		return getItem(book ? SlimefunGuideLayout.BOOK: SlimefunGuideLayout.CHEST);
 	}
-	
+
 	@Deprecated
 	public static ItemStack getDeprecatedItem(boolean book) {
 		return new CustomItem(new ItemStack(Material.ENCHANTED_BOOK), "&eSlimefun Guide &7(Right Click)", (book ? "": "&2"), "&rThis is your basic Guide for Slimefun", "&rYou can see all Items added by this Plugin", "&ror its Addons including their Recipes", "&ra bit of information and more");
 	}
-	
+
 	public static void openSettings(Player p, final ItemStack guide) {
 		final ChestMenu menu = new ChestMenu("Settings / Info");
-		
+
 		menu.setEmptySlotsClickable(false);
 		menu.addMenuOpeningHandler(
 				pl -> pl.playSound(pl.getLocation(), Sound.BLOCK_NOTE_BLOCK_HARP, 0.7F, 0.7F)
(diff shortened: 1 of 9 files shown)


src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -231,60 +231,69 @@ public static void openCredits(Player p, final ItemStack guide) {
...
+		if (last == null)
+			openMainMenuAsync(p, true, book, 1);
+		else
 			handleHistory(p, last, book, false);
-		}
 	}
 
-	public static void openMainMenu(final Player p, final boolean survival, final boolean book, final int selected_page) {
+	private static void openMainMenuAsync(final Player player, final boolean survival, final boolean book, final int selected_page) {
+		Slimefun.getLocal().sendMessage(player, "messages.opening-guide");
```
TheBusyBiscuit: I think this message should only be displayed when the Person is opening the guide for the first time.
Otherwise there won't be any delay anyway.

src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
```diff
@@ -116,30 +117,34 @@ private void openBackpack(ItemStack sfItem, int size, PlayerInteractEvent e, Pla
...
+                        }
+                    }
+
+                    if(!SlimefunPlugin.getUtilities().backpack.containsValue(item)) {
+                        PlayerProfile.getBackpack(item).open(p);
+                        p.playSound(p.getLocation(), Sound.ENTITY_HORSE_ARMOR, 1F, 1F);
+                        SlimefunPlugin.getUtilities().backpack.put(p.getUniqueId(), item);
+                    }
+                    else SlimefunPlugin.getLocal().sendMessage(p, "backpack.already-open", true);
+				});
```
TheBusyBiscuit: That indentation has gotten a bit out of hand, hasn't it?

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
WalshyDev: Was thinking of the best way to do this, got a suggestion?

src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
WalshyDev: Welcome to SF code :D

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -726,7 +735,7 @@ public void run(final Player p) {
 		}
 	}
 
-	public static void openSearch(Player player, String searchTerm, boolean cheat, boolean addToHistory) {
+	private static void openSearch(Player player, String searchTerm, boolean cheat, boolean addToHistory) {
```
TheBusyBiscuit: This needs to remain public as it won't compile otherwise.

src/me/mrCookieSlime/Slimefun/SlimefunGuide.java
TheBusyBiscuit: BiConsumer with a Boolean? XD

WalshyDev: Also thought about it but... it'd be useless for 90% of cases so I'd need another method

src/me/mrCookieSlime/Slimefun/listeners/BackpackListener.java
```diff
@@ -82,62 +83,61 @@ else if (e.getCurrentItem() != null && e.getCurrentItem().getType() != Material.
...
-		
+
 		if (item.getAmount() == 1) {
-			if (Slimefun.hasUnlocked(p, sfItem, true)) {
-				Slimefun.getLocal().sendMessage(p, "messages.opening-backpack");
-				PlayerProfile.get(p, profile -> {
-                    for (int line = 0; line < item.getItemMeta().getLore().size(); line++) {
-                        if (item.getItemMeta().getLore().get(line).equals(ChatColor.translateAlternateColorCodes('&', "&7ID: <ID>"))) {
-                            BackpackInventory backpack = profile.createBackpack(size);
+			if (Slimefun.hasUnlocked(p, item, true)) {
```
TheBusyBiscuit: Could be improved if you use the SlimefunItem variant here, otherwise another call to SlimefunItem.getByItem() is made.
Therefore you should pass the newly added SlimefunItem as a parameter.

PR #1175, opened 2019-10-13
Fixed NullPointerException when using Pickaxe of the Seeker

AlexLander123:
## Description

Fixed NullPointerException caused by Java evaluating:
```Java
if (p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && closest == null || p.getLocation().distance(closest.getLocation()) < p.getLocation().distance(p.getLocation().getBlock().getRelative(x, y, z).getLocation()))
```
as
```Java
if ((p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && closest == null) || p.getLocation().distance(closest.getLocation()) < p.getLocation().distance(p.getLocation().getBlock().getRelative(x, y, z).getLocation()))
```
causing the location distance check after the or operator to generate a NullPointerException because closest is null.

## Changes

Added brackets to allow proper evaluation of the checks.

## Related Issues

None

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/PickaxeOfTheSeeker.java
+++ b/src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/PickaxeOfTheSeeker.java
@@ -27,7 +27,7 @@ public class PickaxeOfTheSeeker extends SimpleSlimefunItem<ItemInteractionHandle
 				for (int x = -4; x <= 4; x++) {
 					for (int y = -4; y <= 4; y++) {
 						for (int z = -4; z <= 4; z++) {
-							if (p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && (closest == null || p.getLocation().distance(closest.getLocation()) < p.getLocation().distance(p.getLocation().getBlock().getRelative(x, y, z).getLocation()))) {
+							if (p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && (closest == null || p.getLocation().distanceSquared(closest.getLocation()) < p.getLocation().distanceSquared(p.getLocation().getBlock().getRelative(x, y, z).getLocation()))) {
 								closest = p.getLocation().getBlock().getRelative(x, y, z);
 							}
 						}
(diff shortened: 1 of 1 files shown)


src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/PickaxeOfTheSeeker.java
```diff
@@ -27,7 +27,7 @@ public ItemInteractionHandler getItemHandler() {
 				for (int x = -4; x <= 4; x++) {
 					for (int y = -4; y <= 4; y++) {
 						for (int z = -4; z <= 4; z++) {
-							if (p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && closest == null || p.getLocation().distance(closest.getLocation()) < p.getLocation().distance(p.getLocation().getBlock().getRelative(x, y, z).getLocation())) {
+							if (p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && (closest == null || p.getLocation().distance(closest.getLocation()) < p.getLocation().distance(p.getLocation().getBlock().getRelative(x, y, z).getLocation()))) {
```
TheBusyBiscuit: Another thing: If you wanna optimize the performance, use distanceSquared instead of distance

src/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/PickaxeOfTheSeeker.java
```diff
@@ -27,7 +27,7 @@ public ItemInteractionHandler getItemHandler() {
 				for (int x = -4; x <= 4; x++) {
 					for (int y = -4; y <= 4; y++) {
 						for (int z = -4; z <= 4; z++) {
-							if (p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && closest == null || p.getLocation().distance(closest.getLocation()) < p.getLocation().distance(p.getLocation().getBlock().getRelative(x, y, z).getLocation())) {
+							if (p.getLocation().getBlock().getRelative(x, y, z).getType().toString().endsWith("_ORE") && (closest == null || p.getLocation().distanceSquared(closest.getLocation()) < p.getLocation().distanceSquared(p.getLocation().getBlock().getRelative(x, y, z).getLocation()))) {
```
TheBusyBiscuit: Now that I am fully awake:
Isn't the < supposed to be a > too?

AlexLander123: That is absolutely right..... I totally followed the original code and overlooked that :sweat_smile:

PR #1176, opened 2019-10-13
Order timings

WalshyDev:
## Description
Order timings when needed by largest time. Also add in an average per machine.

## Changes
Basically what I said in description

## Related Issues
N/A

## Testability

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/api/TickerTask.java
+++ b/src/me/mrCookieSlime/Slimefun/api/TickerTask.java
@@ -6,6 +6,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.logging.Level;
@@ -20,6 +21,7 @@ import org.bukkit.block.Block;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 
+import io.github.thebusybiscuit.cscorelib2.chat.ChatColors;
 import me.mrCookieSlime.CSCoreLibPlugin.general.Chat.TellRawMessage;
 import me.mrCookieSlime.CSCoreLibPlugin.general.Chat.TellRawMessage.HoverAction;
 import me.mrCookieSlime.Slimefun.SlimefunPlugin;
@@ -200,20 +202,20 @@ public class TickerTask implements Runnable {
 		sender.sendMessage(ChatColor.translateAlternateColorCodes('&', "&6Skipped Machines: &e" + skipped));
 		sender.sendMessage("");
 		sender.sendMessage(ChatColor.translateAlternateColorCodes('&', "&6Ticking Machines:"));
-		
+
+		List<Map.Entry<String, Long>> timings = machineCount.keySet().stream()
+				.map(key -> new AbstractMap.SimpleEntry<>(key, machineTimings.getOrDefault(key, 0L)))
+				.sorted((o1, o2) -> o2.getValue().compareTo(o1.getValue()))
+				.collect(Collectors.toList());
+
 		if (sender instanceof Player) {
 			TellRawMessage tellraw = new TellRawMessage();
 			tellraw.addText("   &7&oHover for more Info");
 			StringBuilder hover = new StringBuilder();
 			int hidden = 0;
 
-			Map<String, Long> timings = machineCount.keySet().stream()
-					.map(key -> new AbstractMap.SimpleEntry<>(key, machineTimings.getOrDefault(key, 0L)))
-					.sorted((o1, o2) -> o2.getValue().compareTo(o1.getValue()))
-					.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
-
-			for (Map.Entry<String, Long> entry : timings.entrySet()) {
-				int count = machineCount.getOrDefault(entry.getKey(), 0);
+			for (Map.Entry<String, Long> entry : timings) {
+				int count = machineCount.get(entry.getKey());
 				if (entry.getValue() > 0)
 				    hover.append("\n&c").append(entry.getKey()).append(" - ")
 						.append(count).append("x &7(").append(entry.getValue()).append("ms, ")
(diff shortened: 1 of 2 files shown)


src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -202,20 +205,27 @@ public void info(CommandSender sender) {
...
 			StringBuilder hover = new StringBuilder();
 			int hidden = 0;
-			
-			for (Map.Entry<String, Integer> entry: machineCount.entrySet()) {
-				long timings = machineTimings.get(entry.getKey());
-				if (timings > 0) hover.append("\n&c").append(entry.getKey()).append(" - ").append(entry.getValue()).append("x &7(").append(timings).append("ms)");
-				else hidden++;
+
+			Map<String, Long> timings = machineCount.entrySet().stream()
+					.sorted(Comparator.comparingLong(machineTimings::get).reversed())
```
TheBusyBiscuit: This is a Stream of Map.Entry<String, Long>
machineTimings::get will always return null as that map's key is not an entry...
You need to pass the entry's key instead of just giving it the function there.

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -202,20 +205,27 @@ public void info(CommandSender sender) {
...
+					.collect(Collectors.toMap(Map.Entry::getKey, machineTimings::get, (e1, e2) -> e1, LinkedHashMap::new));
```
TheBusyBiscuit: Why do you use machineTimings:: get as the value?
Aren't you discarding the actual values from the map there...

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -202,20 +205,27 @@ public void info(CommandSender sender) {
...
+
+			for (Map.Entry<String, Long> entry : timings.entrySet()) {
+				if (entry.getValue() > 0)
+				    hover.append("\n&c").append(entry.getKey()).append(" - ")
+						.append(entry.getValue()).append("x &7(").append(entry.getValue()).append("ms, ")
```
TheBusyBiscuit: entry.getValue() no longer refers to the value it is supposed to refer to.

WalshyDev: Yep noticed and fixed that already

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -202,20 +206,29 @@ public void info(CommandSender sender) {
...
-				if (timings > 0) hover.append("\n&c").append(entry.getKey()).append(" - ").append(entry.getValue()).append("x &7(").append(timings).append("ms)");
-				else hidden++;
+
+			Map<String, Long> timings = machineCount.keySet().stream()
+					.map(key -> new AbstractMap.SimpleEntry<>(key, machineTimings.getOrDefault(key, 0L)))
+					.sorted((o1, o2) -> o2.getValue().compareTo(o1.getValue()))
+					.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
+
+			for (Map.Entry<String, Long> entry : timings.entrySet()) {
+				int count = machineCount.getOrDefault(entry.getKey(), 0);
```
TheBusyBiscuit: No need for "getOrDefault" here since you literally took the contents of that map to create this map, so you can safely use .get() here

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -202,20 +206,29 @@ public void info(CommandSender sender) {
...
-			
-			for (Map.Entry<String, Integer> entry: machineCount.entrySet()) {
-				long timings = machineTimings.get(entry.getKey());
-				if (timings > 0) hover.append("\n&c").append(entry.getKey()).append(" - ").append(entry.getValue()).append("x &7(").append(timings).append("ms)");
-				else hidden++;
+
+			Map<String, Long> timings = machineCount.keySet().stream()
+					.map(key -> new AbstractMap.SimpleEntry<>(key, machineTimings.getOrDefault(key, 0L)))
+					.sorted((o1, o2) -> o2.getValue().compareTo(o1.getValue()))
+					.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
```
TheBusyBiscuit: If you care about it being sorted and need the entries themselves anyway, why even collect this as a map and not a List of Entries?

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -202,20 +206,29 @@ public void info(CommandSender sender) {
...
 			int hidden = 0;
-			
-			for (Map.Entry<String, Integer> entry: machineCount.entrySet()) {
-				long timings = machineTimings.get(entry.getKey());
-				if (timings > 0) hover.append("\n&c").append(entry.getKey()).append(" - ").append(entry.getValue()).append("x &7(").append(timings).append("ms)");
-				else hidden++;
+
+			Map<String, Long> timings = machineCount.keySet().stream()
+					.map(key -> new AbstractMap.SimpleEntry<>(key, machineTimings.getOrDefault(key, 0L)))
+					.sorted((o1, o2) -> o2.getValue().compareTo(o1.getValue()))
```
TheBusyBiscuit: Couldn't you just use Map.Entry.comparingByValue here too?

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -238,14 +251,19 @@ public void info(CommandSender sender) {
 			tellraw.addText("   &7&oHover for more Info");
 			StringBuilder hover = new StringBuilder();
 			int hidden = 0;
-			
-			for (Map.Entry<String, Long> entry: chunkTimings.entrySet()) {
+
+			Map<String, Long> timings = chunkTimings.entrySet().stream()
+                    .sorted(Map.Entry.comparingByValue(Comparator.reverseOrder()))
+                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue, (e1, e2) -> e1, LinkedHashMap::new));
```
TheBusyBiscuit: If you care about it being sorted and need the entries themselves anyway, why even collect this as a map and not a List of Entries?

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
WalshyDev: Yeah I had it as that but good ol' Java generics got quite messed up and (despite being fine) failed compilation.

src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
```diff
@@ -211,6 +211,8 @@ public void onRightClick(ItemUseEvent e) {
 		final Player p = e.getPlayer();
 		ItemStack item = e.getItem();
 
+		if (slimefunItem == null) return;
```
TheBusyBiscuit: This would prevent the Guide from opening and the debug fish from working, just remove it.
I already fixed it via a commit.

src/me/mrCookieSlime/Slimefun/api/TickerTask.java
```diff
@@ -196,56 +202,73 @@ public void info(CommandSender sender) {
 		sender.sendMessage(ChatColor.translateAlternateColorCodes('&', "&6Skipped Machines: &e" + skipped));
 		sender.sendMessage("");
 		sender.sendMessage(ChatColor.translateAlternateColorCodes('&', "&6Ticking Machines:"));
-		
+
+		List<Map.Entry<String, Long>> timings = machineCount.keySet().stream()
+				.map(key -> new AbstractMap.SimpleEntry<>(key, machineTimings.getOrDefault(key, 0L)))
+				.sorted((o1, o2) -> o2.getValue().compareTo(o1.getValue()))
```
TheBusyBiscuit: Again: Couldn't you just use "Map.Entry.comparingByValue" here too?

WalshyDev: Again, feel free to try it yourself. At least with my javac despite it seeming completly fine the generics casue it to fail. It thinks it's like <String, Object> and then I'd need to cast and then it fails. This works fine regardless

TheBusyBiscuit: Oh sorry, didn't see your previous reponse to it. That's fine then.

PR #1181, opened 2019-10-14
Ignore if sfitem is null, instead of cancelling it

### LINKED ISSUE #1180: Players unable to place/interact
## Description (Required)
Players are unable to build/interact in one of the latest dev builds. My guess is that it's a cancelled PlayerInteractEvent, but I am currently unable to check the code myself. I have tested on various spigot/paper versions on a few machines and the issue always persists. Downgrading to the latest stable build fixed it.

Install [dev build 303](https://thebusybiscuit.github.io/builds/TheBusyBiscuit/Slimefun4/master/#303)

PlayerInteractEvent wouldn't be cancelled.

## Server Log / Error Report
**There are no error messages whatsoever.**

## Environment (Required)
Paper v1_14_R1
 - Minecraft Version: 1.14.4
 - Slimefun Version: DEV 303 (git 090f45d9)
 - CS-CoreLib Version: 1.6.1

huynhqtienvtag:
## Description

When sfitem if null, the event is cancelled and players can't interact with anything
This PR will move the check seperately so it could just ignore if sfitem if null
Didn't test it yet

## Changes

Ignore if sfitem if null

## Related Issues

Fixes #1180

## Testability

* [ ] I have fully tested the proposed changes and promise that they will not break everything into chaos.
* [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.

### DIFF
--- a/src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
+++ b/src/me/mrCookieSlime/Slimefun/listeners/ItemListener.java
@@ -227,7 +227,10 @@ public class ItemListener implements Listener {
 		else if (SlimefunManager.isItemSimiliar(item, SlimefunItems.DEBUG_FISH, true)) {
 			// Ignore the debug fish in here
 		}
-		else if (slimefunItem != null && Slimefun.hasUnlocked(p, slimefunItem, true)) {
+		else if (slimefunItem != null) {
+			// Ignore null sfItem
+		}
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
1. Empty code is really really bad practice.
2. You got your null check wrong, you are ignoring all Slimefun Items, instead of ignoring non-Slimefun Items, that would make it much worse

PR #1222, opened 2019-10-31
Changed method of accessing reactor from access port

AlexLander123:
## Description

Changed method of accessing reactor from access port. The change is made to stop inventory.no-access message when opening access port due to canOpen(Block, Player) returning false.

## Changes

Added a button to access the reactor from within the access port instead of automatically opening the reactor when it is not empty.

## Related Issues

None

## Testability

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
@@ -50,17 +50,18 @@ public class ReactorAccessPort extends SlimefunItem {
 				if (reactor != null) {
 					menu.replaceExistingItem(INFO_SLOT, new CustomItem(new ItemStack(Material.GREEN_WOOL), "&7Reactor", "", "&6Detected", "", "&7> Click to view Reactor"));
 					menu.addMenuClickHandler(INFO_SLOT, (p, slot, item, action) -> {
-						reactor.open(p);
+						if(reactor != null) {
+							reactor.open(p);
+						}
 						newInstance(menu, b);
 
 						return false;
 					});
 				} 
 				else {
-					menu.replaceExistingItem(INFO_SLOT, new CustomItem(new ItemStack(Material.RED_WOOL), "&7Reactor", "", "&cNot detected", "", "&7Reactor must be", "&7placed 3 blocks below", "&7a access port!"));
+					menu.replaceExistingItem(INFO_SLOT, new CustomItem(new ItemStack(Material.RED_WOOL), "&7Reactor", "", "&cNot detected", "", "&7Reactor must be", "&7placed 3 blocks below", "&7the access port!"));
 					menu.addMenuClickHandler(INFO_SLOT, (p, slot, item, action) -> {
 						newInstance(menu, b);
-						menu.open(p);
 						return false;
 					});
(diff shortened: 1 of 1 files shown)


src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -39,39 +41,29 @@ public void init() {
...
+					menu.replaceExistingItem(INFO_SLOT, new CustomItem(new ItemStack(Material.GREEN_WOOL), "&7Reactor", "", "&6Detected", "", "&7> Click to view Reactor"));
+					menu.addMenuClickHandler(INFO_SLOT, (p, slot, item, action) -> {
+						reactor.open(p);
+						newInstance(menu, b);
+
+						return false;
+					});
+				} 
+				else {
+					menu.replaceExistingItem(INFO_SLOT, new CustomItem(new ItemStack(Material.RED_WOOL), "&7Reactor", "", "&cNot detected", "", "&7Reactor must be", "&7placed 3 blocks below", "&7a access port!"));
```
TheBusyBiscuit: Little typo there:

"a access port".
Should rather be "an" or I'd even suggest to use "the" or "this".

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -39,39 +41,29 @@ public void init() {
...
+				return p.hasPermission("slimefun.inventory.bypass") || SlimefunPlugin.getProtectionManager().hasPermission(p, b.getLocation(), ProtectableAction.ACCESS_INVENTORIES);
+			}
+			
+			@Override
+			public void newInstance(final BlockMenu menu, final Block b) {
+				BlockMenu reactor = getReactorMenu(b.getLocation());
+				if (reactor != null) {
+					menu.replaceExistingItem(INFO_SLOT, new CustomItem(new ItemStack(Material.GREEN_WOOL), "&7Reactor", "", "&6Detected", "", "&7> Click to view Reactor"));
+					menu.addMenuClickHandler(INFO_SLOT, (p, slot, item, action) -> {
+						reactor.open(p);
```
TheBusyBiscuit: This could potentially cause a NullPointerException if the Reactor was broken.
So there needs to be a null-check here!

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/ReactorAccessPort.java
```diff
@@ -39,39 +41,29 @@ public void init() {
...
+						newInstance(menu, b);
+
+						return false;
+					});
+				} 
+				else {
+					menu.replaceExistingItem(INFO_SLOT, new CustomItem(new ItemStack(Material.RED_WOOL), "&7Reactor", "", "&cNot detected", "", "&7Reactor must be", "&7placed 3 blocks below", "&7a access port!"));
+					menu.addMenuClickHandler(INFO_SLOT, (p, slot, item, action) -> {
+						newInstance(menu, b);
+						menu.open(p);
```
TheBusyBiscuit: Why would you open the menu to the Player here? The Player must be viewing this Menu already...

PR #1240, opened 2019-11-11
Call BlockBreakEvent in ExplosivePickaxe and ExplosiveShovel

js6pak:
## Description
Made explosive tools compatible with other plugins listening to BlockBreakEvent

## Changes
Added BlockBreakEvent call in ExplosivePickaxe and ExplosiveShovel

## Testability

### DIFF
--- a/.github/pull.yml
+++ b/.github/pull.yml
@@ -1,0 +1,9 @@
+version: "1"
+rules:
+  - base: master
+    upstream: TheBusyBiscuit:master
+    mergeMethod: hardreset
+  - base: explosives-blockbreakevent
+    upstream: TheBusyBiscuit:master
+    assignees:
+      - js6pak
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/ExplosivePickaxe.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/ExplosivePickaxe.java
@@ -1,17 +1,8 @@
 package me.mrCookieSlime.Slimefun.Objects.SlimefunItem.items;
 
-import java.util.List;
-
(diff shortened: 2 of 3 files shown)


WalshyDev:
Not sure that firing BlockBreakEvent is a good idea. Some plugins may do quite a few operations here and firing it 9/18/27 times is just ew.

TheBusyBiscuit:
Not to mention that this will trigger a LOT of Anti-Cheat plugins out there and noone wants to add compatibility to all of those...

WalshyDev:
If you want this then make a custom event

js6pak:
But still no one wants to add custom event compatibility to every plugin using block break event, maybe make it optional in config or something?

WalshyDev:
It will break in more plugins than it will help. I can't see a valid use case which couldn't just implement a custom event. 

Please explain your usage

js6pak:
Custom block drops, block logger not supported by CSCoreLib

WalshyDev:
Can do it with event priority and "block logger" isn't that helpful but again, priority solves it.

TheBusyBiscuit:
CSCoreLib DOES actually support block loggers, just only LogBlock and CoreProtect, you can add an implementation for the plugin you are using yourself.
https://github.com/TheBusyBiscuit/CS-CoreLib2/tree/master/src/main/java/io/github/thebusybiscuit/cscorelib2/protection/loggers

js6pak:
I know, that's why I said 

> block logger **not supported by CSCoreLib**

and if block logger is closed source then it's just easier to add block break event here that implementing it in cscorelib
But custom block drops are still most important for me

WalshyDev:
Event Priority!

js6pak:
Sorry but I don't know what you mean by event priority

PR #1241, opened 2019-11-12
[WIP] Fix changing block ID

creator3:
## Description
```
Ignoring duplicate block @ -549, 66, -47665
Old block data: {"id":"IRON_PLANT"}
New block data (SurvivalWorld;-549;66;-47665): {"id":"IRON_ESSENCE"}
```

The server I play on has had duplicate block warnings every startup for as long as I can remember.
I guess the warnings must seem pretty low severity because neither I nor anyone else seems to have really looked into this much. Basically, the key point is that the id has changed. Apparently I didn't understand that changing the id was a possibility when I made #718 ("Parse block info at startup") because I removed a call to `_integrated_removeBlockInfo(location, false);` from `setBlockInfo`. That call was responsible for cleaning up the old entry (under the previous id) from the `blocksCache`. This (untested) PR should restore the old behavior, but I don't know how well that worked. It looks like the ticker didn't get updated, which may or may not cause problems in practice. If a BlockStorage rewrite is still in the works, it might be okay to just slap a quick fix together. Otherwise, we should probably be more deliberate to avoid making BlockStorage even more complicated and hard to maintain.

## Changes
This PR mostly adds up to adding back the call to `_integrated_removeBlockInfo(location, false);`
More precisely, it just copies the call to refreshCache that actually cleared the entry

To support this, we need access to the previous block ID. BlockStorage will now return copies of BlockInfoConfig instead of the same mutable object that's in the storage map.
There is some extra copying, but I don't think it will have too big of a performance impact. As you can see, there's code for a "lazy" copy that avoids copying the Map, which might reduce the impact on block info reads without associated writes. Or more likely, it's a bad micro-optimization, and we should just copy the whole thing all the time.

## Testability
I haven't tested it yet because I thought there might be some discussion about what the fix should look like.

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockInfoConfig.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockInfoConfig.java
@@ -13,22 +13,46 @@ import com.google.gson.GsonBuilder;
 import me.mrCookieSlime.CSCoreLibPlugin.Configuration.Config;
 
 public class BlockInfoConfig extends Config {
-	
+
 	private Map<String, String> data;
-	
+	private boolean isMapShared;
+	private boolean readOnly = false;
+
 	public BlockInfoConfig() {
-		this(new HashMap<>());
+		this(new HashMap<>(), false);
 	}
-	
+
 	public BlockInfoConfig(Map<String, String> data) {
+		this(data, false);
+	}
+
+	private BlockInfoConfig(Map<String, String> data, boolean isMapShared) {
 		super(null, null);
 		this.data = data;
+		this.isMapShared = isMapShared;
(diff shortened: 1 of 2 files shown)


ajan-12:
This issue is actually caused by Exotic Garden...

BluGhostYT:
@creator3 , if you can, could you please test this with ExoticGarden? The reason being is because these issues mainly occurred in that specific add-on.

creator3:
Closing this for now since there doesn't seem to be any reason to bother restoring the old behavior now that ExoticGarden doesn't rely on it (fixed in [ExoticGarden#94](https://github.com/TheBusyBiscuit/ExoticGarden/pull/94))

PR #1244, opened 2019-11-16
AutoDisenchantEvent

poma123:
## Description

This is a new custom async event, which is very necessary for servers using McMMO and Slimefun. If you use McMMO, you can add +5 value of Efficiency enchant to axes/pickaxes/spades. If you put it in an item frame, and break it in the way like the enchanted item can fall into a hopper, then with a cargo it goes into the disenchanter and finally you can duplicate infinite amount of enchants.

## Changes

- Added DisenchanterEvent, which is a cancellable event. With this, a new feature could be made, with you can configure whether an item could or couldn't be disenchanted.
- Added a check in AutoDisenchanter, so if the event was cancelled then the function returns.

## Testability

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
@@ -5,7 +5,7 @@ import org.bukkit.event.Event;
 import org.bukkit.event.HandlerList;
 import org.bukkit.inventory.ItemStack;
 
-public class DisenchanterEvent extends Event implements Cancellable {
+public class AutoDisenchantEvent extends Event implements Cancellable {
 	
 	private static final HandlerList handlers = new HandlerList();
 
@@ -20,7 +20,7 @@ public class DisenchanterEvent extends Event implements Cancellable {
         return handlers;
     }
 	
-	public DisenchanterEvent(ItemStack sfitem){
+	public AutoDisenchantEvent(ItemStack sfitem){
 		super(true);
 		this.sfItem = sfitem;
 	}
--- a/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
@@ -12,7 +12,7 @@ public class AutoDisenchantEvent extends Event implements Cancellable {
 	private final ItemStack item;
 	private boolean cancelled;
 
-	public AutoDisenchantEvent(ItemStack item){
+	public AutoDisenchantEvent(ItemStack item) {
 		super(true);
 		this.item = item;
 	}
--- a/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
@@ -12,18 +12,19 @@ public class AutoDisenchantEvent extends Event implements Cancellable {
 	private final ItemStack item;
 	private boolean cancelled;
 
-	public HandlerList getHandlers() {
-	    return handlers;
-	}
-	
-	public static HandlerList getHandlerList() {
-        return handlers;
-    }
-	
 	public AutoDisenchantEvent(ItemStack item){
 		super(true);
 		this.item = item;
 	}
+
+	public HandlerList getHandlers() {
+	    return handlers;
+	}
+
+	public static HandlerList getHandlerList() {
+        return handlers;
+    }
+    
 	
 	public ItemStack getItem() {
 		return this.item;
--- a/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
@@ -9,7 +9,7 @@ public class AutoDisenchantEvent extends Event implements Cancellable {
 	
 	private static final HandlerList handlers = new HandlerList();
 
-	private final ItemStack sfItem;
+	private final ItemStack item;
 	private boolean cancelled;
 
 	public HandlerList getHandlers() {
@@ -20,13 +20,13 @@ public class AutoDisenchantEvent extends Event implements Cancellable {
         return handlers;
     }
 	
-	public AutoDisenchantEvent(ItemStack sfitem){
+	public AutoDisenchantEvent(ItemStack item){
 		super(true);
-		this.sfItem = sfitem;
+		this.item = item;
 	}
 	
-	public ItemStack getSfItem() {
-		return this.sfItem;
+	public ItemStack getItem() {
+		return this.item;
 	}
 
 	@Override
--- a/src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
@@ -1,20 +1,15 @@
 package me.mrCookieSlime.Slimefun.Events;
 
-import me.mrCookieSlime.Slimefun.Objects.MultiBlock;
-
-import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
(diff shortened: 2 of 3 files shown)


J3fftw1:
Wouldnt it be better to check if the item has the lore that mcmmo gives to cancel the disenchanting/ enchanting proccses?

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutoDisenchanter.java
```diff
@@ -94,6 +97,12 @@ protected void tick(Block b) {
 					return;
 				}
 
+				DisenchanterEvent event = new DisenchanterEvent(item);
+				Bukkit.getPluginManager().callEvent(event);
+				if ((event).isCancelled()) {
```
WalshyDev: why are there brackets around `event`?

poma123:
You're talking about the handling of the event? If yes, the code above just an example, it's up to you how you handle it. The more important part is the calling of the event I think

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+	public HandlerList getHandlers() {
+	    return handlers;
+	}
+	
+	public static HandlerList getHandlerList() {
+        return handlers;
+    }
+	
+	public DisenchanterEvent(ItemStack sfitem){
```
WalshyDev: Constructor goes before normal methods

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.bukkit.inventory.ItemStack;
+
+public class DisenchanterEvent extends Event implements Cancellable {
+	
+	private static final HandlerList handlers = new HandlerList();
+
+	private ItemStack sfItem;
```
WalshyDev: This should be final as it cannot change.

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+	public HandlerList getHandlers() {
+	    return handlers;
+	}
+	
+	public static HandlerList getHandlerList() {
+        return handlers;
+    }
+	
+	public DisenchanterEvent(ItemStack sfitem){
+		super(true);
```
WalshyDev: Is this async? I don't think it is so why specify it as so?

J3fftw1:
Mhm i think you miss understood me let me give a example.

If tool has lore 'mcMMO Ability Tool' do not disenchant or enchant

poma123:
Yeah, that's a solution too :D

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
poma123: It's async, it have to bee async.

WalshyDev:
Also, for my sake of mind

```java
Map<Enchantment, Integer> enchants = localMeta.getEnchants();
for (Enchantment ench : enchants.keySet()) {
    // If enchantment value is greater than five
    if (enchants.get(ench) > 5) {
        e.setCancelled(true);
    }
}
```

The enchants variable is redundant and you get the keys just to then use the key for the value. This means you're creating a set and THEN also doing a lookup. This can be improved perfoemance wise with simply using `values()` or even `entrySet()` if you still wanted the Enchantment.

J3fftw1:
It would cut out any form of duping if you check for set lores or pull it from mcmmo since people can change it to there own langauge i think

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+		this.sfItem = sfitem;
+	}
+	
+	public ItemStack getSfItem() {
```
TheBusyBiscuit: Don't use abbreviations in the method name here, just do getItem()
Especially because this is not a SlimefunItem, it just an ItemStack.

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
```diff
@@ -0,0 +1,42 @@
...
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.bukkit.inventory.ItemStack;
+
+public class DisenchanterEvent extends Event implements Cancellable {
+	
+	private static final HandlerList handlers = new HandlerList();
+
+	private final ItemStack sfItem;
```
TheBusyBiscuit: This field should just be called "item" since it represents an ItemStack and not a Slimefunitem.
It's misleadingly named.

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
TheBusyBiscuit: ^ What he said.

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
```diff
@@ -0,0 +1,42 @@
+package me.mrCookieSlime.Slimefun.Events;
+
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.bukkit.inventory.ItemStack;
+
+public class DisenchanterEvent extends Event implements Cancellable {
```
TheBusyBiscuit: I'd rather rename this to "AutoDisenchantEvent", since it represents the "disenchantment", not the "disenchanter" and the machine is also called "Auto-Disenchanter".

WalshyDev:
@TheBusyBiscuit Opinion about putting this in an `event` package rather than putting it in the current name-convention breaking package. Bit confusing but a step towards just generally a more compliant plugin.

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
poma123: I'm sorry, but I don't know what you mean. I'm not a native English speaker, I've never used this literally, I've always created events like this. Can you tell me, without any judgment, what should I change about this?

WalshyDev: Literally move this above the other methods.

TheBusyBiscuit:
@WalshyDev Feel free to do it in a follow-up PR, since the other Events should be moved too.

src/main/java/me/mrCookieSlime/Slimefun/Events/DisenchanterEvent.java
poma123: Okay, thanks

src/main/java/me/mrCookieSlime/Slimefun/Events/AutoDisenchantEvent.java
```diff
@@ -0,0 +1,42 @@
...
+
+	public HandlerList getHandlers() {
+	    return handlers;
+	}
+	
+	public static HandlerList getHandlerList() {
+        return handlers;
+    }
+	
+	public AutoDisenchantEvent(ItemStack item){
```
TheBusyBiscuit: As stated earlier, the constructor should come first. Before the methods.

poma123: Done

TheBusyBiscuit:
@poma123 Also, please give your commits appropriate names, not just "Update AutoDisenchantEvent" all the time...

andris155:
Thanks @poma123

PR #1246, opened 2019-11-16
added wiki to the slimefun guide

J3fftw1:
## Description

added wiki link to the slimefun guide in shift click setting
## Changes

added wiki link to the slimefun guide in shift click setting
## Testability


src/main/java/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -157,6 +157,15 @@ else if (SlimefunManager.isItemSimiliar(guide, getItem(SlimefunGuideLayout.CHEAT
 			return false;
 		});
 
+		menu.addItem(22, new CustomItem(new ItemStack(Material.ENCHANTED_BOOK), "&eSlimefun Wiki", "", "&7\u21E8 Click to go to the Slimefun Wiki"));
+		menu.addMenuClickHandler(22, (pl, slot, item, action) -> {
+			pl.closeInventory();
+			pl.sendMessage("");
+			pl.sendMessage(ChatColor.translateAlternateColorCodes('&', "&7&ohttps://github.com/TheBusyBiscuit/Slimefun4/wiki"));
```
WalshyDev: First off... why. Such a long line for literally just `ChatColor.GRAY.toString() + ChatColor.ITALIC()` but... why even italic?

Not sure I like the empty lines above and below a random link either. If it was like `[Click here] to go to the Slimefun4 wiki` it would be a little better.

J3fftw1: i did it the same as the bug tracker code but without showing open bug reports and open pr's

WalshyDev:
I still maintain it shouldn't just be the link plonked in chat. A message like `[Click here] to go to the Slimefun4 wik` would be much nicer imo.

PR #1248, opened 2019-11-16
Fix #1229

### LINKED ISSUE #1229: Talismen of the Miner and Talismen of the Hunter sending the "doubled drops" message too often.
## Description (Required)
If you are holding a Talismen of the Miner in your inventory, mining an ore block will gives the message "Your Talismen has just doubled your drops" message nearly every time. After some in-game testing, the message seems to appear approximately 90% of time time when mining an ore, which is much larger than the 20% occurrence rate the Talismen says should happen in the item's description.
The Talismen of the Hunter has a similar issue, with the same 90% occurrence rate of the "doubled drops" message when killing a mob.
Note that this issue also occurs for the Ender versions of these Talismens as well.

Demonstration Video for Talismen of the Miner: [Screenshot from Gyazo](https://gyazo.com/f2e275a2355913a18055893fdd18a33a)
Steps for the Talismen of the Miner:
1. Obtain a pickaxe and Talismen of the Miner
2. Find an ore and mine it
Steps for Talismen of the Hunter:
1. Obtain a Talismen of the Hunter
2. Kill a mob 
The only items used in the demonstration video are a diamond pickaxe enchanted with Efficiency 4, Unbreaking 3, Fortune 3, and Mending, and a Talismen of the Miner. Of the 9 coal ore blocks that were broken in the video, 8 of them gave the "doubled drops" message.
It's unreasonable to believe that this occurring is statistically plausible, since even with the small sample size shown in the video we can determine that the odds of this happening are less than 0.002%.
image

The "doubled drops" message should appear roughly 20% of the time and only when the drops have actually been doubled.

## Server Log / Error Report
N/A

## Environment (Required)
Versions are available in the below screenshot as well as being written below:
image

 - Minecraft Version: 1.14.4
 - Slimefun Version: vDEV - 328
 - CS-CoreLib Version: vDEV - 77

WalshyDev:
## Description
Huge changes to solve this simple issue.

## Changes
I hacked the mainframe, disabled the firewall and re-routed all the interweb traffic to my 7 layer proxy.

## Related Issues
Resolves #1229

## Testability

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/Talisman.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/Talisman.java
@@ -123,7 +123,7 @@ public class Talisman extends SlimefunItem {
         }
 
         Talisman talisman = (Talisman) item;
-        if (ThreadLocalRandom.current().nextInt(100) < talisman.getChance()) {
+        if (ThreadLocalRandom.current().nextInt(100) > talisman.getChance()) {
             return false;
         }
 


creator3:
Should probably be >= honestly, not that it matters much

PR #1254, opened 2019-11-22
add flint to cobblestone recipe

firebl0od:
add flint to cobblestone recipe so we have an electric version of the recipe flint to cobble in the compressor

## Description
just added the flint to cobble recip

## Changes
the electric version of the compressor will take 2 less flint to convert to a cobblestone, but it will take 30sec to do so.
so 6 flint --> 1 cobblestone every 30 sec

## Testability
i didn't test anything


TheBusyBiscuit:
The Carbon Press is really not the machine to add this to...

PR #1262, opened 2019-11-26
added cscorelib version to the guide

J3fftw1:
## Description
added cscorelib to sf guide info panel.
people who have no op can now view the versions to make a proper report
also fixed occurred

## Changes
added cscorelib version to the guide

## Related Issues
NA

## Testability


src/main/java/me/mrCookieSlime/Slimefun/SlimefunGuide.java
```diff
@@ -143,7 +144,15 @@ else if (SlimefunManager.isItemSimiliar(guide, getItem(SlimefunGuideLayout.CHEAT
 			});
 		}
 
-		menu.addItem(1, new CustomItem(new ItemStack(Material.WRITABLE_BOOK), "&aCredits", "", "&7Version: &a" + SlimefunPlugin.instance.getDescription().getVersion(), "&7Contributors: &e" + SlimefunPlugin.getUtilities().contributors.size(), "", "&7\u21E8 Click to see the people behind this Plugin"));
+		menu.addItem(1, new CustomItem(new ItemStack(Material.WRITABLE_BOOK),
+				"&aCredits",
+				"",
+				"&7Slimefun Version: &a" + SlimefunPlugin.instance.getDescription().getVersion(),
+				"&7CSCore-Lib Version: &a" + CSCoreLib.getLib().getDescription().getVersion(),
```
TheBusyBiscuit: You placed the hyphon at the wrong spot.

And I purposefully didn't include that as CS-CoreLib2 is used in most classes now, so we can correctly identify the CS-CoreLib version based on the Slimefun version.

CS-CoreLib itself will be faded out as a dependency and is not needed for future bug reports, even now it shouldn't be a source of error anymore.

J3fftw1: we still require it for bug reports though
and people that use older version are having bugs with it for example the ancient altar bugs on older versions

TheBusyBiscuit: We don't support older versions anyway...

J3fftw1: i mean older versions of CS-CoreLib

WalshyDev:
and merged 😂

PR #1263, opened 2019-11-26
AndroidMineEvent

poma123:
## Description

Added AndroidMineEvent so in a thirdy party plugin we can cancel the mining if it's e.g. a creative placed block.

## Changes

- Added AndroidMineEvent
- Added calling the event inside ProgrammableAndroid#mine()

## Related Issues

NA
## Testability

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Events/AndroidMineEvent.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Events/AndroidMineEvent.java
@@ -1,19 +1,30 @@
 package me.mrCookieSlime.Slimefun.Events;
 
+import me.mrCookieSlime.Slimefun.androids.ProgrammableAndroid;
 import org.bukkit.block.Block;
 import org.bukkit.event.Cancellable;
 import org.bukkit.event.Event;
 import org.bukkit.event.HandlerList;
 
+/**
+ * This event is fired before a miner android mines a block.
+ * If this event is cancelled, the block will not be mined.
+ */
 public class AndroidMineEvent extends Event implements Cancellable {
 
     private static final HandlerList handlers = new HandlerList();
 
     private final Block block;
+    private final ProgrammableAndroid android;
     private boolean cancelled;
 
-    public AndroidMineEvent(Block block) {
+    /**
+     * @param block - mined block
+     * @param android - android
+     */
+    public AndroidMineEvent(Block block, ProgrammableAndroid android) {
         this.block = block;
+        this.android = android;
     }
 
     public static HandlerList getHandlerList() {
@@ -25,12 +36,16 @@ public class AndroidMineEvent extends Event implements Cancellable {
     }
 
     /**
-     * @return the broken block
+     * @return the mined block
      */
     public Block getBlock() {
         return this.block;
     }
 
+    /**
+     * @return the android
+     */
+    public ProgrammableAndroid getAndroid() { return this.android;}
     @Override
     public boolean isCancelled() {
         return this.cancelled;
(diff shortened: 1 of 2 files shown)


src/main/java/me/mrCookieSlime/Slimefun/androids/ProgrammableAndroid.java
```diff
@@ -499,7 +500,11 @@ private void mine(Block b, BlockMenu menu, Block block) {
 		Collection<ItemStack> drops = block.getDrops();
 		if (!blockblacklist.contains(block.getType()) && !drops.isEmpty() && SlimefunPlugin.getProtectionManager().hasPermission(Bukkit.getOfflinePlayer(UUID.fromString(BlockStorage.getLocationInfo(b.getLocation(), "owner"))), block.getLocation(), ProtectableAction.BREAK_BLOCK)) {
 			String item = BlockStorage.checkID(block);
-
+			AndroidMineEvent event = new AndroidMineEvent(block);
+			Bukkit.getPluginManager().callEvent(event);
```
TheBusyBiscuit: Shouldn't this also be called in the "movedig" function?

poma123: Oh right!

src/main/java/me/mrCookieSlime/Slimefun/Events/AndroidMineEvent.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static HandlerList getHandlerList() {
+        return handlers;
+    }
+
+    public HandlerList getHandlers() {
+        return handlers;
+    }
+
+    public Block getBlock() {
```
WalshyDev: Is this the block broken or the android? Could we get the android if not?

You should add javadocs to this since it's an API class.

src/main/java/me/mrCookieSlime/Slimefun/Events/AndroidMineEvent.java
```diff
@@ -0,0 +1,44 @@
+package me.mrCookieSlime.Slimefun.Events;
+
+import org.bukkit.block.Block;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+public class AndroidMineEvent extends Event implements Cancellable {
```
TheBusyBiscuit: As @WalshyDev already pointed out, a method to get the android would be nice.

And granted that this plugin has shown the middle finger to "documentation" so far, it's not a priority, but some javadocs for this class would be really awesome.

src/main/java/me/mrCookieSlime/Slimefun/Events/AndroidMineEvent.java
```diff
@@ -0,0 +1,59 @@
...
+     * @return the mined block
+     */
+    public Block getBlock() {
+        return this.block;
+    }
+
+    /**
+     * @return the android
+     */
+    public ProgrammableAndroid getAndroid() { return this.android;}
```
TheBusyBiscuit: First off, please format this like the other methods.

Second, this would only return the instance of SlimefunItem, you should better return the Block of the android instead to correctly identify the android.

poma123: > Second, this would only return the instance of SlimefunItem, you should better return the Block of the android instead to correctly identify the android.

So if I understand correctly, it should return the Block b parameter of the ProgrammableAndroid#mine function?

TheBusyBiscuit: Correct

Issue #1276, opened 2019-12-02
Guide glitch again...


poma123:
## Description (Required)

The Lava generator is now okay, we can not take that out. But the Nuclear reactor still do that.
If players go to Energy and Electricity section, they can take out Nuclear Reactor and Nether Star Reactor from the gui. If they close the Guide, they are unable to open it anymore. (just after a restart or some glitch with '/sf search a')

## Steps to reproduce the Issue (Required)

1. Open SF Guide
2. Go to Energy and Electricity section
3. Find Nuclear Reactor or Nether Star Reactor
4. You can take out them from the gui to your inventory.
5. Close the Guide.
6. And try to open it again.
<details><summary><b>Spoiler: Images (Click me!)</b></summary>
<br>

  ![image](https://user-images.githubusercontent.com/25465545/69966950-b565cc00-1517-11ea-8d21-69919dd7b702.png)
![image](https://user-images.githubusercontent.com/25465545/69966984-c57dab80-1517-11ea-8afc-1c97be155042.png)

</details>

## Expected behavior (Required)

In a normal situation I shouldn't be able to take out anything from the Slimefun Guide, the items must jump back into the Guide.
## Server Log / Error Report

When I took out the generators: https://pastebin.com/dr4Y30WB
After I closed the guide and tried to open it again: https://pastebin.com/8D6mRejF
## Environment (Required)

![image](https://user-images.githubusercontent.com/25465545/69967134-18576300-1518-11ea-842d-8e71995bb798.png)

 - Minecraft Version: git-Paper-226 (MC: 1.14.4)
 - Slimefun Version: Slimefun vDEV - 350 (`git 8183a1a5`)
 - CS-CoreLib Version: CS-CoreLib vDEV - 78 (`git 199e1b1e`)

OpticFusion1:
#EasyFix
Chances are (like a huge chunk of this plugin) guides are coded poorly, causing this issue

OpticFusion1:
*coughs*workingonmyownversionofthepluginthat'snotapoorlycodedmess*coughs*

WalshyDev:
[...]

PR #1277, opened 2019-12-02
actually fixes it now

J3fftw1:
## Description

Fixes this by not adding a null item to the recipes anymore. Not sure if there was any ACTUAL use from this but from testing there aren't any issues.

## Changes

removed if (list.size() % 2 != 0) list.add(null);
## Related Issues
#1276 

## Testability


J3fftw1:
i did ctrl+shift+f if (list.size() % 2 != 0) list.add(null); and ctrl+shift+f  list.add(null);

nothing left :)

PR #1279, opened 2019-12-04
Added a new constructor for HandledBlock

SoSeDiK:
## Description
It will be used in ExoticGarden.

## Changes
Added a new constructor that allows using a new item system.

## Related Issues
N/A

## Testability


TheBusyBiscuit:
I am gonna merge it, but will point out that this HandledBlock class should maybe be deprecated at some point.

PR #1283, opened 2019-12-06
Updated Teleport Sounds and Fixed The “color codes are supported text” for elevators 

coder4321:
## Description
Changed the final teleportation sound to the beacon activate sound and changed the ambient teleportation sound the the beacon ambient sound

Added the end bracket to the “color codes are supported” thing for elevators (could not find the file that Holograms use)

## Changes
As described above 

## Related Issues
Couldn't find any

## Testability

I can’t seem to check the boxes, but yes I did both.


coder4321:
I forgot to do something so I will resubmit this later

TheBusyBiscuit:
Please give your commits proper names...
"Updated <file>" is not really conveying much.
The commits should reflect their changes.

coder4321:
Sorry about that, next time I’ll make sure to put a proper name.

PR #1287, opened 2019-12-06
Allow for different GH and MC names + small improvements

WalshyDev:
## Description
Improved some code, blocked any name containing "bot" and added a way to have different GH and MC names so that people like myself and others actually aren't broken in the credits.

Yay me no broken

## Testability

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
@@ -74,12 +74,10 @@ public class ContributionsConnector extends GitHubConnector {
 	    	int commits = object.get("contributions").getAsInt();
 	    	String profile = object.get("html_url").getAsString();
 
-	    	final String alias = aliases.getOrDefault(name, name);
-
 	    	if (nameFormat.matcher(name).matches() && !blacklist.contains(name)) {
 	    		Contributor contributor = SlimefunPlugin.getUtilities().contributors.computeIfAbsent(
-	    				alias,
-						key -> new Contributor(alias, profile)
+	    				name,
+						key -> new Contributor(aliases.getOrDefault(name, name), profile)
 				);
 	    		contributor.setContribution(role, commits);
 	    	}
--- a/src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
@@ -74,10 +74,12 @@ public class ContributionsConnector extends GitHubConnector {
 	    	int commits = object.get("contributions").getAsInt();
 	    	String profile = object.get("html_url").getAsString();
 
+	    	final String alias = aliases.getOrDefault(name, name);
+
 	    	if (nameFormat.matcher(name).matches() && !blacklist.contains(name)) {
 	    		Contributor contributor = SlimefunPlugin.getUtilities().contributors.computeIfAbsent(
-	    				name,
-						key -> new Contributor(aliases.getOrDefault(name, name), profile)
+	    				alias,
+						key -> new Contributor(alias, profile)
 				);
 	    		contributor.setContribution(role, commits);
 	    	}
--- a/src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
@@ -1,23 +1,29 @@
 package me.mrCookieSlime.Slimefun.hooks.github;
 
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
 import me.mrCookieSlime.Slimefun.SlimefunPlugin;
-import org.bukkit.Bukkit;
 
 public class ContributionsConnector extends GitHubConnector {
-	
+
+	private static final Pattern nameFormat = Pattern.compile("[\\w_]+");
+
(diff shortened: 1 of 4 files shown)


src/main/java/me/mrCookieSlime/Slimefun/hooks/github/Contributor.java
```diff
@@ -16,7 +16,7 @@
 	
 	private String name;
 	private String profile;
-	private Optional<String> headTexture;
+	private Optional<String> headTexture = Optional.empty();
```
TheBusyBiscuit: The optional was intentionally left null.

Null = The texture has not been retrieved yet (or tooManyrequests, so try next time.
Empty = No mc account with that name exists.

src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
```diff
@@ -53,13 +61,16 @@ public String getURLSuffix() {
...
 	    	int commits = object.get("contributions").getAsInt();
 	    	String profile = object.get("html_url").getAsString();
-	    	
-	    	if (!blacklist.contains(name)) {
-	    		Contributor contributor = SlimefunPlugin.getUtilities().contributors.computeIfAbsent(name, key -> new Contributor(name, profile));
+
+	    	if (!name.toLowerCase().contains("bot") && !blacklist.contains(name)) {
+	    		Contributor contributor = SlimefunPlugin.getUtilities().contributors.computeIfAbsent(
+	    				name,
+						key -> new Contributor(aliases.getOrDefault(name, name), profile)
```
TheBusyBiscuit: The alias should better be treated seperately.

We want the GitHub name to display correctly...

src/main/java/me/mrCookieSlime/Slimefun/hooks/github/Contributor.java
WalshyDev: NotLikeThis

src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
WalshyDev: Hmm, not sure I fully agree. I mean, it's showing the contributors but displaying their heads. Having their MC names there so they can be identified makes more sense. Especially since you can click on it to get their GitHub profile anyway.

Imagine someone just having that skin and that name in MC, they didn't contribute but it looks like they did now. Doubt anyone would do that just to show off but... people are weird.

TheBusyBiscuit: Then add a "(MC: username)" or whatever to it if they differ.

But replacing the name is just an incorrect Model.

src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
```diff
@@ -53,13 +61,16 @@ public String getURLSuffix() {
...
-	    	
+
 	    	String name = object.get("login").getAsString();
 	    	int commits = object.get("contributions").getAsInt();
 	    	String profile = object.get("html_url").getAsString();
-	    	
-	    	if (!blacklist.contains(name)) {
-	    		Contributor contributor = SlimefunPlugin.getUtilities().contributors.computeIfAbsent(name, key -> new Contributor(name, profile));
+
+	    	if (!name.toLowerCase().contains("bot") && !blacklist.contains(name)) {
```
TheBusyBiscuit: What if I call myself
TheBotcher3000...

You probably just should have updated the blacklist and added a check for URL-invalid characters.

src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
```diff
@@ -10,15 +13,26 @@
 import me.mrCookieSlime.Slimefun.SlimefunPlugin;
 
 public class ContributionsConnector extends GitHubConnector {
-	
-	// GitHub Bots that do not count as Contributors
+
+	private static final Pattern nameFormat = Pattern.compile("[\\w_]+");
+
+	// All names including "bot" are automatically blacklisted. But, others can be too right here.
```
TheBusyBiscuit: You should adjust that comment.

ajan-12:
Me and @LinoxGH's mc name is ajan_12. Please add us too.

src/main/java/me/mrCookieSlime/Slimefun/hooks/github/ContributionsConnector.java
```diff
@@ -53,13 +69,16 @@ public String getURLSuffix() {
...
 	    	String name = object.get("login").getAsString();
 	    	int commits = object.get("contributions").getAsInt();
 	    	String profile = object.get("html_url").getAsString();
-	    	
-	    	if (!blacklist.contains(name)) {
-	    		Contributor contributor = SlimefunPlugin.getUtilities().contributors.computeIfAbsent(name, key -> new Contributor(name, profile));
+
+	    	if (nameFormat.matcher(name).matches() && !blacklist.contains(name)) {
+	    		Contributor contributor = SlimefunPlugin.getUtilities().contributors.computeIfAbsent(
+	    				name,
```
TheBusyBiscuit: If we now start mapping multiple gh accounts to a single mc account we should probably use the alias as the key too.

TheBusyBiscuit: Nevermind, I think the double keys is fine. We have the (MC: x) suffix now.

TheBusyBiscuit:
Revert the last commit, thatd cause more issues in the end.

WalshyDev:
I'm losing the will to live, can we just be happy and get everyone's name?

WalshyDev:
I'll take the silence as a yes.

NihilistBrew:
Hey hey, I saw the announcment that I should post my name here

Ya boi is called "ma1yang2" in-game (without the quotes)

TheBusyBiscuit:
We should maybe consider switching to storing the UUID instead at some point.
Just saying.

PR #1294, opened 2019-12-09
Added recipes

BurningBrimstone:
## Description

Changes are for the electric press.
Added recipes for iron and gold nuggets into their respective ingots. I also added a recipe for quartz into quartz blocks. 

https://discordapp.com/channels/565557184348422174/627048923122499584/652563773793763343

## Changes

Same as above. 

## Testability


WalshyDev:
Build is failing, the material isn't called `NETHER_QUARTZ`, please look at the Javadocs and correct this.

BurningBrimstone:
> 
> 
> Build is failing, the material isn't called `NETHER_QUARTZ`, please look at the Javadocs and correct this.

My apologies, changed to "QUARTZ". I seem to recall the namespace ID being NETHER_QUARTZ but the wikia denies that.

TheBusyBiscuit:
I think that all other block recipes should be added too if we already have Quartz blocks.

BurningBrimstone:
I would be inclined to agree. But the addition of quartz blocks serves to assist in automation of solar panels and silicon, a major electric resource.

I was discussing with panda about adding the resource blocks and they made a valid point that adding to much more functionality to the electric press would qualify it to be a small addon.

Nonetheless, if you want me to add them, I will.

TheBusyBiscuit:
Normally I am always for putting stuff rather in an addon.
But this is a very specfic use-case that fits to this machine.

And since we already have quartz blocks, it'd be weird to make the other require an addon...
There will probably come requests to add them anyway at some point.

PR #1300, opened 2019-12-11
Support 1.15

WalshyDev:
## **Requires merge of [this](https://github.com/TheBusyBiscuit/CS-CoreLib/pull/122)**

1.15 barely changed anything so all we need to do is update this array and then the CS-CoreLib dep. So, once the PR above is merged I will push the dep change.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,9 @@
         				<goals>
          					<goal>shade</goal>
         				</goals>
+                        <configuration>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -79,6 +79,9 @@
         				<goals>
          					<goal>shade</goal>
         				</goals>
+                        <configuration>
+                            <outputFile>C:/Users/Walshy/Documents/Servers/Slimefun-1.15/plugins/${project.artifactId}.jar</outputFile>
```
TheBusyBiscuit: Bad Walshy, bad,

WalshyDev: OOPS

Issue #1306, opened 2019-12-16
Explosive Pickaxe issues with 1.15 snapshot


Aearost:
## Description (Required)
I have come to notice two issues regarding the explosive pickaxe! It has full functionality (breaks all expected blocks at once), though it takes durability damage, even with the damage-on-use property set to false. On top of this, only the single block that was destroyed will be dropped (not all others in the 3x3x3 space).

I was also having issues when I had ExoticGarden installed, where the fruit tree heads would simply drop the actual CS-CoreLib head instead of the fruit. I removed this plugin and the issue persisted.

I believe the issue lays in CS-CoreLib (the source of the error log below), but the Explosive Pickaxe is something else altogether, as there is no error log when blocks are destroyed.

## Server Log
This error occurs as the very last line once the server has started up.
https://pastebin.com/h1p3L2y9

## Environment (Required)
Using: spigot-1.15.jar
CS-CoreLib: Dev 79
Slimefun: Dev 367

TheBusyBot:
> Your issue has been closed!

If your issue was closed **and not resolved** by one of our developers it may fall under one or more of these categories:

* You did not follow our template. Please follow the Issue template to help us identify your issue more quickly.
* You did not provide any information about your versions (We absolutely need your exact versions, \"latest\" is not helpful)
* Your issue was already reported, it is a duplicate. Check the other open issues first before posting!
* You posted an error without using pastebin. Please always post errors via pastebin.
* Your issue has already been fixed in a later version of Slimefun or CS-CoreLib, you should update.
* You are using an outdated version of Slimefun / CS-CoreLib, again, you should update.
* You are using an unsupported version of Minecraft. We only provide support for the latest version of Minecraft, old versions are not supported anymore.
* You are using a \"stable\" version of Slimefun (prefixed with \"RC - \"), your issue may have been fixed in a development build, so we only accept bug reports from those.
* You are on the wrong issue tracker. We would like to remind you that this Issue Tracker is **only for Slimefun**, to report bugs on any addons, head to the corresponding issue tracker.

Make sure to check out our article on [How to report bugs](https://github.com/TheBusyBiscuit/Slimefun4/wiki/How-to-report-bugs) for even more information.

Issue #1304, opened 2019-12-14
Android Miner Can't Dig Cobblestone.


xMikux:
## Description (Required)
Android Miner Can't Dig Cobblestone.

## Evidence / Proof (Required)
Put android miner in front of cobblestone and set script dig forward.
And you will see cobblestone not dig.
## Environment (Required)
[01:52:08 INFO]: Paper v1_14_R1
[01:52:08 INFO]:
[01:52:08 INFO]: CS-CoreLib vDEV - 79 (git ec5b4c24)
[01:52:08 INFO]: Slimefun vDEV - 367 (git aa9652cf)
[01:52:08 INFO]:
[01:52:08 INFO]: Installed Addons (12)
[01:52:08 INFO]:  HotbarPets vDEV - 20 (git cd87bc60)
[01:52:08 INFO]:  ExtraGear vDEV - 3 (git de222ab7)
[01:52:08 INFO]:  ExoticGarden vDEV - 22 (git a855f5b7)
[01:52:08 INFO]:  Barrels vDEV - 7 (git 0d14dd9f)
[01:52:08 INFO]:  ElectricSpawners vDEV - 5 (git c3f8cc9d)
[01:52:08 INFO]:  ExtraHeads v1.1
[01:52:08 INFO]:  SoulJars vDEV - 4 (git 4a600aa2)
[01:52:08 INFO]:  PrivateStorage vDEV - 6 (git b012877e)
[01:52:08 INFO]:  ChestTerminal vDEV - 19 (git ebf13b83)
[01:52:08 INFO]:  DyedBackpacks vDEV - 1 (git 4b4d1f0f)
[01:52:08 INFO]:  SlimefunOreChunks vDEV - 3 (git 7ddf966e)
[01:52:08 INFO]:  SlimeXpansion vDEV - 4 (git 5bf10c29)

BurningBrimstone:
Could you provide a video of this bug? Just tested on the latest slimefun build and the android miner appears to be working fine. Make sure your android has fuel, an empty inventory, and that you manually start the script.

xMikux:
https://youtu.be/vbWNDBJG6u4
And this video will remove if bug fix or is not a bug.

J3fftw1:
Got reported on my server aswell

'
[...]

Issue #1303, opened 2019-12-12
Explosive Tools dropping incorrect block count.


mrcoffee1026:
## Description (Required)
Explosive Shovels and Pickaxes mine a 3x3x3 area when striking one block, the number of blocks mined should be 18.  For Shovels: the count is 17.  For Pickaxes: the count is 1.
## Steps to reproduce the Issue (Required)
Empty inventory for ease of count and mine a solid 3x3x3 area with each tool and count blocks mined.  (missing blocks should number 18 for each if you have mined a solid surface).

## Expected behavior (Required)
Number of mined (dropped) blocks should match the number of blocks removed (18).

## Server Log / Error Report
No log activity.

## Environment (Required)
[15:41:24 INFO]: Paper v1_14_R1 (236)
[15:41:24 INFO]:
[15:41:24 INFO]: CS-CoreLib vDEV - 78 (git 199e1b1e)
[15:41:24 INFO]: Slimefun vDEV - 365 (git 460270f2)
[15:41:24 INFO]:
[15:41:24 INFO]: Installed Addons (11)
[15:41:24 INFO]:  Barrels vDEV - 7 (git 0d14dd9f)
[15:41:24 INFO]:  SlimeXpansion vDEV - 4 (git 5bf10c29)
[15:41:24 INFO]:  ExtraGear vDEV - 3 (git de222ab7)
[15:41:24 INFO]:  ChestTerminal vDEV - 17 (git 933a7d36)
[15:41:24 INFO]:  HotbarPets vDEV - 20 (git cd87bc60)
[15:41:24 INFO]:  ExoticGarden vDEV - 22 (git a855f5b7)
[15:41:24 INFO]:  SoulJars vDEV - 4 (git 4a600aa2)
[15:41:24 INFO]:  SlimefunOreChunks vDEV - 3 (git 7ddf966e)
[15:41:24 INFO]:  ElectricSpawners vDEV - 5 (git c3f8cc9d)
[15:41:24 INFO]:  PrivateStorage vDEV - 6 (git b012877e)
[15:41:24 INFO]:  DyedBackpacks vDEV - 1 (git 4b4d1f0f)

TheBusyBot:
> Your issue has been closed!

If your issue was closed **and not resolved** by one of our developers it may fall under one or more of these categories:

* You did not follow our template. Please follow the Issue template to help us identify your issue more quickly.
[...]

PR #1308, opened 2019-12-16
Fix drops in 1.15

creator3:
## Description
Fix drops in 1.15

## Changes
Replace calls to `block.getDrops()` with calls specifying the correct tool ItemStack

## Related Issues
#1306 #1304 #1303
Some of these issues describe multiple problems with 1.15, not just this one

## Testability
I only did basic testing, but it seems to have fixed the issue. I don't think it'll break any addons, but some of them will probably need the same fix applied

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/ExplosivePickaxe.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/ExplosivePickaxe.java
@@ -26,7 +26,10 @@ import me.mrCookieSlime.Slimefun.api.Slimefun;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
 public class ExplosivePickaxe extends SimpleSlimefunItem<BlockBreakHandler> implements NotPlaceable, DamageableItem {
-	
+
+	// Determines the drops an ExplosivePickaxe will get
+	private static final ItemStack EFFECTIVE_PICKAXE = new ItemStack(Material.DIAMOND_PICKAXE);
+
 	private String[] blacklist;
 	private boolean damageOnUse;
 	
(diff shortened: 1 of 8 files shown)


src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/HerculesPickaxe.java
```diff
@@ -13,6 +13,9 @@
 
 public class HerculesPickaxe extends SimpleSlimefunItem<BlockBreakHandler> {
 
+	// Determines the drops a HerculesPickaxe will get
+	private static final ItemStack EFFECTIVE_PICKAXE = new ItemStack(Material.DIAMOND_PICKAXE);
```
TheBusyBiscuit: I think it should be an Iron Pick...

WalshyDev:
God I hate Spigot...

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/items/ExplosivePickaxe.java
```diff
@@ -77,7 +80,7 @@ else if (b.getType().name().endsWith("_SHULKER_BOX")) {
 									b.breakNaturally();
 								}
 								else {
-									for (ItemStack drop: b.getDrops()) {
+									for (ItemStack drop: b.getDrops(EFFECTIVE_PICKAXE)) {
```
TheBusyBiscuit: Please add a space before the :
to follow our new conventions.

(Would be awesome if you could change that in all the for loops of the other files you edited too)

WalshyDev:
Are there other getDrops around the plugin that also need to be changed. How do androids work in this regard?

creator3:
The mining talisman also uses getDrops(). I think it's currently broken, though, so maybe a second PR is in order there. I seem to recall I tracked down a bug where the code had been relocated and it ended up that nothing ever happens after it calculates the drops.

CS-CoreLib (v1) also seems to have a call in the deprecated BlockBreaker class. I didn't see anything in CS-CoreLib2, but I was just using a text search for "getDrops()"

creator3:
Oh, also CS-CoreLib2 probably needs a version bump. When I compiled Slimefun, I got a version of CS-CoreLib2 that was incompatible with 1.15 because of Tag.DIRT_LIKE, which was fixed in https://github.com/TheBusyBiscuit/CS-CoreLib2/commit/0f82b794e07ea8ccbe46e03932eca2b56d26dbc9.

PR #1339, opened 2019-12-31
Enable Automated Crafting Chamber To Craft Last Shaped Recipe

mohkamfer:
## Description
The changes below enable players to craft the last remaining shaped recipe in an Automated Crafting Chamber instead of having to pull out items, find an Enhanced Crafting Table, put the items back in the same order and craft them then pull the resulting output out. With the help of a single button press, these 5 steps are just two. Press the button, and take the output!

## Changes
- A new method, tickOnce() that does the same work of tick(), except that it doesn't return at ItemStack amount of 1, actually it especially targets amounts of 1.
- Refactored the common code between the two methods tick() and tickOnce() into
  - getSerializedInput(BlockMenu, boolean) and
  - testInputAgainstRecipes(Block, String);

## Related Issues
New Addition.

## Testability
- [ X ] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ X ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
@@ -214,6 +214,7 @@ public abstract class AutomatedCraftingChamber extends SlimefunItem implements I
 		BlockMenu menu = BlockStorage.getInventory(block);
 		StringBuilder builder = new StringBuilder();
 		int i = 0;
+		boolean lastIteration = false;
 		for (int j = 0; j < 9; j++) {
 			if (i > 0) {
 				builder.append(" </slot> ");
@@ -224,7 +225,7 @@ public abstract class AutomatedCraftingChamber extends SlimefunItem implements I
 				// we're only executing the last possible shaped recipe
 				// we don't want to allow this to be pressed instead of the default timer-based
 				//   execution to prevent abuse and auto clickers
-				if (item != null && item.getAmount() > 1) return "";
+				if (item != null && item.getAmount() == 1) lastIteration = true;
 			} else {
 				if (item != null && item.getAmount() == 1) return "";
 			}
@@ -233,6 +234,8 @@ public abstract class AutomatedCraftingChamber extends SlimefunItem implements I
 
 			i++;
 		}
+
+		if (craftLast && !lastIteration) return "";
 
 		return builder.toString();
 	}
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
@@ -71,7 +71,7 @@ public abstract class AutomatedCraftingChamber extends SlimefunItem implements I
 
 				menu.replaceExistingItem(7, new CustomItem(new ItemStack(Material.CRAFTING_TABLE), "&7Craft Last", "", "&e> Click to craft the last shaped recipe", "&cOnly works with the last one"));
 				menu.addMenuClickHandler(7, (p, slot, item, action) -> {
-					tickOnce(b);
+					tick(b, true);
 					return false;
 				});
 			}
@@ -189,7 +189,7 @@ public abstract class AutomatedCraftingChamber extends SlimefunItem implements I
 			
 			@Override
 			public void tick(Block b, SlimefunItem sf, Config data) {
-				AutomatedCraftingChamber.this.tick(b);
+				AutomatedCraftingChamber.this.tick(b, false);
 			}
 
 			@Override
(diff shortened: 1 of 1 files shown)


src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
```diff
@@ -192,40 +198,65 @@ public boolean isSynchronized() {
...
+	}
+
+	protected void tickOnce(Block b) {
+		if (ChargableBlock.getCharge(b) < getEnergyConsumption()) return;
+
+		BlockMenu menu = BlockStorage.getInventory(b);
+
+		String input = getSerializedInput(menu, true);
+		testInputAgainstRecipes(b, input);
+	}
```
TheBusyBiscuit: This method is basically a way to do the exact same as tick() but without consuming power...

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
```diff
@@ -68,6 +68,12 @@ public void newInstance(BlockMenu menu, Block b) {
 						return false;
 					});
 				}
+
+				menu.replaceExistingItem(7, new CustomItem(new ItemStack(Material.CRAFTING_TABLE), "&7Craft Last", "", "&e> Click to craft the last shaped recipe", "&cOnly works with the last one"));
+				menu.addMenuClickHandler(7, (p, slot, item, action) -> {
+					tickOnce(b);
```
TheBusyBiscuit: You could pretty much spam-click this button and craft a bunch of items very fast and without consuming power.

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
```diff
@@ -192,40 +198,65 @@ public boolean isSynchronized() {
...
+	private void testInputAgainstRecipes(Block block, String input) {
+		BlockMenu menu = BlockStorage.getInventory(block);
+
 		if (SlimefunPlugin.getUtilities().automatedCraftingChamberRecipes.containsKey(input)) {
 			ItemStack output = SlimefunPlugin.getUtilities().automatedCraftingChamberRecipes.get(input).clone();
-			
+
 			if (menu.fits(output, getOutputSlots())) {
 				menu.pushItem(output, getOutputSlots());
-				ChargableBlock.addCharge(b, -getEnergyConsumption());
```
mohkamfer: @TheBusyBiscuit power consumption is handled here, the method is called for both tick() and tickOne()

TheBusyBiscuit: Yeah, you can tell that I just woke up...

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
TheBusyBiscuit: So, now...

These methods are almost identical, so I'd suggest merging them and just adding a parameter to the tick() method instead.

mohkamfer:
Done. And fixed a small bug that escaped me ^_^

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
```diff
@@ -192,40 +198,61 @@ public boolean isSynchronized() {
...
-	protected void tick(Block b) {
-		if (BlockStorage.getLocationInfo(b.getLocation(), "enabled").equals("false")) return;
-		if (ChargableBlock.getCharge(b) < getEnergyConsumption()) return;
-		
-		BlockMenu menu = BlockStorage.getInventory(b);
-		
+
+	protected void tick(Block block, boolean craftLast) {
+		if (!craftLast) {
+			if (BlockStorage.getLocationInfo(block.getLocation(), "enabled").equals("false")) return;
```
TheBusyBiscuit: Please merge these two statements with an AND connector.

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/machines/electric/AutomatedCraftingChamber.java
```diff
@@ -192,40 +198,61 @@ public boolean isSynchronized() {
...
 			}
-			
+
 			ItemStack item = menu.getItemInSlot(getInputSlots()[j]);
-			if (item != null && item.getAmount() == 1) return;
+			if (craftLast) {
+				// we're only executing the last possible shaped recipe
+				// we don't want to allow this to be pressed instead of the default timer-based
+				//   execution to prevent abuse and auto clickers
+				if (item != null && item.getAmount() == 1) lastIteration = true;
```
TheBusyBiscuit: I don't really get why you changed this and introduced this variable.
It does pretty much the same as it did before.

mohkamfer: Previously it would return from the whole method if it encountered an ItemStack with amount 1. However what I was after is getting at least an ItemStack with amount 1 so I could count this as a craftable, last iteration.

mohkamfer: I also refactored them to look better @ 667ec59

TheBusyBiscuit: I see, alright then!

PR #1351, opened 2020-01-05
Nuclear reactor displayed wrong tool tips

J3fftw1:
## Description
Nuclear reactor displayed wrong info

## Changes
changed display item from uranium to a nether star
changed tool tip of the display item to say it uses nether stars instead of uranium and Neptunium
Changed tooltip that said "Without any Coolant Cells, your Reactor" to "Without any &6Nether Ice Coolant Cells"

## Related Issues
N/A

## Testability


TheBusyBiscuit:
You edited the super class of all Reactors...

J3fftw1:
ffs

PR #1387, opened 2020-01-26
Added support for furnaces and brewing stands to cargo nets.

### LINKED ISSUE #114: Cargo Input Nodes Input into Furnace Fuel and Output Slots
## Description
With a Cargo Input Node, items inserted into Furnaces will be inserted into the output slot and non-fuel items into the fuel slot instead of just the ingredients slot.

## Evidence / Proof
Screenshot: https://i.gyazo.com/8118ca4fc6346f44a64b012576ccf0a9.png

## Environment
MC version: 1.11.2 (git-Paper-1032)
CS-CoreLib version: 1.5.12
Slimefun version: 4.1.1

BigBadE:
## Description

Added support for Furnaces and Brewing Stands to cargo nets. 
## Changes

Added code to check the inventory to the insert and withdraw methods CargoManager.java (Intellij is still acting up and showing weird change files, all changes are at lines 47-68, 92-109, 148-197)
 
## Related Issues

Resolves #114 

## Testability

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/item_transport/CargoManager.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/item_transport/CargoManager.java
@@ -72,7 +72,6 @@ public final class CargoManager {
     }
 
     public static ItemSlot withdraw(Block node, Block target, int index) {
-        System.out.println(1);
         DirtyChestMenu menu = getChestMenu(target);
 
         if (menu != null) {
@@ -86,8 +85,6 @@ public final class CargoManager {
             }
         } else {
             BlockState state = target.getState();
-
-            System.out.println(3);
 
             if (state instanceof InventoryHolder) {
                 Inventory inv = ((InventoryHolder) state).getInventory();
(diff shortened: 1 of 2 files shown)


src/main/java/me/mrCookieSlime/Slimefun/api/item_transport/CargoManager.java
```diff
@@ -1,228 +1,259 @@
...
+                    }
+                }
+
+                for (int slot = minSlot; slot < maxSlot; slot++) {
+                    ItemStack is = inv.getContents()[slot];
+
+                    if (is == null) {
+                        inv.setItem(slot, ChestManipulator.trigger(target, slot, null, stack.clone()));
+                        stack.setAmount(0);
+                        return stack;
```
TheBusyBiscuit: Why did you not return null there?

BigBadE: I originally had a really bad system where I moved that to another method and that was the only way I could figure out if the item was placed fully or not placed at all. Should functionally be the same, but replaced it with a null anyways.

TheBusyBiscuit:
Awesome btw, nice to see a three year old issue get resolved XD

PR #1388, opened 2020-01-26
Add Hungarian translation

andris155:
## Description

Add Hungarian translation

## Changes

src/main/resources/languages/ + messages_hu.yml

## Testability


TheBusyBiscuit:
Do not submit translations via GitHub.
Also, translations are not possible yet, so this will not work...

PR #1433, opened 2020-02-02
Sync up cache stuff in BlockStorage

WalshyDev:
## Description
Title

## Related Issues
#1428 and others

## Testability


TheBusyBiscuit:
Testability...

TheBusyBiscuit:
Awesome lack of description too

bitmoji

WalshyDev:
It compiles, doesn't need to be tested. It's also gonna be a pain to actually test this since I'd need to make it read and write to the cache at the exact same time

PR #1458, opened 2020-02-05
formatting

NihilistBrew:


NihilistBrew:
I changed it to Material.ELYTRA to make it coherent with everything else

PR #1512, opened 2020-02-10
Implement a '/sf report' command!

WalshyDev:
## Description
This implements a `/sf report` command! This allows for server admins to report issues to the GitHub with a lot more data for us to look at. We no longer need to rely on people to give versions, logs or plugin list. At least, it's not a complete requirement (though still may be needed).

I have a NodeJS app to run the redirect of this, slimefun.dev (which you can DM me and I can transfer if you want) is just a redirection. I looked at other link shortners but you have limited ability before you have to pay. This allows us to do anything we want (for example, https://slimefun.dev/github already exists to get here).

There is a rate-limit on the new endpoint, this may be changed around but it's just to stop people spamming 1k new reports a min. 

anyway, enough of the boring stuff, here is a preview:
Message in chat
The issue it creates

## Changes
* There is a new sub-command - `/sf report`
* ErrorReport has a way to store errors/warnings - this is currently capped at 10

## Related Issues
N/A

## Testability

## Side note
I'm sure there's gonna be a lot of changes you want so feel free to send 'em over! I just did this for now so no localization of the like 2 strings.
I can give you an auth token to create custom redirects with set names (such as github) rather than a random ID. Just DM me.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
@@ -5,11 +5,13 @@ import java.io.FileNotFoundException;
 import java.io.PrintStream;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import org.bukkit.Bukkit;
@@ -26,10 +28,15 @@ import me.mrCookieSlime.Slimefun.api.PlayerProfile;
 import me.mrCookieSlime.Slimefun.api.Slimefun;
 
 public class ErrorReport {
-	
+
+	private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss");
+	private static final List<String> errors = new ArrayList<>();
+
 	private File file;
 	
 	public ErrorReport(Throwable throwable, Consumer<PrintStream> printer) {
+		logError(Level.SEVERE, "", throwable);
+
 		Slimefun.runSync(() -> {
 			file = getNewFile();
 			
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
```diff
@@ -26,10 +28,15 @@
 import me.mrCookieSlime.Slimefun.api.Slimefun;
 
 public class ErrorReport {
-	
+
+	private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss");
```
TheBusyBiscuit: Date Formats are afaik not thread safe and shouldn't be static, should they...?

src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
```diff
@@ -188,4 +195,30 @@ public static void tryCatch(Function<Exception, ErrorReport> function, Runnable
 		}
 	}
 
+	public static void error(String message, Throwable throwable) {
+		logError(Level.SEVERE, message, throwable);
+		SlimefunPlugin.instance.getLogger().log(Level.SEVERE, message, throwable);
```
TheBusyBiscuit: You can just do Slimefun.getLogger()

src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
```diff
@@ -26,10 +28,15 @@
 import me.mrCookieSlime.Slimefun.api.Slimefun;
 
 public class ErrorReport {
-	
+
+	private static final SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss");
+	private static final List<String> errors = new ArrayList<>();
```
TheBusyBiscuit: I have a CS-CoreLib2 implementation for that, CappedLinkedList or so. It auto-removes elements already.

src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
WalshyDev: Not the biggest fan of it having to be initialized every time... performance wise it makes me want to cry. Not sure there'd ever be much of a case where it'd have an issue thread wise but we can synchronize the use anyway if need be.

src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
WalshyDev: Ah cool!

WalshyDev:
Gave you access to the site itself, it isn't anything large and can be open-sourced. If I was to open source though I would need to change some thinks since right now it has stuff like hard-coded https://slimefun.dev. I can just add a URL value in config etc but yeah, just noting this.

PR #1539, opened 2020-02-15
Spelling correction

koloksk:
## Testability


TheBusyBiscuit:
Please use gitlocalize for translations in the future...

PR #1584, opened 2020-02-26
Add New Rainbow Blocks

ThatPreston:
## Description

Added Rainbow Concrete, Rainbow Concrete Powder, and Rainbow Glazed Terracotta

## Changes

No changes to existing content

## Testability

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
@@ -2406,11 +2406,6 @@ public final class SlimefunItemSetup {
 		new CustomItem(SlimefunItems.RAINBOW_CONCRETE, 8), new RainbowTicker(MaterialCollections.getAllConcreteColors()))
 		.register(plugin);
 		
-		new RainbowBlock(Categories.MAGIC, (SlimefunItemStack) SlimefunItems.RAINBOW_CONCRETE_POWDER, RecipeType.ANCIENT_ALTAR,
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
```diff
@@ -2400,6 +2400,21 @@ public int getTier() {
...
+		new CustomItem(SlimefunItems.RAINBOW_CONCRETE, 8), new RainbowTicker(MaterialCollections.getAllConcreteColors()))
+		.register(plugin);
+		
+		new RainbowBlock(Categories.MAGIC, (SlimefunItemStack) SlimefunItems.RAINBOW_CONCRETE_POWDER, RecipeType.ANCIENT_ALTAR,
+		new ItemStack[] {new ItemStack(Material.WHITE_CONCRETE_POWDER), new ItemStack(Material.WHITE_CONCRETE_POWDER), new ItemStack(Material.WHITE_CONCRETE_POWDER), new ItemStack(Material.WHITE_CONCRETE_POWDER), SlimefunItems.RUNE_RAINBOW, new ItemStack(Material.WHITE_CONCRETE_POWDER), new ItemStack(Material.WHITE_CONCRETE_POWDER), new ItemStack(Material.WHITE_CONCRETE_POWDER), new ItemStack(Material.WHITE_CONCRETE_POWDER)}, 
+		new CustomItem(SlimefunItems.RAINBOW_CONCRETE_POWDER, 8), new RainbowTicker(MaterialCollections.getAllConcretePowderColors()))
+		.register(plugin);
```
TheBusyBiscuit: I don't think Rainbow Concrete Powder is a good idea since you can easily mess up these blocks by turning them into solid concrete...

ThatPreston: You're right, removed it

PR #1596, opened 2020-02-28
Create recipes_pl.yml

koloksk:
Polish recipes have been added.

## Testability


TheBusyBiscuit:
Translations via GitHub are not allowed.

PR #1605, opened 2020-03-02
fair alloy amounts

EOT3000:
## Description
Ever since I played slimefun, I thought that the amount of resources to make ONE ingot are just too much. Did I really just use 30 dusts and ingots just to make ONE reinforced alloy ingot? Well this pr fixes the issue

## Changes
The amount of alloy you get from a smeltery is equal to the amount of resources put in

## Related Issues
no

## Testability

I just moved one line into a for loop that already existed. How should that break anything? Nonetheless, I tested with exotic garden, and have found that I can grow, harvest, and eat fruit properly even with my proposed changes.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/Smeltery.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/Smeltery.java
@@ -87,11 +87,12 @@ public class Smeltery extends MultiBlockMachine {
 					if (outputInv != null) {
 						for (ItemStack removing : inputs.get(i)) {
 							if (removing != null) {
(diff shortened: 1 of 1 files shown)


WalshyDev:
SF in my personal opinion (and I know others share this view) is too easy alright. Too easy to finish and create fully-automated farms for everything. 

Therefore, I think this (ignoring the code...) is wrong. 

If anything it should be harder to reach endgame not easier but hey ho. 

Now to this code,
What are you attempting to do.... you make it so the same output is sent multiple times. If this is outputting let's say 8 diamonds and the recipe is 9 slots. This is now outputting 72 diamonds. 
If you wanted to make this "easier" change the recipes.... this.... makes 0 sense.

PR #1667, opened 2020-03-11
Add getItem() to SlimefunItemStack

WalshyDev:
## Description
Add getItem() to SlimefunItemStack

## Changes
Added a nullable getItem() to SlimefunItemStack

## Related Issues
N/A

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
@@ -150,6 +150,15 @@ public class SlimefunItemStack extends CustomItem {
         return id;
     }
 
+    /**
+     * Gets the {@link SlimefunItem} associated for this SlimefunItemStack. Null if no item is found.
+     *
+     * @return The {@link SlimefunItem} for this SlimefunItemStack, null if not found.
+     */
+    public SlimefunItem getItem() {
+        return SlimefunItem.getById(this.id);
+    }
+
     public ImmutableItemMeta getImmutableMeta() {
         return immutableMeta;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
```diff
@@ -150,6 +151,15 @@ public String getItemID() {
         return id;
     }
 
+    /**
+     * Gets the {@link SlimefunItem} associated for this SlimefunItemStack. Null if no item is found.
+     *
+     * @return The {@link SlimefunItem} for this SlimefunItemStack, null if not found.
+     */
+    public SlimefunItem getItem() {
```
TheBusyBiscuit: I'd probably even name it "getSlimefunItem()".
Also why is this build failing?

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
```diff
@@ -150,6 +151,15 @@ public String getItemID() {
         return id;
     }
 
+    /**
+     * Gets the {@link SlimefunItem} associated for this SlimefunItemStack. Null if no item is found.
+     *
+     * @return The {@link SlimefunItem} for this SlimefunItemStack, null if not found.
```
TheBusyBiscuit: Nitpicky but you could also mark "SlimefunItemStack" as a link, even though its the same file...

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
WalshyDev: Very stupid things, fixed now. 

The reason I didn't name it that is, this is already an ItemStack and getting the ID is getItemID. Follows the pattern of using getItem. 

Personally, it should be both changed or keeping this like so.

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
WalshyDev: Yeah... reee

PR #1760, opened 2020-03-24
Pull Request: 4K Carrat Gold able to be used as Vanilla Ingot

SlimeyDerp:
Added:
 * Lore for Gold Ingot 4K suggesting it can be used as a Vanilla Crafting Material. It uses same description as Syntetic Diamonds and Emeralds do, to avoid inconsistencies
* Ability to craft using 4K Gold Ingots

PS: If you see .DS_Store files, They are files that tell MacOS how folders are placed. They are hidden, so they don't make a mess, and they do not interfere with the code. I couldn't remove them, although they probably won't be shown.


WalshyDev:
Please add DS_Store to the gitignore.

SlimeyDerp:
Will do

PR #1761, opened 2020-03-24
Pull Request: 4K Carrat Gold able to be Used for Vanilla Crafting

SlimeyDerp:
Added:
* Lore to 4K Carrat Gold Specifying it can now be used for Vanilla crafting. This lore is equal to the one used in Synthetic Diamonds, changing the diamond for "ingot".
* Ability to craft using 4K Gold Ingots
* Update to .gitignore to add DS_Store, to prevent macOS from adding extra garbage files.


src/main/java/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
```diff
@@ -441,7 +441,7 @@ private SlimefunItems() {}
     public static final ItemStack COBALT_INGOT = new SlimefunItemStack("COBALT_INGOT", Material.IRON_INGOT, "&9Cobalt Ingot");
 
     /* Gold */
-    public static final ItemStack GOLD_4K = new SlimefunItemStack("GOLD_4K", Material.GOLD_INGOT, "&rGold Ingot &7(4-Carat)");
+    public static final ItemStack GOLD_4K = new SlimefunItemStack("GOLD_4K", Material.GOLD_INGOT, "&rGold Ingot &7(4-Carat)", "", "&rThis item can be used in a", "&rworkbench and acts like a normal ingot");
```
TheBusyBiscuit: Don't change the lore here, this would break all existing items

PR #1772, opened 2020-03-30
Added World support to GEO

SlimeyDerp:
Added:

* GEO Resources can now use "World" argument to check how many resources should be given. This allows better implementation of extra Slimefun resources in add-ons, while keeping the original resources without modification.


SlimeyDerp:
Build failed, changing again

SlimeyDerp:
I am stupid, I forgot to import the World object so it failed. It should build properly now.   > o >

TheBusyBiscuit:
You cannot just change the methods like that...
You would be breaking a lot of addons with this and there is really no point in doing this.

PR #1790, opened 2020-04-05
Cargo Network - optimization

DNx5:
## Description
This PR add performance improvements in some part of cargo network.

## Changes
- Reduced calls for ``Inventory#getContents()``, avoiding creating the clone of the inventory's item stacks on each iteration
- Added utilities to check if a block has inventory, avoiding unnecessary calls to ``Block#getState()`` at cost for maintaining the list of inventory blocks
- Added tick delay option for Cargo Network.

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
```diff
@@ -0,0 +1,44 @@
...
+            case ENDER_CHEST:
+            case BARREL:
+            case BLAST_FURNACE:
+            case LECTERN:
+            case SMOKER:
+                return true;
+            default:
+                return false;
+        }
+    }
```
TheBusyBiscuit, line 42: Keeping a hardcoded list for this is not really the best idea.
I get that this will save a lot of performance though...

By using a switch case here with the values SMOKER, LECTERN, etc... you will be breaking 1.13 compatibility though since those constants do not exist then.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
```diff
@@ -0,0 +1,44 @@
+package io.github.thebusybiscuit.slimefun4.utils;
+import org.bukkit.block.Block;
+public final class BlockUtils {
+
+    public static boolean hasInventory(Block block) {
```
TheBusyBiscuit, line 5: Please create a private empty constructor here to prevent instantiation.

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
```diff
@@ -14,6 +15,8 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import com.google.gson.*;
```
TheBusyBiscuit, line 18: Never use wildcard imports...

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
```diff
@@ -378,14 +376,21 @@ private static BlockInfoConfig parseBlockInfo(Location l, String json) {
...
+            JsonWriter jsonWriter = new JsonWriter(stringWriter);
+            jsonWriter.setLenient(true);
+            jsonWriter.beginObject();
+            for (String key : cfg.getKeys()) {
+                jsonWriter.name(key).value(cfg.getString(key));
+            }
+            jsonWriter.endObject();
+            return stringWriter.toString();
+        } catch (IOException e) {
+            throw new AssertionError(e);
```
TheBusyBiscuit, line 392: An AssertionError is completely inappropriate here.
No Assertion was made and catching an exception just to throw another one serves no benefit here.
You hsould instead properly log the exception rather than creating a new one that makes no real sense here.

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
```diff
@@ -378,14 +376,21 @@ private static BlockInfoConfig parseBlockInfo(Location l, String json) {
...
+    private static String serializeBlockInfo(Config cfg) {
+        try {
+            StringWriter stringWriter = new StringWriter();
+            JsonWriter jsonWriter = new JsonWriter(stringWriter);
+            jsonWriter.setLenient(true);
+            jsonWriter.beginObject();
+            for (String key : cfg.getKeys()) {
+                jsonWriter.name(key).value(cfg.getString(key));
+            }
+            jsonWriter.endObject();
```
TheBusyBiscuit, line 389: What exactly is the advantage of using this right here?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
```diff
@@ -0,0 +1,44 @@
...
+            case PINK_SHULKER_BOX:
+            case GRAY_SHULKER_BOX:
+            case LIGHT_GRAY_SHULKER_BOX:
+            case CYAN_SHULKER_BOX:
+            case PURPLE_SHULKER_BOX:
+            case BLUE_SHULKER_BOX:
+            case BROWN_SHULKER_BOX:
+            case GREEN_SHULKER_BOX:
+            case RED_SHULKER_BOX:
+            case BLACK_SHULKER_BOX:
```
TheBusyBiscuit, line 32: All of that could be reduced to one if-statement. other than having 17 cases here.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
```diff
@@ -0,0 +1,44 @@
+package io.github.thebusybiscuit.slimefun4.utils;
```
TheBusyBiscuit, line 1: Is this of any use to non-cargo related classes? If no, then it may be moved to the cargo package.
Otherwise it is fine to stay here.

TheBusyBiscuit:
Also, please adhere to the checklist in your PR.
A Pull request will not be merged until you tested your changes and can confirm that they work without any issues.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
```diff
@@ -15,32 +16,26 @@ public static boolean hasInventory(Block block) {
...
                 return true;
             default:
-                return false;
+                return isShulkerBox(block);
         }
     }
 
+    public static boolean isShulkerBox(Block block) {
+        Material type = block.getType();
+        return type == Material.SHULKER_BOX || type == Material.WHITE_SHULKER_BOX || type == Material.ORANGE_SHULKER_BOX
```
WalshyDev, line 32: I'd just do name().endsWith("SHULKER_BOX") saves a lot of code

DNx5: Should I remove the method?

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
DNx5: It's same as ``JsonObject#toString()``, but because we don't need the object, we can use JsonWriter directly without creating a new JsonObject.
This part of code is get called pretty often and keep showing on profiler. I need to measure it more to see if it could help with performance.

TheBusyBiscuit: Alright, I see. Some actual performance figures would be cool but I will mark this one as resolved nonetheless. 👍

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
TheBusyBiscuit: I agree, this all can be reduced to the simple .endsWith statement which would be much simpler.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
andris155: Only a few server running on 1.13 versions. I don't know why keep this.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoUtils.java
```diff
@@ -204,26 +208,28 @@ else if (stack.getType() == Material.BLAZE_POWDER) {
...
+                    int amount = is.getAmount() + stack.getAmount();
 
-                inv.setItem(slot, is);
-                return stack;
+                    if (amount > is.getType().getMaxStackSize()) {
+                        is.setAmount(is.getType().getMaxStackSize());
+                        stack.setAmount(amount - is.getType().getMaxStackSize());
+                    } else {
+                        is.setAmount(amount);
+                        stack = null;
```
WalshyDev, line 227: This could be reduced in size to just Math.min(amount, getMaxStackSize

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
DNx5: It looks like only cargo needed this, I think I will move it.

pom.xml
```diff
@@ -216,7 +216,7 @@
         <dependency>
             <groupId>com.github.thebusybiscuit</groupId>
             <artifactId>CS-CoreLib2</artifactId>
-            <version>0.12</version>
+            <version>0b974b0</version>
```
DNx5, line 219: I don't know why, but 0.13 just didn't work. Seems like the repo is not up to date

TheBusyBiscuit: 0.13 works fine for me and also for our GitHub CI. Maybe delete your .m2 maven cache and try to build it again.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
```diff
@@ -0,0 +1,32 @@
+package io.github.thebusybiscuit.slimefun4.utils;
+import io.github.thebusybiscuit.cscorelib2.reflection.ReflectionUtils;
+import org.bukkit.Material;
+import org.bukkit.block.Block;
+public final class BlockUtils {
+
+    private static final boolean is_1_14 = ReflectionUtils.isVersion("v1_14_");
```
TheBusyBiscuit, line 7: Slimefun has a built-in method for that. And there is no need for making that a field here.

You can do SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_14) instead.
Your method would break for all 1.15 servers since you check for the exact version, not the minimum.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/BlockUtils.java
```diff
@@ -0,0 +1,32 @@
...
+        Material type = block.getType();
+        switch (type) {
+            case CHEST:
+            case TRAPPED_CHEST:
+            case FURNACE:
+            case DISPENSER:
+            case DROPPER:
+            case HOPPER:
+            case BREWING_STAND:
+            case ENDER_CHEST:
```
TheBusyBiscuit, line 23: I don't think Ender Chests have Inventories that you can interact with...
Since those are per-player not per-block.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/OreCrusher.java
```diff
@@ -38,18 +39,21 @@ public OreCrusher() {
...
 				BlockFace.SELF
 		);
-		
 		addItemSetting(doubleOres);
-		
+	}
+
+	@Override
+	public void register(SlimefunAddon addon) {
+		super.register(addon);
```
TheBusyBiscuit, line 47: 1. Never override the .register() method.

TheBusyBiscuit: You should use the .postRegister() method instead for this.

pom.xml
DNx5: I mean, its not compatible with minecraft version 1.13.2

TheBusyBiscuit: Oh that is fine.
Backwards compatibility with 1.13 is a bit broken in current versions, that will be fixed.
You don't have to worry about that right now though, since it is a different project.
Simply build against the latest version.

andris155:
https://github.com/TheBusyBiscuit/Slimefun4/issues/1801

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/BlockUtils.java
```diff
@@ -0,0 +1,30 @@
+package io.github.thebusybiscuit.slimefun4.core.networks.cargo;
+import io.github.thebusybiscuit.slimefun4.api.MinecraftVersion;
```
WalshyDev, line 2: Space between package and imports

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/BlockUtils.java
```diff
@@ -0,0 +1,30 @@
...
+import me.mrCookieSlime.Slimefun.SlimefunPlugin;
+import org.bukkit.Material;
+import org.bukkit.block.Block;
+final class BlockUtils {
```
WalshyDev, line 6: Space between imports and class.

TheBusyBiscuit:
Not really sure when you want to have this merged but this PR is still marked as a Draft by you, so make sure to mark it as ready when you are ready.

There is still a few unresolved comments though which you should check out, I am also gonna mark some things since you updated your fork.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoUtils.java
```diff
@@ -139,48 +141,50 @@ public static ItemStack insert(Block node, Block target, ItemStack stack, int in
...
 
-                    if (amount > is.getType().getMaxStackSize()) {
-                        is.setAmount(is.getType().getMaxStackSize());
-                        stack.setAmount(amount - is.getType().getMaxStackSize());
-                    }
-                    else {
-                        is.setAmount(amount);
-                        stack = null;
-                    }
+            int inSlotMaxSize = itemInSlot.getType().getMaxStackSize();
```
TheBusyBiscuit, line 163: maxSize would probably be a more sensible name for this variable here. "inSlotMaxSize" is rather unpleasant to read.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoUtils.java
```diff
@@ -139,48 +141,50 @@ public static ItemStack insert(Block node, Block target, ItemStack stack, int in
...
+            int inSlotAmount = itemInSlot.getAmount();
```
TheBusyBiscuit, line 164: currentAmount  would probably be a more sensible name for this variable here. "inSlotAmount" is rather unpleasant to read.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoUtils.java
```diff
@@ -139,48 +141,50 @@ public static ItemStack insert(Block node, Block target, ItemStack stack, int in
...
+            if (SlimefunUtils.isItemSimilar(itemInSlot, wrapper, true, false) && inSlotAmount < inSlotMaxSize) {
+                int amount = inSlotAmount + stack.getAmount();
 
-                    menu.replaceExistingItem(slot, is);
-                    return stack;
+                itemInSlot.setAmount(Math.max(amount, inSlotMaxSize));
```
TheBusyBiscuit, line 168: This should me Math.min, not Math.max
since now you are taking the maximum of the two which will make the amount potentially exceed the max size.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoUtils.java
```diff
@@ -139,48 +141,50 @@ public static ItemStack insert(Block node, Block target, ItemStack stack, int in
...
         }
 
         return stack;
     }
 
     private static ItemStack insertIntoVanillaInventory(ItemStack stack, Inventory inv) {
-        int minSlot = 0;
-        int maxSlot = inv.getContents().length;
+        ItemStack[] contents = inv.getContents();
+    	int minSlot = 0;
```
TheBusyBiscuit, line 186: Looks like you are missing a space there.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -32,6 +32,8 @@
     private final Set<Location> outputNodes = new HashSet<>();
 
     private final Map<Location, Integer> roundRobin = new HashMap<>();
+    private final int tickDelay = SlimefunPlugin.getCfg().getInt("URID.cargo-network-tick-delay");
```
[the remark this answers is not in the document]
DNx5: This is specifically made only for Cargo network. I don't think NetworkManager class should contain this, or should I made it for all types of Network?
How about moving it into:
```java
if (tickDelayThreshold < SlimefunPlugin.getCfg().getInt("URID.cargo-network-tick-delay")) {
```
I see this kind of usage in some place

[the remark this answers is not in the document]
TheBusyBiscuit: Retrieving the value directly from the config should definitely be avoided, that would just contradict the prupose of optimizing that method anyway.

Even if the value is only for cargo nets, the NetworkManager may still be an appropriate place for that, if you simply call it "cargoTickDelay" accordingly.
To elaborate: The NetworkManager is just a singleton that holds references to all Networks and some config values.

Alternatively, if you really think it should stay in the CargoNet you should definitely make that field static, as it is not cargo network specific but rather a static constant.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -175,6 +176,14 @@ private void run(Block b, Set<Location> destinations, Map<Integer, List<Location
             display();
         }
 
+        // Skip ticking if the threshold is not reached. The delay is not same as minecraft tick,
+        // but it's based on 'custom-ticker-delay' config.
+        if (tickDelayThreshold < SlimefunPlugin.getCfg().getInt("URID.cargo-network-tick-delay")) {
```
TheBusyBiscuit, line 181: I just told you to NOT do that...

andris155:
Very big thanks for this!

PR #1800, opened 2020-04-11
AncientAltarOutputEvent added

TweepCoding:
Added:

AncientAltarOutputEvent: This event will trigger before the ancient altar gives you an item. This will allow you to modify which item you want it to drop, get the location of the ancient altar block or even cancel it, so that the item isn't dropped and the drop animation for the item is cancelled.

Also, documented it, in case someone wanted to yell at me because of that.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
@@ -35,9 +35,9 @@ public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
      *            The altar {@link Block}
      * @param output
      *            The {@link ItemStack} that would be dropped by the ritual
-     *
+     * @param player
+     *            The {@link Player} that started the ritual.
      */
-
     public AncientAltarCraftEvent(ItemStack output, Block block, Player player) {
         super(player);
         this.block = block;
@@ -56,7 +56,6 @@ public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
      *
      * @return the main altar's block {@link Block}
      */
-
     public Block getAltarBlock() {
         return block;
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
@@ -32,7 +32,7 @@ public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
 
     /**
      * @param block
-     *            The mined {@link Block}
+     *            The altar {@link Block}
      * @param output
      *            The {@link ItemStack} that would be dropped by the ritual
      *
@@ -56,7 +56,8 @@ public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
      *
      * @return the main altar's block {@link Block}
      */
-    public Block getBlock() {
+
+    public Block getAltarBlock() {
         return block;
     }
 
@@ -69,7 +70,8 @@ public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
         return output;
     }
 
-    /** This method will change the item that would be dropped by the {@link AncientAltar}
+    /**
+     * This method will change the item that would be dropped by the {@link AncientAltar}
      *
      * @param output being the {@link ItemStack} you want to change the item to.
      */
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
@@ -22,7 +22,6 @@ import org.bukkit.inventory.ItemStack;
  * @see AncientAltarTask
  * @see AncientAltarListener
  */
-
 public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
 
     private static final HandlerList handlers = new HandlerList();
@@ -51,6 +50,7 @@ public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
     public HandlerList getHandlers() {
         return handlers;
     }
+    
     /**
      * This method returns the main altar's block {@link Block}
      *
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
@@ -2,6 +2,8 @@ package io.github.thebusybiscuit.slimefun4.api.events;
 
 
 import io.github.thebusybiscuit.slimefun4.implementation.items.altar.AncientAltar;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.AncientAltarListener;
+import io.github.thebusybiscuit.slimefun4.implementation.tasks.AncientAltarTask;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
 import org.bukkit.event.Cancellable;
@@ -16,9 +18,12 @@ import org.bukkit.inventory.ItemStack;
  *
  * @author Tweep
  *
+ * @see AncientAltar
+ * @see AncientAltarTask
+ * @see AncientAltarListener
  */
 
-public class AncientAltarOutputEvent extends PlayerEvent implements Cancellable {
+public class AncientAltarCraftEvent extends PlayerEvent implements Cancellable {
 
     private static final HandlerList handlers = new HandlerList();
     private final Block block;
@@ -33,7 +38,7 @@ public class AncientAltarOutputEvent extends PlayerEvent implements Cancellable 
      *
      */
 
-    public AncientAltarOutputEvent(ItemStack output, Block block, Player player) {
+    public AncientAltarCraftEvent(ItemStack output, Block block, Player player) {
         super(player);
         this.block = block;
         this.output = output;
@@ -43,31 +48,43 @@ public class AncientAltarOutputEvent extends PlayerEvent implements Cancellable 
         return handlers;
     }
 
-    public HandlerList getHandlers() { return handlers; }
+    public HandlerList getHandlers() {
+        return handlers;
+    }
     /**
      * This method returns the main altar's block {@link Block}
      *
      * @return the main altar's block {@link Block}
      */
-    public Block getBlock() { return block; }
+    public Block getBlock() {
+        return block;
+    }
 
     /**
-     * This method returns the item that would be dropped by the altar. {@link ItemStack}
+     * This method returns the {@link ItemStack} that would be dropped by the {@link AncientAltar }
      *
-     * @return the item that would be dropped by the altar. {@link ItemStack}
+     * @return the {@link ItemStack} that would be dropped by the {@link AncientAltar}
      */
     public ItemStack getItem() {
         return output;
     }
 
+    /** This method will change the item that would be dropped by the {@link AncientAltar}
+     *
+     * @param output being the {@link ItemStack} you want to change the item to.
+     */
     public void setItem(ItemStack output) {
         this.output = output;
     }
 
     @Override
-    public boolean isCancelled() { return cancelled; }
+    public boolean isCancelled() {
+        return cancelled;
+    }
 
     @Override
-    public void setCancelled(boolean b) { cancelled = b; }
+    public void setCancelled(boolean cancel) {
+        cancelled = cancel;
+    }
 
 }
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * This {@link Event} is fired before an item is dropped by an {@link AncientAltar}.
+ * Cancelling this event will make the {@link AncientAltar} drop no item after the recipe is finished.
+ *
+ * @author Tweep
+ *
+ */
+
+public class AncientAltarOutputEvent extends Event implements Cancellable {
```
WalshyDev, line 19: This should probably be a player event. I'd like to get the player here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
```diff
@@ -140,10 +143,14 @@ private void abort() {
 
     private void finish() {
         if (running) {
-            dropLocation.getWorld().playSound(dropLocation, Sound.ENTITY_ZOMBIE_VILLAGER_CURE, 1F, 1F);
-            dropLocation.getWorld().playEffect(dropLocation, Effect.STEP_SOUND, Material.EMERALD_BLOCK);
-            dropLocation.getWorld().dropItemNaturally(dropLocation.add(0, -0.5, 0), output);
 
+            AncientAltarOutputEvent ancientAltarOutputEvent = new AncientAltarOutputEvent(output, altar);
+            Bukkit.getPluginManager().callEvent(ancientAltarOutputEvent);
```
WalshyDev, line 148: Hmm. Maybe we have a pre and post event or something. I don't really like this running entirely and then not doing anything. 

Hmm.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
TweepCoding: Sure thing, hang on a second

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
TweepCoding: Huh? What do you mean?

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,73 @@
...
+        super(player);
+        this.block = block;
+        this.output = output;
+    }
+
+    public static HandlerList getHandlerList() {
+        return handlers;
+    }
+
+    public HandlerList getHandlers() { return handlers; }
```
TheBusyBiscuit, line 46: No one-line methods.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * This method returns the main altar's block {@link Block}
+     *
+     * @return the main altar's block {@link Block}
+     */
+    public Block getBlock() { return block; }
```
TheBusyBiscuit, line 52: No one-line methods.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @return the item that would be dropped by the altar. {@link ItemStack}
+     */
+    public ItemStack getItem() {
+        return output;
+    }
+
+    public void setItem(ItemStack output) {
+        this.output = output;
+    }
```
TheBusyBiscuit, line 65: Missing documentation

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public boolean isCancelled() { return cancelled; }
```
TheBusyBiscuit, line 68: No one-line methods.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public void setCancelled(boolean b) { cancelled = b; }
```
TheBusyBiscuit, line 71: No one-line methods.
Also rename the parameter "b" to something more meaningful like "cancel".

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
```diff
@@ -140,10 +146,14 @@ private void abort() {
...
-            dropLocation.getWorld().playSound(dropLocation, Sound.ENTITY_ZOMBIE_VILLAGER_CURE, 1F, 1F);
-            dropLocation.getWorld().playEffect(dropLocation, Effect.STEP_SOUND, Material.EMERALD_BLOCK);
-            dropLocation.getWorld().dropItemNaturally(dropLocation.add(0, -0.5, 0), output);
 
+            AncientAltarOutputEvent ancientAltarOutputEvent = new AncientAltarOutputEvent(output, altar, player);
+            Bukkit.getPluginManager().callEvent(ancientAltarOutputEvent);
+            if (!ancientAltarOutputEvent.isCancelled()) {
+                dropLocation.getWorld().playSound(dropLocation, Sound.ENTITY_ZOMBIE_VILLAGER_CURE, 1F, 1F);
+                dropLocation.getWorld().playEffect(dropLocation, Effect.STEP_SOUND, Material.EMERALD_BLOCK);
+                dropLocation.getWorld().dropItemNaturally(dropLocation.add(0, -0.5, 0), output);
```
TheBusyBiscuit, line 155: Since you are able to modify the output of the item you should also use the output item from that event rather than the original one.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
```diff
@@ -140,10 +146,14 @@ private void abort() {
 
     private void finish() {
         if (running) {
-            dropLocation.getWorld().playSound(dropLocation, Sound.ENTITY_ZOMBIE_VILLAGER_CURE, 1F, 1F);
-            dropLocation.getWorld().playEffect(dropLocation, Effect.STEP_SOUND, Material.EMERALD_BLOCK);
-            dropLocation.getWorld().dropItemNaturally(dropLocation.add(0, -0.5, 0), output);
 
+            AncientAltarOutputEvent ancientAltarOutputEvent = new AncientAltarOutputEvent(output, altar, player);
```
TheBusyBiscuit, line 150: That variable name is rather long, might just be "event".

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
```diff
@@ -47,13 +51,15 @@
 
     private boolean running;
     private int stage;
+    private Player player;
```
TheBusyBiscuit, line 54: This can be final.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * This {@link Event} is fired before an item is dropped by an {@link AncientAltar}.
+ * Cancelling this event will make the {@link AncientAltar} drop no item after the recipe is finished.
+ *
+ * @author Tweep
+ *
+ */
+
+public class AncientAltarOutputEvent extends PlayerEvent implements Cancellable {
```
TheBusyBiscuit, line 21: I would rather name it "AncientAltarCraftEvent" instead of Output.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarOutputEvent.java
```diff
@@ -0,0 +1,73 @@
...
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.player.PlayerEvent;
+import org.bukkit.inventory.ItemStack;
+
+/**
+ * This {@link Event} is fired before an item is dropped by an {@link AncientAltar}.
+ * Cancelling this event will make the {@link AncientAltar} drop no item after the recipe is finished.
+ *
+ * @author Tweep
+ *
```
TheBusyBiscuit, line 18: You can also add some "@see" tags here to the AncientAltar task and listener class etc... to mark that they belong to each other.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
```diff
@@ -50,13 +51,15 @@
...
 
-    public AncientAltarTask(Block altar, ItemStack output, List<Block> pedestals, List<ItemStack> items) {
+    public AncientAltarTask(Block altar, ItemStack output, List<Block> pedestals, List<ItemStack> items, Player p) {
         this.dropLocation = altar.getLocation().add(0.5, 1.3, 0.5);
         this.altar = altar;
         this.output = output;
         this.pedestals = pedestals;
         this.items = items;
+        player = p;
```
WalshyDev, line 62: Be consistent with the other this.'s

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * This method returns the {@link ItemStack} that would be dropped by the {@link AncientAltar }
+     *
+     * @return the {@link ItemStack} that would be dropped by the {@link AncientAltar}
+     */
+    public ItemStack getItem() {
+        return output;
+    }
+
+    /** This method will change the item that would be dropped by the {@link AncientAltar}
```
TheBusyBiscuit, line 72: This comment needs to be one line further down. You can't have it start on the same line as /**

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public HandlerList getHandlers() {
+        return handlers;
+    }
+    /**
+     * This method returns the main altar's block {@link Block}
+     *
+     * @return the main altar's block {@link Block}
+     */
+    public Block getBlock() {
```
TheBusyBiscuit, line 59: Maybe rename this to `getAltarBlock()` to avoid potential confusion.
Other than that it is ready to be merged!

WalshyDev:
I still maintain that it could be a bit of a pain losing all items and then you don't get a result because it got cancelled...

I think making this not cancellable or having a pre-post is best. Opinions Cookie?

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
TweepCoding: :D

TheBusyBiscuit:
> I still maintain that it could be a bit of a pain losing all items and then you don't get a result because it got cancelled...
> 
> I think making this not cancellable or having a pre-post is best. Opinions Cookie?

You should not cancel the event unless you do something else with the event output, in Tweep's case it's spawning a boss.
And if we were to remove Cancellable, there'd still be the option of people doing .setItem(dirt); to disrupt the crafting outcome. It is a rather unique use case and I doubt plugins will use it so commonly that it will cause rage because of vanishing items.
I personally see no reason for splitting this into a pre and post event.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
```diff
@@ -0,0 +1,91 @@
...
+ * This {@link Event} is fired before an item is dropped by an {@link AncientAltar}.
+ * Cancelling this event will make the {@link AncientAltar} drop no item after the recipe is finished.
+ *
+ * @author Tweep
+ *
+ * @see AncientAltar
+ * @see AncientAltarTask
+ * @see AncientAltarListener
+ */
+
```
TheBusyBiscuit, line 25: This line needs to go

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AncientAltarCraftEvent.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public static HandlerList getHandlerList() {
+        return handlers;
+    }
+
+    public HandlerList getHandlers() {
+        return handlers;
+    }
+    /**
```
TheBusyBiscuit, line 54: And you need a new line before this one.

Issue #1819, opened 2020-04-23
Cargo system voids items transferred at specifc times


Liruxo:
## Description
The cargo system can transfer items to blocks which technically do not exist anymore
More specific in the time window between the block being destroyed on the main thread (dropping the current content of the inventory) and the actual removal from the BlockStorage, the cargo system can transfer items to the inventory which will be voided

## Steps to reproduce the Issue
1. Create a simple cargo setup consisting of one cargo manager, one cargo input node and one cargo output node
2. Configure the cargo system in such way that it is transferring items into a slimefun managed inventory container, such as the "Electric Furnace"
3. Destroy the Electric Furnace exactly after the TickerTask has processed Block removals and before it starts processing the tick of the cargo manager
4. Observe how items are transferred into the destroyed Electric Furnace and are destroyed immediately afterwards as the TickerTask processes the removal of the Electric Furnace 

From my testing this is a very rare occasion, and as such a very hard bug to reproduce.
However, with the help of a debugger which is capable of pausing the async TickerTask, it is reproducible every time

## Expected behavior
The item does not transfer, or drops in the world

## Environment
 - Minecraft Version: v1_15_R1
 - Slimefun Version: DEV 489 (git d8964)
 - CS-CoreLib Version: v1.6.3

TheBusyBiscuit:
Also to bring this discussion onto here now, a quick and easy fix for this would be to mark any BlockMenu as "obsolete" upon block breaks and then have a BlockMenu#isObsolete() check in the cargo ticker to prevent such inventories from being manipulated.

Liruxo:
## Current Problems
There is more or less no synchronization in Slimefun whatsoever. Currently the workload of ticking blocks is offloaded in the asynchronous TickerTask which commits many Tasks back onto the main thread as synchronous task. However important tasks such as the EnergyNet are still processed asynchronously, which improves performance and poses synchronization challenges.

## Current Solutions
Most of the problems are solved by scheduling BlockStorage removals onto the TickerTask to ensure the TickerTask only works on a (mostly) consistent state of the BlockStorage
Additionally most block changes while or after a block removal is in process are sudo made consistent by the constant check for the SlimefunItem-ID, and the following disregard of any information if it is not present

## Currently Proposed Solution
This is my suggestions:
Make the BlockStorage and/or the accesses to it thread-safe, meaning it would be safe to process block ticks with multiple threads not just one asynchronously running TickerTask
[...]

PR #1821, opened 2020-04-23
Added comment to BlockStorage

Liruxo:
## Description
Currently none

## Changes
Currently none

## Related Issues
Goal is to solve #1819 

## Checklist

# Current Status
Currently this PR is goal is to provide space for a discussion about the BlockStorage thread-safety and/or the thread-safety of Slimefun in general

## Current Problems
There is more or less no synchronization in Slimefun whatsoever. Currently the workload of ticking blocks is offloaded in the asynchronous TickerTask which commits many Tasks back to the main thread as synchronous task. However important tasks such as the EnergyNet are still processed asynchronously, which improves performance and poses synchronization challenges.

## Current Solutions
Most of the problems are solved by scheduling BlockStorage removals onto the TickerTask to ensure the TickerTask only works on a (mostly) consistent state of the BlockStorage
Additionally most block changes while or after a block removal is in process are sudo made consistent by the constant check for the SlimefunItem-ID, and the following disregard of any information if it is not present

## Currently Proposed Solution
This is my suggestions:
Make the BlockStorage and/or the accesses to it thread-safe, meaning it would be safe to process block ticks with multiple threads not just one asynchronously running TickerTask
I would propose some sort of lock system, to ensure only one thread modifies a block at one time.
Alternatively it should be possible largely to realize this with the "compute" functions of the already used ConcurrentHashMap
### Cons
- These changes may worsen the overall performance of slimefun both on the main thread and the async thread because they both spend more time waiting on each other.
- It introduces the risk of deadlocks
### Pros
- The performance may increase, as it would now be possible to use more than 1 thread to process ticks. However this requires a very large workload, test on this should be performed.
- This should fix all the current race condition bugs, and eliminated the need for the current "Hack" to only process block removals asynchronously. I only mentioned one bug but i am positive there are many more bugs related to the current circumstances

## Closing Words
I am willing to produce these changes largely myself, but due to the enormous affected areas and the scale of this undertaking this would be very hard to realize only by myself.
Feel free to post other solution or problems with my current proposal
The current commit is only a thought for a hotfix and will be force-pushed out (or new PR)

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
@@ -495,6 +495,7 @@ public class BlockStorage {
         }
 
         if (destroy) {
+//          Check again for drops here
             if (storage.hasInventory(l)) storage.clearInventory(l);
 
             UniversalBlockMenu universalInventory = getUniversalInventory(l);


TheBusyBiscuit:
I think it is needless to say that creating a Pull Request just to add a single line of commenting is pointless.

I understand that this should just be the foundation of a discussion but creating a Pull Request solely for that makes no sense.

This discussion should probably be held under the actual issue.

Liruxo:
Is there any reason, you want to explain, on why this was closed without comment?
I find it very immature to close this without any mentioned reason or comment, as it took me quite some effort and time to create and research this post. If you have any issues with this, please be at least so kind to mention these before completely disregarding hard work of others.

TheBusyBiscuit:
I did post a comment...

Liruxo:
I excuse myself for the accusations, it seems i had some timing problems as you first closed and the posted your answer, sorry for that

Should i repost the main points of my proposal or is just referencing this PR fine?
Since we are already here, i might ask if there are currently any plans to refactor the BlockStoarge anyways, as it is currently in the, to my understanding, legacy package. It contains many methods referencing the old outdated CSCoreLib Config too.

Again, sorry for the accusations, i got a bit emotional and did not wait long enough for your answer

TheBusyBiscuit:
No offense taken.

I would suggest you to just copy-paste your comments as a new comment on the issue.
I don't really see much of a point in having two threads and jumping back and forfth.

Also I appreciate your engagement. Yes, all the me.mrCookieSlime... packages are very old and should be replaced at some point.
For the BlockStorage things I have no intentions of tampering with these, they were altered so many times that I have barely any idea of their workings anymore, so I better leave them as they are until someone tackles the problem.

I heard that Walshy was onto that at some point, so you could see what he has to say about this.
But anyway, let's start this discussion on the actual issue post then.

PR #1826, opened 2020-04-25
Easier to register items

WalshyDev:
## Description
Instead of having to do registerResearch(Research, ItemStack...) you can just do new Research(..., items).register()

Makes this more consistent with the other registering too.

## Changes
Varargs in constructor calling addItems

## Related Issues
N/A

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/Research.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/Research.java
@@ -63,11 +63,14 @@ public class Research implements Keyed {
      *            The Cost in XP levels to unlock this {@link Research}
      * 
      */
-    public Research(NamespacedKey key, int id, String name, int defaultCost) {
+    public Research(NamespacedKey key, int id, String name, int defaultCost, SlimefunItem... items) {
         this.key = key;
         this.id = id;
         this.name = name;
         this.cost = defaultCost;
+        
+        if (items.length > 0)
+            this.addItems(items);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/me/mrCookieSlime/Slimefun/Objects/Research.java
```diff
@@ -63,11 +63,14 @@
      *            The Cost in XP levels to unlock this {@link Research}
      * 
      */
-    public Research(NamespacedKey key, int id, String name, int defaultCost) {
+    public Research(NamespacedKey key, int id, String name, int defaultCost, SlimefunItem... items) {
```
TheBusyBiscuit, line 66: Changing this would break all addons at once because of this parameter change.

src/main/java/me/mrCookieSlime/Slimefun/Objects/Research.java
```diff
@@ -63,11 +63,14 @@
...
         this.key = key;
         this.id = id;
         this.name = name;
         this.cost = defaultCost;
+        
+        if (items.length > 0)
```
TheBusyBiscuit, line 72: This length check seems rather unnecessary since an empty array will simply not cause the loop to trigger.

WalshyDev:
> I don't really see much of a benefit of why this should be added. Since this will cram it all in one line rather than the current method:
I mean I can make the addItems return the instance so it can be chained instead but there's really no point making a variable JUST to add items. It's redundant. A construtor or chained addItems is definitely best I'd say.

TheBusyBiscuit:
I'm fine with chaining it. It'd at least be less intrusive and not break every addon in existence within a single commit XD

And depending on what we do with Researches in the future, a variable may be important to hold onto, got no plans for that though so feel free to chain it.

WalshyDev:
Chaining it is

PR #1844, opened 2020-05-01
Resolves #1843

### LINKED ISSUE #1843: Talisman of the Anvil doesn't repair armor if previous slots are empty
## Description (Required)
when you have empty slots in your armors talisman of the anvil doesn't repair broken armor piece.
Specifically, when there is an empty slot 

equip some boots and a talisman of the anvil, break the boots, a message appear saying "your talisman saved your tool from breaking" but the item dissapeared

Expect boots to be back in inventory at full durability

## Server Log / Error Report
https://pastebin.com/kJsjCHv4

## Environment (Required)

 - Minecraft Version: 1.15.2
 - Slimefun Version: vDEV - 496 (git 085d10be)
 - CS-CoreLib Version: vDEV - 85 (git 4a7fce52)

clementgallois:
## Description
The issue come from calling the "equals" method on item slots for the armor. those return null when no armor is present, raising an exception.

## Changes
Calling the method equals on the broken item to compare to the armor slot

## Related Issues
Resolves #1843

## Checklist

- [ x] I followed the existing code standards and didn't mess up the formatting.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
@@ -149,7 +149,7 @@ public class TalismanListener implements Listener {
             // Did the tool in our hand broke or was it an Armorpiece?
             if (!inv.getItem(inv.getHeldItemSlot()).equals(e.getBrokenItem())) {
                 for (int s : armorSlots) {
-                    if (inv.getItem(s).equals(e.getBrokenItem())) {
+                    if (e.getBrokenItem().equalst(inv.getItem(s))) {
                         slot = s;
                         break;
                     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
@@ -149,7 +149,7 @@ public class TalismanListener implements Listener {
             // Did the tool in our hand broke or was it an Armorpiece?
             if (!inv.getItem(inv.getHeldItemSlot()).equals(e.getBrokenItem())) {
                 for (int s : armorSlots) {
-                    if (e.getBrokenItem().equalst(inv.getItem(s))) {
+                    if (e.getBrokenItem().equals(inv.getItem(s))) {
                         slot = s;
                         break;
                     }


WalshyDev:
You should just do a null check... also you typo'd so this would fail to compile.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
```diff
@@ -149,7 +149,7 @@ public void onItemBreak(PlayerItemBreakEvent e) {
             // Did the tool in our hand broke or was it an Armorpiece?
             if (!inv.getItem(inv.getHeldItemSlot()).equals(e.getBrokenItem())) {
                 for (int s : armorSlots) {
-                    if (inv.getItem(s).equals(e.getBrokenItem())) {
+                    if (e.getBrokenItem().equalst(inv.getItem(s))) {
```
TheBusyBiscuit, line 152: That's a typo right there... "equalst".
You should always test your changes before commiting them. Especially seeing if it even compiles properly.

TheBusyBiscuit:
@WalshyDev A null Check is not even really needed since the broken item will never be null anyway.

WalshyDev:
@TheBusyBiscuit that's the entire point of his PR though. Something may return null so he's moved it. He should just null check that instead which let's face it is a better solution.

clementgallois:
Kind of disagree with the null check, the event happen when an item break, so it's that item that we're trying to compare to some value that could be null but that shouldn't matter.

Will gladly change to a null check if deemed better, let me know

WalshyDev:
I'm actually wondering why the for even is like this, there's an int array just for the slots however PlayerInventory has [getArmorContents](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/inventory/PlayerInventory.html#getArmorContents--)
That for should really just be removed and the variable removed for looping the contents, checking if it isn't null and then doing the equals.

Means if they change slots (which let's face it it's Mojang, can happen) it doesn't break.

TheBusyBiscuit:
> Kind of disagree with the null check, the event happen when an item break, so it's that item that we're trying to compare to some value that could be null but that shouldn't matter.
> 
> Will gladly change to a null check if deemed better, let me know

I personally wouldn't really insist on a null check, it isn't needed.
But from a theoretical standpoint, ItemStack#equals(...) is a third-party method and third-party methods must be treated like a black box, meaning we may never know what happens inside. By adding a null check you would protect the codebase against potential changes to this method which could break the software if a null parameter is passed.

This may not be very likely and again, your solution is definitely a fix for the underlying issue and I wouldn't necessarily insist on a null check.
But a null check is justifiable as one of good practices.

TheBusyBiscuit:
Null check or not, I am merging this now since it does fix the issue.
Thanks @clementgallois

PR #1853, opened 2020-05-04
Patch 8

WalshyDev:
## Description
Quick edit to add location data to the CargoNet parse error message. Then people can actually FIND the broken one.

This was done on GitHub web so yes I could have used Java Loggers horrible variable system but it's stupid so yeah no, not on web.

## Changes
Adds world name, x, y, z to NumberFormatException error

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
@@ -294,7 +294,8 @@ public class CargoNet extends ChestTerminalNetwork {
             return Integer.parseInt(str);
         }
         catch (Exception x) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -294,7 +294,8 @@ private static int getFrequency(Location l) {
             return Integer.parseInt(str);
         }
         catch (Exception x) {
-            Slimefun.getLogger().log(Level.SEVERE, "An Error occured while parsing a Cargo Node Frequency", x);
+            Slimefun.getLogger().log(Level.SEVERE, "An Error occured while parsing Cargo Node Frequency "
```
TheBusyBiscuit, line 297: Should probably use a deferred error message by using a lambda here.
Otherwise Sonarcloud will kill ya

PR #1892, opened 2020-05-09
Added Unit Tests via MockBukkit

### LINKED ISSUE #1891: Items stop smelting after a short time in Carbonado Edged Furnace
## Description (Required)

When smelting items in a Carbonado Edged Furnace after smelting a couple blocks it freezes and discontinues smelting. As per attached screenshot is what occurs

Smelting in a Carbonado Edged Furnace

At the time of occurence the furnace had a hopper inputing items and a hopper outputting items from the furnace, I tried breaking and replacing the furnace and also tried placing the furnace on it's own with no surrounding blocks. The same issue occured

I expected the items to be smelted

## Environment (Required)

 - Minecraft Version: 1.15.2
 - Slimefun Version: DEV - 499 (git c3dbe8f3)
 - CS-CoreLib Version: DEV - 85 (git 4a7fce52)

Mcmmo is installed on the server if this could have anything to do with it conflicting. I have no levels in the 'Smelting' skill on the server so personally don't believe it could affect me if it was a Mcmmo related confliction as it was recently added to the server. I had not made the Carbonado Edged Furnace before Mcmmo was added

TheBusyBiscuit:
## Description
This PR adds Unit tests to Slimefun and also updates the `pom.xml` with some normalizing changes

## Changes
* Implemented JUnit, MockBukkit, Mockito and JaCoco
* Relocated `LockedCategory ` and `Research` classes
* Added over 60 Unit Tests and a few mock classes
* The `PlayerProfile` class is now completely callback-based and ready for concurrent implementations
* Adjusted some implementations to be more resilient and added more `Validate...` statements to ensure integrity
* Added more documentation
* Relocated `NetworkManager` class
* Deprecated a few things
* Updated GitHubService repository urls
* `PerWorldSettings` now uses `World#getUID()` internally over `World#getName()`
* A bunch of refactoring

## Related Issues
Resolves #1891

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/SoulboundRune.java
```diff
@@ -104,6 +102,21 @@ public ItemDropHandler getItemHandler() {
         };
     }
 
+    /**
+     * This method applies the {@link Soulbound} effect onto a given {@link ItemStack}.
+     * 
+     * @param item
+     *            The {@link ItemStack} to apply this effect to
+     */
+    public void apply(ItemStack item) {
```
WalshyDev, line 111: While this is being moved to a method, should add a persistent data tag for this rather than relying on the lore.

src/test/java/io/github/thebusybiscuit/slimefun4/tests/items/TestCategories.java
```diff
@@ -0,0 +1,187 @@
...
+        Category category = new Category(new NamespacedKey(plugin, "items_test"), new CustomItem(Material.DIAMOND_AXE, "&6Testing"));
+        SlimefunItem item = SlimefunMocks.mockSlimefunItem(plugin, "CATEGORY_ITEMS_TEST_ITEM", new CustomItem(Material.BAMBOO, "&6Test Bamboo"));
+        item.setCategory(category);
+        item.register(plugin);
+        item.load();
+
+        Assertions.assertEquals(1, category.getItems().size());
+
+        // Size must still be 1 since we disallow duplicates
+        item.setCategory(category);
```
WalshyDev, line 65: You should also check size after this first one. You're checking the intended initial behaviour and then a checked behaviour (dupe)

src/test/java/io/github/thebusybiscuit/slimefun4/tests/services/TextCustomTextureService.java
```diff
@@ -0,0 +1,75 @@
...
+        Assertions.assertThrows(IllegalArgumentException.class, () -> service.getModelData(null));
+    }
+
+    @Test
+    public void testSetTexture() throws NoSuchFieldException, IllegalAccessException {
+        CustomTextureService service = new CustomTextureService(plugin);
+        SlimefunItem item = SlimefunMocks.mockSlimefunItem(plugin, "TEXTURE_TEST", new ItemStack(Material.LANTERN));
+        String version = "Unit Test v1.0";
+
+        Config config = (Config) ReflectionUtils.getFieldValue(service, "config");
```
WalshyDev, line 59: Not the biggest fan of using reflection in unit tests. You should be able to access the things you're making sure to test.

src/main/java/io/github/thebusybiscuit/slimefun4/core/categories/LockedCategory.java
```diff
@@ -0,0 +1,159 @@
...
+     * @param p
+     *            The {@link Player} to check
+     * @param profile
+     *            The {@link PlayerProfile} that belongs to the given {@link Player}
+     * @return Whether the {@link Player} has fully completed all parent categories, otherwise false
+     */
+    public boolean hasUnlocked(Player p, PlayerProfile profile) {
+        for (Category category : parents) {
+            for (SlimefunItem item : category.getItems()) {
+                // Should we replace this all with Slimefun.hasUnlocked() ?
```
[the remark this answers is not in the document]
TheBusyBiscuit: That comment is from last year or so... It only shows as a diff because the class was relocated

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/SoulboundRune.java
TheBusyBiscuit: Definitely agree with this, not my code, just refactored it.

src/test/java/io/github/thebusybiscuit/slimefun4/tests/items/TestCategories.java
TheBusyBiscuit: I know I could, doing it twice was just an exaggeration

src/test/java/io/github/thebusybiscuit/slimefun4/tests/services/TextCustomTextureService.java
TheBusyBiscuit: Yeah me neither but we shouldn't expose that Config though, so Reflection is the best we can do to inject some values there

PR #1898, opened 2020-05-10
Backpack command

Sfiguz7:
## Description
Added a command that allows for backpack retrieval in the event they are lost. The command accepts a name and id, if those match up it spawns a Medium Backpack with the correct lore set in the sender's inventory - if not it sends them a message about which of the two isn't found.

## Changes
Added a new SubCommand class, edited messages-en.yml to include 2 new error messages and and the command description, edited plugin.yml to add the command

## Related Issues
N/A

## Checklist
- [ V ] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ = ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
Only ExtraTools and TranscEndence were loaded with it, it does not mess with anything Addon related though.
- [ X ] I followed the existing code standards and didn't mess up the formatting.
Under Walshy's pressure (so blame him) I rewrote the SubCommand to have if/return checks instead of the nested ifs other SubCommands follow.
- [ X ] I did my best to add documentation to any public classes or methods I added.
Not sure what "proper" documentation would imply for this class so I assume not.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
@@ -1,0 +1,4 @@
+package io.github.thebusybiscuit.slimefun4.core.commands.subcommands;
+
+public class BackpackCommand {
+}
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
@@ -1,16 +1,23 @@
 package io.github.thebusybiscuit.slimefun4.core.commands.subcommands;
 
+import io.github.thebusybiscuit.slimefun4.api.player.PlayerBackpack;
 import io.github.thebusybiscuit.slimefun4.api.player.PlayerProfile;
 import io.github.thebusybiscuit.slimefun4.core.commands.SlimefunCommand;
 import io.github.thebusybiscuit.slimefun4.core.commands.SubCommand;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.BackpackListener;
 import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
 import me.mrCookieSlime.Slimefun.Lists.SlimefunItems;
 import me.mrCookieSlime.Slimefun.SlimefunPlugin;
 import me.mrCookieSlime.Slimefun.api.Slimefun;
 import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
 import org.bukkit.OfflinePlayer;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.ItemMeta;
+
+import java.util.List;
 
 class BackpackCommand extends SubCommand {
 
@@ -30,7 +37,7 @@ class BackpackCommand extends SubCommand {
 
     @Override
     public void onExecute(CommandSender sender, String[] args) {
-        if (!(sender instanceof Player) || !(sender.hasPermission("slimefun.command.backpack"))) {
+        if (!(sender instanceof Player) || !sender.hasPermission("slimefun.command.backpack")) {
             SlimefunPlugin.getLocal().sendMessage(sender, "messages.no-permission", true);
             return;
         }
@@ -41,13 +48,12 @@ class BackpackCommand extends SubCommand {
         }
 
         final Player p = (Player) sender;
-        final String ownerName = args[1];
-        if (!PatternUtils.NUMERIC.matcher(args[2]).matches()){
+        if (!PatternUtils.NUMERIC.matcher(args[2]).matches()) {
             SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.invalid-id");
             return;
         }
         final int id = Integer.parseInt(args[2]);
-        final OfflinePlayer owner = Bukkit.getOfflinePlayer(ownerName);
+        final OfflinePlayer owner = Bukkit.getOfflinePlayer(args[1]);
         if (!owner.hasPlayedBefore()) {
             SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.player-never-joined");
             return;
@@ -59,7 +65,13 @@ class BackpackCommand extends SubCommand {
                 return;
             }
             Slimefun.runSync(() -> {
-                p.getInventory().addItem(SlimefunItems.RESTORED_BACKPACK);
+                ItemStack item = SlimefunItems.RESTORED_BACKPACK;
+                ItemMeta meta = item.getItemMeta();
+                List lore = meta.getLore();
+                lore.set(2, ChatColor.GRAY + "ID: " + profile.getUUID().toString() + "#" + id);
+                meta.setLore(lore);
+                item.setItemMeta(meta);
+                p.getInventory().addItem(item);
                 SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.restored-backpack-given");
             });
         });
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
@@ -1,21 +1,16 @@
 package io.github.thebusybiscuit.slimefun4.core.commands.subcommands;
 
-import io.github.thebusybiscuit.slimefun4.api.player.PlayerBackpack;
 import io.github.thebusybiscuit.slimefun4.api.player.PlayerProfile;
 import io.github.thebusybiscuit.slimefun4.core.commands.SlimefunCommand;
 import io.github.thebusybiscuit.slimefun4.core.commands.SubCommand;
+import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
 import me.mrCookieSlime.Slimefun.Lists.SlimefunItems;
 import me.mrCookieSlime.Slimefun.SlimefunPlugin;
 import me.mrCookieSlime.Slimefun.api.Slimefun;
 import org.bukkit.Bukkit;
-import org.bukkit.ChatColor;
 import org.bukkit.OfflinePlayer;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
-import org.bukkit.inventory.ItemStack;
-import org.bukkit.inventory.meta.ItemMeta;
-
-import java.util.List;
 
 class BackpackCommand extends SubCommand {
 
@@ -40,21 +35,21 @@ class BackpackCommand extends SubCommand {
             return;
         }
         if (args.length != 3) {
-            SlimefunPlugin.getLocal().sendMessage(sender, "messages.usage", true, msg -> msg.replace("%usage%", "/sf backpack <Player> <Number>"));
+            SlimefunPlugin.getLocal().sendMessage(sender, "messages.usage", true, msg -> msg.replace("%usage%", "/sf backpack <Player> <ID>"));
             return;
         }
 
-        Player p = (Player) sender;
-        String ownerName = args[1];
-        int id = Integer.parseInt(args[2]);
-        OfflinePlayer owner = Bukkit.getPlayer(ownerName);
-
-        if (owner == null) {
-            owner = Bukkit.getOfflinePlayer(ownerName);
-            if (!owner.hasPlayedBefore()) {
-                SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.player-never-joined");
-                return;
-            }
+        final Player p = (Player) sender;
+        final String ownerName = args[1];
+        if (!PatternUtils.NUMERIC.matcher(args[2]).matches()){
+            SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.invalid-id");
+            return;
+        }
+        final int id = Integer.parseInt(args[2]);
+        final OfflinePlayer owner = Bukkit.getOfflinePlayer(ownerName);
+        if (!owner.hasPlayedBefore()) {
+            SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.player-never-joined");
+            return;
         }
 
         PlayerProfile.get(owner, profile -> {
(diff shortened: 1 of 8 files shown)


WalshyDev:
First off, checkboxes can be ticked with `[X]`.

Secondly,
> Under Walshy's pressure
If it was under my pressure you'd also have all those variables as final. 

You also should really give the correct backpack. You can't give Soulbound but that's fine as the admin can get a Soulbound rune anyway. However, they can't just change it from a medium to large or small.

Sfiguz7:
Boi you could've said that when I told you I was ditching the switch :(
Checkboxes I ticked differently to differentiate between yes no and kindof, as I explained in a one liner just below. I'll happily make the variables final too for you <3

TheBusyBiscuit:
Alternatively (since the rest of the itemmeta doesn't matter) you could give them a "Restored Backpack" or "Generic Backpack" item which simply serves as a reference to a lost backpack, can't upgrade that one though of course.

Also going by size could lead to issues since coolers or dyed backpacks also classify as backpacks but share the same size as other backpacks.

WalshyDev:
really should just be an id field in the backpack section of data

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,76 @@
...
+            return;
+        }
+
+        Player p = (Player) sender;
+        String ownerName = args[1];
+        int id = Integer.parseInt(args[2]);
+        OfflinePlayer owner = Bukkit.getPlayer(ownerName);
+
+        if (owner == null) {
+            owner = Bukkit.getOfflinePlayer(ownerName);
```
TheBusyBiscuit, line 53: This is a little redundant. Bukkit.getOfflinePlayer(...) will return an online player if possible anyway, so potentially calling this twice makes little sense, can pretty much just call Bukkit.getOfflinePlayer(...) right away

src/main/resources/plugin.yml
```diff
@@ -37,6 +37,9 @@ permissions:
   slimefun.command.versions:
     description: Allows you to do /sf versions
     default: op
+  slimefun.command.backpack:
+    description: Allows you to do /sf backpack
+    default: op
```
TheBusyBiscuit, line 42: Also make sure to add the new permission and command to the wiki

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/Commands.java
```diff
@@ -25,5 +25,6 @@ public static void addCommands(SlimefunCommand cmd, Collection<SubCommand> comma
         commands.add(new OpenGuideCommand(plugin, cmd));
         commands.add(new SearchCommand(plugin, cmd));
         commands.add(new DebugFishCommand(plugin, cmd));
+        commands.add(new BackpackCommand(plugin,cmd));
```
TheBusyBiscuit, line 28: Please add a space after the comma

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,76 @@
...
+                return;
+            }
+            PlayerBackpack bp = profile.getBackpack(id).get();
+            //No point switching for size, would still miss soulbound
+            ItemStack item = SlimefunItems.BACKPACK_MEDIUM;
+            ItemMeta meta = item.getItemMeta();
+            List lore = meta.getLore();
+            lore.set(2, ChatColor.GRAY + "ID: " + profile.getUUID().toString() + "#" + id);
+            meta.setLore(lore);
+            item.setItemMeta(meta);
```
TheBusyBiscuit, line 72: Pretty sure you can just use BackpackListener#setBackpackId(...) or so here, might have to look into that

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,76 @@
...
+            Slimefun.runSync(() -> p.getInventory().addItem(item));
```
TheBusyBiscuit, line 73: Should probably also send a message like "Heres your backpack" or so

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,66 @@
...
+            SlimefunPlugin.getLocal().sendMessage(sender, "messages.no-permission", true);
+            return;
+        }
+        if (args.length != 3) {
+            SlimefunPlugin.getLocal().sendMessage(sender, "messages.usage", true, msg -> msg.replace("%usage%", "/sf backpack <Player> <ID>"));
+            return;
+        }
+
+        final Player p = (Player) sender;
+        final String ownerName = args[1];
```
WalshyDev, line 44: Variable is only ever used once, no point in it

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public void onExecute(CommandSender sender, String[] args) {
+        if (!(sender instanceof Player) || !(sender.hasPermission("slimefun.command.backpack"))) {
+            SlimefunPlugin.getLocal().sendMessage(sender, "messages.no-permission", true);
+            return;
+        }
+        if (args.length != 3) {
+            SlimefunPlugin.getLocal().sendMessage(sender, "messages.usage", true, msg -> msg.replace("%usage%", "/sf backpack <Player> <ID>"));
```
WalshyDev: Looks very long, I assume this is over column limit?

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public boolean isHidden() {
+        return false;
+    }
+
+    @Override
+    public void onExecute(CommandSender sender, String[] args) {
+        if (!(sender instanceof Player) || !(sender.hasPermission("slimefun.command.backpack"))) {
```
WalshyDev, line 33: Redundant brackets around hasPermission statement

src/main/java/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
```diff
@@ -72,6 +72,8 @@ private SlimefunItems() {}
     public static final ItemStack RADIANT_BACKPACK = new SlimefunItemStack("RADIANT_BACKPACK", "40cb1e67b512ab2d4bf3d7ace0eaaf61c32cd4681ddc3987ceb326706a33fa", "&eRadiant Backpack", "", "&7Size: &e54 (Double chest)", "&7ID: <ID>", "", "&7&eRight Click&7 to open");
     public static final SlimefunItemStack BOUND_BACKPACK = new SlimefunItemStack("BOUND_BACKPACK", "2a3b34862b9afb63cf8d5779966d3fba70af82b04e83f3eaf6449aeba", "&cSoulbound Backpack", "", "&7Size: &e36", "&7ID: <ID>", "", "&7&eRight Click&7 to open");
     public static final SlimefunItemStack COOLER = new SlimefunItemStack("COOLER", "d4c1572584eb5de229de9f5a4f779d0aacbaffd33bcb33eb4536a6a2bc6a1", "&bCooler", "&rAllows you to store Juices/Smoothies", "&rand automatically consumes them when you are hungry", "&rand you have this in your Inventory", "", "&7Size: &e27", "&7ID: <ID>", "", "&7&eRight Click&7 to open");
+    public static final ItemStack RESTORED_BACKPACK = new SlimefunItemStack("RESTORED_BACKPACK", "40cb1e67b512ab2d4bf3d7ace0eaaf61c32cd4681ddc3987ceb326706a33fa", "&eRestored Backpack", "", "&7Retrieve your lost items", "", "&7&eRight Click&7 to open");
```
WalshyDev, line 75: Make this a SlimefunItemStack

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,67 @@
...
+            return;
+        }
+
+        PlayerProfile.get(owner, profile -> {
+            if (!profile.getBackpack(id).isPresent()) {
+                SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.backpack-does-not-exist");
+                return;
+            }
+            Slimefun.runSync(() -> {
+                p.getInventory().addItem(SlimefunItems.RESTORED_BACKPACK);
```
WalshyDev, line 62: Never set the backpack ID or anything...

src/main/java/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
Sfiguz7: Why? All of them are ItemStack, it would not be coherent

WalshyDev: They're being moved to SlimefunItemStack. ItemStack is the old shit. You're right now just causing stupidly useless casts.

WalshyDev:
Be careful formatting SF... you did did over 6000 line changes

Sfiguz7:
How the...? What did I do/can it be prevented?

WalshyDev:
Just dont press your formatting key on a class you know for a fact isn't formatted well. 

You'll need to revert the last commit on that file.

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,77 @@
...
+                SlimefunPlugin.getLocal().sendMessage(sender, "guide.backpack.backpack-does-not-exist");
+                return;
+            }
+            Slimefun.runSync(() -> {
+                ItemStack item = SlimefunItems.RESTORED_BACKPACK;
+                ItemMeta meta = item.getItemMeta();
+                List lore = meta.getLore();
+                lore.set(2, ChatColor.GRAY + "ID: " + profile.getUUID().toString() + "#" + id);
+                meta.setLore(lore);
+                item.setItemMeta(meta);
```
TheBusyBiscuit, line 71: Pretty sure you can do BackpackListener#setBackpackId(...) her... wait a second didn't I already say that?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HiddenBackpack.java
```diff
@@ -0,0 +1,14 @@
+package io.github.thebusybiscuit.slimefun4.implementation.items.tools;
+
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+import org.bukkit.inventory.ItemStack;
+
+public class HiddenBackpack extends SlimefunBackpack {
```
TheBusyBiscuit, line 8: Should maybe rather be called "RestoredBackpack". Also please add adequate javadocs here.

src/main/resources/languages/messages_en.yml
```diff
@@ -42,6 +43,12 @@ guide:
 
   cheat:
     no-multiblocks: '&4You cannot cheat in Multiblocks, you have to build them!'
+
+  backpack:
+    invalid-id: '&4The id must be a non-negative integer!'
+    player-never-joined: '&4No player with that name has ever joined the server!'
+    backpack-does-not-exist: '&4That backpack does not exist!'
+    restored-backpack-given: 'Backpack restored successfully! Added to your inventory!'
```
TheBusyBiscuit, line 51: Why exactly is this in the "guide" section? Doesn't seem to have anything to do with the Slimefun guide to me.
Should probably rather be in the "commands" section.

src/main/resources/languages/messages_en.yml
```diff
@@ -42,6 +43,12 @@ guide:
 
   cheat:
     no-multiblocks: '&4You cannot cheat in Multiblocks, you have to build them!'
+
+  backpack:
+    invalid-id: '&4The id must be a non-negative integer!'
```
TheBusyBiscuit, line 48: Rather use the word "number", "integer" does not appear to be so user-friendly. Also maybe call it "backpack id" rather than just "id"

src/main/resources/languages/messages_en.yml
TheBusyBiscuit, line 51: Maybe some color? :eyes:

Sfiguz7: RED! 😈

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
Sfiguz7: I think I mistakenly resolved it and didn't read it heh now it's resolved

src/main/resources/languages/messages_en.yml
```diff
@@ -15,6 +15,12 @@ commands:
     description: Unlock/Reset researches for a player
     reset: '&cYou have reset %player%''s Knowledge'
     reset-target: '&cYour Knowledge has been reset'
+
+  backpack:
```
TheBusyBiscuit, line 19: This is actually a duplicate key. You already placed the description under "commands.backpack".

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/BackpackCommand.java
```diff
@@ -0,0 +1,72 @@
...
+            return;
+        }
+
+        PlayerProfile.get(owner, profile -> {
+            if (!profile.getBackpack(id).isPresent()) {
+                SlimefunPlugin.getLocal().sendMessage(sender, "commands.backpack.backpack-does-not-exist");
+                return;
+            }
+            Slimefun.runSync(() -> {
+                ItemStack item = SlimefunItems.RESTORED_BACKPACK;
```
TheBusyBiscuit, line 66: Hold up...
Very dangerous here, you need to add a .clone() otherwise you are modifying the original and not a copy.

src/main/resources/languages/messages_en.yml
```diff
@@ -17,6 +16,7 @@ commands:
     reset-target: '&cYour Knowledge has been reset'
 
   backpack:
+    description: Retrieve an existing backpack
```
TheBusyBiscuit, line 19: You updated the key though but you don't seme to use it actually :?

Sfiguz7: Doesn't /sf help retrieve that field(?)

TheBusyBiscuit: How would it know what field to retrieve though...?
You just changed the standard location for it so you need to override getDescription() too.

Sfiguz7: Hmm idk I tried /sf help and it showed the correct one so I assumed it worked. I'll review it then

Sfiguz7: I see it now, overridden getDescription(); I'll wipe my folder just in case and test, then push

Sfiguz7:
When I changed it to BackpackListener#setBackpackID I passed the wrong player; that method requires a Player though and owner is only an OfflinePlayer so IntelliJ gets mad.
Should I revert to having the 5 lines explicitly? Would you prefer if I changed the required parameter to an OfflinePlayer instead?

PR #1909, opened 2020-05-14
Addition: Bee Armor

beSn0w:
## Description

I have created Bee Armor, basically the honey equivalent of the Slime Armor.
Because bees fly I also replaced the chestplate with an Elytra (I was really debating whether to put it in the "Magical Gadgets" tab but it Probably wouldn't be obvious that it was a part of the bee set)
I also added some features to the armor to make life easier for those players playing on high latency servers, taking off is much easier and landing (because of the slow fall effect) is a piece of cake, the boots are also fall damage resistant and give you a jump boost (at that point i realized the similarity's are a bit to much so i stopped there with features)
Also the ballance may need to be tweaked a bit, i havent tested this extencively in survival yet.

Side note: I'm relatively new to coding so if I messed something up in a major way please go easy on me.

## Changes

Added BeeWingListener which is responsible for the slow falling effect given to the player when nearing the ground.

Added the Bee Boots to the SlimefunBootsListener to make them negate damage and play a the "Honey Fall" sound effect when they "absorb" the damage.

Added the armor to the ReserchSetup, SlimefunItemSetup and SlimefunItems

## Related Issues

N/A

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author beSnow
+ */
+public class BeeWingListener implements Listener {
+
+    public BeeWingListener(SlimefunPlugin plugin) {
+        plugin.getServer().getPluginManager().registerEvents(this, plugin);
+    }
+
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    public void onApproachGround(PlayerMoveEvent e) {
```
TheBusyBiscuit, line 31: Using the PlayerMoveEvent is very risky as it can cause a ton of lag.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,54 @@
...
+        Player player = e.getPlayer();
+        ItemStack helmet = e.getPlayer().getInventory().getChestplate();
+        if (!SlimefunUtils.isItemSimilar(helmet, SlimefunItems.BEE_WINGS, true) && !Slimefun.hasUnlocked(e.getPlayer(), helmet, true)) return;
+        if (!player.isGliding()) return;
+        if (getDistanceToGround(player) == 0) return; //More accurate than Entity#isOnGround()
```
TheBusyBiscuit, line 36: Exclusion checks like these should be at the top for performance reasons.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/ResearchSetup.java
```diff
@@ -257,6 +257,7 @@ public static void setupResearches() {
         register("kelp_cookie", 254, "Tasty Kelp", 4, SlimefunItems.KELP_COOKIE);
         register("makeshift_smeltery", 255, "Improvised Smeltery", 6, SlimefunItems.MAKESHIFT_SMELTERY);
         register("tree_growth_accelerator", 256, "Faster Trees", 18, SlimefunItems.TREE_GROWTH_ACCELERATOR);
+        register("bee_armor", 257, "Bee Armor", 24, SlimefunItems.BEE_HELMET, SlimefunItems.BEE_WINGS, SlimefunItems.BEE_LEGGINGS, SlimefunItems.BEE_BOOTS);
```
TheBusyBiscuit, line 260: Bees/Honey blocks have only been introduced in 1.15. Slimefun however is for 1.13, 1.14 and 1.15. So you will need to do adequate checks to make these version-exclusive...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
```diff
@@ -3352,6 +3352,23 @@ public int getSpeed() {
...
+        .register(plugin);
+
+        new SlimefunItem(categories.magicalArmor, SlimefunItems.BEE_LEGGINGS, RecipeType.ARMOR_FORGE,
+        new ItemStack[]{SlimefunItems.GOLD_8K, new ItemStack(Material.HONEY_BLOCK), SlimefunItems.GOLD_8K, new ItemStack(Material.HONEYCOMB_BLOCK), null, new ItemStack(Material.HONEYCOMB_BLOCK), new ItemStack(Material.HONEYCOMB_BLOCK), null, new ItemStack(Material.HONEYCOMB_BLOCK)})
+        .register(plugin);
+
+        new SlimefunArmorPiece(categories.magicalArmor, SlimefunItems.BEE_BOOTS, RecipeType.ARMOR_FORGE,
+        new ItemStack[]{null, null, null, SlimefunItems.GOLD_8K, null, SlimefunItems.GOLD_8K, new ItemStack(Material.HONEY_BLOCK), null, new ItemStack(Material.HONEY_BLOCK)},
+        new PotionEffect[]{new PotionEffect(PotionEffectType.JUMP, 300, 2)})
+        .register(plugin);
```
TheBusyBiscuit, line 3371: Bees/Honey blocks have only been introduced in 1.15. Slimefun however is for 1.13, 1.14 and 1.15. So you will need to do adequate checks to make these version-exclusive...

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -237,6 +238,7 @@ else if (getServer().getPluginManager().isPluginEnabled("CS-CoreLib")) {
             new FireworksListener(this);
             new WitherListener(this);
             new IronGolemListener(this);
+            new BeeWingListener(this);
```
TheBusyBiscuit, line 241: Bees/Honey blocks have only been introduced in 1.15. Slimefun however is for 1.13, 1.14 and 1.15. So you will need to do adequate checks to make these version-exclusive...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
WalshyDev, line 31: This is ran on every little movement, it can run hundreds or thousands of times a second. Doing an item lookup EACH time is gonna REALLY hurt performance. You should do a ton of checks before ANYTHING to make sure this is exactly what you want!

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/ResearchSetup.java
TheBusyBiscuit, line 260: You added a new research but didnt update the language files...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,54 @@
...
+
+        if (getDistanceToGround(player) <= 6)
+            player.addPotionEffect(new PotionEffect(PotionEffectType.SLOW_FALLING, 30, 0));
+    }
+
+    private static int getDistanceToGround(Entity e) {
```
WalshyDev, line 42: You should use the heightmaps (getHighestBlockAt, getHighestBlockYAt) which would be much more efficient.

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -9,6 +9,38 @@
...
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.SlimefunGuideListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.SlimefunItemConsumeListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.SlimefunItemListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.SoulboundListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.TalismanListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.VampireBladeListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.VanillaMachinesListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.WitherListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.WorldListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.BeeWingListener;
```
WalshyDev, line 43: Should probably just revert all this. It makes no real change but now imports are quite unorganised

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @author beSnow
+ */
+public class BeeWingListener implements Listener {
+
+    public BeeWingListener(SlimefunPlugin plugin) {
+        plugin.getServer().getPluginManager().registerEvents(this, plugin);
+    }
+
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
```
WalshyDev, line 30: Also you shouldn't ignore cancelled. There's really no point.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
beSn0w: Yeah, i should have been more concernded about the performance.
I'm currently looking for an alternative to PlayerMoveEvent.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
beSn0w: A bit folish of me, will fix it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
beSn0w: Fixed and tested with my latest commit.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
beSn0w: Looking into it, good suggestion.

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
beSn0w: Fixed. I don't know why my auto formater even puts stuff in Wildcards, I'll have to see if i can turn that off.

src/main/java/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
```diff
@@ -267,6 +266,35 @@ private SlimefunItems() {}
...
     public static final SlimefunItemStack BOOTS_OF_THE_STOMPER = new SlimefunItemStack("BOOTS_OF_THE_STOMPER", Material.LEATHER_BOOTS, Color.AQUA, "&bBoots of the Stomper", "", "&9All Fall Damage you receive", "&9will be applied to nearby Mobs/Players", "", "&9+ No Fall Damage");
 
+    public static final SlimefunItemStack BEE_HELMET;
+    public static final SlimefunItemStack BEE_WINGS;
+    public static final SlimefunItemStack BEE_LEGGINGS;
+    public static final SlimefunItemStack BEE_BOOTS;
+
+static {
+
+    if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
```
TheBusyBiscuit, line 276: These items actually don't use any 1.15 Materials, so there isn't really much of a point in initializing them conditionally here. It's fine for these items to exist on 1.13/1.14 since you already handle that in SlimefunItemSetup

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -9,6 +9,7 @@
 import java.util.logging.Level;
 import java.util.stream.Collectors;
 
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.*;
```
TheBusyBiscuit, line 12: Please do not use wildcard imports.

beSn0w: I thought I fixed that, I must have accidentally used the auto formater.
Fixed in my latest Commit.

src/main/java/me/mrCookieSlime/Slimefun/Lists/SlimefunItems.java
beSn0w: Good point, I'll just go ahead and replace the recipe with something else for the versions below 1.15.X

TheBusyBiscuit: nononononononoononnoonoononono.

The way you handled it in SlimefunItemSetup was totally correct.
It's just redundant at this spot since you have already done it in SlimefunItemSetup.
Do **not** introduce alternate recipes because that will just be a total mess and frankly, having bee armor prior to 1.15 doesn't really make much sense anyway.

beSn0w: Ah, I understand. Its indeed a bit wierd to have bee themed armor in a version without bees.

src/main/resources/languages/researches_en.yml
```diff
@@ -230,4 +230,4 @@ slimefun:
   magnesium_generator: Power from Magnesium
   kelp_cookie: Tasty Kelp
   makeshift_smeltery: Improvised Smeltery
-  tree_growth_accelerator: Faster Trees
+  tree_growth_accelerator: Faster Trees
```
TheBusyBiscuit, line 233: Aside from all of the above issues that are still waiting, this is also still missing an entry for your new research. There are no file changes.

beSn0w: I thought i added that, pushing all changes now.
(my IDE has been acting up recently, either that or I'm just not good with version control)

TheBusyBiscuit (dismissed):
Alright, all major issues seem resolved now.
To me this is looking good, sorry for the delayed response btw, I currently find myself very stressed inbetween work and university.

Code-wise this is looking good to go, I would still like to spark a discussion about balancing and performance improvements though before I am merging this.
Anyway, thanks for making the pull request in the first place :)

WalshyDev:
I still think the event fires way too many times. Every little yaw, pitch, x, y and z fire it. Eliminating a lot of those and also something basic like a Material check would help a lot.

beSn0w:
I was thinking of maybe instaid of using PlayeMoveEvent I could maybe trigger the event on pressing shift, doing that will probably make it perform a lot better but might make it a bit overpowered if you can always just toggle Slowfalling at will.

TheBusyBiscuit:
Pressing shift to activate Slow falling would be very similar - if not identical to the parachute actually.
So I don't think it would be overpowered. But it would definitely be a lot better performance-wise.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
LinoxGH: you can give the potion effect slow falling?

LinoxGH:
What about adding the feature of negating damage from angry bees?

WalshyDev:
Is this now a stale PR?

TheBusyBiscuit:
I mean it's been stale since some time, however @poma123 just resolved all merge conflicts. If someone would address the performance concerns and at least two members approve it all, it could still be merged of course.

LinoxGH:
What about adding a counter to BeeWingListener? So like every 5th or 10th event fired sill be processed. I think this would improve performance.

TheBusyBiscuit:
The build is failing, compilation errors are probably more important first :D

LinoxGH:
I miss using an IDE.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
```diff
@@ -219,8 +220,7 @@ else if (getServer().getPluginManager().isPluginEnabled("CS-CoreLib")) {
             if (config.getBoolean("options.auto-update")) {
                 getLogger().log(Level.INFO, "Starting Auto-Updater...");
                 updaterService.start();
-            }
-            else {
+            } else {
```
TheBusyBiscuit, line 223: } and else on seperate lines please.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
```diff
@@ -350,6 +355,15 @@ private SlimefunItems() {}
         SLIME_CHESTPLATE_STEEL.addUnsafeEnchantments(slime);
         SLIME_LEGGINGS_STEEL.addUnsafeEnchantments(slime);
         SLIME_BOOTS_STEEL.addUnsafeEnchantments(slime);
+
+        Map<Enchantment, Integer> bee = new HashMap<>();
```
TheBusyBiscuit, line 359: I was gonna point out the irresponsibly chosen name, should have rather been "beeArmorEnchantments".
But then I realized the above armorsets use the exact same pattern ._.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @EventHandler(priority = EventPriority.LOWEST)
+    public void onApproachGround(PlayerMoveEvent e) {
+        Player player = e.getPlayer();
+        
+        if (!player.isGliding()) return;
+        if (player.isOnGround()) return;
+        
+        ItemStack chestplate = player.getInventory().getChestplate();
+        if (chestplate.getType() != Material.ELYTRA) return;
```
TheBusyBiscuit, line 42: This check is pretty redundant since the material check is one of the first lines in `SlimefunUtils.isItemSimilar` anyway. So this adds no real benefit to it, neither readability, efficiency nor performance.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
```diff
@@ -3047,19 +3047,38 @@ public int getSpeed() {
...
+
+    			  new SlimefunItem(categories.magicalArmor, SlimefunItems.BEE_LEGGINGS, RecipeType.ARMOR_FORGE,
+    			  new ItemStack[]{SlimefunItems.GOLD_8K, new ItemStack(Material.HONEY_BLOCK), SlimefunItems.GOLD_8K, new ItemStack(Material.HONEYCOMB_BLOCK), null, new ItemStack(Material.HONEYCOMB_BLOCK), new ItemStack(Material.HONEYCOMB_BLOCK), null, new ItemStack(Material.HONEYCOMB_BLOCK)})
+    			  .register(plugin);
+
+    			  new SlimefunArmorPiece(categories.magicalArmor, SlimefunItems.BEE_BOOTS, RecipeType.ARMOR_FORGE,
+    			  new ItemStack[]{null, null, null, SlimefunItems.GOLD_8K, null, SlimefunItems.GOLD_8K, new ItemStack(Material.HONEY_BLOCK), null, new ItemStack(Material.HONEY_BLOCK)},
+    			  new PotionEffect[]{new PotionEffect(PotionEffectType.JUMP, 300, 2)})
+    			  .register(plugin);
+    		}
```
TheBusyBiscuit, line 3074: Wrong indentation

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * This {@link Listener} is responsible for the slow falling effect given to the player
+ * when nearing the ground while using the Bee Wings.
+ *
+ * @author beSnow
+ * 
+ */
+public class BeeWingListener implements Listener {
+
+    public BeeWingListener(SlimefunPlugin plugin) {
```
TheBusyBiscuit, line 29: Also since we now enforce these: Nonnull please

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,70 @@
...
+import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
+import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
+import me.mrCookieSlime.Slimefun.api.Slimefun;
+
+/**
+ * This {@link Listener} is responsible for the slow falling effect given to the player
+ * when nearing the ground while using the Bee Wings.
+ *
+ * @author beSnow
+ *
```
TheBusyBiscuit, line 25: @LinoxGH You can pretty much add yourself as an author here below too at this point.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,70 @@
...
+            player.addPotionEffect(new PotionEffect(PotionEffectType.SLOW_FALLING, 40, 0));
+            return;
+        }
+
+        if (playerDistanceToHighestBlock <= 6) {
+            player.addPotionEffect(new PotionEffect(PotionEffectType.SLOW_FALLING, 40, 0));
+        }
+    }
+
+    private static int getDistanceToGround(Entity e) {
```
TheBusyBiscuit, line 59: Nonnull and documentation please.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,70 @@
...
+        Location loc = e.getLocation().clone();
+        double y = loc.getBlockY();
+        int distance = 0;
+        for (double i = y; i >= 0; i--) {
```
TheBusyBiscuit, line 63: This should rather be an int since you only decrement by whole numbers anyway

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeWingListener.java
```diff
@@ -0,0 +1,70 @@
...
+            loc.setY(i);
+            if (loc.getBlock().getType().isSolid()) break;
+            distance++;
+        }
+        return distance;
```
TheBusyBiscuit, line 68: Not really any need for this `distance` variable here, it is a bit redundant. You can just return y - (the new loc y)

TheBusyBiscuit (changes requested):
Just some small nitpickings. Other than that this is only really a matter of performance-balancing now.

TheBusyBiscuit (changes requested):
This has already seen big improvements from when we left off.
But I am still hesitant in terms of performance as the `PlayerMoveEvent` is really a problem in of itself, ideally something similar like with Jetpacks or Parachutes may be useful, a `PlayerTask` which runs as long as the Player is gliding and in the air etc...

PR #1919, opened 2020-05-18
Add location data to caro frequency error message

WalshyDev:
## Description
Couldn't repro this issue and changing frequency in file causes it to fail on load rather than tick so... yeah. I assume this works as intended.
(Also, yes, done on laptop this time not browser so Sonar wont bitch)

## Changes
Added world and x,y,z to the error so that someone can fix it more easily.

Not sure if a String#format would be better here (Or better yet, taking one of my utils and using that for formatting instead).

## Related Issues
N/A

## Checklist


TheBusyBiscuit:
>  I added sufficient Unit Tests to cover my code.
:eyes:

PR #1924, opened 2020-05-22
Fix compatibility with other plugins / Explosive Pickaxe

FabioZumbi12:
## Description
Fix compatibility with other plugins firing bukkit explosion event to do not drop near blocks on pickaxe area if the block was removed from explosion event.

## Changes
Instead to check only for block break, fire explosion event using the blocks from explosive pickaxe area.

## Related Issues
Not open already, as i am a developer of the plugin usim together SlimeFun and done fixing before open an issue.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
@@ -10,6 +10,7 @@ import org.bukkit.Sound;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
 import org.bukkit.event.block.BlockBreakEvent;
+import org.bukkit.event.block.BlockExplodeEvent;
 import org.bukkit.event.entity.EntityExplodeEvent;
 import org.bukkit.inventory.ItemStack;
 
@@ -72,12 +73,10 @@ public class ExplosivePickaxe extends SimpleSlimefunItem<BlockBreakHandler> impl
                             }
                         }
 
-                        EntityExplodeEvent entityExplodeEvent = new EntityExplodeEvent(e.getPlayer(), e.getBlock().getLocation(), blocks, 0);
-                        Bukkit.getServer().getPluginManager().callEvent(entityExplodeEvent);
-                        if (!entityExplodeEvent.isCancelled()) {
-                            entityExplodeEvent.blockList().forEach(b -> {
-                                breakBlock(e.getPlayer(), item, b, fortune, drops);
-                            });
+                        BlockExplodeEvent blockExplodeEvent = new BlockExplodeEvent(e.getBlock(), blocks, 0);
+                        Bukkit.getServer().getPluginManager().callEvent(blockExplodeEvent);
+                        if (!blockExplodeEvent.isCancelled()) {
+                            blockExplodeEvent.blockList().forEach(b -> breakBlock(e.getPlayer(), item, b, fortune, drops));
                         }
                     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
@@ -1,13 +1,16 @@
 package io.github.thebusybiscuit.slimefun4.implementation.items.tools;
 
+import java.util.ArrayList;
 import java.util.List;
 
+import org.bukkit.Bukkit;
 import org.bukkit.Effect;
 import org.bukkit.Material;
 import org.bukkit.Sound;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
 import org.bukkit.event.block.BlockBreakEvent;
+import org.bukkit.event.entity.EntityExplodeEvent;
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
```diff
@@ -63,11 +67,18 @@ public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List
...
                                     }
 
-                                    Block b = e.getBlock().getRelative(x, y, z);
-                                    breakBlock(e.getPlayer(), item, b, fortune, drops);
+                                    blocks.add(e.getBlock().getRelative(x, y, z));
                                 }
                             }
                         }
+
+                        EntityExplodeEvent entityExplodeEvent = new EntityExplodeEvent(e.getPlayer(), e.getBlock().getLocation(), blocks, 0);
```
WalshyDev, line 75: Not a fan of EntityExplodeEvent, this should b BlockExplodeEvent with the actual broken block.

This is gonna lead to weird behaviour by plugins.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
```diff
@@ -63,11 +67,18 @@ public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List
...
+                        Bukkit.getServer().getPluginManager().callEvent(entityExplodeEvent);
+                        if (!entityExplodeEvent.isCancelled()) {
+                            entityExplodeEvent.blockList().forEach(b -> {
```
WalshyDev, line 78: Shouldn't use curly brace for lambda with one line.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
FabioZumbi12: Thats right, ill change and test.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosivePickaxe.java
FabioZumbi12: IDE completing! Ill change ;)

FabioZumbi12:
I have to do the same with ExplosiveShovel... testing now!

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosiveShovel.java
```diff
@@ -57,31 +61,38 @@ public boolean onBlockBreak(BlockBreakEvent e, ItemStack item, int fortune, List
...
+                        List<Block> blocks = new ArrayList<>();
                         for (int x = -1; x <= 1; x++) {
                             for (int y = -1; y <= 1; y++) {
                                 for (int z = -1; z <= 1; z++) {
                                     if (x == 0 && y == 0 && z == 0) {
                                         continue;
                                     }
 
-                                    Block b = e.getBlock().getRelative(x, y, z);
+                                   blocks.add(e.getBlock().getRelative(x, y, z));
```
WalshyDev, line 72: Missing space at start of line

TheBusyBiscuit:
Looking good. Some documentation and maybe a toggle for whether the event should be fired or not would be nice but approved it nonetheless.

TheBusyBiscuit:
Hi, just a heads up that I will be adding a toggle for this feature which will be **off** by default.
The addition of this has lead to a lot of unforeseen side effects, formally #1933 and tons of players complaining on our discord server about the explosive tools no longer working because they have build permissions but explosions are disabled in their region.

So yeah, while you and probably a lot of server owners will find use in this feature, I think it will be in the best interest of the majority of players to turn this into an opt-in feature as it seems to disrupt a lot of servers at this moment.

FabioZumbi12:
That's cools and you its the best guy to do whit settings/options than me, as i started to work with SF because this thing.

In fact the best way its to add this option, to allow people to choose the best behavior for your servers, but just to know the benefits (as i see for now on my server), fire epxlosion can previne an important bug related to drops:

- People have control over explosions on their regions, then they need to enable its a good thing;
- I am using the plugin FlyingCarpet (like MagicCarpet), and without explosion event, the blocks of carpet it not break, but drops;
- The same with protected regions. If the protected blocks are on range of the tool explosion, the blocks is not broke, but we can get their drops... infinitelly... Some server owners can do this test and ask for your player to make this test too... probably some players is using this as a bug to have infinite resources.

FabioZumbi12:
Sorry but I didn't give more details before, as it still doesn't know the seriousness of the bug.

To my surprise, I only found out after I fixed the bug, and the players started to complain that they could no longer duplicate the blocks.

The bug: Only happen with plugin not hooking with SF (like little plugins) and as example the FlyingCarpet:
I have this carpet with some tools:
image

When i use the explosive pickaxe near my carpet tools, i can get the tools drops (like obsidian) but the block still there (because listen for blockbreak event).
image

image

With the option you added it solves the explosion on regions with explosion disabled, but can cause BUG with other plugins.

Just let you (and other server owners to know) and se whats better for all.

PR #1946, opened 2020-06-01
remove metadata on entity death, and when items are removed from pede…

PhanaticD:
…stal. Also remove references to "item_placed" I do not see anywhere where this key is actually set, maybe leftover from old code?

I am currently testing this on my server, can wait a few days if you do not trust it for me to notice any harm

the metadata removal is necessary otherwise it will be a memory leak

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
@@ -129,8 +129,7 @@ public class AncientAltarTask implements Runnable {
 
             itemLock.remove(item);
             item.remove();
-
-            pedestal.removeMetadata("item_placed", SlimefunPlugin.instance);
+            item.removeMetadata("no_pickup", SlimefunPlugin.instance);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/AncientAltarTask.java
```diff
@@ -129,8 +129,7 @@ private void checkPedestal(Block pedestal) {
 
             itemLock.remove(item);
             item.remove();
-
-            pedestal.removeMetadata("item_placed", SlimefunPlugin.instance);
```
TheBusyBiscuit, line 133: Why was this removed?

PhanaticD: I could not find anywhere that this was actually added

TheBusyBiscuit: Hmm true. Weird...
Since there hasn't been any bug reports or complaints about the altars recently I will just assume that this is safe to remove, some probably added this at some point and removed it half-heartedly, leaving these things behind. Guess it wasn't that important then.

PR #1953, opened 2020-06-05
Added Industrial Miner & Removed Digital Miner

TheBusyBiscuit:
## Description

The Digital Miner has been deprecated for a very long time, now it's finally been removed.

This Pull Request proposes a new machine that functions similar but is much more performance-friendly. It is also much more balanced as it consumes fuel to run.
It features beautiful animations and particle effects.
When operating this machine, there will be a short warm up animation with the Pistons moving up and down.

Here's a video showcase:
https://youtu.be/1hvv6hHNCiI

## Changes

* Removed Digital Miner
* Added Industrial Miner
* Added Advanced Industrial Miner
* Also added Dried Kelp blocks to the Coal Generator

## Checklist


TheBusyBiscuit:
@WalshyDev I summon thou for you shall review.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/ActiveMiner.java
```diff
@@ -0,0 +1,362 @@
...
+
+    private int fuel = 0;
+    private int ores = 0;
+    private boolean running = false;
+
+    private final Block chest;
+    private final Block[] pistons;
+
+    private final Location start;
+    private final Location end;
```
WalshyDev, line 51: There's BlockPosition in CS-CoreLib2 which would be better for this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/ActiveMiner.java
```diff
@@ -0,0 +1,362 @@
...
+                    }
+                }
+
+                nextColumn();
+            });
+
+            queue.execute(SlimefunPlugin.instance);
+        }
+        catch (Exception e) {
+            Slimefun.getLogger().log(Level.SEVERE, "An Error occured while running an Industrial Miner", e);
```
WalshyDev, line 202: We should post location info like in cargo. Better yet, have a util method in ErrorCatcher to handle it for us.

Think we should always post coords if possible, server owners may need them depending on the issue.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/ActiveMiner.java
```diff
@@ -0,0 +1,362 @@
...
+                    stop("machines.INDUSTRIAL_MINER.piston-space");
+                }
+            }
+            else {
+                // The piston has been destroyed
+                stop("machines.INDUSTRIAL_MINER.destroyed");
+            }
+        }
+        catch (Exception e) {
+            Slimefun.getLogger().log(Level.SEVERE, "An Error occured while moving a Piston for an Industrial Miner", e);
```
WalshyDev, line 357: same as said above

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/IndustrialMiner.java
```diff
@@ -0,0 +1,213 @@
...
+ * @author TheBusyBiscuit
+ * 
+ * @see AdvancedIndustrialMiner
+ * @see ActiveMiner
+ *
+ */
+public class IndustrialMiner extends MultiBlockMachine {
+
+    protected final Map<Location, ActiveMiner> activeMiners = new HashMap<>();
+    protected final List<MachineFuel> fuelTypes = new ArrayList<>();
```
WalshyDev, line 46: does this need to be a list? If you need order a LinkedHashSet may be better.

TheBusyBiscuit: It does need to be consistently ordered, don't really see the benefits of using a LinkedHashSet though here, an ArrayList is just fine for this purpose.

PR #1963, opened 2020-06-10
Adding recipe for vanilla Gold Ingot -> Gold Dust conversion

siaynoq:
## Description
Adding a recipe to Electric Ingot Pulverizer for converting vanilla Gold Ingot to Gold Dust.

## Changes
Added `registerDefaultRecipes()` - most of the recipe addition for Smeltery & Pulverizer are handled in `PostSetup.loadSmelteryRecipes()` - so it made sense to create this method here as a "special case".

## Related Issues
No issues created for this - there was a separate conversation with TheBusyBiscuit in Slimefun-Wiki and I think this is a good addition to the Pulverizer anyway.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/ElectricIngotPulverizer.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/ElectricIngotPulverizer.java
@@ -56,7 +56,7 @@ public class ElectricIngotPulverizer extends AContainer implements RecipeDisplay
         // this is an extra recipe on top of PostSetup.loadSmelteryRecipes() for converting Vanilla Gold Ingot to Slimefun gold dust
         registerRecipe(3,
                 new ItemStack(Material.GOLD_INGOT),
-                new ItemStack(SlimefunItems.GOLD_DUST));
+                SlimefunItems.GOLD_DUST);
     }
 
     @Override
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/ElectricIngotPulverizer.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/ElectricIngotPulverizer.java
@@ -55,8 +55,8 @@ public class ElectricIngotPulverizer extends AContainer implements RecipeDisplay
     protected void registerDefaultRecipes() {
         // this is an extra recipe on top of PostSetup.loadSmelteryRecipes() for converting Vanilla Gold Ingot to Slimefun gold dust
         registerRecipe(3,
-                new ItemStack[] { new ItemStack(Material.GOLD_INGOT) },
-                new ItemStack[] { new ItemStack(SlimefunItems.GOLD_DUST) });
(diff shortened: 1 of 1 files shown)


siaynoq:
I couldn't find any unit test specifically targeting `ElectricIngotPulverizer` or recipes in general. If you want, I can put something together around and start the good habit - just let me know.

TheBusyBiscuit:
I think we should probably hold on a bit with per-item unit tests since I could really see a proper framework for this being quite useful.

TheBusyBiscuit:
Also what about the OreCrusher?
And just fyi: These recipes are *not* handled in PostSetup.loadSmelteryRecipes();

siaynoq:
> Also what about the OreCrusher?

I thought it makes more sense to add the conversion to an _ingot_ to the Pulverizer.

> And just fyi: These recipes are _not_ handled in PostSetup.loadSmelteryRecipes();

I was thinking about this part:

https://github.com/TheBusyBiscuit/Slimefun4/blob/919c86aebeb334442bf2641cfdf57e31af451462/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java#L218-L227

The "magic" happens in `if` branch, where the smeltery recipes are pushed into the Ingot Factory and Pulverizer (reversed). Pretty clever, but took me some time to find it the first place ;)

TheBusyBiscuit:
> >Also what about the OreCrusher?
>I thought it makes more sense to add the conversion to an ingot to the Pulverizer.

Right... the normal ore crusher doesn't even allow the crushing of ingots, forgot about that.

> > And just fyi: These recipes are _not_ handled in PostSetup.loadSmelteryRecipes();
> 
> I was thinking about this part:
> 
> https://github.com/TheBusyBiscuit/Slimefun4/blob/919c86aebeb334442bf2641cfdf57e31af451462/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java#L218-L227
> 
> The "magic" happens in `if` branch, where the smeltery recipes are pushed into the Ingot Factory and Pulverizer (reversed). Pretty clever, but took me some time to find it the first place ;)

Oh, you're right. I was very sure that there was a seperate part for it, although thinking about it, it does make sense to have it there since it's the reverse of the above.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/ElectricIngotPulverizer.java
```diff
@@ -50,6 +51,14 @@ public ItemStack getProgressBar() {
         return displayRecipes;
     }
 
+    @Override
+    protected void registerDefaultRecipes() {
+        // this is an extra recipe on top of PostSetup.loadSmelteryRecipes() for converting Vanilla Gold Ingot to Slimefun gold dust
+        registerRecipe(3,
+                new ItemStack[] { new ItemStack(Material.GOLD_INGOT) },
+                new ItemStack[] { new ItemStack(SlimefunItems.GOLD_DUST) });
```
TheBusyBiscuit, line 59: No need to do `new ItemStack()` around the Slimefun Item, it would just create a wasteful and object there.
Other than that, good for merging!

PR #1981, opened 2020-06-12
allow null meta

PhanaticD:
## Description

allow null meta for the itemstackwrapper, since hasitemmeta is checked everywhere first anyways, creating meta for every item that the cargo net checks is pretty expensive
## Checklist

- [x ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [ x] I followed the existing code standards and didn't mess up the formatting.
- [ x] I did my best to add documentation to any public classes or methods I added.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
@@ -45,6 +45,9 @@ public class CustomItemDataService implements PersistentDataService, Keyed {
     }
 
     public Optional<String> getItemData(ItemStack item) {
+        if (!item.hasItemMeta()) {
+            return Optional.empty();
+        }
         ItemMeta meta = item.getItemMeta();
         if (meta == null) {
             return Optional.empty();
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
@@ -45,6 +45,9 @@ public class CustomItemDataService implements PersistentDataService, Keyed {
     }
 
     public Optional<String> getItemData(ItemStack item) {
+        if (item.getItemMeta() == null) {
+            return Optional.empty();
+        }
         return getItemData(item.getItemMeta());
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
@@ -45,10 +45,11 @@ public class CustomItemDataService implements PersistentDataService, Keyed {
     }
 
     public Optional<String> getItemData(ItemStack item) {
-        if (item.getItemMeta() == null) {
+        ItemMeta meta = item.getItemMeta();
+        if (meta == null) {
             return Optional.empty();
         }
-        return getItemData(item.getItemMeta());
+        return getItemData(meta);
     }
 
     public Optional<String> getItemData(ItemMeta meta) {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
@@ -45,14 +45,10 @@ public class CustomItemDataService implements PersistentDataService, Keyed {
     }
 
     public Optional<String> getItemData(ItemStack item) {
-        if (!item.hasItemMeta()) {
+        if (item == null || !item.hasItemMeta()) {
             return Optional.empty();
         }
-        ItemMeta meta = item.getItemMeta();
-        if (meta == null) {
-            return Optional.empty();
-        }
-        return getItemData(meta);
+        return getItemData(item.getItemMeta());
     }
 
     public Optional<String> getItemData(ItemMeta meta) {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/itemstack/ItemStackWrapper.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/itemstack/ItemStackWrapper.java
@@ -45,7 +45,11 @@ public final class ItemStackWrapper extends ItemStack {
         // Since this class is immutable, we can simply let the super class create one copy
         // and then store that instead of creating a clone everytime.
         // This will significantly speed up any loop comparisons if used correctly.
-        return meta;
+        if (meta == null) {
+            throw new UnsupportedOperationException("#hasItemMeta() must be checked prior to this call");
+        } else {
+            return meta;
+        }
     }
 
     @Override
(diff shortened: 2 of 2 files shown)


WalshyDev:
Have you checked the usages of this? Do any assume not-null?

Also, should add a @Nullable tag
IIRC SF has findbugs in it

TheBusyBiscuit:
> IIRC SF has findbugs in it

Nope. It's a longtime todo but when we introduce this, we need to introduce that everywhere not just at this one specific place.

PhanaticD:
does that look good?

PhanaticD:
I am not sure what should be returned there, I am not that familiar with slimefun workings

TheBusyBiscuit:
It expects an Optional to be returned.
https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html

So in the event that the ItemStack does not have any ItemMeta, an empty Optional should just be returned there. Nothing crazy nor Slimefun-specific here, just a simple if/else addition which will prevent any issues. Could also add a null-check for the actual ItemStack there too, if you wanna be extra-safe.

But either way, the method just redirects it to the one with an ItemMeta, the result is an Optional and if the item or item meta does not exist, the Optional should just be empty here.

PhanaticD:
there

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
```diff
@@ -45,6 +45,9 @@ public void setItemData(ItemMeta im, String id) {
     }
 
     public Optional<String> getItemData(ItemStack item) {
+        if (item.getItemMeta() == null) {
```
TheBusyBiscuit, line 48: You should do .hasItemMeta() here (and maybe also add a null-check for the given ItemStack parameter as pointed out before), otherwise you won't gain any performance benefit from this here, since #getItemMeta() (for normal ItemStacks) will create a new instance of ItemMeta and perform a clone operation which is very costly.

WalshyDev, line 48: You should check hasItemMeta, just like your exception in getItemMeta says ;)

PhanaticD:
':D

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
```diff
@@ -45,7 +45,14 @@ public void setItemData(ItemMeta im, String id) {
...
 
     public Optional<String> getItemData(ItemStack item) {
-        return getItemData(item.getItemMeta());
+        if (!item.hasItemMeta()) {
+            return Optional.empty();
+        }
+        ItemMeta meta = item.getItemMeta();
+        if (meta == null) {
+            return Optional.empty();
+        }
```
TheBusyBiscuit, line 54: This is redundant, ItemStack#getItemMeta() will **never** return null, except for ItemStackWrappers but we already handled these in the previous condition. The null-check I suggested was about the ItemStack, not the meta.
`if (item == null || !item.hasItemMeta())`

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
```diff
@@ -45,7 +45,14 @@ public void setItemData(ItemMeta im, String id) {
     }
 
     public Optional<String> getItemData(ItemStack item) {
-        return getItemData(item.getItemMeta());
+        if (!item.hasItemMeta()) {
+            return Optional.empty();
+        }
+        ItemMeta meta = item.getItemMeta();
+        if (meta == null) {
```
WalshyDev, line 52: This isn't ever null

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
PhanaticD: why would the item be null

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
PhanaticD: it could be for wrappers

WalshyDev: No in the wrapper you throw an exception now

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
TheBusyBiscuit: If I do `.getItemData(null)` the ItemStack will be null.

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
TheBusyBiscuit: In that case this code won't even be reached though since we prevent this from ever being called due to the #hasItemMeta() check from before.

PhanaticD: busybiscuit told me to explicitly add a check here, that is the only reason I put it

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
PhanaticD: it did not have a null check for however long this class has existed before, that seems out of the scope of this PR to change

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
TheBusyBiscuit: You only seem to read half of our comments then, I already told you this is redundant and that it isn't nullable. Only the actual ItemStack is.

PhanaticD: I am reading all your comments, you are just confusing, edits are enabled for maintainers, please change the PR to be as you see fit

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
TheBusyBiscuit: Something being broken for a long time doesn't mean you shouldn't fix it.
And it is very well in scope of this PR since your change broke this method. It's only logical to fully fix this then, rather than doing a half-hearted job and fixing it only halfway through, otherwise the issue will only be delayed to a more distant point in time.

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/CustomItemDataService.java
TheBusyBiscuit: Tell us what's unclear then. There is nothing that can't be discussed or cleared up.

PhanaticD:
is this good now?

Issue #178, opened 2017-03-05
Can you add an Energy Transporter?


gabe56f:
Please add one!

PR #2014, opened 2020-06-23
Added Magic Pills.

LinoxGH:
## Description

Added Magic Pills that cures zombie villagers.

## Changes

Added a new magical gadget called Magical Pills.
A new recipe for Magic Workbench.
From approved suggestion: #178

## Related Issues

N/A

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicPills.java
```diff
@@ -0,0 +1,64 @@
...
+                zombieVillager.setConversionTime(1);
+                if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
+                    zombieVillager.setConversionPlayer(p);
+                }
+
+                p.playSound(p.getLocation(), Sound.ENTITY_ZOMBIE_VILLAGER_CONVERTED, 1, 1);
+
+                int amount = item.getAmount();
+                amount -= 1;
+                item.setAmount(amount);
```
WalshyDev, line 52: Why is this not just getAmount() - 1...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicPills.java
```diff
@@ -0,0 +1,64 @@
...
+
+                if (e.getHand() == EquipmentSlot.OFF_HAND) {
```
WalshyDev, line 54: Not a big fan of the order here. I would prefer the other way around but I'm picky

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/PlayerInteractEntityListener.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class PlayerInteractEntityListener implements Listener {
+
+    public PlayerInteractEntityListener(SlimefunPlugin plugin) {
+        plugin.getServer().getPluginManager().registerEvents(this, plugin);
+    }
+
+    @EventHandler
+    public void onInteractEntity(PlayerInteractAtEntityEvent e) {
+
```
WalshyDev, line 32: Empty line

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -1,632 +1,634 @@
-package me.mrCookieSlime.Slimefun;
```
WalshyDev, line 1: Please don't reformat the entire class

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicPills.java
LinoxGH: Wdym other way around? Amount Reduction - Sound - Curing?

LinoxGH: I didn't want to reduce item amount before curing just in case if the item gets used for nothing

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
LinoxGH: Actually I only deleted the * import and added single class imports but the github detected it as the whole file is changed.

PR #2019, opened 2020-06-24
Fortune cookies

Neon72:
a few more quotes for fortune cookies

## Description
Added a few quotes to fortune cookies to pull from

## Changes
6 new lines of dialogue for Fortune cookies

## Related Issues
None

## Checklist

### DIFF
--- a/src/main/resources/languages/messages_en.yml
+++ b/src/main/resources/languages/messages_en.yml
@@ -156,7 +156,7 @@ messages:
   - '&7A Walshy a day will keep the troubles away.'
   - '&7Never dig straight down!'
   - '&7Tis but a flesh wound!'
-  - '&7I'm &4NOT &7the Messiah!'
+  - '&7I am &4NOT &7the Messiah!'
   - '&7Always look on the bright side of life!'
   - '&7African or European Swallow?'
   - '&7This one was actually a Biscuit and not a Cookie'
--- a/src/main/resources/languages/messages_en.yml
+++ b/src/main/resources/languages/messages_en.yml
@@ -155,6 +155,12 @@ messages:
   - '&742. The answer is 42.'
   - '&7A Walshy a day will keep the troubles away.'
   - '&7Never dig straight down!'
+  - '&7Tis but a flesh wound!'
+  - '&7I'm &4NOT &7the Messiah!'
+  - '&7Always look on the bright side of life!'
+  - '&7African or European Swallow?'
+  - '&7This one was actually a Biscuit and not a Cookie'
+  - '&7Neon signs are LIT!'
 
 machines:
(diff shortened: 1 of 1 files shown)


src/main/resources/languages/messages_en.yml
```diff
@@ -155,6 +155,12 @@ messages:
   - '&742. The answer is 42.'
   - '&7A Walshy a day will keep the troubles away.'
   - '&7Never dig straight down!'
+  - '&7Tis but a flesh wound!'
+  - '&7I'm &4NOT &7the Messiah!'
```
WalshyDev, line 159: You need to "escape" the single quote. So, do `''` (two back to back)

WalshyDev:
Cookie, we need to run this through a YAML validator if a YAML file is edited.

TheBusyBiscuit:
@WalshyDev I agree, workflow file or unit test?

WalshyDev:
Definitely workflow

TheBusyBiscuit:
Why definitely?

WalshyDev:
Well, unit test we'd need to recursively check the resources and stuff. For a workflow we can just see if a changed file is YAML and validate that. It's a lot lighter, makes the checks fail in a more verbose way and also would validate issues in other workflow files.

TheBusyBiscuit:
Fair, workflow validation would also be cool.
Feel free to dig one up and make a PR, would gladly add that then.

src/main/resources/languages/messages_en.yml
```diff
@@ -155,6 +155,12 @@ messages:
   - '&742. The answer is 42.'
   - '&7A Walshy a day will keep the troubles away.'
   - '&7Never dig straight down!'
+  - '&7Tis but a flesh wound!'
+  - '&7I am &4NOT &7the Messiah!'
+  - '&7Always look on the bright side of life!'
+  - '&7African or European Swallow?'
```
TheBusyBiscuit, line 161: I would probably not include this one.

src/main/resources/languages/messages_en.yml
```diff
@@ -155,6 +155,12 @@ messages:
   - '&742. The answer is 42.'
   - '&7A Walshy a day will keep the troubles away.'
   - '&7Never dig straight down!'
+  - '&7Tis but a flesh wound!'
+  - '&7I am &4NOT &7the Messiah!'
```
TheBusyBiscuit, line 159: Also wouldn't exactly include this either.

src/main/resources/languages/messages_en.yml
Neon72: you marked all 4 which one out of the 4 you mean?

TheBusyBiscuit: I only marked one, the last one. All above is only automatically added context by GitHub.

Neon72: I figured 5 seconds after

PR #2020, opened 2020-06-24
Auto brewer

LinoxGH:
## Description

Added the Auto Brewer from an old approved suggestion of TheBusyBiscuit.

## Changes

A new electric machine and corresponding research.

## Related Issues

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/AutoBrewer.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/AutoBrewer.java
@@ -116,8 +116,8 @@ public class AutoBrewer extends AContainer {
 
         if (input1 == null || input2 == null) return null;
 
-        if (input1.getType().name().endsWith("POTION") || input2.getType().name().endsWith("POTION")) {
-            boolean slot = input1.getType().name().endsWith("POTION");
+        if (isPotion(input1.getType()) || isPotion(input2.getType())) {
+            boolean slot = isPotion(input1.getType());
             ItemStack pItem = slot ? input1 : input2;
             ItemStack iItem = slot ? input2 : input1;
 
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/AutoBrewer.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    private MachineRecipe findRecipe(BlockMenu menu) {
+        ItemStack input1 = menu.getItemInSlot(getInputSlots()[0]);
+        ItemStack input2 = menu.getItemInSlot(getInputSlots()[1]);
+
+        if (input1 == null || input2 == null) return null;
+
+        if (input1.getType().name().endsWith("POTION") || input2.getType().name().endsWith("POTION")) {
+            boolean slot = input1.getType().name().endsWith("POTION");
```
TheBusyBiscuit, line 120: You should probably create an "isPotion(ItemStack)" method for this instead of doing this multiple times, also

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/AutoBrewer.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    @Override
+    public String getInventoryTitle() {
+        return "&6Auto-Brewer";
+    }
+
+    @Override
+    public ItemStack getProgressBar() {
+        return new ItemStack(Material.CARROT_ON_A_STICK);
```
TheBusyBiscuit, line 179: Why a carrot on a stick?

LinoxGH: because I thought it looks similar to brewing stand. any better idea?

TheBusyBiscuit: Nah, not really :/

LinoxGH: if you don't have a better idea why not keep it?

PR #2027, opened 2020-06-25
 Improve the handling of non-registered worlds

SoSeDiK:
## Description & 

1. Some plugins may create fake worlds. For example, AWE can do that after the first usage of `//regen`. Those ghost worlds aren't handled by Slimefun (they do not exist during startup and world load event is not fired for them too), so there was an error when the plugin was trying to save them.
2. There's a minor chance for NPE when saving block data. See related issues.
3. Fixed a tiny typo that was annoying me.

## Changes

1. Slimefun now tries to save data only for registered worlds.
2. Slimefun now handles possible NPEs. There will be a warning instead.
3. `occured` -> `occurred`.

## Related Issues

See https://github.com/TheBusyBiscuit/ExoticGarden/issues/123 as an example of point 2 above.

## Checklist
*Wow, there are more checkboxes now! I wasn't contributing to Slimefun code-wise for quite a while :o*

- [x] I followed the existing code standards and didn't mess up the formatting. (*I hope so xD*)

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
@@ -1,11 +1,7 @@
 package me.mrCookieSlime.Slimefun;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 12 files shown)


TheBusyBiscuit:
One day I will now how to spell "occurred" :NotLikeThis:

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -391,21 +391,15 @@ public void onDisable() {
             }
         });
 
-        for (World world : Bukkit.getWorlds()) {
+        // Save all registered Worlds
+        getRegistry().getWorlds().forEach((world, storage) -> {
```
TheBusyBiscuit, line 395: Please don't use .forEach(...), it can have quite a negative impact on performance. It is advised to use the standard for loop format instead.

SoSeDiK: Normally it should be done only once when shutting down the server without any players already but I see the point. Agreed, should be resolved now.

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -1,11 +1,7 @@
 package me.mrCookieSlime.Slimefun;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
TheBusyBiscuit, line 4: Please do not use the star import.

SoSeDiK: Yeah. I'm trying to find a solution for that :P 
My IntelliJ IDEA does that automatically. I guess I'll have to set them manually, as well as add new imports for entries.

PR #2030, opened 2020-06-25
1.16 support

WalshyDev:
## Description
1.16 support is added.

## Changes
Not many changes required for this. Spigot have removed Bukkit from the repo now so switched the dep in the pom. Due to Bukkit being removed I have also improved the timing message to use a TextComponent since everyone will have that and it isn't a pain.

**Blocked by:** https://github.com/TheBusyBiscuit/CS-CoreLib2/pull/79 & https://github.com/TheBusyBiscuit/CS-CoreLib/pull/137

Need the hash of both for dependency.

## Related Issues
N/A

## Checklist


WalshyDev:
Having SF & CSLib really slows development down man. I have to host the shit on my repo or commit to my fork and then change the dep and change it all back once PRs get merged.

Move this to the org soon plssss, GH redirecrs anyway for a while.

PR #2045, opened 2020-06-25
Explosive Bow Area of Effect

LinoxGH:
## Description

Added Area of Effect functionality to Explosive Bow.

## Changes

Better knockback and damage calculation for Explosive Bow and area of effect.

## Related Issues

N/A

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/SlimefunBowListener.java
```diff
@@ -69,6 +70,7 @@ public void onArrowHit(ProjectileHitEvent e) {
     @EventHandler
     public void onArrowSuccessfulHit(EntityDamageByEntityEvent e) {
         if (e.getDamager() instanceof Arrow && e.getEntity() instanceof LivingEntity) {
+            if (e.getCause() == EntityDamageEvent.DamageCause.ENTITY_EXPLOSION) return;
```
TheBusyBiscuit, line 73: You should just negate it and include this in the if-statement above.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
+                        Bukkit.getPluginManager().callEvent(damageEvent);
+                    }
+                }
+            }
         };
     }
 
+    private double calculateDamage(double distance, double ogDamage) {
+
+        if (distance == 0D) return ogDamage;
```
TheBusyBiscuit, line 61: Never use == when comparing doubles, due to floating point errors, this is a very unreliable operation.
You should instead do sth like distance <= 0.05 or so.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
+                        entityL.damage(damage);
+                        EntityDamageByEntityEvent damageEvent = new EntityDamageByEntityEvent(e.getDamager(), entityL, EntityDamageEvent.DamageCause.ENTITY_EXPLOSION, damage);
+                        Bukkit.getPluginManager().callEvent(damageEvent);
+                    }
+                }
+            }
         };
     }
 
+    private double calculateDamage(double distance, double ogDamage) {
```
TheBusyBiscuit, line 59: The name "ogDamage" seems rather un-intuitive.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
+
+        if (distance == 0D) return ogDamage;
+        double damage = ogDamage - Math.pow((distance / range.getValue()), 2.5);
+        if (Math.round(damage) == 0D) damage += 1D;
```
TheBusyBiscuit, line 63: Why are you doing this?

Also seems quite wasteful to round it and then compare it with ==, should instead use a threshold.
Or since you seem to want to always have at least a damage of 1 you can just use Math.max(1, damage) instead.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
     public BowShootHandler onShoot() {
         return (e, n) -> {
-            Vector vector = n.getVelocity();
-            vector.setY(0.6);
-            n.setVelocity(vector);
-            n.getWorld().createExplosion(n.getLocation(), 0F);
-            n.getWorld().playSound(n.getLocation(), Sound.ENTITY_GENERIC_EXPLODE, 1F, 1F);
+            Collection<Entity> entites = n.getWorld().getNearbyEntities(n.getLocation(), range.getValue(), range.getValue(), range.getValue(), entity -> entity instanceof LivingEntity);
+            for (Entity entity : entites) {
+                if (entity.isValid() && entity instanceof LivingEntity) {
```
TheBusyBiscuit, line 34: This "instanceof LivingEntity" check is redundant, you are already doing this in the Predicate above.

TheBusyBiscuit, line 34: Also this .IsValid() method should be moved to the Predicate

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
+                    LivingEntity entityL = (LivingEntity) entity;
+
+                    double distance = entityL.getLocation().distance(n.getLocation());
```
TheBusyBiscuit, line 37: .distance() is a rather expensive operation and not really needed for this purpose, should rather use .distanceSquared() and then calculate the damage based on that.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
+                    double damage = calculateDamage(distance, e.getDamage());
+
+                    Vector distanceVector = entityL.getLocation().toVector().subtract(n.getLocation().toVector());
+                    distanceVector.setY(distanceVector.getY() + 0.6D);
+                    Vector entityVelocity = entityL.getVelocity();
+                    Vector knockback = entityVelocity.add(distanceVector.normalize().multiply((int) (e.getDamage() / Math.round(damage))));
```
TheBusyBiscuit, line 43: Why are you rounding the damage here?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
+                    entityL.setVelocity(knockback);
+
+                    entityL.getWorld().spawnParticle(Particle.EXPLOSION_LARGE, entityL.getLocation(), 1);
+                    entityL.getWorld().playSound(entityL.getLocation(), Sound.ENTITY_GENERIC_EXPLODE, 1F, 1F);
+
+                    if (!entityL.getUniqueId().equals(n.getUniqueId())) {
```
TheBusyBiscuit, line 49: This check should probably happen wayyyyyyyy earlier. Best would be in the Predicate itself.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
LinoxGH: Because a double value can be infinite, and from what I've heard doing equations with those cause problems(exceptions).

TheBusyBiscuit: You are already performing a min operation on that, so this being infinite is infinitely impossible.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,67 @@
...
+import io.github.thebusybiscuit.slimefun4.api.items.ItemSetting;
 import me.mrCookieSlime.Slimefun.Objects.Category;
 import me.mrCookieSlime.Slimefun.Objects.handlers.BowShootHandler;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
+import java.util.Collection;
+
 public class ExplosiveBow extends SlimefunBow {
 
+    private final ItemSetting<Integer> range = new ItemSetting<>("explosion-range", 3);
```
TheBusyBiscuit, line 22: Also should probably override the validation method for this, since we don't want negative or zero ranges.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
LinoxGH: I have no idea how can I use squared distance instead

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
LinoxGH: Nah it shouldn't because we want knockback and explosion effect for the main target too.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,70 @@
...
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
+import java.util.Collection;
+
 public class ExplosiveBow extends SlimefunBow {
 
+    private final ItemSetting<Integer> range = new ItemSetting<Integer>("explosion-range", 3) {
+        @Override
+        public boolean validateInput(Integer input) {
+            return input > 0;
```
WalshyDev, line 25: This is a boxed int, null check needed.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,70 @@
...
+                distanceVector.setY(distanceVector.getY() + 0.6D);
+                Vector entityVelocity = entityL.getVelocity();
+
+                double distanceSquared = distanceVector.lengthSquared();
+                double damage = calculateDamage(distanceSquared, e.getDamage());
+
+                Vector knockback = entityVelocity.add(distanceVector.normalize().multiply((int) (e.getDamage() / damage)));
+                entityL.setVelocity(knockback);
+
+                entityL.getWorld().spawnParticle(Particle.EXPLOSION_LARGE, entityL.getLocation(), 1);
```
WalshyDev, line 51: Explosion particle for every entity in the AOE? Surely this should just be in the actual centre.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,70 @@
...
+    };
+
     public ExplosiveBow(Category category, SlimefunItemStack item, ItemStack[] recipe) {
         super(category, item, recipe);
+        addItemSetting(range);
     }
 
     @Override
     public BowShootHandler onShoot() {
         return (e, n) -> {
```
WalshyDev, line 36: Can you rename this "n" variable. It's very nondescript.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,70 @@
...
+
+                Vector knockback = entityVelocity.add(distanceVector.normalize().multiply((int) (e.getDamage() / damage)));
+                entityL.setVelocity(knockback);
+
+                entityL.getWorld().spawnParticle(Particle.EXPLOSION_LARGE, entityL.getLocation(), 1);
+                entityL.getWorld().playSound(entityL.getLocation(), Sound.ENTITY_GENERIC_EXPLODE, 1F, 1F);
+
+                if (!entityL.getUniqueId().equals(n.getUniqueId())) {
+                    entityL.damage(damage);
+                    EntityDamageByEntityEvent damageEvent = new EntityDamageByEntityEvent(e.getDamager(), entityL, EntityDamageEvent.DamageCause.ENTITY_EXPLOSION, damage);
```
WalshyDev, line 56: This is a cancellable event. Also, like most events, this is a pre-damage event not post. So, move the damage down, make sure it isn't cancelled and then make sure the damage is actually correct.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -1,28 +1,70 @@
...
+                    Bukkit.getPluginManager().callEvent(damageEvent);
+                }
+            }
         };
     }
 
+    private double calculateDamage(double distanceSquared, double originalDamage) {
+
```
WalshyDev, line 64: Empty line.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
LinoxGH: I'm not sure I understand this fully. Can you elaborate a little?

WalshyDev: If a plugin wants to change the damage or even cancel then this is what should happen. Right now you aren't letting this happen.

You need to fire event, check if cancelled, if not, then damage(e.getDamage());

PR #2057, opened 2020-06-26
Added Bee Protection to Hazmat Suit.

LinoxGH:
## Description

Now bees cannot harm you while wearing the full hazmat suit.

## Changes

Added BeeListener.
Added CustomProtection ItemAttribute.
Added ProtectionType enum.
Added HazmatArmorPiece.
Improved ArmorTask.
Now full hazmat suit prevents from bee damage.

## Related Issues

N/A

## Checklist


J3fftw1:
hazmat suit = bee suit?

LinoxGH:
> hazmat suit = bee suit?

Well the hazmat suit can protect you from lava, and you can breathe underwater with it. But somehow bee sting passes through? Didn't make much sense so I added this feature. Also this only happens when the full set is equipped which gives the players a reason to equip the whole set.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/HazmatSuitListener.java
```diff
@@ -0,0 +1,59 @@
...
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
+import me.mrCookieSlime.Slimefun.SlimefunPlugin;
+
+/**
+ * The listener for Hazmat Suit's {@link Bee} sting protection.
+ * Only applied if the whole set is worn.
+ *
+ * @author Linox
+ *
+ * @see Bee
```
TheBusyBiscuit, line 21: The `see` attribute shouldn't really be used for external classes, please only use it for Slimefun classes.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/HazmatSuitListener.java
```diff
@@ -0,0 +1,59 @@
...
+ *
+ */
+public class HazmatSuitListener implements Listener {
```
TheBusyBiscuit, line 24: Should probably rename this to BeeListener, since it only checks for Bees and has rather few to do with the actual mechanics of the HazmatSuit.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/HazmatSuitListener.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public HazmatSuitListener(SlimefunPlugin plugin) {
+        plugin.getServer().getPluginManager().registerEvents(this, plugin);
+    }
+
+    @EventHandler
+    public void onDamage(EntityDamageByEntityEvent e) {
+        if (!SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) return;
```
TheBusyBiscuit, line 32: Performance-wise this is a rather careless design choice. You should put this if-statement in front of initializing this Listener in the first place (in SlimefunPlugin.java), otherwise you are registering a completely unnecessary Listener.

And let's be honest... the Minecraft version won't change from one minute to another.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/HazmatSuitListener.java
```diff
@@ -0,0 +1,59 @@
...
+                for (ItemStack armor : p.getInventory().getArmorContents()) {
+                    SlimefunItem sfItem = SlimefunItem.getByItem(armor);
+                    if (sfItem == null) return;
+
+                    String id = sfItem.getID();
+                    if (id.equals("SCUBA_HELMET")) hazmatCount++;
+                    if (id.equals("HAZMAT_CHESTPLATE")) hazmatCount++;
+                    if (id.equals("HAZMAT_LEGGINGS")) hazmatCount++;
+                    if (id.equals("RUBBER_BOOTS")) hazmatCount++;
+                }
```
TheBusyBiscuit, line 48: This is a really expensive check and should not be done this way.
Check the ArmorTask class, there are already optimizations for the Hazmat Suit in place.

src/main/java/me/mrCookieSlime/Slimefun/SlimefunPlugin.java
```diff
@@ -237,6 +238,7 @@ else if (getServer().getPluginManager().isPluginEnabled("CS-CoreLib")) {
             new WitherListener(this);
             new IronGolemListener(this);
             new PlayerInteractEntityListener(this);
+            new HazmatSuitListener(this);
```
TheBusyBiscuit, line 241: This should be renamed and gated behind an if statement (see my review comments on this Listener class)

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
```diff
@@ -245,4 +246,20 @@ else if (PatternUtils.ALPHANUMERIC.matcher(texture).matches()) {
...
 
+    /**
+     * Adds additional lores for the {@link ItemStack}.
+     *
+     * @param newLore
+     *                New lores to be added to the {@link ItemStack}
+     */
+    public void addLore(String... newLore) {
+        ItemMeta meta = getItemMeta();
+        if (meta == null) return;
```
TheBusyBiscuit, line 257: ItemMeta cannot be null.

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
```diff
@@ -245,4 +246,20 @@ else if (PatternUtils.ALPHANUMERIC.matcher(texture).matches()) {
...
     }
 
+    /**
+     * Adds additional lores for the {@link ItemStack}.
+     *
+     * @param newLore
+     *                New lores to be added to the {@link ItemStack}
+     */
+    public void addLore(String... newLore) {
+        ItemMeta meta = getItemMeta();
```
TheBusyBiscuit, line 256: You should check .hasItemMeta() before this, otherwise you are computing wasteful objects.

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
```diff
@@ -245,4 +246,20 @@ else if (PatternUtils.ALPHANUMERIC.matcher(texture).matches()) {
...
+        if (meta == null) return;
+        List<String> oldLore = meta.getLore();
+        if (oldLore == null) return;
```
TheBusyBiscuit, line 259: This is not good. Use .hasLore() instead.

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
```diff
@@ -245,4 +246,20 @@ else if (PatternUtils.ALPHANUMERIC.matcher(texture).matches()) {
         }
     }
 
+    /**
+     * Adds additional lores for the {@link ItemStack}.
+     *
+     * @param newLore
+     *                New lores to be added to the {@link ItemStack}
+     */
+    public void addLore(String... newLore) {
```
TheBusyBiscuit, line 255: I don't really think this method is useful here, I'd argue this should be removed and just done manually or via the Consumer in the constructor.

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
LinoxGH: I'm almost sure it could be in 1.13. Or it's my bad memory

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,47 @@
...
+    public void onDamage(EntityDamageByEntityEvent e) {
+        if (e.getDamager() instanceof Bee) {
+            if (e.getEntity() instanceof Player) {
+                Player p = (Player) e.getEntity();
+
+                // Check for a Hazmat Suit
+                if (!SlimefunUtils.isItemSimilar(SlimefunItems.SCUBA_HELMET, p.getInventory().getHelmet(), true) &&
+                        !SlimefunUtils.isItemSimilar(SlimefunItems.HAZMAT_CHESTPLATE, p.getInventory().getChestplate(), true) &&
+                        !SlimefunUtils.isItemSimilar(SlimefunItems.HAZMAT_LEGGINGS, p.getInventory().getLeggings(), true) &&
+                        !SlimefunUtils.isItemSimilar(SlimefunItems.RUBBER_BOOTS, p.getInventory().getBoots(), true)) {
```
TheBusyBiscuit, line 38: This statement is the wrong way around, you are checking for the opposite, so remove the negations here.
Also the armor item and the SlimefunItem need to be swapped too.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,47 @@
...
+                    e.setDamage(0D);
```
TheBusyBiscuit, line 39: Why are you doing this instead of cancelling the event?

src/main/java/me/mrCookieSlime/Slimefun/api/SlimefunItemStack.java
```diff
@@ -2,6 +2,7 @@
 
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.Arrays;
```
TheBusyBiscuit, line 5: Random unused import

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
LinoxGH: Right I didn't notice the !s there it's a copy-paste lol

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
LinoxGH: When you cancel the event bee just keeps trying. And doesn't die

LinoxGH:
I've also did some performance improvements to ArmorTask.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
```diff
@@ -119,9 +120,26 @@ private boolean hasSunlight(Player p) {
         return (world.getTime() < 12300 || world.getTime() > 23850) && p.getEyeLocation().getBlock().getLightFromSky() == 15;
     }
 
-    private void checkForRadiation(Player p) {
+    private void checkForRadiation(PlayerProfile profile) {
+        HashedArmorpiece[] armor = profile.getArmor();
+        Player p = profile.getPlayer();
```
TheBusyBiscuit, line 125: Just pass the Player as an argument, this will perform a lengthy UUID lookup.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
```diff
@@ -119,9 +120,26 @@ private boolean hasSunlight(Player p) {
...
+        for (HashedArmorpiece armorPiece : armor) {
+
+            Optional<SlimefunArmorPiece> sfArmor = armorPiece.getItem();
+            if (!sfArmor.isPresent()) continue;
+
+            if (sfArmor.get().getID().equals("SCUBA_HELMET") ||
+                    sfArmor.get().getID().equals("HAZMAT_CHESTPLATE") ||
+                    sfArmor.get().getID().equals("HAZMAT_LEGGINGS") ||
+                    sfArmor.get().getID().equals("RUBBER_BOOTS")) {
+                hasHazmat = true;
```
TheBusyBiscuit, line 138: I very much like the transition of the Hazmat suit to the SlimefunArmorPiece system!
But I am not the greatest fan of id-comparison. 
Best would be to create a new `ItemAttribute`, like "HazmatProtection" and then create a HazmatSuit class which extends SlimefunArmorpiece and implements that attribute.
This way we don't have to rely on hardcoded ids and are flexible in terms of adding more hazmat suits in the future.

TheBusyBiscuit, line 138: Also since you are only checking for ANY hazmat piece, you can break the loop here.
Or even better, make a seperate method with a boolean return type for this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,68 @@
...
+        for (HashedArmorpiece armor : armors) {
+            Optional<SlimefunArmorPiece> armorPiece = armor.getItem();
+            if (!armorPiece.isPresent()) return false;
+
+            if (armorPiece.get().getID().equals("SCUBA_HELMET") ||
+                    armorPiece.get().getID().equals("HAZMAT_CHESTPLATE") ||
+                    armorPiece.get().getID().equals("HAZMAT_LEGGINGS") ||
+                    armorPiece.get().getID().equals("RUBBER_BOOTS")) {
+                hazmatCount++;
+            }
```
TheBusyBiscuit, line 63: Same as below, in regards to the hardcoded ids.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,68 @@
...
+            if (e.getEntity() instanceof Player) {
+                Player p = (Player) e.getEntity();
+                PlayerProfile.get(p, profile -> {
+
+                    HashedArmorpiece[] armors = profile.getArmor();
+                    if (hasFullHazmat(armors)) {
+                        for (ItemStack armor : p.getInventory().getArmorContents()) {
+                            ItemUtils.damageItem(armor, 1, false);
+                        }
+                        e.setDamage(0D);
```
WalshyDev, line 43: This is in an async operation, this will happen AFTER the event has fired and therefore have no change.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,68 @@
...
+public class BeeListener implements Listener {
+
+    public BeeListener(SlimefunPlugin plugin) {
+        plugin.getServer().getPluginManager().registerEvents(this, plugin);
+    }
+
+    @EventHandler
+    public void onDamage(EntityDamageByEntityEvent e) {
+        if (e.getDamager() instanceof Bee) {
+            if (e.getEntity() instanceof Player) {
```
WalshyDev, line 34: Useless nested if. &&

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,68 @@
...
+                PlayerProfile.get(p, profile -> {
+
+                    HashedArmorpiece[] armors = profile.getArmor();
+                    if (hasFullHazmat(armors)) {
+                        for (ItemStack armor : p.getInventory().getArmorContents()) {
+                            ItemUtils.damageItem(armor, 1, false);
+                        }
+                        e.setDamage(0D);
+                    }
+                });
```
TheBusyBiscuit, line 45: This is an asynchronous operation, so you should NOT do `e.setDamage()` here, as it will have no effect when this is delayed.
You need to the following instead:

```java
Optional<PlayerProfile> optional = PlayerProfile.find(p);
if (!optional .isPresent()) {
  PlayerProfile.request(p);
  return;
}

PlayerProfile profile = optional.get();
// Do stuff...
```

This will ensure that we don't access the PlayerProfile asynchronously, which will lead to problems.
And only execute the code if it's synced.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
WalshyDev, line 138: This should only be true if all items are hazmat gear. Right now, this is not the case.

LinoxGH: It never worked like that though. Only 1 piece always protected you from radiation. Idk if it's a bug but players could just do half slime armor and half hazmat suit for:
 + Water Breathing
 + Fire Resistance
 + Radiation Resistance
 + Jump Boost
 + Speed Boost

Even I use that set when I play.

TheBusyBiscuit: This is the radiation protection check though. And for that you ALWAYS needed the full set, as seen here, before you changed this behaviour:

https://github.com/TheBusyBiscuit/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java#L124

If any piece is missing, the radiation check continues.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/CustomProtection.java
```diff
@@ -0,0 +1,48 @@
...
+     * @return The {@link ProtectionType}s
+     */
+    boolean requireFullSet();
+
+    enum ProtectionType {
+
+        RADIATION,
+
+        BEES;
+    }
```
TheBusyBiscuit, line 47: Please don't declare an enum inside the interface. Make a seperate file for this...

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/CustomProtection.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    ProtectionType[] getProtectionTypes();
+
+    /**
+     * This returns the {@link ProtectionType}s this {@link ItemAttribute}
+     * prevents the assigned {@link SlimefunArmorPiece} to be damaged by.
+     *
+     * @return The {@link ProtectionType}s
+     */
+    boolean requireFullSet();
```
TheBusyBiscuit, line 40: This is a bit misleading, should probably be "isFullSetRequired()".

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/SlimefunArmorPiece.java
```diff
@@ -29,4 +36,12 @@ public SlimefunArmorPiece(Category category, SlimefunItemStack item, RecipeType
...
     }
 
+    /**
+     * This returns the armor set ID of this {@link SlimefunArmorPiece}.
+     *
+     * @return The set ID, <code>null</code> if no set ID is found.
+     */
+    public String getSetID() {
+        return setID;
+    }
```
TheBusyBiscuit, line 46: Please use a NamespacedKey for this rather than a String, otherwise you could run into ID conflicts.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/SlimefunArmorPiece.java
```diff
@@ -29,4 +36,12 @@ public SlimefunArmorPiece(Category category, SlimefunItemStack item, RecipeType
         return effects;
     }
 
+    /**
+     * This returns the armor set ID of this {@link SlimefunArmorPiece}.
+     *
+     * @return The set ID, <code>null</code> if no set ID is found.
+     */
+    public String getSetID() {
```
TheBusyBiscuit, line 44: Also: "Id", not "ID".

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,85 @@
...
+            if (shouldProtect(armors)) {
+                for (ItemStack armor : p.getInventory().getArmorContents()) {
+                    ItemUtils.damageItem(armor, 1, false);
+                }
+                e.setDamage(0D);
+            }
+        }
+    }
+
+    private boolean shouldProtect(HashedArmorpiece[] armors) {
```
TheBusyBiscuit, line 54: "isProtected" sounds more appropriate.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
```diff
@@ -119,18 +121,19 @@ private boolean hasSunlight(Player p) {
...
-                if (isRadioactive(p, item)) {
+                if (checkAndApplyRadioactive(p, item)) {
                     break;
                 }
             }
         }
     }
 
-    private boolean isRadioactive(Player p, ItemStack item) {
+    private boolean checkAndApplyRadioactive(Player p, ItemStack item) {
```
TheBusyBiscuit, line 136: *Radiation, not Radioactive.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
```diff
@@ -148,4 +151,35 @@ private boolean isRadioactive(Player p, ItemStack item) {
         return false;
     }
 
+    private boolean shouldProtect(HashedArmorpiece[] armors) {
```
TheBusyBiscuit, line 154: "isProtected" sounds better here.

Also since you are reusing this method twice... It may make sense to make a more general method for this inside PlayerProfile.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
```diff
@@ -148,4 +151,35 @@ private boolean isRadioactive(Player p, ItemStack item) {
...
+        String setID = null;
+        for (HashedArmorpiece armor : armors) {
+            Optional<SlimefunArmorPiece> armorPiece = armor.getItem();
+            if (!armorPiece.isPresent()) return false;
+
+            if (armorPiece.get() instanceof CustomProtection) {
+                CustomProtection protectedArmor = (CustomProtection) armorPiece.get();
+
+                if (first) {
+                    if (protectedArmor.requireFullSet()) setID = armorPiece.get().getSetID();
```
TheBusyBiscuit, line 167: Is the id even required in the first place?
I think it would make more sense to have a "ProtectionType" argument for this method and check if all 4 protect against this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
LinoxGH: thats how it was

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/ArmorTask.java
LinoxGH: I wanted to allow people to create sets that only required 1 part for the protection

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
```diff
@@ -0,0 +1,8 @@
+package io.github.thebusybiscuit.slimefun4.core.attributes;
+
+public enum ProtectionType {
+
+    RADIATION,
+    BEES;
+
+}
```
TheBusyBiscuit, line 8: Some documentation here would be nice.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/HazmatArmorPiece.java
```diff
@@ -0,0 +1,31 @@
...
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+
+public class HazmatArmorPiece extends SlimefunArmorPiece implements CustomProtection {
+
+    private final ProtectionType[] types;
+
+    public HazmatArmorPiece(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, PotionEffect[] effects, String setID) {
+        super(category, item, recipeType, recipe, effects, setID);
```
TheBusyBiscuit, line 17: Is the setID here really required? shouldn't it be `slimefun:hazmat` by default for this class?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/HazmatArmorPiece.java
```diff
@@ -0,0 +1,31 @@
...
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.potion.PotionEffect;
+
+import io.github.thebusybiscuit.slimefun4.core.attributes.CustomProtection;
+import io.github.thebusybiscuit.slimefun4.core.attributes.ProtectionType;
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+
+public class HazmatArmorPiece extends SlimefunArmorPiece implements CustomProtection {
```
TheBusyBiscuit, line 12: Missing documentation.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/SlimefunArmorPiece.java
```diff
@@ -19,6 +22,12 @@ public SlimefunArmorPiece(Category category, SlimefunItemStack item, RecipeType
         this.effects = effects == null ? new PotionEffect[0] : effects;
     }
 
+    public SlimefunArmorPiece(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, PotionEffect[] effects, String setId) {
```
TheBusyBiscuit, line 25: Please use a NamespacedKey instead of a String for the setId...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/SlimefunArmorPiece.java
```diff
@@ -19,6 +22,12 @@ public SlimefunArmorPiece(Category category, SlimefunItemStack item, RecipeType
...
+        this(category, item, recipeType, recipe, effects);
+
+        this.id = new NamespacedKey(SlimefunPlugin.instance, setId);
```
TheBusyBiscuit, line 28: Using Slimefun's instance here defeats the entire purpose of a NamespacedKey...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/HazmatArmorPiece.java
LinoxGH: Thought an addon may extend this but mkay changing it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/SlimefunArmorPiece.java
```diff
@@ -29,4 +37,12 @@ public SlimefunArmorPiece(Category category, SlimefunItemStack item, RecipeType
...
     }
 
+    /**
+     * This returns the armor set {@link NamespacedKey} of this {@link SlimefunArmorPiece}.
+     *
+     * @return The set {@link NamespacedKey}, <code>null</code> if none is found.
+     */
+    public NamespacedKey getSetId() {
+        return id;
+    }
```
TheBusyBiscuit, line 47: This method should probably be moved to your `CustomProtection` interface and then implemented in the Hazmat class.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/SlimefunArmorPiece.java
```diff
@@ -1,16 +1,19 @@
...
 
+import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import me.mrCookieSlime.Slimefun.Lists.RecipeType;
 import me.mrCookieSlime.Slimefun.Objects.Category;
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
 public class SlimefunArmorPiece extends SlimefunItem {
 
+    private NamespacedKey id = null;
```
TheBusyBiscuit, line 16: This being null here isn't really the best ideas.
I mean, we only use this key for the CustomProtection armor, so it might make sense to just have the getter in that interface and force the developer to implement this field in their implementation class instead.

This way we avoid unnecessary null fields, since those aren't needed in this class here.

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
```diff
@@ -448,6 +453,39 @@ public static void getBackpack(ItemStack item, Consumer<PlayerBackpack> callback
...
+        NamespacedKey setId = null;
+        for (HashedArmorpiece armor : armor) {
+            Optional<SlimefunArmorPiece> armorPiece = armor.getItem();
+            if (!armorPiece.isPresent()) return false;
+
+            if (armorPiece.get() instanceof CustomProtection) {
+                CustomProtection protectedArmor = (CustomProtection) armorPiece.get();
+
+                if (first) {
+                    if (protectedArmor.isFullSetRequired()) setId = armorPiece.get().getSetId();
```
WalshyDev, line 469: You made a variable for the armour, use it. Also, if you're gonna do a one line if, it's generally only for a short return. This is bad.

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
```diff
@@ -448,6 +453,39 @@ public static void getBackpack(ItemStack item, Consumer<PlayerBackpack> callback
...
+                    first = false;
```
WalshyDev, line 470: Instead of a first variable you could literally just check if the ID is null and set it.

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
```diff
@@ -448,6 +453,39 @@ public static void getBackpack(ItemStack item, Consumer<PlayerBackpack> callback
...
+                }
+
+                for (ProtectionType protectionType : protectedArmor.getProtectionTypes()) {
+                    if (protectionType == type) {
+                        if (setId == null) {
+                            return true;
+                        } else if (setId.equals(armorPiece.get().getSetId())) {
```
WalshyDev, line 477: Again, there's a variable. I don't like all these get()s

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/CustomProtection.java
```diff
@@ -0,0 +1,41 @@
...
+import io.github.thebusybiscuit.slimefun4.implementation.items.armor.HazmatArmorPiece;
+import io.github.thebusybiscuit.slimefun4.implementation.items.armor.SlimefunArmorPiece;
+
+/**
+ * Implement this interface for any {@link SlimefunArmorPiece} to prevent
+ * the {@link Player} wearing that {@link SlimefunArmorPiece}
+ *
+ * <b>Important</b>: This will not cancel any {@link EntityDamageEvent}.
+ * It will simply prevent Slimefun from ever applying {@link ProtectionType}
+ * to this {@link SlimefunArmorPiece}'s wearer.
```
WalshyDev, line 15: I don't understand what this class does at all from this comment. It doesn't let people put on this armour? But also (?) stops "applying protection", what does that mean?

This really needs to be reworded.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/CustomProtection.java
```diff
@@ -0,0 +1,41 @@
...
+ * @see SlimefunArmorPiece
+ * @see HazmatArmorPiece
+ * @see ItemAttribute
+ *
+ */
+public interface CustomProtection extends ItemAttribute {
+
+    /**
+     * This returns the {@link ProtectionType}s this {@link ItemAttribute}
+     * prevents the assigned {@link SlimefunArmorPiece} to be damaged by.
```
WalshyDev, line 28: Does this just stop the armour from being damaged or the player too?

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/CustomProtection.java
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @return The {@link ProtectionType}s
+     */
+    ProtectionType[] getProtectionTypes();
+
+    /**
+     * This returns the {@link ProtectionType}s this {@link ItemAttribute}
+     * prevents the assigned {@link SlimefunArmorPiece} to be damaged by.
+     *
+     * @return The {@link ProtectionType}s
```
WalshyDev, line 38: Copied comment. Wasn't changed.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Represents the {@link ProtectionType} that a {@link CustomProtection}
+ * prevents the damage from.
+ *
+ * @author Linox
+ *
+ * @see CustomProtection
+ *
+ */
+public enum ProtectionType {
```
WalshyDev, line 12: Not sure I'm a fan of this being an enum. I'd rather addons be able to add their own as they wish. This could be a Predicate with the damage event (or such object) and then it's tested.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
```diff
@@ -295,6 +298,41 @@ private SlimefunItems() {}
         REINFORCED_ALLOY_LEGGINGS.addUnsafeEnchantments(reinforced);
         REINFORCED_ALLOY_BOOTS.addUnsafeEnchantments(reinforced);
 
+        if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
+
```
WalshyDev, line 302: Empty line

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
```diff
@@ -295,6 +298,41 @@ private SlimefunItems() {}
...
+            ItemMeta scubaHelmetMeta = SCUBA_HELMET.getItemMeta();
+            List<String> scubaHelmetMetaLore = scubaHelmetMeta.getLore();
+            scubaHelmetMetaLore.addAll(Arrays.asList("",
+                    ChatColor.translateAlternateColorCodes('&', "&7Equip the full set for:"),
```
WalshyDev, line 306: ChatColors.color

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/HazmatArmorPiece.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ */
+public class HazmatArmorPiece extends SlimefunArmorPiece implements CustomProtection {
+
+    private final ProtectionType[] types;
+
+    public HazmatArmorPiece(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, PotionEffect[] effects) {
+        super(category, item, recipeType, recipe, effects, new NamespacedKey(SlimefunPlugin.instance, "hazmat_suit"));
+
+        types = new ProtectionType[] {ProtectionType.BEES, ProtectionType.RADIATION};
```
WalshyDev, line 31: Just initialise right away

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
```diff
@@ -0,0 +1,51 @@
...
+                PlayerProfile.request(p);
+                return;
+            }
+
+            PlayerProfile profile = optional.get();
+            if (profile.isProtected(ProtectionType.BEES)) {
+                for (ItemStack armor : p.getInventory().getArmorContents()) {
+                    ItemUtils.damageItem(armor, 1, false);
+                }
+                e.setDamage(0D);
```
WalshyDev, line 46: Should cancel. This will mess with other plugins.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
LinoxGH: I don't think I understand this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
LinoxGH: Cannot when cancelled bee doesn't stop from trying to attack you.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
WalshyDev: Instead of accepting an enum, having like a canProtect(Predicate<Event>) and then test the predicate for if it can protect.

This way, other addons can actually add their own protection with this system.

TheBusyBiscuit: I agree that it would be really awesome if addons could add their own types.
But I don't really see how a Predicate would be useful here.
I would rather suggest a Namespaced-key based approach here too.

However I would also recommend to delay this discussion, we can always change it and improve this API later.
We probably bombarded Linox with change requests more than enough for now :LUL:
I'd say this is ready to be merged and we can always make the API more accessible later down the line.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BeeListener.java
TheBusyBiscuit: Hmmm, what about setting the anger level or target or such?

LinoxGH: That would cause more issues most likely.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
WalshyDev: I'm ok with this.

TheBusyBiscuit:
I'll be merging this in a seperate branch for now since I may change a few tids and bits before merging it.
I'll also accumulate your PRs in that branch in order to not fire 10 translations update notifications after another.

PR #1101, opened 2019-09-08
Code Cleanup, Addition of Electric Compressor(s) and Head Cracker!

LinoxGH:
## Description
Added Head Cracker, Electric Compressor and cleaned up some code.

## Changes
- Added Head Cracker.
  - An item that breaks head/skull blocks on right click.
- Fixed indent in CargoNet.java
- Moved Icy Bow, Wind Staff, Easter Egg and Explosive Bow to separate classes in items package.
- Added "messages.cannot-break" message and added it to Head Cracker, Explosive Pickaxe and Explosive Shovel.
- Added Electric compressor 1, 2 and 3.
  - Compresses reinforced alloy ingots and steel ingots into reinforced alloy plates and steel plates.
  - Compresses stone chunks into cobblestones.
- Optimised the code of ElectricSmeltery.java a bit.
- Changed SlimefunBow from a SlimefunItem to SimpleSlimefunItem<BowShootHandler>.

## Related Issues
Resolves #1098 (Note: But it is also resolved in #1099)

## Testability

- [X] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [X] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.

### NOTE: It is recommended that you merge #1099 first as it fixes #1098 alone.

PR #2061, opened 2020-06-27
Added the Head Cracker.

LinoxGH:
## Description

Added the Head Cracker from my old pr #1101.

## Changes

Added Head Cracker and relevant research, Enhanced Crafting Table recipe.
Added EntityDamageListener and EntityDamageHandler.

## Related Issues

N/A

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HeadCracker.java
```diff
@@ -0,0 +1,100 @@
...
+
+    // This is kind of an Easter Egg.
+    // Don't add this to the item lore.
+    // I implemented this because I thought it was funny.
+    // ~Linox
+    private EntityDamageHandler getEntityDamageHandler() {
+        return (e, entity, item1) -> {
+            if (entity instanceof Skeleton  || entity instanceof SkeletonHorse) {
+                ((Creature) entity).damage(e.getFinalDamage() + 4D);
+                e.setCancelled(true);
```
WalshyDev, line 84: You shouldn't damage and then cancel. You should set damage and continue. You would break all sorts of protection plugins and all sorts with this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HeadCracker.java
```diff
@@ -0,0 +1,100 @@
...
+            Optional<Block> block = e.getClickedBlock();
+            Player p = e.getPlayer();
+
+            if (block.isPresent()) {
+                Block b = block.get();
+                if (!MaterialCollections.getAllHeads().contains(b.getType())) return;
+
+                if (SlimefunPlugin.getProtectionManager().hasPermission(p, b.getLocation(), ProtectableAction.BREAK_BLOCK)) {
+                    BlockBreakEvent event = new BlockBreakEvent(b, p);
+                    Bukkit.getPluginManager().callEvent(event);
```
WalshyDev, line 46: Why? You aren't even breaking a block?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HeadCracker.java
```diff
@@ -0,0 +1,100 @@
...
+import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SimpleSlimefunItem;
+import me.mrCookieSlime.Slimefun.Objects.handlers.BlockBreakHandler;
+import me.mrCookieSlime.Slimefun.Objects.handlers.EntityDamageHandler;
+import me.mrCookieSlime.Slimefun.Objects.handlers.ItemUseHandler;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+
+public class HeadCracker extends SimpleSlimefunItem<ItemUseHandler> implements DamageableItem {
```
WalshyDev, line 28: What does this item do exactly...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HeadCracker.java
LinoxGH: you are...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HeadCracker.java
LinoxGH: Right click on a sf machine or any head really to pick it up.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HeadCracker.java
WalshyDev: Where exactly? This is a use handler, right click. You're just firing a break event on right click.

TheBusyBiscuit:
I don't really like this addition too much to be honest.
A "Head Cracker" seems more random than useful to me.

I know people suggested this but I think it should rather be in an addon than the core plugin.

I also think the "Head Cracker" item should rather be something more accessible, like a Wrench which can only pick up machines though.
Iirc Panda or Walshy was working on something like that in LiteXpansion, so Id much prefer having this only for machines and in that addon, rather than having this in the core plugin.

LinoxGH:
mmkay.

PR #2071, opened 2020-07-04
Added Chance Drop for MOB_DROP recipe items. References #980

dniym:
## Description
This was in reference to pull request #980 I finally got around to making the requested changes

## Changes
Added interface for ChanceDrop items
Added a check for chance drop items in the EntityDeathListener
Adjusted the basic circuit board from 100% chance down to 75%

## Related Issues
#980 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
@@ -9,18 +9,24 @@ import io.github.thebusybiscuit.slimefun4.core.handlers.ItemUseHandler;
 import me.mrCookieSlime.Slimefun.Lists.RecipeType;
 import me.mrCookieSlime.Slimefun.Objects.Category;
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SimpleSlimefunItem;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.interfaces.ChanceDrop;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
-public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implements NotPlaceable {
+public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implements NotPlaceable, ChanceDrop {
 
     private final ItemSetting<Boolean> dropSetting = new ItemSetting<>("drop-from-golems", true);
-
+    private final ItemSetting<Integer> chance = new ItemSetting<> ("golem-drop-chance", 75);
+    
     public BasicCircuitBoard(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
         super(category, item, recipeType, recipe);
 
         addItemSetting(dropSetting);
+        addItemSetting(chance);
     }
 
+    public int getChance() {
+    	return chance.getValue();
+    }
     public boolean isDroppedFromGolems() {
         return dropSetting.getValue();
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
@@ -16,7 +16,7 @@ public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implem
 
     private final ItemSetting<Boolean> dropSetting = new ItemSetting<>("drop-from-golems", true);
     private final ItemSetting<Integer> chance = new ItemSetting<> ("golem-drop-chance", 75);
-    
+
     public BasicCircuitBoard(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
         super(category, item, recipeType, recipe);
 
@@ -26,9 +26,9 @@ public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implem
 
     @Override
     public int getDropChance() {
-    	return chance.getValue();
+        return chance.getValue();
     }
-    
+
     public boolean isDroppedFromGolems() {
         return dropSetting.getValue();
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
@@ -28,6 +28,7 @@ public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implem
     public int getChance() {
     	return chance.getValue();
     }
+    
     public boolean isDroppedFromGolems() {
         return dropSetting.getValue();
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
@@ -9,10 +9,10 @@ import io.github.thebusybiscuit.slimefun4.core.handlers.ItemUseHandler;
 import me.mrCookieSlime.Slimefun.Lists.RecipeType;
 import me.mrCookieSlime.Slimefun.Objects.Category;
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SimpleSlimefunItem;
-import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.interfaces.ChanceDrop;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.interfaces.RandomMobDrop;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
-public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implements NotPlaceable, ChanceDrop {
+public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implements NotPlaceable, RandomMobDrop {
 
     private final ItemSetting<Boolean> dropSetting = new ItemSetting<>("drop-from-golems", true);
     private final ItemSetting<Integer> chance = new ItemSetting<> ("golem-drop-chance", 75);
@@ -25,7 +25,7 @@ public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implem
     }
 
     @Override
-    public int getChance() {
+    public int getDropChance() {
     	return chance.getValue();
     }
     
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
@@ -24,6 +24,7 @@ public class BasicCircuitBoard extends SimpleSlimefunItem<ItemUseHandler> implem
         addItemSetting(chance);
     }
 
+    @Override
     public int getChance() {
     	return chance.getValue();
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
@@ -3,7 +3,6 @@ package io.github.thebusybiscuit.slimefun4.implementation.listeners;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
-import java.util.concurrent.ThreadLocalRandom;
 
 import org.bukkit.Material;
 import org.bukkit.entity.Player;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
@@ -55,7 +55,7 @@ public class MobDropListener implements Listener {
                 if (sfi instanceof RandomMobDrop && ((RandomMobDrop)sfi).getDropChance() <= random) 
                     continue;
 
-                if (sfi instanceof BasicCircuitBoard  && ((BasicCircuitBoard)sfi).isDroppedFromGolems()) 
+                if (sfi instanceof BasicCircuitBoard  && !((BasicCircuitBoard)sfi).isDroppedFromGolems()) 
                     continue;
 
                 drops.add(drop.clone());
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
@@ -2,6 +2,7 @@ package io.github.thebusybiscuit.slimefun4.implementation.listeners;
 
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
 
 import org.bukkit.Material;
 import org.bukkit.entity.Player;
@@ -14,6 +15,7 @@ import io.github.thebusybiscuit.slimefun4.core.handlers.EntityKillHandler;
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.implementation.items.electric.BasicCircuitBoard;
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.interfaces.ChanceDrop;
 import me.mrCookieSlime.Slimefun.api.Slimefun;
 
 public class MobDropListener implements Listener {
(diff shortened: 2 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -34,7 +36,17 @@ public void onEntityKill(EntityDeathEvent e) {
 
             Set<ItemStack> customDrops = SlimefunPlugin.getRegistry().getMobDrops(e.getEntityType());
             if (customDrops != null && !customDrops.isEmpty()) {
-                addDrops(p, customDrops, e.getDrops());
+            	for(ItemStack is:customDrops)
+            	{
+            		SlimefunItem sfi = SlimefunItem.getByItem(is);
+            		if(sfi != null && sfi instanceof ChanceDrop) {
+            			if(((ChanceDrop)sfi).getChance() <= ThreadLocalRandom.current().nextInt(0,100))
```
WalshyDev, line 43: Should make a variable for the random since it could be called multiple times and this isn't exactly the least expensive thing in the world

WalshyDev: No need to specify the 0 as origin

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -34,7 +36,17 @@ public void onEntityKill(EntityDeathEvent e) {
 
             Set<ItemStack> customDrops = SlimefunPlugin.getRegistry().getMobDrops(e.getEntityType());
             if (customDrops != null && !customDrops.isEmpty()) {
-                addDrops(p, customDrops, e.getDrops());
+            	for(ItemStack is:customDrops)
+            	{
+            		SlimefunItem sfi = SlimefunItem.getByItem(is);
+            		if(sfi != null && sfi instanceof ChanceDrop) {
+            			if(((ChanceDrop)sfi).getChance() <= ThreadLocalRandom.current().nextInt(0,100)) 
+                			continue;
```
WalshyDev, line 44: Useless continue, there's no other logic in this for loop

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/ChanceDrop.java
```diff
@@ -0,0 +1,16 @@
...
+
+
+public interface ChanceDrop extends ItemAttribute {
+
+    /**
+     * This field sets the % chance for an item to drop when a creature is killed.
+     */
+    int chance = 100;  
+
+    public default int getChance() {
```
WalshyDev, line 13: The entire point of this class is to have a chance. Why would there be a default for it? 

Remove the constant and make this a standard non-default, interface method

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/BasicCircuitBoard.java
```diff
@@ -9,18 +9,24 @@
...
+    private final ItemSetting<Integer> chance = new ItemSetting<> ("golem-drop-chance", 75);
+    
     public BasicCircuitBoard(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
         super(category, item, recipeType, recipe);
 
         addItemSetting(dropSetting);
+        addItemSetting(chance);
     }
 
+    public int getChance() {
```
WalshyDev, line 27: Missing @Override

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -31,11 +33,18 @@ public void onEntityKill(EntityDeathEvent e) {
         if (e.getEntity().getKiller() != null) {
             Player p = e.getEntity().getKiller();
             ItemStack item = p.getInventory().getItemInMainHand();
-
+            Random rnd = new Random();
```
WalshyDev, line 36: Always use ThreadLocalRandom

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -31,11 +33,18 @@ public void onEntityKill(EntityDeathEvent e) {
...
+            
             Set<ItemStack> customDrops = SlimefunPlugin.getRegistry().getMobDrops(e.getEntityType());
-            if (customDrops != null && !customDrops.isEmpty()) {
-                addDrops(p, customDrops, e.getDrops());
-            }
+            if (customDrops != null && !customDrops.isEmpty()) 
+            	for(ItemStack is:customDrops)
+            	{
```
WalshyDev, line 41: On the line above

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -31,11 +33,18 @@ public void onEntityKill(EntityDeathEvent e) {
...
             ItemStack item = p.getInventory().getItemInMainHand();
-
+            Random rnd = new Random();
+            
             Set<ItemStack> customDrops = SlimefunPlugin.getRegistry().getMobDrops(e.getEntityType());
-            if (customDrops != null && !customDrops.isEmpty()) {
-                addDrops(p, customDrops, e.getDrops());
-            }
+            if (customDrops != null && !customDrops.isEmpty()) 
+            	for(ItemStack is:customDrops)
```
WalshyDev, line 40: Spacing between is the colon and custom drops

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -31,11 +33,18 @@ public void onEntityKill(EntityDeathEvent e) {
...
+            	{
+            		SlimefunItem sfi = SlimefunItem.getByItem(is);
+            		if(sfi != null && sfi instanceof ChanceDrop)
```
WalshyDev, line 43: Space after if and below too

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/ChanceDrop.java
```diff
@@ -0,0 +1,13 @@
...
+
+import io.github.thebusybiscuit.slimefun4.core.attributes.ItemAttribute;
+
+
+public interface ChanceDrop extends ItemAttribute {
+	
+    /**
+     * This method returns the % chance for an item to drop when a creature is killed.
+     */
+    public int getChance() ;
```
WalshyDev, line 11: Remove the space before the semicolon.

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/ChanceDrop.java
```diff
@@ -0,0 +1,13 @@
+package me.mrCookieSlime.Slimefun.Objects.SlimefunItem.interfaces;
+
+import io.github.thebusybiscuit.slimefun4.core.attributes.ItemAttribute;
+
+
+public interface ChanceDrop extends ItemAttribute {
```
TheBusyBiscuit, line 6: Also: Missing documentation and @author tag.
Should probably also mark this as a @FunctionalInterface too.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -31,11 +33,18 @@ public void onEntityKill(EntityDeathEvent e) {
...
             Set<ItemStack> customDrops = SlimefunPlugin.getRegistry().getMobDrops(e.getEntityType());
-            if (customDrops != null && !customDrops.isEmpty()) {
-                addDrops(p, customDrops, e.getDrops());
-            }
+            if (customDrops != null && !customDrops.isEmpty()) 
+            	for(ItemStack is:customDrops)
+            	{
+            		SlimefunItem sfi = SlimefunItem.getByItem(is);
+            		if(sfi != null && sfi instanceof ChanceDrop) 
+            			if(((ChanceDrop)sfi).getChance() >= rnd.nextInt(100))
```
TheBusyBiscuit, line 44: I think you can merge these two statements.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
TheBusyBiscuit, line 43: Also: An instanceof check always returns false for null, so the null check is redundant here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
TheBusyBiscuit, line 36: ThreadLocalRandom

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
dniym: not sure what you're asking for with "On the line above"

WalshyDev: Move the curly brace to the line above. One true brace style

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -31,11 +33,16 @@ public void onEntityKill(EntityDeathEvent e) {
...
+            
             Set<ItemStack> customDrops = SlimefunPlugin.getRegistry().getMobDrops(e.getEntityType());
-            if (customDrops != null && !customDrops.isEmpty()) {
-                addDrops(p, customDrops, e.getDrops());
-            }
+            if (customDrops != null && !customDrops.isEmpty()) 
+            	for(ItemStack is : customDrops) {
+            		SlimefunItem sfi = SlimefunItem.getByItem(is);
+            		if (sfi instanceof ChanceDrop && ((ChanceDrop)sfi).getChance() >= random) 
+            				addDrops(p, customDrops, e.getDrops());
```
WalshyDev, line 43: Many spaces on this line

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -31,11 +33,16 @@ public void onEntityKill(EntityDeathEvent e) {
...
             ItemStack item = p.getInventory().getItemInMainHand();
-
+            int random = ThreadLocalRandom.current().nextInt(100);
+            
             Set<ItemStack> customDrops = SlimefunPlugin.getRegistry().getMobDrops(e.getEntityType());
-            if (customDrops != null && !customDrops.isEmpty()) {
-                addDrops(p, customDrops, e.getDrops());
-            }
+            if (customDrops != null && !customDrops.isEmpty()) 
+            	for(ItemStack is : customDrops) {
```
WalshyDev, line 40: Space after for

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/ChanceDrop.java
```diff
@@ -0,0 +1,27 @@
+package me.mrCookieSlime.Slimefun.Objects.SlimefunItem.interfaces;
+
+import io.github.thebusybiscuit.slimefun4.core.attributes.ItemAttribute;
+
+/**
+ * This interface, when attached to a {@link SlimefunItem}, provides an easy method for adding 
+ * a % chance to drop for an {@link SlimefunItem} on {@link entityDeathEvent}, this chance is 0-100
+ * and used in conjunction with the MOB_DROP {@link RecipeType}. 
+ * see {@link BasicCircuitBoard} and {@link MobDropListener}.
```
TheBusyBiscuit, line 9: There is an @ see which you should use here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
TheBusyBiscuit, line 43: Also, since this is in a for loop... wouldn't this add the entire drops list multiple times instead of the items one by one?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -48,14 +50,20 @@ public void onEntityKill(EntityDeathEvent e) {
...
+    	int random = ThreadLocalRandom.current().nextInt(100);
+    	
         for (ItemStack drop : customDrops) {
             if (Slimefun.hasUnlocked(p, drop, true)) {
-                if (circuitBoard != null && circuitBoard.isItem(drop) && !circuitBoard.isDroppedFromGolems()) {
+            	SlimefunItem sfi = SlimefunItem.getByItem(drop);
+            	if (sfi instanceof RandomMobDrop && ((RandomMobDrop)sfi).getDropChance() <= random) 
+            		continue;
+            	
+                if (circuitBoard != null && circuitBoard.isItem(drop) && !circuitBoard.isDroppedFromGolems())
```
TheBusyBiscuit, line 61: No need to even have this CircuitBoard variable at this point anymore. You can just check if `sfi` is an instance of your Circuit Board, would be much better performance wise since you already done these operations earlier.

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/RandomMobDrop.java
```diff
@@ -0,0 +1,27 @@
+package me.mrCookieSlime.Slimefun.Objects.SlimefunItem.interfaces;
+
+import io.github.thebusybiscuit.slimefun4.core.attributes.ItemAttribute;
+
+/**
+ * This interface, when attached to a {@link SlimefunItem}, provides an easy method for adding 
+ * a % chance to drop for an {@link SlimefunItem} on {@link entityDeathEvent}, this chance is 0-100
+ * and used in conjunction with the MOB_DROP {@link RecipeType}. 
+ * @see BasicCircuitBoard and @see MobDropListener.
```
TheBusyBiscuit, line 9: No "and" here. The syntax is as follow:

@ see x
@ see y

And we usually place it below the @ author tag.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
dniym: I see what you're saying here, however; in testing I did not ever receive any extra drops no matter how many other items the creature in question dropped.   It may have become an issue if the creature dropped multiple slimefun items...   Moved the logic to the addDrops() method to just skip them if they fail the chance check.

TheBusyBiscuit: Yes, you are looping through "customDrops", so this issue would only appear when multiple Slimefun items were at play here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
```diff
@@ -48,14 +47,20 @@ public void onEntityKill(EntityDeathEvent e) {
...
+    	
         for (ItemStack drop : customDrops) {
             if (Slimefun.hasUnlocked(p, drop, true)) {
-                if (circuitBoard != null && circuitBoard.isItem(drop) && !circuitBoard.isDroppedFromGolems()) {
+            	SlimefunItem sfi = SlimefunItem.getByItem(drop);
+            	if (sfi instanceof RandomMobDrop && ((RandomMobDrop)sfi).getDropChance() <= random) 
+            		continue;
+            	
+                if (sfi instanceof BasicCircuitBoard  && ((BasicCircuitBoard)sfi).isDroppedFromGolems()) 
                     continue;
```
WalshyDev, line 59: Why is this being continued? We want this to drop. You need to negate the isDropped...

dniym:
> 
> 
> One last formatting request, remove these bloody tabs and use spaces. It's 2020 dammit, no one should use tabs anymore.
> 
> Also, fix this thing and then LGTM

Also...  Pretty sure my IDE adds tabs rather than single spaces...  I can write a block of code with zero indents, then do 
Source->Correct Indentation and it formats it with tabs instead of spaces...

Also isn't pressing tab once or twice way faster than the spacebar multiple times to get the same indentation?

WalshyDev:
There'll be a setting in your IDE to use spaces instead of tabs. And yes it's quicker, hence why we all press tab. It's just, that tab key does 4 spaces. Just like how everyone in JS presses tab for 2 chars.

dniym:
> 
> 
> There'll be a setting in your IDE to use spaces instead of tabs. And yes it's quicker, hence why we all press tab. It's just, that tab key does 4 spaces. Just like how everyone in JS presses tab for 2 chars.

Alright done.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MobDropListener.java
WalshyDev: this issue still exists

dniym: Fixed

PR #2109, opened 2020-07-22
Fix some regex, finals and replace isPresent with ifPresent

WalshyDev:
## Description
Title and also fixed a type. NUMBER_SEPERATOR -> NUMBER_SEPARATOR.

## Changes
* Fix some useless escapes in regex
* Change the `,|\\.` to a simple `[,.]`
* Fixed typo
* Added some finals
* Changed `if (isPresent) execute()` to a simple `ifPresent(execute())`

## Related Issues
N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/RechargeableHelper.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/RechargeableHelper.java
@@ -29,7 +29,7 @@ final class RechargeableHelper {
 
     private static final NamespacedKey CHARGE_KEY = new NamespacedKey(SlimefunPlugin.instance(), "item_charge");
     private static final String LORE_PREFIX = ChatColors.color("&8\u21E8 &e\u26A1 &7");
-    private static final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "[0-9\\.]+ \\/ [0-9\\.]+ J");
+    private static final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "[0-9.]+ / [0-9.]+ J");
 
     private RechargeableHelper() {}
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/PatternUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/PatternUtils.java
@@ -26,6 +26,5 @@ public final class PatternUtils {
     public static final Pattern ALPHANUMERIC = Pattern.compile("[A-Fa-f0-9]+");
     public static final Pattern NUMERIC = Pattern.compile("[0-9]+");
 
-    public static final Pattern NUMBER_SEPERATOR = Pattern.compile(",|\\.");
-
+    public static final Pattern NUMBER_SEPARATOR = Pattern.compile("[,.]");
 }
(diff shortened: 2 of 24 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/RechargeableHelper.java
```diff
@@ -29,7 +29,7 @@
 
     private static final NamespacedKey CHARGE_KEY = new NamespacedKey(SlimefunPlugin.instance(), "item_charge");
     private static final String LORE_PREFIX = ChatColors.color("&8\u21E8 &e\u26A1 &7");
-    private static final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "[0-9\\.]+ \\/ [0-9\\.]+ J");
+    private static final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "[0-9.]+ / [0-9.]+ J");
```
TheBusyBiscuit, line 32: This will break the regex though, wouldn't it?
Since . must be escaped otherwise it is treated as a wildcard for any character...

src/main/java/io/github/thebusybiscuit/slimefun4/utils/PatternUtils.java
```diff
@@ -26,6 +26,5 @@ private PatternUtils() {}
     public static final Pattern ALPHANUMERIC = Pattern.compile("[A-Fa-f0-9]+");
     public static final Pattern NUMERIC = Pattern.compile("[0-9]+");
 
-    public static final Pattern NUMBER_SEPERATOR = Pattern.compile(",|\\.");
-
+    public static final Pattern NUMBER_SEPARATOR = Pattern.compile("[,.]");
```
TheBusyBiscuit, line 29: Same here, wouldn't . be treated as a wildcard?

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/RechargeableHelper.java
WalshyDev: No. In a character class it is treated as a literal. Only defined groups such as `\w`, `\s`, `\d` and ranges like `0-9`, `a-z` are not.

Here's a repl too see for yourself: https://repl.it/repls/ComposedJoyfulUnderstanding

src/main/java/io/github/thebusybiscuit/slimefun4/utils/PatternUtils.java
WalshyDev: See comment above

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/RechargeableHelper.java
TheBusyBiscuit: What's with the slash though? Since that is not in a character class. Did you test this with a charging bench?

WalshyDev: Forward slashes don't need to be escaped in Java.

TheBusyBiscuit: Everything I learned was a lie :CryingPepe:

PR #2116, opened 2020-07-23
Change to how ExplosiveTool calculates drops to allow for silk touch

kylehassall27:
## Description

Changed ExplosiveTool to use the vanilla ItemStack of the tool (with correct enchantments) rather than the SlimeFunItemStack (which does not contain the enchantments) when calculating what drops.
<strike>I'm not sure why this wasn't done this way in the first place, so I might be missing something but it seems that the Explosive Pickaxe and Explosive Shovel now work as expected with fortune and silk touch enchantments.</strike>

## Changes

Changed from getItem() (which returns a SlimeFunItemStack) to item (which is an ItemStack) when calculating the drops for explosive tools. This made the fortune calculation redundant (since the game calculates all drops based on the enchantments that are actually on the pick) so I removed that.

## Related Issues

N/A

## Checklist


kylehassall27:
After testing, it seems that the functionality of getDrops() was changed somewhere between 1.13 and 1.16, so this actually needs to  be done differently

Issue #2097, opened 2020-07-17
The piglins accept the Slimefun Gold Ingots


andris155:
## Description (Required)

The piglins accept the Slimefun Gold Ingots.
The Gold Ingot (4-Carat) very easy to make. Piglins give valuable items for it.

## Steps to reproduce the Issue (Required)

Drop the Gold Ingot to the piglin and then will trade with you.

## Expected behavior (Required)

The piglins don't accept the Slimefun Gold Ingots

## Environment (Required)

 - Minecraft Version: 1.16.1
 - Slimefun Version: vDEV - 584 (git 9af90830)
 - CS-CoreLib Version: vDEV - 88 (git 6a39a9e7)

andris155:
Piglins accept all gold Slimefun items.

PR #2121, opened 2020-07-25
Fixed piglin SlimefunItem pickup

poma123:
## Description

Piglins now can not pick up items which are SlimefunItems.
A low-budget video about the fix: https://imgur.com/G0PDuZl

## Changes

* Added a check when piglins are about to pick up Slimefun items. When this happens, we cancel EntityPickupItemEvent.

## Related Issues

#2097 

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ItemPickupListener.java
```diff
@@ -30,6 +34,12 @@ public void onEntityPickup(EntityPickupItemEvent e) {
         else if (e.getItem().getItemStack().hasItemMeta() && e.getItem().getItemStack().getItemMeta().hasDisplayName() && e.getItem().getItemStack().getItemMeta().getDisplayName().startsWith(ITEM_PREFIX)) {
             e.setCancelled(true);
             e.getItem().remove();
+        } else if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_16)) {
+            if (e.getEntityType().equals(EntityType.PIGLIN)) {
```
TheBusyBiscuit, line 38: Enum comparisons should be done using == instead of .equals() as it allows the JVM to do premature eliminations during compiler-time rather than runtime.

poma123: Got it

TheBusyBiscuit:
What should probably also be noted is that you can also hand Piglins gold by right clicking them.
So this would also need to be patched too. I would suggest making a new Listener class for Piglins and loading that only if 1_16 is used.
This Listener class could then handle the pickup AND right-click.

Also from the video it seems like the Piglin will still try to target the gold. Can you maybe check if there is a way to prevent that? Perhaps an EntityTargetEvent or so is called?

poma123:
> What should probably also be noted is that you can also hand Piglins gold by right clicking them.
> So this would also need to be patched too. I would suggest making a new Listener class for Piglins and loading that only if 1_16 is used.
> This Listener class could then handle the pickup AND right-click.

Yeah I probably forgot that one, on it

> Also from the video it seems like the Piglin will still try to target the gold. Can you maybe check if there is a way to prevent that? Perhaps an EntityTargetEvent or so is called?

Will look into it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
```diff
@@ -10,9 +10,12 @@
 import java.util.logging.Level;
 import java.util.stream.Collectors;
 
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.*;
```
poma123, line 13: I know I know don't blame me!!!

TheBusyBiscuit: No star imports please D:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
```diff
@@ -10,9 +10,12 @@
...
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
 import org.bukkit.Bukkit;
 import org.bukkit.Server;
 import org.bukkit.command.Command;
+import org.bukkit.entity.EntityType;
```
TheBusyBiscuit, line 17: Also what is that import doing here?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/PiglinListener.java
```diff
@@ -0,0 +1,56 @@
...
+        if (e.getHand() == EquipmentSlot.OFF_HAND) {
+            itemStack = e.getPlayer().getInventory().getItemInOffHand();
+        } else {
+            itemStack = e.getPlayer().getInventory().getItemInMainHand();
+        }
+
+        SlimefunItem sfItem = SlimefunItem.getByItem(itemStack);
+
+        if (sfItem != null) {
+            e.setCancelled(true);
```
TheBusyBiscuit, line 53: We should probably also add a message saying that you cannot barter with piglins using Slimefun items.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/PiglinListener.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        ItemStack itemStack;
+
+        if (e.getHand() == EquipmentSlot.OFF_HAND) {
+            itemStack = e.getPlayer().getInventory().getItemInOffHand();
+        } else {
+            itemStack = e.getPlayer().getInventory().getItemInMainHand();
+        }
+
```
TheBusyBiscuit, line 49: Also just a small performance optimization: We should check if the item is a gold ingot, that way we won't have to do any checks that are not needed.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
poma123: No idea lol

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/PiglinListener.java
poma123: Is this resolved in the last commit?

TheBusyBiscuit: Looks good to me

poma123: Yay

PR #2133, opened 2020-07-25
A more efficient way to prevent brewing with Slimefun items

### LINKED ISSUE #2122: The potion brewing works from Slimefun dusts when using shift click
## Description (REQUIRED)

The potion brewing works from Slimefun dusts when using shift click.

1. /sf cheat - give Lead dust.
2. Place Awkward Potion and Blaze Powder to Brewing Stand.
3. Place Lead dust to Brewing Stand with the shift + click.
4. The Brewing Stand start create the potion.
Annotation 2020-07-25 132049

The potion brewing not work from Slimefun dusts when using shift click

## Environment (REQUIRED)

 - Minecraft Version: 1.16.1
 - Slimefun Version: vDEV - 593 (git abc47af5)
 - CS-CoreLib Version: vDEV - 88 (git 6a39a9e7)

poma123:
## Description

Changed the InventoryClickEvent check to BrewEvent. This way we only listen to the BrewEvent which fires when a Brewing Stand is about to complete brewing. If the ingredient is a Slimefun item, then we cancel the brewing event.

Video: https://imgur.com/3nXWxQ6

## Changes

- Changed InventoryClickEvent to the actual BrewEvent

## Related Issues

Fixes #2122

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VanillaMachinesListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VanillaMachinesListener.java
@@ -92,7 +92,7 @@ public class VanillaMachinesListener implements Listener {
         }
     }
 
-    @EventHandler(ignoreCancelled = true)
+    @EventHandler
     public void onBrew(BrewEvent e) {
         e.setCancelled(isUnallowed(SlimefunItem.getByItem(e.getContents().getIngredient())));
     }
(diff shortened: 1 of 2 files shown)


poma123:
Working on the Unit Test fix now.

poma123:
I don't have an actual idea of what am I missing here... It is working live on the server when using Slimefun items, the event gets cancelled.

TheBusyBiscuit:
Hmm, not too sure what to think of this approach. I think from a user-design standpoint this is rather... ambigious or weird since it would basically loop there forever.

I think a more appropriate fix would be to extend the current implementation to cover all slots and even shift, that way the Player will get a message telling them **why** it doesn't work.

poma123:
Well, You have a point. I agree. On user-side, this is a bit weird. I will spend some time with the inv clicks then and will open a new PR.

PR #2134, opened 2020-07-25
Backwards-compatibility mode disabled on new servers

poma123:
## Description

This will make backwards-compatibility mode disabled for new servers. (approved suggestion # 281 made by Walshy the king)

## Changes

See description.

## Related Issues

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
@@ -137,7 +137,7 @@ public final class SlimefunPlugin extends JavaPlugin implements SlimefunAddon {
     private ProtectionManager protections;
 
     // Important config files for Slimefun
-    private final Config config = new Config(this);
+    private Config config;
     private final Config items = new Config(this, "Items.yml");
     private final Config researches = new Config(this, "Researches.yml");
 
@@ -161,6 +161,7 @@ public final class SlimefunPlugin extends JavaPlugin implements SlimefunAddon {
         instance = this;
 
         if (minecraftVersion == MinecraftVersion.UNIT_TEST) {
+            config = new Config(this);
             local = new LocalizationService(this, "", null);
             gpsNetwork = new GPSNetwork();
             command.register();
@@ -174,6 +175,15 @@ public final class SlimefunPlugin extends JavaPlugin implements SlimefunAddon {
             if (isVersionUnsupported()) {
                 getServer().getPluginManager().disablePlugin(this);
                 return;
+            }
+
+            // Disabling backwards-compatibility fresh Slimefun installs
+            if (!new File("plugins/" + getName().replace(" ", "_"), "config.yml").exists()) {
+                config = new Config(this);
+                config.setValue("options.backwards-compatibility", false);
+                config.save();
+            } else {
+                config = new Config(this);
             }
 
             // Creating all necessary Folders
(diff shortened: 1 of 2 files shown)


TheBusyBiscuit:
Hmm, you see changing the default value here isn't exactly the best approach. The default value should still remain true, for... compatibility reasons, as this setting suggests.
What Walshy meant is having it as "false" if and only if the server has installed Slimefun for the very first time, in that case we can be sure that we won't need backwards compatibility.

Simply setting the default value in the config won't do the trick, as servers who were 1 or 2 RC builds behind and update will suddenly experience a ton of items breaking. We can only safely set it to false on confirmed fresh installations. Otherwise we would just provoke a huge wave of angry server owners and bug reports.

TheBusyBiscuit:
Also just to back this up with numbers... Only 2 in 3 servers run a Slimefun version that actually supports this config setting. It's a surprisingly good portion, but every third server will most definitely break when they finally update and this is left as false.

poma123:
Soo we should set this false when it's confirmed that the config was firstly created?

TheBusyBiscuit:
Either of that would probably be best, yes.
Also, @WalshyDev is probably gonna love you after this PR XD

poma123:
> Also, WalshyDev is probably gonna love you after this PR XD
Yeah I guess xd

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
```diff
@@ -176,6 +177,15 @@ else if (getServer().getPluginManager().isPluginEnabled("CS-CoreLib")) {
                 return;
             }
 
+            // Disabling backwards-compatibility for fresh Slimefun installs
+            if (!new File("plugins/" + getName().replace(" ", "_"), "config.yml").exists()) {
+                config = new Config(this);
```
WalshyDev, line 182: This is always set to the same thing. It makes no sense for this not to be final.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
```diff
@@ -176,6 +177,15 @@ else if (getServer().getPluginManager().isPluginEnabled("CS-CoreLib")) {
                 return;
             }
 
+            // Disabling backwards-compatibility for fresh Slimefun installs
+            if (!new File("plugins/" + getName().replace(" ", "_"), "config.yml").exists()) {
```
WalshyDev, line 181: JavaPlugin#getDataFolder

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
```diff
@@ -176,6 +177,15 @@ else if (getServer().getPluginManager().isPluginEnabled("CS-CoreLib")) {
...
+                config = new Config(this);
+                config.setValue("options.backwards-compatibility", false);
+                config.save();
```
WalshyDev, line 184: I'm thinking maybe data storage should be checked. Since that's a good indication of a new server and it's w good way to see if there's existing items.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
poma123: Sure, but when declaring the config, it just creates itself, so where can we check if it was created for the first time?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
poma123: Hmm, good idea

poma123:
Yeah, I tested with the last change

PR #2173, opened 2020-08-02
Allow Ancient Debris to be mined by Industrial Miners

lleyton:
## Description
Currently Ancient Debris is not able to be mined by Industrial Miners. This PR allows for that.

## Changes
- Ancient Debris can now be mined by Industrial Miners.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/IndustrialMiner.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/IndustrialMiner.java
@@ -212,7 +212,7 @@ public class IndustrialMiner extends MultiBlockMachine {
      * @return Whether this {@link IndustrialMiner} is capable of mining this {@link Material}
      */
     public boolean canMine(Material type) {
-        return type.name().endsWith("_ORE") || (canMineAncientDebris.getValue() && type == Material.ANCIENT_DEBRIS);
+        return type.name().endsWith("_ORE") || (type == Material.ANCIENT_DEBRIS && canMineAncientDebris.getValue());
     }
 
 }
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
While I am all for the addition of this feature, there has been a hefty debate about this addition of ancient debris for the miner on discord in the past.
Many were concerned about this being a big disruptance and unbalanced, as Players could easily clear out large quantities of debris within minutes.
And with the Advanced Miner being able to run on Lava... fuel consumption isn't really a counter-balance to this either.

I do think that these arguments hold some value against it, so we should really think about this in a balanced manner.
Maybe an ItemSetting for this, maybe a third tier which can mine Ancient Debris but requires expensive fuel, maybe radioactive even?
Or that mining Ancient Debris does some sort of damage to the machine?

lleyton:
Yeah, that's fair. I added an ItemSetting for it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/IndustrialMiner.java
```diff
@@ -207,7 +212,7 @@ public void onInteract(Player p, Block b) {
      * @return Whether this {@link IndustrialMiner} is capable of mining this {@link Material}
      */
     public boolean canMine(Material type) {
-        return type.name().endsWith("_ORE");
+        return type.name().endsWith("_ORE") || (canMineAncientDebris.getValue() && type == Material.ANCIENT_DEBRIS);
```
TheBusyBiscuit, line 215: Just a small micro-optimization here: If you swap the ItemSetting and the Material check, then this statement may execute just a tiny bit quicker as Enum comparisons are sped up by the compiler.

PR #2055, opened 2020-06-26
Lots of refactoring and rewriting of Item Energy

TheBusyBiscuit:
## Description

Item Energy uses lore to store energy which is really not good.
So I switched over to Persistent Data and updated all associated classes using this old system to use the new.
Next step would be to add Unit Tests for this.

@WalshyDev requesting your review.

## Changes

* Item Energy uses Persistent Data now
* Added a proper "Rechargeable" interface which can be added to items to safely define what is chargeable and what isn't
* Fixed the long-time going on typo of "Chargable" with this
* Fully deprecated the classes ChargableItem and ItemEnergy now.

## Checklist

- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [x] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [x] I followed the existing code standards and didn't mess up the formatting.
- [x] I did my best to add documentation to any public classes or methods I added.
- [x] I added sufficient Unit Tests to cover my code.

PR #2191, opened 2020-08-04
Refactored Block Energy API

TheBusyBiscuit:
## Description

Similar to PR #2055 this refactores the energy system.
This time it's the turn of the old class "ChargableBlock" (finally getting rid of that typo...).

In a similar fashion to the Rechargeable interface, the methods were moved to the interface `EnergyNetComponent`.
This removes the need of an id lookup and speeds up performance, the EnergyNet has been rewritten accordingly.

The old system should still work, it's just deprecated.

@WalshyDev requesting your review.

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
```diff
@@ -36,4 +43,104 @@
...
+     * @param l
+     *            The target {@link Location}
+     * @param charge
+     *            The new charge
+     */
+    default void setCharge(Location l, int charge) {
+        int capacity = getCapacity();
+
+        // This method only makes sense if we can actually store energy
+        if (capacity > 0) {
```
WalshyDev, line 89: isChargeable should be used here

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
```diff
@@ -36,4 +43,104 @@
...
+            }
+        }
+    }
+
+    default void addCharge(Location l, int charge) {
+        Validate.isTrue(charge > 0, "You can only add a positive charge!");
+        int capacity = getCapacity();
+
+        // This method only makes sense if we can actually store energy
+        if (capacity > 0) {
```
WalshyDev, line 109: Again, isChargeable

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
```diff
@@ -36,4 +43,104 @@
...
+            }
+        }
+    }
+
+    default void removeCharge(Location l, int charge) {
+        Validate.isTrue(charge > 0, "The charge to remove must be greater than zero!");
+        int capacity = getCapacity();
+
+        // This method only makes sense if we can actually store energy
+        if (capacity > 0) {
```
WalshyDev, line 130: isChargeable

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
```diff
@@ -73,9 +72,9 @@ public static EnergyNet getNetworkFromLocationOrCreate(Location l) {
         }
     }
 
-    private final Set<Location> generators = new HashSet<>();
-    private final Set<Location> storage = new HashSet<>();
-    private final Set<Location> consumers = new HashSet<>();
+    private final Map<Location, EnergyNetComponent> generators = new HashMap<>();
+    private final Map<Location, EnergyNetComponent> capacitors = new HashMap<>();
+    private final Map<Location, EnergyNetComponent> consumers = new HashMap<>();
```
WalshyDev, line 78: Global variables... not at the top... NotLikeThis

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
```diff
@@ -139,138 +149,141 @@ public void tick(Block b) {
...
-                        }
+                    if (provider.isChargeable()) {
+                        energy += provider.getCharge(l);
                     }
 
-                    if (provider.willExplode(source, config)) {
-                        exploded.add(source);
-                        BlockStorage.clearBlockInfo(source);
+                    if (provider.willExplode(l, config)) {
+                        exploded.add(l);
```
WalshyDev, line 251: Make a variable name less verbose, that's not like you Cookie

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
TheBusyBiscuit: That would just be the same call to getCapacity() again. The point of this is to not get the capacity twice.

TheBusyBiscuit: Also, isChargeable() could be overridden. We don't know the internals of that method in that scenario and neither do we know the computational time of that operation.
For these methods we assume they are chargeable, otherwise that method shouldn't have been called in the first place, we don't need to know if it actually should be charged or not, this decision should have taken place in advance. We should only care if a valid number was given to work with because of maths...

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
TheBusyBiscuit: Not my code, I only refactored this. But I can change it

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
TheBusyBiscuit: Do you prefer loc over l...

TheBusyBiscuit:
As already explained on discord, the benefit is stability and a massive performance improvement over the previous method.
With the old method we had to obtain the reference to `EnergyNetComponent` on every single electricity-related call, for a simple "Do I have enough energy, if so then consume x energy" operation, this lookup could run up to 8 times.
By bundling it in the HashMap, since that info/reference is there at that time anyway... we save a ton of processing power for electric machines. 

And even memory-wise the impact will be insignificant as only the reference is passed, we are saving a ton more memory by not temporarily creating up to 8 references for each location.

TheBusyBiscuit:
@WalshyDev :eyes:

WalshyDev:
This is on my review list. Also it's half 2 for you, go to sleep. 

GitHub <notifications@github.com> wrote:
“@WalshyDev 👀”

TheBusyBiscuit:
> This is on my review list.

Alright 👍 

> Also it's half 2 for you, go to sleep. 

"Proper sleep schedule" is foreign to me.
Also it's not half 2, it's actually an hour later, almost 3 am by now :LUL:

TheBusyBiscuit:
@WalshyDev review pls

PR #2256, opened 2020-08-27
Fixed Fluid Pump dupe.

LinoxGH:
## Description

There's a very old infinite lava duping bug made by fluid pump, dispenser and a timer. This fixes it by only checking for source blocks.

## Changes

Made fluid pump consume only source blocks.

## Related Issues

N/A on Github.

## Checklist

Note: I didn't test these changes and made them from my mobile phone, so yeah.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
@@ -33,6 +33,14 @@ import me.mrCookieSlime.Slimefun.api.energy.ChargableBlock;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenuPreset;
 
+/**
+ * This machine draws liquids from the world and puts them
+ * into buckets provided to the machine by using energy.
+ *
+ * @author TheBusyBiscuit
+ * @author Linox
+ *
+ */
 public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements InventoryBlock, EnergyNetComponent {
 
     private static final int ENERGY_CONSUMPTION = 32;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -149,6 +149,14 @@ else if (fluid.getType() == Material.WATER) {
 
         return Optional.empty();
     }
+    
+    private boolean isLiquid(Block block) {
+        BlockData data = block.getBlockData();
```
TheBusyBiscuit, line 154: This may be a bit performance-intense, so a prior Material check would be very adequate here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -130,12 +130,12 @@ protected void tick(Block b) {
     }
 
     private void consumeFluid(Block fluid) {
-        if (fluid.getType() == Material.WATER) {
+        if (fluid.getType() == Material.STATIONARY_WATER) {
```
TheBusyBiscuit, line 133: I dont think STATIONARY_WATER exists anymore.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
LinoxGH: > This may be a bit performance-intense, so a prior Material check would be very adequate here.

Done for both fluid in consumeFluid and that.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
LinoxGH: > I dont think STATIONARY_WATER exists anymore.

Oops fixed. I thought it did lol.

TheBusyBiscuit (approved):

PR #2264, opened 2020-08-30
Fixes #2263

### LINKED ISSUE #2263: Cooler fails to restore hunger when your hunger bar reaches absolute 0.
## Description (REQUIRED)

When your hunger level reaches 0 to the point where you start taking starvation damage, despite having filled Cooler with juices, it fails to feed/restore hunger points. As soon as you eat something to stop the starvation damage it starts working.
NOTE: It doesn't consume the juices from the cooler either.

Empty your hunger bar either using the Wind Staff or normally. While you are taking damage fillup your Cooler or place a filled Cooler with juices in your inventory.

It consumes the juice and starts restoring hunger points.

## Environment (REQUIRED)

 - Minecraft Version: 1.16.2-R0.1-SNAPSHOT
 - Slimefun Version: DEV - 647
 - CS-CoreLib Version: DEV - 79

LinoxGH:
## Description

Made coolers also saturate the player when they get hunger damage.

## Changes

Made it so when a player gets hunger damage they use their cooler.

## Related Issues

Fixes #2263 

## Checklist

Note: I did these changes ftom mobile. So yeah... it needs testing and I can't do that.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
@@ -31,6 +31,11 @@ public class WindStaff extends SimpleSlimefunItem<ItemUseHandler> {
                 if (p.getInventory().getItemInMainHand().getType() != Material.SHEARS && p.getGameMode() != GameMode.CREATIVE) {
                     FoodLevelChangeEvent event = new FoodLevelChangeEvent(p, p.getFoodLevel() - 2);
                     Bukkit.getPluginManager().callEvent(event);
+                    
+                    if (event.isCancelled()) {
+                        SlimefunPlugin.getLocalization().sendMessage(p, "messages.wind-staff-fail", true);
+                        return;
+                    }
                     p.setFoodLevel(event.getFoodLevel());
                 }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
@@ -32,11 +32,9 @@ public class WindStaff extends SimpleSlimefunItem<ItemUseHandler> {
                     FoodLevelChangeEvent event = new FoodLevelChangeEvent(p, p.getFoodLevel() - 2);
                     Bukkit.getPluginManager().callEvent(event);
                     
-                    if (event.isCancelled()) {
-                        SlimefunPlugin.getLocalization().sendMessage(p, "messages.wind-staff-fail", true);
-                        return;
+                    if (!event.isCancelled()) {
+                        p.setFoodLevel(event.getFoodLevel());
                     }
-                    p.setFoodLevel(event.getFoodLevel());
                 }
 
                 p.setVelocity(p.getEyeLocation().getDirection().multiply(4));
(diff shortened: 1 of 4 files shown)


TheBusyBiscuit (approved):
This is a sufficient fix for this issue.
However I would suggest adding an additional safeguard to the Wind Staff, as it seems to neglect cancelled FoodLevelChangeEvents, this would prevent any further issues from occuring in this regard.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CoolerListener.java
```diff
@@ -47,14 +49,29 @@ public void onHungerLoss(FoodLevelChangeEvent e) {
...
+        }
+    }
+    
+    @EventHandler
+    public void onHungerDamage(EntityDamageEvent e) {
+        if (cooler == null || cooler.isDisabled()) {
+            return;
+        }
+        
+        if (e.getCause() == EntityDamageEvent.DamageCause.STARVATION) {
```
TheBusyBiscuit, line 62: You can remove the "EntityDamageEvent" here, it's already imported.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CoolerListener.java
```diff
@@ -47,14 +49,29 @@ public void onHungerLoss(FoodLevelChangeEvent e) {
...
+            checkAndConsume((Player) e.getEntity());
```
TheBusyBiscuit, line 63: You will need to check whether the Entity is actually a Player first before casting.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
```diff
@@ -31,6 +31,11 @@ public ItemUseHandler getItemHandler() {
                 if (p.getInventory().getItemInMainHand().getType() != Material.SHEARS && p.getGameMode() != GameMode.CREATIVE) {
                     FoodLevelChangeEvent event = new FoodLevelChangeEvent(p, p.getFoodLevel() - 2);
                     Bukkit.getPluginManager().callEvent(event);
+                    
+                    if (event.isCancelled()) {
+                        SlimefunPlugin.getLocalization().sendMessage(p, "messages.wind-staff-fail", true);
+                        return;
+                    }
```
TheBusyBiscuit, line 38: I don't think we should send a message here, if the FoodLevelChangeEvent was cancelled, then we should simply not consume any hunger points. The wind staff itself did not fail, if the event was cancelled, this is most likely because that server doesn't use the hunger system...

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
LinoxGH: Yeah but still we need to feedback to the player as this is not lag or sth.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CoolerListener.java
```diff
@@ -19,10 +21,11 @@
 import me.mrCookieSlime.Slimefun.api.Slimefun;
 
 /**
- * This {@link Listener} listens for a {@link FoodLevelChangeEvent} and consumes a {@link Juice}
- * from any {@link Cooler} that can be found in the {@link Inventory} of the given {@link Player}.
+ * This {@link Listener} listens for a {@link FoodLevelChangeEvent} and a {@link EntityDamageEvent} for starvation damage
```
TheBusyBiscuit, line 24: * "or an", not "and a"

TheBusyBiscuit (changes requested):
One last typo

TheBusyBiscuit (approved):

PR #2272, opened 2020-08-31
Added /sf charge

NCBPFluffyBear:
## Description
This command charges the item that the sender is holding to max. It is intended to be used for admins on servers or testing items.

## Changes
Added ChargeCommand subcommand class, added command description, a success message, and a fail message to messages_en.yml, and added command to plugin.yml.

## Related Issues
N/A

## Checklist

*This PR only adds a small change, so the entire thing is essentially a Unit Test.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/ChargeCommand.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/ChargeCommand.java
@@ -1,0 +1,44 @@
+package io.github.thebusybiscuit.slimefun4.core.commands.subcommands;
+
+import io.github.thebusybiscuit.slimefun4.core.attributes.Rechargeable;
+import io.github.thebusybiscuit.slimefun4.core.commands.SlimefunCommand;
+import io.github.thebusybiscuit.slimefun4.core.commands.SubCommand;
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/ChargeCommand.java
```diff
@@ -0,0 +1,44 @@
...
+
+    protected String getDescription() {
+        return "commands.charge.description";
+    }
+
+    @Override
+    public void onExecute(CommandSender sender, String[] args) {
+        if (sender instanceof Player) {
+            if (sender.hasPermission("slimefun.charge.command")) {
+                Player p = ((Player) sender).getPlayer();
```
TheBusyBiscuit, line 26: Player#getPlayer() is very redundant, you already cast sender to a Player.

TheBusyBiscuit (changes requested):
Lots of formatting issues. And quite a few inconsistencies/redundancies that should be adressed.
Also missing documentation and author attribution.

This command is also quite small and should definitely be unit-tested!

TheBusyBiscuit (changes requested):
All issues I have adressed before were fixed.
However like I pointed out in my last review: There should definitely be a Unit Test added for this.

PR #2285, opened 2020-09-04
Multitool offhand hunger consumption fix

NCBPFluffyBear:
## Description
There was a bug that caused the multitool to consume hunger when a player used it on windstaff mode while having it in their offhand. This fix prevents that from happening.

## Changes
I added a check for if the multitool was in the player's offhand in WindStaff.java

## Related Issues
Mentioned on discord.

## Checklist


TheBusyBiscuit:
Do not check anything on the checkboxes if it isn't true...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
```diff
@@ -28,7 +28,7 @@ public ItemUseHandler getItemHandler() {
             Player p = e.getPlayer();
 
             if (p.getFoodLevel() >= 2) {
-                if (p.getInventory().getItemInMainHand().getType() != Material.SHEARS && p.getGameMode() != GameMode.CREATIVE) {
+                if (p.getInventory().getItemInMainHand().getType() != Material.SHEARS && p.getInventory().getItemInOffHand().getType() != Material.SHEARS && p.getGameMode() != GameMode.CREATIVE) {
```
TheBusyBiscuit, line 31: I know this has been here before but this can easily exploited by holding normal shears or even a multi tool in your off hand and then using a normal Wind staff.
It would probably be better to check if a wind staff is held in any of the two hands, you can use `isItem(...)` for this.

TheBusyBiscuit (changes requested):

NCBPFluffyBear:
Did proposed changes, but when testing I found that this causes the player to loose 4 hunger instead of 2 because the multi tool is no longer caught. I added an additional if statement instead that cancels the multitool if there is a wind staff in the other hand.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/WindStaff.java
```diff
@@ -28,7 +30,7 @@ public ItemUseHandler getItemHandler() {
             Player p = e.getPlayer();
 
             if (p.getFoodLevel() >= 2) {
-                if (p.getInventory().getItemInMainHand().getType() != Material.SHEARS && p.getInventory().getItemInOffHand().getType() != Material.SHEARS && p.getGameMode() != GameMode.CREATIVE) {
+                if ((SlimefunUtils.isItemSimilar(p.getInventory().getItemInMainHand(), SlimefunItems.STAFF_WIND, false, false) || SlimefunUtils.isItemSimilar(p.getInventory().getItemInOffHand(), SlimefunItems.STAFF_WIND, false, false)) && p.getGameMode() != GameMode.CREATIVE) {
```
TheBusyBiscuit, line 33: Why... I told you to just use `isItem(...)`

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -67,8 +69,11 @@ protected ItemUseHandler getItemUseHandler() {
             int index = selectedMode.getOrDefault(p.getUniqueId(), 0);
 
             if (!p.isSneaking()) {
+                SlimefunItem sfItem = modes.get(index).getItem();
+                if (sfItem.getID().equals(SlimefunItems.STAFF_WIND.getItemId()) && (SlimefunUtils.isItemSimilar(p.getInventory().getItemInMainHand(), SlimefunItems.STAFF_WIND, false, false) || SlimefunUtils.isItemSimilar(p.getInventory().getItemInOffHand(), SlimefunItems.STAFF_WIND, false, false))) {
+                    return;
+                }
```
TheBusyBiscuit, line 75: Also: Why did you add this? It doesn't make any sense to add that here

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
NCBPFluffyBear: > but when testing I found that this causes the player to loose 4 hunger instead of 2 because the multi tool is no longer caught.

TheBusyBiscuit:
So I have been looking at this for quite some time and something still feels off.
I think I found what's so weird about this:

The MultiTool calls the right click handler of the currently selected tool using this line of code:
> sfItem.callItemHandler(ItemUseHandler.class, handler -> handler.onRightClick(e));

As you can see the original PlayerRightClickEvent is passed onto the handler.

So that means, we wouldn't even need to check whether the main-hand or the off-hand is a Wind Staff, the Event itself stores the item that was used and since that hasn't been altered it still stores the multi tool.
https://github.com/TheBusyBiscuit/Slimefun4/pull/2285/files#diff-40c01e80feeba895d77f3128ccc775d1R34
This can all basically be reduced to one simple call: `isItem(e.getItem())`.

About that special case for the MultiTool I am still not exactly sure why you added it. Under what circumstances did that happen to you?
https://github.com/TheBusyBiscuit/Slimefun4/pull/2285/files#diff-f158efcc1d11c7229a2a6c692d691ed9R73

NCBPFluffyBear:
Nice, with this replacement the special case is no longer relevant.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -66,13 +66,10 @@ protected ItemUseHandler getItemUseHandler() {
...
-                    SlimefunItem sfItem = modes.get(index).getItem();
-
-                    if (sfItem != null) {
-                        sfItem.callItemHandler(ItemUseHandler.class, handler -> handler.onRightClick(e));
-                    }
+            if (!p.isSneaking() && removeItemCharge(item, COST)) {
+                SlimefunItem sfItem = modes.get(index).getItem();
+                if (sfItem != null) {
+                    sfItem.callItemHandler(ItemUseHandler.class, handler -> handler.onRightClick(e));
                 }
```
TheBusyBiscuit, line 73: Now you changed the logic of this...
`removeItemCharge` needs to be in a seperate if, otherwise the mode-changing else block will run if you are out of energy.

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):

PR #2290, opened 2020-09-05
Performance improvements on TickerTask <Tested>

LinoxGH:
## Description

A slimefun tick is now processed in seperate game ticks. This way lag spikes will be way less common but instead the lag will we spread out to several ticks/seconds.

## Changes

Many improvements in TickerTask.

## Related Issues

N/A

## Checklist


CHANGELOG.md
```diff
@@ -62,6 +62,7 @@
 * Small performance improvements for radiation
 * Small performance improvements for Auto Disenchanters
 * Magnesium Salt in Magnesium-Salt generators now lasts longer
+* Performance improvements for TickerTask(Slimefun ticking)
```
TheBusyBiscuit, line 65: Please remove this from the change log.
There is no way such a drastic change, especially when it's about threading and concurrency... will make it into the next RC.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -57,66 +74,169 @@
...
+                () -> this.tick(syncTasks)
+        );
+
+        // Process the synchronous ticker tasks on the main thread.
+        processSyncTasks(syncTasks);
     }
 
-    @Override
-    public void run() {
+    private void processSyncTasks(final @Nonnull BlockingQueue<Runnable> tasks) {
```
TheBusyBiscuit, line 110: No need to mark this parameter as final really.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -57,66 +74,169 @@
...
+
         try {
-            // If this method is actually still running... DON'T
-            if (running) {
-                return;
+            while (endNs > System.nanoTime()) {
+
+                // Don't wait on an empty queue for too long.
+                long s = System.nanoTime();
+                Runnable r = tasks.poll(250_000L, TimeUnit.NANOSECONDS);
```
TheBusyBiscuit, line 124: This magic number should definitely be a static final constant for easier tweaking.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -57,66 +74,169 @@
...
+        long endNs = System.nanoTime() + 1_500_000L;
+
         try {
-            // If this method is actually still running... DON'T
-            if (running) {
-                return;
+            while (endNs > System.nanoTime()) {
+
+                // Don't wait on an empty queue for too long.
+                long s = System.nanoTime();
```
TheBusyBiscuit, line 123: Please give your variable a more meaningful name.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
TheBusyBiscuit, line 124: Please give your variable a more meaningful name.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -136,7 +256,9 @@ private void tickChunk(@Nonnull Set<BlockTicker> tickers, @Nonnull String chunk)
         }
     }
 
-    private void tickLocation(@Nonnull Set<BlockTicker> tickers, @Nonnull Location l) {
+    private void tickLocation(@Nonnull List<BlockTicker> tickers, @Nonnull Location l, @Nonnull BlockingQueue<Runnable> syncTasks) {
+
+
```
TheBusyBiscuit, line 261: What's with the two empty lines here...?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -215,12 +341,28 @@ public void halt() {
 
     @ParametersAreNonnullByDefault
     public void queueMove(Location from, Location to) {
-        movingQueue.put(from, to);
+        // This collection is iterated over in a different thread. Need to lock it.
+        new Throwable().printStackTrace();
```
TheBusyBiscuit, line 345: Why are you printing a stacktrace here...?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -215,12 +341,28 @@ public void halt() {
...
+        finally {
+            queueLock.readLock().unlock();
+        }
     }
 
     @ParametersAreNonnullByDefault
     public void queueDelete(Location l, boolean destroy) {
-        deletionQueue.put(l, destroy);
+        // This collection is iterated over in a different thread. Need to lock it.
+        new Throwable().printStackTrace();
```
TheBusyBiscuit, line 358: Why are you printing a stacktrace here...?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -215,12 +341,28 @@ public void halt() {
...
+        queueLock.readLock().lock();
+        try {
+            deletionQueue.put(l, destroy);
+        }
+        finally {
+            queueLock.readLock().unlock();
```
TheBusyBiscuit, line 364: Also is it really sensible to re-use the same lock for moving and deleting operations?

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
LinoxGH: Oops that's from debugging

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
LinoxGH: This is from debugging. I just forgot to remove it

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
LinoxGH: Traces from removes debug messages

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
LinoxGH: It is needed. We use it in a lambda which requires it as final.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
LinoxGH: Since we lock and then unlock in finally blocks the lock won't be used by both of them at the same time.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
TheBusyBiscuit: Ah okay, can you at least put the annotation before the final then?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -92,6 +92,10 @@ public final void run() {
         if (!Bukkit.isPrimaryThread()) {
             return;
         }
+        
+        if (!SlimefunPlugin.isEnabled()) {
```
poma123, line 96:
```suggestion
        if (!SlimefunPlugin.instance().isEnabled()) {
```

poma123 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
TheBusyBiscuit: instance() is nullable. So a static isEnabled() method might be a good idea, check if it isnt null and then do the normal isEnabled call.

LinoxGH: instance will never be null, because before we set it to null, we cancel tasks anyway.

TheBusyBiscuit: Cancelling will only cancel the scheduling of new tasks, it won't terminate tasks which have already been started. And since this is running asynchronously, it may run just in time when instance is null and tasks have already been cancelled.

LinoxGH: > Cancelling will only cancel the scheduling of new tasks, it won't terminate tasks which have already been started. And since this is running asynchronously, it may run just in time when instance is null and tasks have already been cancelled.

This isn't run asynchronously.

LinoxGH:
Can someone test if this still functions properly?

poma123:
> Can someone test if this still functions properly?

Yep, maybe tomorrow.

poma123:
Just tested it ingame, works so far.

TheBusyBiscuit:
Can you test if `/sf timings` still works correctly, that's my biggest concern right now.

poma123:
Almost the same timings as yesterday.
image

TheBusyBiscuit:
I was more concerned about it breaking the profiler itself rather than the actual timings.
But thanks for testing it out!

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -333,8 +424,8 @@ public void enableTicker(@Nonnull Location l) {
     public void disableTicker(@Nonnull Location l) {
         Validate.notNull(l, "Location cannot be null!");
 
-        ChunkPosition chunk = new ChunkPosition(l.getWorld(), l.getBlockX() >> 4, l.getBlockZ() >> 4);
-        Set<Location> locations = tickingLocations.get(chunk);
+        String chunk = l.getWorld().getName() + ";" + l.getChunk().getX() + ";" + l.getChunk().getZ();
+        Set<Location> locations = activeTickers.get(chunk);
```
TheBusyBiscuit, line 428: :NotLikeThis:

LinoxGH:
I saw that the pr doesn't build, so I decided that maybe I should fix it, and fixed it in my little free time.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
LinoxGH: I don't have time to non-fatal beauty changes, feel free to do it yourself though.

TheBusyBiscuit:
Alright, I have fixed everything up I think and it should now be back to the non-String and non-Regex handling it had before.
I haven't tested it though.

@WalshyDev I think this is ready for review now and we can finally start getting to this. However I would like to compile and test this first.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -68,76 +128,162 @@ public void start(@Nonnull SlimefunPlugin plugin) {
...
         try {
-            // If this method is actually still running... DON'T
-            if (running) {
-                return;
+            while (endNs > System.nanoTime() || !SlimefunPlugin.instance().isEnabled()) {
+                // Don't wait on an empty queue for too long.
+                Runnable task = syncTasks.poll(MAX_POLL_NS, TimeUnit.NANOSECONDS);
+
+                if (task == null) {
+                    if (SlimefunPlugin.instance().isEnabled()) {
```
J3fftw1, line 176: cant these if's be combined with &&

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
TheBusyBiscuit: No...

TheBusyBiscuit (approved):
Merging into seperate branch for further actions.

PR #2293, opened 2020-09-06
Added gold nugget to gold pan drops

Seggan:
Added gold nugget to gold pan drops with 1% chance of dropping

## Checklist


TheBusyBiscuit:
Gold Nuggets can already be obtained, that's what the Nether Gold Pan is for.

PR #2295, opened 2020-09-06
Fixes #2280.

### LINKED ISSUE #2280: Solar Generators not generating any power
## Description (REQUIRED)

Solar Generator (Basic) Advanced Solar Generator and Carbonado Solar Generator not generating any power
Energized Solar Generator only generates 64J during Daytime

Versions of plugins:

Above testing made on 03 Sept 2020.

The above gif should well detailed how to reproduce it.
gamerule doDaylightCycle set to false
gamerule doWeather set to false
and time locked at Noon

Solar Generators expected to produce the respective J as printed in the machine lore.

## Server Log

https://pastebin.com/ZkzkvEvB

## Error Reports

the said folder is empty

## Environment (REQUIRED)

 - Minecraft Version: 1.16
 - Slimefun Version: vDEV - 660
 - CS-CoreLib Version: vDEV - 90
(Screenshot see above)

LinoxGH:
## Description

Just removed the bit shift, and it seems to fix the issue.

## Changes

Changed the way how we get the chunk of solar panel.

## Related Issues

Fixes #2280 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/generators/SolarGenerator.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/generators/SolarGenerator.java
@@ -23,6 +23,7 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
  * {@link #getNightEnergy()}.
  * 
  * @author TheBusyBiscuit
+ * @author Linox
  * 
  * @see EnergyNet
  * @see EnergyNetProvider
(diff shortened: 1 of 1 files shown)


WalshyDev:
Unless some server software broke this then there is no reason it'd break due to this.

Did you test that with bitshift it returned false and isChunkLoaded(Chunk) returned true?

Can you also send your server software version.

LinoxGH:
> Unless some server software broke this then there is no reason it'd break due to this.
> 
> Did you test that with bitshift it returned false and isChunkLoaded(Chunk) returned true?
> 
> Can you also send your server software version.

No I didn't do such tests, and it was Paper 1.16.2

TheBusyBiscuit:
I am closing this now.
In the issue you referenced the generated energy output was not zero, it was effectively the night-time value, as shown via the energized solar generator.

If the `isChunkLoaded` check fails, it will always return a zero. So we can be certain that whatever causes this issue must happen here:

`return isDaytime ? getDayEnergy() : getNightEnergy();`

So the issue effectively lies in the method `isDaytime()` which this PR does not even touch. Furthermore any inconsistency between the method with numeric and Chunk arguments would be the fault of the Server implementation used and not our problem to fix.

PR #2298, opened 2020-09-07
Implemented Tags

TheBusyBiscuit:
## Description

This Pull Request proposes a new mechanic to check if a Material is valid for a certain use case.
We previously mostly relied on MaterialCollections which is just a pain to maintain.

This PR gives us the ability to fine-tune what materials are allowed where and combine tags into larger ones.
The file format follows Minecraft's tag format for datapacks: https://minecraft.gamepedia.com/Tag#JSON_format
Most of them have a Slimefun-specific use case but there are also some tags (like ores, concrete powders or torches) which are just needed in Minecraft pretty badly anyway.

This also allows us to define "optional" elements without any ternary operators or ugly conditions. If one would want to add a new block to something, it will also be much easier to maintain and more beginner-friendly.

Unit Tests and docs were added accordingly.
There is also a "cyclic" check for tags in Unit Tests

## Checklist


WalshyDev:
Before I do a review on this I need to ask, are you only creating ones which don't exist in base MC? If not (which I think is the right path), are you generating these or is it manual?

I think there should definitely be some automation on this, we can generate them with the server jar super easily so all we'd need is GH action to push here.

TheBusyBiscuit:
These are only tags which do NOT exist in Minecraft itself and I simply wrote a short script to convert the old MaterialCollections into this format.

But I don't see much point in automating them, most of them are very very small and should remain very much static or are simply aggregations of existing tags. This should really not change frequently at all.

WalshyDev:
I think we should really be consistent with a tag usage. If we're adding a bunch of our own then I think we should also just include the MC ones. Since this is in tag format we can simply just check for a new MC build, gen tags and commit them. It's super simple.

With them all in one it means we can just have one unified way to identify blocks and it can be easier for new devs to understand (rather than figuring out when to use one over the other).

Bukkit when I last checked didn't even have all the MC ones for some reason so that'd also be a positive.

Side note, if glass isn't included please make a glass one.

Sfiguz7 (approved):

LinoxGH (dismissed):
From what I could see there aren't any problems.

Sfiguz7 (approved):

LinoxGH:
A tag for here would be nice I think.

poma123 (approved):
LGTM. Tested ingame, and can confirm it's fine.

There are still a few MaterialCollections here where we could use SlimefunTag instead.
https://github.com/Slimefun/Slimefun4/blob/tags/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ClimbingPick.java#L69-L79

TheBusyBiscuit:
@poma123 There is actually a `CLIMBING_PICK`-specific tag already. However each value has a modified assigned to it, so I decided to leave it as it is and worry about it later on

Sfiguz7 (approved):

poma123 (dismissed):

TheBusyBiscuit:
Can anyone try the unit tests locally? They are all passing for me but sonehow GitHub disagrees with that, can't seem to find out why

poma123:
Yep, will try it out.

poma123:
The same tests are failing for me as well. :/

TheBusyBiscuit:
@poma123 Perfect, can you do a coverage analysis and send a screenshot of the `ClimbingPick` and `TestClimbingPick` classes?
Because it all runs without failures for me, can't find it because of that.

poma123:
I've never done that before, where can I find it in IntelliJ? Or should I do that on Sonarcloud?

poma123:
Oh found it!

poma123:
Oh found it! But I couldn't make one on ClimbingPick.

TestClimbingPick.java:
image

TheBusyBiscuit:
@poma123 No, I mean a Coverage Analysis, preferabbly of the `ClimbingPick#climb()` method.
It should look like this:
63ae8bb70984a64b2b48544d85c237c5

As you can see, the code in green was executed, the code in red was not.
That's how we could find out what went wrong.

Here. this might help you:
https://www.jetbrains.com/help/idea/running-test-with-coverage.html

poma123:
I mean, there are some colors on the left, but the code isn't highlighted.
image

TheBusyBiscuit:
That's what I was looking for, thanks. 
That's interesting though... so it fails at the if-statement...

poma123:
image

TheBusyBiscuit:
Thanks :heart: Could you maybe show me the `getClimbingSpeed(item, surface)` method?

poma123:
image

TheBusyBiscuit:
How is this all green then :NotLikeThis:
What about the normal `getClimbingSpeed()` method?

poma123:
Same
image

TheBusyBiscuit:
:NotLikeThis:

TheBusyBiscuit:
OH! I think I found it

TheBusyBiscuit:
oh come on...

poma123:
image
image
image

poma123:
Oh crap XD

TheBusyBiscuit:
```
0.6
0.6
0.6
1.0
1.0
1.0
```
These values are somehow logged correctly, yet `if (power > 0.05)` still returns false?
This is so cumbersome to debug since it works just fine for me.

poma123:
Added a debug for the power, and guess what...
image

TheBusyBiscuit:
But why is it zero? 🤔

poma123:
The speed is zero as well
image

TheBusyBiscuit:
Well, they're effectively the same. But still why are they zero for you and GitHub but not for me?

poma123:
That's the big question

TheBusyBiscuit:
Can you maybe sysout the surfaces map's contents?

poma123:
image

TheBusyBiscuit:
aaaaha, I seee

poma123:
Maybe this helps
image

TheBusyBiscuit:
Yeah I think I know what's wrong.

TheBusyBiscuit:
...

Well, from your image I could at least see that it's due to missing default tags since all the manual values seem to be there.
So my conclusion was that somehow due to Java version differences between you, me and GitHub, the order of execution might vary, so that the MethodSource providers get invoked before the `@BeforeAll` method, making the default tags go missing.
Thanks for providing all this info thus far though, now we just gotta find out where to register the default tags.

TheBusyBiscuit:
Hmmm, could you maybe check the `nameLookup` map in `SlimefunTag.java` when `getTag(String)` is called?

poma123:
Yep

poma123:
image

TheBusyBiscuit:
Well yeah that's not it.
I'm lost.

poma123:
Cookie... you won't believe this
image

poma123:
I added this line, and now all passing.
image

TheBusyBiscuit:
> I added this line, and now all passing. I have no idea why.
> Why do we reload these in the first place?
> image

OMG you are a genius.
Of course we'd need to reload the child tags too, I completely forgot about that XD

poma123:
They are empty otherwise right?

TheBusyBiscuit:
Yes

TheBusyBiscuit:
Still not quite understanding why all tests passed on my computer, but oh well, hopefully this is fixed now.
Thanks a lot @poma123 ❤️

poma123:
Wait till the build fails again XD Just kidding.

poma123:
Lol

TheBusyBiscuit:
XD Yeah that one was just a dumb typo though

poma123:
Yep. Hopefully all fixed now.

TheBusyBiscuit:
Hooray :tada:

TheBusyBiscuit:
@Slimefun/code-reviewers I would like to merge this relatively soon, can everyone please have a final look on this and submit a review?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -47,7 +47,7 @@
  */
 public class BlockPlacer extends SlimefunItem {
 
-    private final ItemSetting<List<String>> blacklist = new ItemSetting<>("unplaceable-blocks", MaterialCollections.getAllUnbreakableBlocks().stream().map(Material::name).collect(Collectors.toList()));
+    private final ItemSetting<List<String>> blacklist = new ItemSetting<>("unplaceable-blocks", SlimefunTag.UNBREAKABLE_MATERIALS.stream().map(Material::name).collect(Collectors.toList()));
```
LinoxGH, line 50: I think we should instead add a constructor accepting SlimefunTag instead.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ClimbableSurface.java
```diff
@@ -0,0 +1,51 @@
+package io.github.thebusybiscuit.slimefun4.implementation.items.tools;
```
LinoxGH, line 1: I'm not sure about the location of this class. We should have a package just for stuff like these.(There is also GrapplingHook class in listener package)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/PickaxeOfTheSeeker.java
```diff
@@ -55,7 +56,9 @@ private Block findClosestOre(Player p) {
         for (int x = -4; x <= 4; x++) {
             for (int y = -4; y <= 4; y++) {
                 for (int z = -4; z <= 4; z++) {
-                    if (MaterialCollections.getAllOres().contains(p.getLocation().add(x, y, z).getBlock().getType()) && (closest == null || p.getLocation().distanceSquared(closest.getLocation()) > p.getLocation().distanceSquared(p.getLocation().add(x, y, z)))) {
+                    Material type = p.getLocation().add(x, y, z).getBlock().getType();
```
LinoxGH, line 59: I think Block.getRelative() should have been used instead.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/PickaxeOfTheSeeker.java
```diff
@@ -55,7 +56,9 @@ private Block findClosestOre(Player p) {
...
+
+                    if (SlimefunTag.PICKAXE_OF_THE_SEEKER_BLOCKS.isTagged(type) && (closest == null || p.getLocation().distanceSquared(closest.getLocation()) > p.getLocation().distanceSquared(p.getLocation().add(x, y, z)))) {
```
LinoxGH, line 61: A bit of refactoring could be of use here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/SmeltersPickaxe.java
```diff
@@ -27,7 +27,7 @@ public SmeltersPickaxe(Category category, SlimefunItemStack item, RecipeType rec
     @Override
     public ToolUseHandler getItemHandler() {
         return (e, tool, fortune, drops) -> {
-            if (MaterialCollections.getAllOres().contains(e.getBlock().getType()) && !BlockStorage.hasBlockInfo(e.getBlock())) {
+            if (SlimefunTag.SMELTERS_PICKAXE_BLOCKS.isTagged(e.getBlock().getType()) && !BlockStorage.hasBlockInfo(e.getBlock())) {
```
LinoxGH, line 30: This could use some refactoring too.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
```diff
@@ -0,0 +1,312 @@
...
+
+    static {
+        for (SlimefunTag tag : valuesCache) {
+            nameLookup.put(tag.name(), tag);
+        }
+    }
+
+    private final NamespacedKey key;
+    private final Set<Material> includedMaterials = new HashSet<>();
+    private final Set<Tag<Material>> additionalTags = new HashSet<>();
```
LinoxGH, line 190: Isn't Tag an enum? So why don't you use EnumSet here?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
```diff
@@ -0,0 +1,312 @@
...
+        new TagParser(this).parse(this, (materials, tags) -> {
+            this.includedMaterials.clear();
+            this.includedMaterials.addAll(materials);
+
+            this.additionalTags.clear();
+            this.additionalTags.addAll(tags);
+        });
+    }
+
+    public static void reloadAll() throws TagMisconfigurationException {
```
LinoxGH, line 217: Javadocs here please.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public NamespacedKey getKey() {
+        return key;
+    }
+
+    @Override
+    public boolean isTagged(@Nonnull Material item) {
```
LinoxGH, line 230: Javadocs here too.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
```diff
@@ -0,0 +1,312 @@
...
+            }
+
+            // Now we can be sure it isn't tagged in any way
+            return false;
+        }
+    }
+
+    @Nonnull
+    @Override
+    public Set<Material> getValues() {
```
LinoxGH, line 248: This also could use some javadocs.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/TagParser.java
```diff
@@ -0,0 +1,199 @@
...
+
+import io.github.thebusybiscuit.slimefun4.api.exceptions.TagMisconfigurationException;
+import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
+import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
+
+/**
+ * The {@link TagParser} is responsible for parsing a JSON input into a {@link SlimefunTag}.
+ * 
+ * @author TheBusyBiscuit
+ *
```
LinoxGH, line 38: Maybe add some `@see` for SlimefunTag and etc. Just an idea.

LinoxGH (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
TheBusyBiscuit: Tag is not an enum.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
TheBusyBiscuit: It's an overridden method, it's just the implementation of the Tag#isTagged() method, the docs are inherited and would be redundant here.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
TheBusyBiscuit: It's an overridden method, it's just the implementation of the Tag#isTagged() method, the docs are inherited and would be redundant here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/SmeltersPickaxe.java
TheBusyBiscuit: Couldn't really think how this would be improved.

LinoxGH: > Couldn't really think how this would be improved.

I mean like separating into multiple lines instead of 1 very long line.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ClimbableSurface.java
TheBusyBiscuit: I relocated the Surface and talisman settings into a dedicated package. But I'll be leaving the grappling hook one for now until we found a suitable place for it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/SmeltersPickaxe.java
TheBusyBiscuit: No.

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/settings/MaterialTagSetting.java
```diff
@@ -0,0 +1,76 @@
...
+        return defaultTag;
+    }
+
+    @Override
+    public boolean validateInput(List<String> input) {
+        if (input != null) {
+            for (String value : input) {
+                Material material = Material.matchMaterial(value);
+
+                // This value is not a valid material, the setting is not valid.
```
LinoxGH, line 53:
```suggestion
                // This value is not a valid material, the setting is invalid.
```

TheBusyBiscuit:
Also blame @LinoxGH for this now being 89 changed files :P

*Nah, some were valid critics and things that should have happened sooner or later anyway*

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
```diff
@@ -12,6 +12,7 @@
 import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.api.items.ItemSetting;
+import io.github.thebusybiscuit.slimefun4.implementation.settings.TalismanEnchantment;
```
LinoxGH, line 15: This is imported but not used? Or github just doesn't feel like showing it?

poma123 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -30,14 +31,7 @@
...
 
-    private final ItemSetting<Integer> range = new ItemSetting<Integer>("explosion-range", 3) {
-
-        @Override
-        public boolean validateInput(Integer input) {
-            return super.validateInput(input) && input > 0;
-        }
-
-    };
+    private final ItemSetting<Integer> range = new IntRangeSetting("explosion-range", 1, 3, Integer.MAX_VALUE);
```
LinoxGH, line 34: Integer.MAX_VALUE should not be used here. We don't want some crazy server owner to make an explosive bow that launches players at the other side of the world.

src/main/resources/tags/sensitive_materials.json
```diff
@@ -0,0 +1,11 @@
+{
+	"values" : [
```
LinoxGH, line 2: Not sure but shouldn't buttons and redstone conponents be included too?

src/main/resources/tags/fluid_sensitive_materials.json
```diff
@@ -0,0 +1,7 @@
+{
+	"values" : [
```
LinoxGH, line 2: Not sure but shouldn't buttons and redstone conponents be included too? As well as cake, pressure plates...

LinoxGH (changes requested):
I think we should also use proper max values instead of Integer.MAX_VALUE just for fool-proofing.

LinoxGH:
Btw I think a customized set for Tag<Material> could be of use. And let us do some performance improvements.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
TheBusyBiscuit: It is used. It was just relocated, previously it was in the same package, thus not needing an import...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
TheBusyBiscuit: Their own fault tbh.

src/main/resources/tags/sensitive_materials.json
TheBusyBiscuit: There are lots of things that aren't included because they aren't used.

src/main/resources/tags/fluid_sensitive_materials.json
TheBusyBiscuit: Not really

TheBusyBiscuit:
What do you mean "customized Set"?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
LinoxGH: > Their own fault tbh.

Well get ready for bug reports of "explosive bow blows from 100k blocks away" then. Lol

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/settings/MaterialTagSetting.java
TheBusyBiscuit: That's pretty much the same thing, and since this is a "validate" method, not valid is more appropriate of a term here.

LinoxGH:
> What do you mean "customized Set"?

A class that operates as Set<Tag<Material>> but overrides Set class's methods to improve the performance for Tag<Material>. This is just an idea though.

TheBusyBiscuit:
Definitely far out of scope for this.
And implementing the Set interface is generally not recommended as the required methods may vary depending on the Java installation of the end user.

TheBusyBiscuit:
And about the max values: Of course Integer.MAX_VALUE is not a reasonabl limit, but the point of validation is not to filter what's reasonable.
The point of this validation is to filter out illegal and invalid values.
Setting a hardcoded limit would be far too subjective, we can only objectively determine what inputs are valid programming-wise.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
TheBusyBiscuit: That would not be a bug though, if that happens then that's very much inline with what they configured.
If you throw a brick at a window, don't be surprised the window breaks...

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
```diff
@@ -0,0 +1,325 @@
...
+/**
+ * This enum contains various implementations of the {@link Tag} interface.
+ * Most of them serve some purpose within Slimefun's implementation, some are just pure
+ * extensions of the default Minecraft tags.
+ * The actual tag files are located in the {@literal /src/main/resources/tags} directory
+ * and follow Minecraft's tags.json format.
+ * 
+ * @author TheBusyBiscuit
+ * 
+ * @see TagParser
```
LinoxGH, line 40:
```suggestion
 * @see TagParser
 * @see Material
```

LinoxGH (approved):
Just another suggestion.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
TheBusyBiscuit: That doesn't really make much sense here, we generally use `@see` to link relevant classes in the same project.
As the main intent behind that parameter is to give additional context to the code to better understand how it plays into the bigger context of the project.

LinoxGH: > That doesn't really make much sense here, we generally use `@see` to link relevant classes in the same project.
> As the main intent behind that parameter is to give additional context to the code to better understand how it plays into the bigger context of the project.

Exactly. SlimefunTag's center is Material's anyway.

TheBusyBiscuit: Material is from Bukkit though, not from Slimefun,...

poma123:
_Perfection_
image

TheBusyBiscuit:
I think I am gonna merge this now.

TheBusyBiscuit:
I just tested it all again and found a small portion of the tags to only exist on 1.14+, hence why they have now been marked as not-required.

Also improved the error message to include the tag's name.
But yeah, the current version is tested to be working just fine and I'm gonna merge this now.
Sorry @poma123 for making this 70 commits now XD

poma123:
> Sorry @poma123 for making this 70 commits now XD

NOOOOO XD

PR #2304, opened 2020-09-09
MachineProcessCompleteEvent

poma123:
## Description

Added MachineProcessCompleteEvent, which fires whenever an AContainer has completed its process. This was the best name I could came up with. If you have a better idea, then tell me.

I can add a Unit Test, if needed.

## Changes

- Added MachineProcessCompleteEvent.class
- Fixed a small typo in GEOResourceGenerationEvent.class
- Added the #callEvent() method to AContainer.class
- Finally a minor impovement in AContainer.class

## Related Issues

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MachineProcessCompleteEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MachineProcessCompleteEvent.java
@@ -23,7 +23,7 @@ public class MachineProcessCompleteEvent extends Event {
     private final MachineRecipe machineRecipe;
 
     @ParametersAreNonnullByDefault
-    public MachineProcessCompleteEvent(@Nonnull Block block, @Nonnull MachineRecipe machineRecipe) {
+    public MachineProcessCompleteEvent(Block block, MachineRecipe machineRecipe) {
         this.block = block;
         this.machineRecipe = machineRecipe;
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MachineProcessCompleteEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MachineProcessCompleteEvent.java
@@ -6,6 +6,7 @@ import org.bukkit.block.Block;
 import org.bukkit.event.Event;
 import org.bukkit.event.HandlerList;
 
+import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 /**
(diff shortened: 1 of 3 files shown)


TheBusyBiscuit:
Oh hmmm, I am not too sure whether the Event API would be the most suitable for this. To be honest the entire logic behind machine processing is long overdue for a rewrite anyway and we could probably implement this in a much better way, as firing this many events can create a bit of a performance disadvantage.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MachineProcessCompleteEvent.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ */
+public class MachineProcessCompleteEvent extends Event {
+
+    private static final HandlerList handlerList = new HandlerList();
+
+    private final Block block;
+    private final MachineRecipe machineRecipe;
+
+    @ParametersAreNonnullByDefault
```
WalshyDev, line 25: You already tag them as Nonnull...

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MachineProcessCompleteEvent.java
poma123: Reee

TheBusyBiscuit: Reeeeee

LinoxGH:
> Oh hmmm, I am not too sure whether the Event API would be the most suitable for this. To be honest the entire logic behind machine processing is long overdue for a rewrite anyway and we could probably implement this in a much better way, as firing this many events can create a bit of a performance disadvantage.

Agreed with the rewrite. But event api wouldn't be a problem, I don't think. Bukkit runs hundreds of events consistently without a problem.

TheBusyBiscuit:
> Bukkit runs hundreds of events consistently without a problem.

Just because they are doing so doesn't mean it is a good idea :LUL:
Even the creation of new objects, like the event objects is already a drawback in terms of memory/performance that doesn't **need** to happen. It won't make such an enormous difference but the problem certainly scales and Slimefun is already pretty packed with features and events.

So I really think a rewrite (with such possibility to tap into the outcome) would be the best approach here.
I may probably still merge it for the time being though until that rewrite has happened.

TheBusyBiscuit (approved):

PR #2306, opened 2020-09-10
Made MachineProcessCompleteEvent async

poma123:
## Description

Made the event async, because it is called asynchronously. I didn't notice this tbh, but here I am with the fix.

## Changes

- Made MachineProcessCompleteEvent async
- Renamed to AsyncMachineProcessCompleteEvent

## Related Issues

Fixed that this sync event was called asynchronously, so I made it async.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
@@ -15,7 +15,7 @@ import javax.annotation.ParametersAreNonnullByDefault;
  * @author poma123
  *
  */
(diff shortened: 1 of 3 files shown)


WalshyDev:
Method should really also be renamed to AsyncMachineProcessCompleteEvent to be consistent with Bukkit

LinoxGH:
Now that the event is asynchronous. Is it really safe for it to have a Block inside? Block's aren't thread-safe, and this may cause issues.

WalshyDev:
Now it's been renamed I'm ok with it.

poma123:
> I'd mark the getBlock method as not thread-safe though. @ NotThreadSafe

It's only-class, so I leave it then.

LinoxGH:
You can add a javadocs explaining it nonetheless.

TheBusyBiscuit (approved):

PR #2312, opened 2020-09-12
Fixed colors & functionality of cheat sheet

TheSilentPro:
## Description
Fixed non-colored string in Cheat Sheet lore
Fixed Shift + Right Click on Cheat Sheet not opening Settings Menu instead 

## Changes
Added missing coloring to lore in SimefunGuideItem
Added check in SlimefunGuideListener for sneaking in order to open Settings Menu

## Related Issues
N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/SlimefunGuideListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/SlimefunGuideListener.java
@@ -62,9 +62,14 @@ public class SlimefunGuideListener implements Listener {
             }
         }
         else if (openGuide(e, SlimefunGuideLayout.CHEAT_SHEET) == Result.ALLOW) {
-            // We rather just run the command here,
-            // all necessary permission checks will be handled there.
-            p.chat("/sf cheat");
+            if (p.isSneaking()) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/SlimefunGuideListener.java
```diff
@@ -62,9 +62,14 @@ else if (openGuide(e, SlimefunGuideLayout.CHEST) == Result.ALLOW) {
             }
         }
         else if (openGuide(e, SlimefunGuideLayout.CHEAT_SHEET) == Result.ALLOW) {
-            // We rather just run the command here,
-            // all necessary permission checks will be handled there.
-            p.chat("/sf cheat");
+            if (!p.isSneaking()) {
```
TheBusyBiscuit, line 65: In all other variants is `p.isSneaking()` not negated.
I think we should keep it uniform and swap it around here too to avoid confusion here.

TheBusyBiscuit (changes requested):

poma123 (approved):
Looking good.

Sfiguz7 (approved):

TheBusyBiscuit (approved):

PR #2323, opened 2020-09-13
Fixing #2320

### LINKED ISSUE #2320: Smelteries dont work with soul fire

When lighting up a fire on soul soil, it will turn ito soul fire. Smelteries don't recognize this fire, thus not smelting stuff.

Set up a the multiblock smeltery, put a block of soul soil under the block where the fire is going to be placed on. Then light up the fire, it should turn blue. Now try to smelt anything in the smeltery, for example dust. It wont even give you an error message, as it doesnt recognize the fire below.
https://imgur.com/lAHafPW

The smeltery should smelt stuff, as there is a form of fire below it.

 - Server Software (Spigot/Paper): Paper v1_16_R2
 - Minecraft Version: 1.16.2
 - Slimefun Version: Slimefun vDev - 674
 - CS-CoreLib Version: CS-Corelib vDev - 674
https://screenshots.skizzles.me/JIWA4/bugUsugo23.png

LinoxGH:
## Description

Made smeltery also accept blue fire.

## Changes

Added a new supported tag, Tag.FIRE.

## Related Issues

Fixes #2320 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlock.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlock.java
@@ -45,6 +45,9 @@ public class MultiBlock {
         if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_14)) {
             SUPPORTED_TAGS.add(Tag.WOODEN_FENCES);
         }
+        if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_16)) {
+            SUPPORTED_TAGS.add(Tag.FIRE);
+        }
     }
 
     @Nonnull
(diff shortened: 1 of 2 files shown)


TheBusyBiscuit (changes requested):
I very much doubt that the FIRE tag even existed before 1.16, I mean why would you even have a fire tag if there is only one Type of fire.
So this change would make it break on 1.13-1.15

I reckon it would be better to pr to the tags branch and add a custom tag with soul fire being optional.

WalshyDev:
> I very much doubt that the FIRE tag even existed before 1.16, I mean why would you even have a fire tag if there is only one Type of fire.

It does not. He does a isAtLeast(1.16) though. Should be fine no?

LinoxGH:
That is correct that's why I added an isAtLeast check.

TheBusyBiscuit (approved):
Maybe I should look into getting glasses... or doing an eyesight test in general. My last one was for my driver's license...

poma123 (approved):

WalshyDev (approved):
LGTM

PR #2328, opened 2020-09-15
Programmable Android Optimization: Infinite Block Generators

TheBusyBiscuit:
## Description

This Pull Request proposes an optimization for Androids which shall reduce block updates and lag.
Instead of continously setting the block to AIR, it will simply play particles and sound effects as if the block had just been "generated". This also calls a `BlockFromToEvent`, so that custom cobble generator plugins won't be disturbed by this.

I will probably add Unit Tests for the various Generators too ~~(or make a Hacktoberfest issue for it ;D )~~

The generator detection will probably not be 100% accurate but it doesn't have to be, this is only to reduce lag for servers who use this excessively. And people with massive cobblestone farms are likely to use a pretty basic set up anyway.

here is a showcase of the detection in action:
https://youtu.be/oe7tu5WyUuw

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/MinerAndroid.java
```diff
@@ -106,4 +99,23 @@ protected void moveAndDig(Block b, BlockMenu menu, BlockFace face, Block block)
         }
     }
 
+    @ParametersAreNonnullByDefault
+    private void breakBlock(BlockMenu menu, Collection<ItemStack> drops, Block block) {
+        // Push our drops to the inventory
+        for (ItemStack drop : drops) {
```
Sfiguz7, line 105: Are we not checking if drops fit anymore? Old code had a if (menu.fits(drop, getOutputSlots())) check, I don't see why we would skip it

TheBusyBiscuit: I intentionally removed it because it seemed redundant. When we push items that dont fit, nothing really happens.

So checking if items don't and then doing nothing instead kinda leads to the same outcome anyway. That's why I removed it.

Sfiguz7 (commented):
Check my comment - also, should valid generator Unit Tests check if the faces given as arguments are the same? I'm not an expert with Tests so I don't know for sure, if the faces were indeed the same the generator wouldn't of course be valid so unless implicitly checked we should test that too

TheBusyBiscuit:
@Sfiguz7 What do you mean with checking faces?

src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/utils/TestInfiniteBlockGenerators.java
```diff
@@ -0,0 +1,162 @@
...
+
+        block.setType(Material.COBBLESTONE);
+        Assertions.assertFalse(InfiniteBlockGenerator.COBBLESTONE_GENERATOR.test(block));
+        Assertions.assertNull(InfiniteBlockGenerator.findAt(block, false));
+    }
+
+    @ParameterizedTest
+    @MethodSource(value = "provideFaces")
+    @DisplayName("Test if a Cobblestone Generator can be detected")
+    void testValidCobblestoneGenerator(BlockFace water, BlockFace lava) {
```
Sfiguz7, line 60: Can't seem to link to this line so I'm just commenting, these two arguments. I'm not sure how Unit Tests handle args but if water and lava refer to the same BlockFace you get no ValidCobblestoneGenerator

TheBusyBiscuit: There is a not-equals check for the arguments. Look at the annotations, they point to "provideFaces", a method found further down. There is a .filter() for this exact case.

Sfiguz7 (dismissed):
Had missed the filter, looks all good to me then

TheBusyBiscuit:
Here is a video showcase of the new vs. old behaviour...
Sorry for the poor audio quality, it's around midnight, I am super tired and not used to talking in videos at all...
https://youtu.be/wlnwrrAict0

LinoxGH (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockPhysicsListener.java
```diff
@@ -84,6 +84,7 @@ public void onPistonRetract(BlockPistonRetractEvent e) {
     @EventHandler(ignoreCancelled = true)
     public void onLiquidFlow(BlockFromToEvent e) {
         Block block = e.getToBlock();
+        Material type = block.getType();
```
svr333, line 87: variable is never used

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockPhysicsListener.java
```diff
@@ -84,6 +84,7 @@ public void onPistonRetract(BlockPistonRetractEvent e) {
...
 
         // Check if this Material can be destroyed by fluids
         if (SlimefunTag.FLUID_SENSITIVE_MATERIALS.isTagged(block.getType())) {
```
svr333, line 90: or you forgot to replace block.getType() here, either one works

src/main/java/io/github/thebusybiscuit/slimefun4/utils/InfiniteBlockGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+     * Similar to {@link #test(Block)} this tests whether this {@link InfiniteBlockGenerator}
+     * exists at the given {@link Block}.
+     * 
+     * @param b
+     *            The {@link Block}
+     * 
+     * @return Whether this {@link InfiniteBlockGenerator} exists at the given {@link Block}
+     */
+    @Override
+    public boolean test(@Nonnull Block b) {
```
svr333, line 77: naming wise I feel like `InfiniteBlockGenerator#checkForAt(Block b)` would fit better than `InfiniteBlockGenerator#test(Block b)`

src/main/java/io/github/thebusybiscuit/slimefun4/utils/InfiniteBlockGenerator.java
```diff
@@ -0,0 +1,177 @@
...
+     * 
+     * @return Our called {@link BlockFormEvent}
+     */
+    @Nonnull
+    public BlockFormEvent callEvent(@Nonnull Block block) {
+        BlockState state = PaperLib.getBlockState(block, false).getState();
+        BlockFormEvent event = new BlockFormEvent(block, state);
+        Bukkit.getPluginManager().callEvent(event);
+        return event;
+    }
```
svr333, line 154: Public method so Validate#notNull here? Or am I still getting this wrong xD

src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/utils/TestInfiniteBlockGenerators.java
```diff
@@ -0,0 +1,168 @@
...
+
+        Assertions.assertEquals(generator, InfiniteBlockGenerator.findAt(block));
+    }
+
+    private static Stream<Arguments> provideFaces() {
+        BlockFace[] faces = { BlockFace.NORTH, BlockFace.EAST, BlockFace.SOUTH, BlockFace.WEST };
+        Stream<BlockFace> stream = Arrays.stream(faces);
+        return stream.flatMap(a -> Arrays.stream(faces).filter(b -> a != b).map(b -> Arguments.of(a, b)));
+    }
+
```
svr333, line 167: extra empty line

svr333 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockPhysicsListener.java
TheBusyBiscuit:
```suggestion
        if (SlimefunTag.FLUID_SENSITIVE_MATERIALS.isTagged(type)) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/InfiniteBlockGenerator.java
TheBusyBiscuit: This method is overridden, as this is a Predicate implementation.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/InfiniteBlockGenerator.java
TheBusyBiscuit:
```suggestion
    public BlockFormEvent callEvent(@Nonnull Block block) {
        Validate.notNull(block, "The Block cannot be null!");
        BlockState state = PaperLib.getBlockState(block, false).getState();
        BlockFormEvent event = new BlockFormEvent(block, state);
        Bukkit.getPluginManager().callEvent(event);
        return event;
    }
```

src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/utils/TestInfiniteBlockGenerators.java
TheBusyBiscuit:
```suggestion
```

TheBusyBiscuit:
@svr333 Please make use of change suggestions XD

src/main/java/io/github/thebusybiscuit/slimefun4/utils/InfiniteBlockGenerator.java
```diff
@@ -0,0 +1,178 @@
...
+
+        for (InfiniteBlockGenerator generator : values) {
+            if (generator.test(b)) {
+                return generator;
+            }
+        }
+
+        return null;
+    }
+
```
svr333, line 177:
```suggestion

```

svr333 (changes requested):

svr333 (dismissed):

svr333:
I broke it by deleting a whiteline

svr333 (approved):
LGTM

svr333 (approved):

Sfiguz7 (approved):
This has been open for so long already? Damn. Sorry for taking a bit to check it again

PR #2331, opened 2020-09-17
Added my current github account to aliases.

LinoxGH:
## Description

Added my current github acc as a contributor alias.

## Changes

Added a new contributor alias.

## Related Issues

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/github/ContributionsConnector.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/github/ContributionsConnector.java
@@ -28,6 +28,7 @@ class ContributionsConnector extends GitHubConnector {
         aliases.put("WalshyDev", "HumanRightsAct");
         aliases.put("J3fftw1", "_lagpc_");
         aliases.put("ajan-12", "ajan_12");
+        aliases.put("LinoxGH", "ajan_12");
         aliases.put("mrcoffee1026", "mr_coffee1026");
         aliases.put("Cyber-MC", "CyberPatriot");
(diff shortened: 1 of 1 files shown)


WalshyDev (approved):
LGTM

poma123 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/github/ContributionsConnector.java
```diff
@@ -28,6 +28,7 @@
         aliases.put("WalshyDev", "HumanRightsAct");
         aliases.put("J3fftw1", "_lagpc_");
         aliases.put("ajan-12", "ajan_12");
+        aliases.put("LinoxGH", "ajan_12");
```
TheBusyBiscuit, line 31: This is redundant since your alias for the GitHub account is already present here:
https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/localization/Translators.java#L153

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/github/ContributionsConnector.java
LinoxGH: Does it make the plugin use ajan_12 head on LinoxGH on Contributors page?

TheBusyBiscuit: Well it should since the code for aliases is the same.

PR #2334, opened 2020-09-19
Added AsyncReactorProcessCompleteEvent and AsyncGeneratorProcessCompleteEvent

poma123:
## Description

Added AsyncGeneratorProcessCompleteEvent which fires whenever an AGenerator has completed its process.
Added AsyncReactorProcessCompleteEvent which fires whenever a Reactor has completed its process.
Also, I changed AsyncMachineProcessCompleteEvent a little bit to be more cleaner.

This time I tested every change carefully.

## Changes

- Added AsyncGeneratorProcessCompleteEvent
- Added AsyncReactorProcessCompleteEvent
- Minor improvement in AsyncMachineProcessCompleteEvent

## Related Issues

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
@@ -1,7 +1,7 @@
 package io.github.thebusybiscuit.slimefun4.api.events;
 
 import javax.annotation.Nonnull;
-import javax.annotation.ParametersAreNonnullByDefault;
+import javax.annotation.Nullable;
 
 import org.bukkit.Location;
 import org.bukkit.event.Event;
@@ -25,8 +25,7 @@ public class AsyncMachineProcessCompleteEvent extends Event {
     private final AContainer container;
     private final MachineRecipe machineRecipe;
 
-    @ParametersAreNonnullByDefault
-    public AsyncMachineProcessCompleteEvent(Location l, AContainer container, MachineRecipe machineRecipe) {
+    public AsyncMachineProcessCompleteEvent(@Nonnull Location l, @Nullable AContainer container, @Nullable MachineRecipe machineRecipe) {
         super(true);
 
         this.location = l;
@@ -49,7 +48,7 @@ public class AsyncMachineProcessCompleteEvent extends Event {
      *
      * @return The {@link SlimefunItem} instance of the machine
      */
-    @Nonnull
+    @Nullable
     public AContainer getMachine() {
         return container;
     }
@@ -59,7 +58,7 @@ public class AsyncMachineProcessCompleteEvent extends Event {
      *
      * @return The {@link MachineRecipe} of the process
      */
-    @Nonnull
+    @Nullable
     public MachineRecipe getMachineRecipe() {
         return machineRecipe;
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
@@ -3,12 +3,13 @@ package io.github.thebusybiscuit.slimefun4.api.events;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import org.bukkit.block.Block;
+import org.bukkit.Location;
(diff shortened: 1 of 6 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
```diff
@@ -20,31 +21,43 @@
...
-     * This method returns the {@link Block} of the machine.
+     * This returns the {@link Location} of the machine.
      *
-     * @return the {@link Block} of the machine
+     * @return The {@link Location} of the machine
      */
     @Nonnull
-    public Block getMachine() {
-        return block;
+    public Location getLocation() {
```
LinoxGH, line 43: Why is this changed?

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncReactorProcessCompleteEvent.java
```diff
@@ -0,0 +1,77 @@
...
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineFuel;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
+
+/**
+ * This {@link Event} is fired whenever a {@link Reactor} has completed its process.
+ *
+ * @author poma123
+ *
+ */
+public class AsyncReactorProcessCompleteEvent extends Event {
```
LinoxGH, line 20: This should probably extend AsyncMachineProcessCompleteEvent

LinoxGH (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
poma123: Because I thought getting the SlimefunItem instance would be better than just the block. If you look a little lower, there is the new getMachine method.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncReactorProcessCompleteEvent.java
poma123: But Reactor doesn't extend AContainer. Also, Reactor has MachineFuel, AContainer has MachineRecipe.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
LinoxGH: I doubt anyone uses this event yet but in the future avoid changing existent api, at least @Deprecate them and add your new methods, but the old ones should at least work for some time.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncReactorProcessCompleteEvent.java
LinoxGH: I'm just saying all these 3 events are very similar, and doing a `event instanceof MachineEvent` or sth. will probably prove usedul to some people.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncMachineProcessCompleteEvent.java
poma123: I know that and agreed. If this wasn't a brand new event, I would've made the old method deprecated. I think I am the only one who uses this rn

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AsyncReactorProcessCompleteEvent.java
poma123: Hmm. Good idea though.

LinoxGH (approved):
Better, but I'd suggest expanding the api a bit while you're at it. Idk what you can do but I'm sure sth can be figured out. Though, it's good to merge as it is rn.

TheBusyBiscuit:
Yeah, like I said before: The item processing stuff is urging for a rewrite quite badly anyway, so this will all need to get overhauled eventually anyway. But for the time being this is okay (Haven't looked at the code yet though)

Sfiguz7 (approved):

PR #2347, opened 2020-09-26
Fixed CargoNetworkTask not checking for other slots before dropping i…

OmerBenGera:
…tems on ground

## Description

I added a check for other slots for the CargoNetworkTask before items are dropped. It was necessary as it conflicted with storage units from WildChests, as they always return the item of the storage for the same slot.

## Changes

Instead of dropping the items, I also added a check to add items to the source inventory before dropping them.

## Related Issues

None

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
@@ -111,7 +111,8 @@ class CargoNetworkTask implements Runnable {
                 if (inv.getItem(previousSlot) == null) {
                     inv.setItem(previousSlot, stack);
                 }
-                else {
+                // Making sure the item cannot be added to another slot before dropping it on ground
+                else if((stack = inv.addItem(stack).get(0)) != null){
                     inputTarget.getWorld().dropItem(inputTarget.getLocation().add(0, 1, 0), stack);
                 }
             }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
@@ -108,12 +108,18 @@ class CargoNetworkTask implements Runnable {
             Inventory inv = inventories.get(inputTarget.getLocation());
 
             if (inv != null) {
+                // Check if the original slot hasn't been occupied in the meantime
                 if (inv.getItem(previousSlot) == null) {
                     inv.setItem(previousSlot, stack);
                 }
-                // Making sure the item cannot be added to another slot before dropping it on ground
-                else if((stack = inv.addItem(stack).get(0)) != null){
-                    inputTarget.getWorld().dropItem(inputTarget.getLocation().add(0, 1, 0), stack);
+                else {
+                    // Try to add the item into another available slot then
+                    ItemStack rest = inv.addItem(stack).get(0);
+                    
+                    if (rest != null) {
+                        // If the item still couldn't be inserted, simply drop it on the ground
+                        inputTarget.getWorld().dropItem(inputTarget.getLocation().add(0, 1, 0), rest);
+                    }
                 }
             }
             else {


src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
```diff
@@ -111,7 +111,8 @@ private void routeItems(Location inputNode, Block inputTarget, int frequency, Ma
                 if (inv.getItem(previousSlot) == null) {
                     inv.setItem(previousSlot, stack);
                 }
-                else {
+                // Making sure the item cannot be added to another slot before dropping it on ground
+                else if((stack = inv.addItem(stack).get(0)) != null){
```
TheBusyBiscuit, line 115: This could lead to an IndexOutOfBoundsException. So doing .isEmpty() on the returned map instead would be much safer.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
WalshyDev, line 115: nitpick: Space after if and after the ending closing bracket

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
OmerBenGera: You are wrong, that's a Map, not a List. Therefore, running Map#get(0) will return null if everything was fine, not an IndexOutOfBoundsException

TheBusyBiscuit: I even wrote Map but yeah I thought of a Collection, still could be simplified to .isEmpty() nonetheless

OmerBenGera: > 
> 
> I even wrote Map but yeah I thought of a Collection, still could be simplified to .isEmpty() nonetheless

It will just make the code longer, this version works fine and can be easily understood

TheBusyBiscuit: I think the opposite is true.
Doing isEmpty() would be significantly shorter and much easier to read than an assignment and a null check.

OmerBenGera: > 
> 
> I think the opposite is true.
> Doing isEmpty() would be significantly shorter and much easier to read than an assignment and a null check.

I will change it if I have the time...

TheBusyBiscuit: Lemme change it to a "suggested code change" then you only need to hit apply.

TheBusyBiscuit, line 115:
```suggestion
                else if (!inv.addItem(stack).isEmpty()) {
```

OmerBenGera: > 

This will lead to duplication. Let's say some of the items were deposited into the source inventory, you are ignoring the leftovers and you deposit the old item, which will result in a possible duplication.

TheBusyBiscuit: Oh you are right. Yeah, in that case your solution would be better.
I think I really need a break or some sleep o_O

LinoxGH, line 115: I think this is a bit messy. Maybe create a variable of inv.addItem().get() and then check if it's null. I don't quite understand how checking if the inventory is empty can duplicate the item though.

TheBusyBiscuit (approved):
I had another look at it and I am approving it now. Someone else still needs to submit their review and can then merge it.

I am sorry @OmerBenGera today seems to be not the right day for code reviews in our team XD

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
OmerBenGera: > 
> 
> I think this is a bit messy. Maybe create a variable of inv.addItem().get() and then check if it's null. I don't quite understand how checking if the inventory is empty can duplicate the item though.

addItem is adding the item, and returns the results. If it tried to add 64 of an item, but only 28 of it could be added, it would return the item with an amount of 36. If you only check if it's empty, and you drop the original item (the 64 one), you drop 64, however 28 were already deposited. Aka, the items was duped.

Instead of checking why the PR is not good, just accept it lmao
It's such a simple additional, I don't understand why you keep refusing to accept it

WalshyDev:
Still waiting for formatting changes to be done. Also, SonarCloud will complain about that initialization in the if statement.

OmerBenGera:
> 
> 
> Still waiting for formatting changes to be done. Also, SonarCloud will complain about that initialization in the if statement.



> 
> 
> Still waiting for formatting changes to be done. Also, SonarCloud will complain about that initialization in the if statement.

The code is good as it is, I don't know what SonarCloud looks for, but that's a known technic that follows coding formats.
I am tired of explaining why the PR should be accepted, I already gave the user that reported the issue a working jar, and I am out of here

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
TheBusyBiscuit: We aren't checking whether the PR is good, we are doing code reviews to ensure changes won't break anything.

There is always a bias with the person who wrote the changes so we require 2 approving reviews from maintainers to ensure nothing goes unnoticed.

Especially in such a messy area of the codebase there really isn't anything like a "simple addition".
I agree today's reviews didnt go as usual, but its still important to review changes before merging and to ask questions and get a better understanding of the situation.

LinoxGH: Because it's still hard to read. Can't you separate it into a couple lines. Like 1 for the variable and other checking if that variable is null. Just add spaces around the brackets and instead of assigning the variable inside if statement, assign it in an upper line.

OmerBenGera: > 
> 
> Because it's still hard to read. Can't you separate it into a couple lines. Like 1 for the variable and other checking if that variable is null. Just add spaces around the brackets and instead of assigning the variable inside if statement, assign it in an upper line.

it makes no sense to separate the lines. If you really want to separate them, go for it.
You are only looking for reasons why to not accept a literally cleaned and working code

LinoxGH: > > 
> > 
> > Because it's still hard to read. Can't you separate it into a couple lines. Like 1 for the variable and other checking if that variable is null. Just add spaces around the brackets and instead of assigning the variable inside if statement, assign it in an upper line.
> 
> it makes no sense to separate the lines. If you really want to separate them, go for it.
> You are only looking for reasons why to not accept a literally cleaned and working code

And you are now looking for ways to start a fight. In nowhere of sf code, it was used. Your code does not follow the code standards of this repo. All changes we have asked for can easily be done even on github website and on mobile. You could have done them already while writing all these comments.

TheBusyBiscuit:
@WalshyDev @LinoxGH If you wanna make formatting or refactoring changes in this PR (remember, maintainers have write access to PRs), go for it, some additional docs to explain the situation would be good too.
But I think it is feature wise good as is, we can always refactor later.
If you need any more context on these changes, check the chat history on discord

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
TheBusyBiscuit: No need to fight, both of you. Mistakes have been made here by multiple people. Formatting and docs can always be improved further. But no need to drag this on even longer.

OmerBenGera: I am not starting a fight. I don't see a point in changing it, and if you really want to do that - go for it.
I don't know what are your coding rules, and I am 100% sure that if I change it you'll have complaints again.
I am not gonna continue this pointless discussion. If you want to change it, do it. If not, then fine - the code is good like this.
And as you said, "All changes we have asked for can easily be done even on github website and on mobile. You could have done them already while writing all these comments." - then just do the changes instead of wasting my time.

OmerBenGera:
> 
> 
> @WalshyDev @LinoxGH If you wanna make formatting or refactoring changes in this PR (remember, maintainers have write access to PRs), go for it, some additional docs to explain the situation would be good too.
> But I think it is feature wise good as is, we can always refactor later.
> If you need any more context on these changes, check the chat history on discord

Finally, a comment that makes sense. It's not a complicated code, it's literally one line of code, which is also commented. If they don't like the way it's written (which I also don't get - it's so clean and nice), then they can do their changes. I would understand why they don't want to make a recode to large changes, however, that's literally a one-line change and it makes no sense to change it - at least not for me.

TheBusyBiscuit:
There's been enough of this discussion and "back and forfth" with who should do what now.
> it's so clean and nice

a bit overconfident to my taste.
**Any** code can always be improved and this is by far no exception from this rule.
Inline variable assignments in sub-expressions are not considered "clean and nice", as @WalshyDev pointed out before, this is referenced as `java:s1121` ([ref](https://jira.sonarsource.com/browse/RSPEC-1121) and discouraged as it creates less-readable code. It is closely related to `CWE-481` (Common Weakness Enumeration Edition 481) ([ref](http://cwe.mitre.org/data/definitions/481.html)) where assignments and comparisons in a boolean sub-expression can be easily mistaken and shouldn't be used alongside each other.

Anyway, my point is the following: Can we improve this code? Absolutely. Is it in scope of this pull request? debatable.
Do we need to argue about who needs to do what? Absolutely not...

As long as one of the other organization members approves the merge of this pull request, it can be changed later.
If anyone thinks this should be adressed before the merge, be my guest to change it now. In the end it doesn't matter who commits here. I think everyone here has learned something from this experience, whether it would be about code quality, not reviewing PRs when you're tired, or whether it's just general improvement of Slimefun as a whole. There's no need to argue any further, so I would suggest not to make any further comments on this PR.
Either someone submits an approving review, the author adds additional commits or one from the team submits their suggested changes.

WalshyDev (approved):
LGTM

LinoxGH (approved):

PR #2348, opened 2020-09-26
Pull request for an additional customizable option in grappling hooks

flaxeneel2:
[+] Added a setting to allow the grappling hook to stay in inventory instead of being dropped on the floor on use

## Description
I added an item setting to allow players to set if they want the grappling hook to drop on the floor on use or just stay in the inventory. I did this because I use slimefun on a small server with friends and they complained about the grappling hooks dropping on the floor as annoying

## Changes
[+] Item Setting named "stay-in-inv-on-use"

## Related Issues
None

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/GrapplingHook.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/GrapplingHook.java
@@ -22,14 +22,14 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
 public class GrapplingHook extends SimpleSlimefunItem<ItemUseHandler> {
 
-    private final ItemSetting<Boolean> stayininvonuse = new ItemSetting<>("stay-in-inv-on-use", false);
+    private final ItemSetting<Boolean> consumeOnUse = new ItemSetting<>("consume-on-use", true);
     private final ItemSetting<Integer> despawnTicks = new ItemSetting<>("despawn-seconds", 60);
 
     public GrapplingHook(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
         super(category, item, recipeType, recipe);
 
         addItemSetting(despawnTicks);
-        addItemSetting(stayininvonuse);
+        addItemSetting(consumeOnUse);
     }
 
     @Override
@@ -37,7 +37,7 @@ public class GrapplingHook extends SimpleSlimefunItem<ItemUseHandler> {
         return e -> {
             Player p = e.getPlayer();
             UUID uuid = p.getUniqueId();
-            Boolean keepininv = stayininvonuse.getValue();
+            boolean consumeOnUseValue = consumeOnUse.getValue();
 
             if (!e.getClickedBlock().isPresent() && !SlimefunPlugin.getGrapplingHookListener().isGrappling(uuid)) {
                 e.cancel();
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/GrapplingHook.java
```diff
@@ -22,19 +22,22 @@
 
 public class GrapplingHook extends SimpleSlimefunItem<ItemUseHandler> {
 
+    private final ItemSetting<Boolean> stayininvonuse = new ItemSetting<>("stay-in-inv-on-use", false);
```
TheBusyBiscuit, line 25: Please follow the naming convention for fields, this should be in "pascal case", so "stayInInvOnUse".
Other than that I think the name is a little confusing, I think something like `consumeOnUse` (which would actually be the opposite here) would make more sense and is much more user-friendly.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/GrapplingHook.java
```diff
@@ -22,19 +22,22 @@
...
         addItemSetting(despawnTicks);
+        addItemSetting(stayininvonuse);
     }
 
     @Override
     public ItemUseHandler getItemHandler() {
         return e -> {
             Player p = e.getPlayer();
             UUID uuid = p.getUniqueId();
+            Boolean keepininv = stayininvonuse.getValue();
```
TheBusyBiscuit, line 40: Please don't use the boxed `Boolean` class, use the primitive `boolean` instead.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/GrapplingHookEntity.java
```diff
@@ -15,13 +15,14 @@
 final class GrapplingHookEntity {
 
     private final boolean dropItem;
-
+    private final boolean keepininv;
```
TheBusyBiscuit, line 18: Pascal case, so `keepInInv`, or with the suggested reversal: `wasConsumed` may be more fitting.

TheBusyBiscuit (changes requested):

LinoxGH (approved):
Other than the changes TheBusyBiscuit requested, it looks fine to merge.

flaxeneel2:
Made requested changes

LinoxGH (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/GrapplingHook.java
```diff
@@ -47,7 +50,9 @@ public ItemUseHandler getItemHandler() {
                 ItemStack item = e.getItem();
 
                 if (item.getType() == Material.LEAD) {
-                    item.setAmount(item.getAmount() - 1);
+                    if (consumeOnUseValue) {
+                        item.setAmount(item.getAmount() - 1); //If consume on use is enabled, this line will take 1 grappling hook out of player's hand
```
TheBusyBiscuit, line 54: Comments should be made on the line above the code, never right next to it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/GrapplingHookEntity.java
```diff
@@ -33,8 +34,10 @@ public Arrow getArrow() {
 
     public void drop(@Nonnull Location l) {
         if (dropItem) {
-            Item item = l.getWorld().dropItem(l, SlimefunItems.GRAPPLING_HOOK.clone());
-            item.setPickupDelay(16);
+            if (wasConsumed) {
+                Item item = l.getWorld().dropItem(l, SlimefunItems.GRAPPLING_HOOK.clone());
+                item.setPickupDelay(16);
+            } //If a grappling hook was consumed, then this will drop one on the floor
```
TheBusyBiscuit, line 40: This comment should be made above the if-statement or inside of it.

TheBusyBiscuit (changes requested):
Just two small mistakes with comment placing, but other than that it's looking good to go!

TheBusyBiscuit (approved):
:+1:

LinoxGH (approved):

PR #2372, opened 2020-09-30
Feature/shareable androids

md5sha256:
## Description
Draft implementation for #2326. 

## Changes
- Adds a new package (accessibility) for access level persistence and handling.
- Adds a new attribute (`TierAccessible`) [Open to suggestion for a better name]

## Checklist


TheBusyBiscuit:
(Just assigning first, this isn't ready for review yet anyway) 
I simply think it'll be a three-man job to review it and that Walshy should definiely be part of it XD

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/AbstractAccessManager.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public void clear() {
+
+    }
+
+    public <I extends SlimefunItem & TierAccessible> void load(@Nonnull I object) {
+        clear();
+        if (!SlimefunPlugin.getMinecraftVersion()
+            .isAtLeast(MinecraftVersion.MINECRAFT_1_14)) { // This fe
```
TheBusyBiscuit, line 33: covfefe

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/TierAccessible.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Represents anything with tiered access.
+ *
+ * @see io.github.thebusybiscuit.slimefun4.core.accessibility.AccessLevel
+ * @see AccessData
+ *
+ * @author md5sha256
+ */
+public interface TierAccessible {
```
TheBusyBiscuit, line 17: Single-Method-Interfaces should be marked as `@FunctionalInterface`, unless you got explicit plans to add more methods.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/TierAccessible.java
```diff
@@ -0,0 +1,22 @@
...
+import io.github.thebusybiscuit.slimefun4.core.accessibility.AccessData;
+import io.github.thebusybiscuit.slimefun4.core.accessibility.AccessManager;
+
+import javax.annotation.Nonnull;
+
+
+/**
+ * Represents anything with tiered access.
+ *
+ * @see io.github.thebusybiscuit.slimefun4.core.accessibility.AccessLevel
```
TheBusyBiscuit, line 12: Please import "AccessLevel" here, no point in writing this out

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Represents an access level where no privileges are allocated.
+     */
+    public static final ConcreteAccessLevel NONE = new ConcreteAccessLevel("&cNone");
+
+    /**
+     * Represents an access level where all privileges are allocated.
+     */
+    public static final ConcreteAccessLevel FULL = new ConcreteAccessLevel("&bFull");
```
TheBusyBiscuit, line 23: This should be turned into an enum, you have a private constructor anyway so this should definitely be an enum here.

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
```diff
@@ -0,0 +1,79 @@
...
+            return -1;
+        } else {
+            return 1;
+        }
+    }
+
+    @Override
+    public String toString() {
+        return this.displayName;
+    }
```
TheBusyBiscuit, line 78: Once this is an enum this can be removed as enums do that in general.

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
```diff
@@ -0,0 +1,79 @@
...
+    @Override
+    public AccessLevel previous() {
+        return this == FULL ? NONE : null;
+    }
+
+    @Nonnull
+    @Override
+    public ConcreteAccessLevel toConcreteAccessLevel() {
+        return this;
+    }
```
TheBusyBiscuit, line 53: What exactly is the point of this?

TheBusyBiscuit (changes requested):
There's still lots of docs missing but you can add it later anyway.
The main changes I would like to request are:
1. I think this is definitely more of an API and lesser of a core mechanic since you are designing this to be very API-centric, so I reckon it should be moved to the `api` package
2. ConcreteAccessLevel should be an enum

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
md5sha256: The issue here lies in the fact that an enum has a `final compareTo` method which would clash. I originally intended for this to be an enum.

TheBusyBiscuit: Hmmm, I see.
Could always add your own compare method to the interface though

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
md5sha256: This was intended for handlers to be able handle the access level at a macro level with far less precision. For example, the android UI handler would (by default) only care if a person has full access to the android or not. However, this may change when more fine-grained access is needed say when specific actions of an android should be limited based off of an access level. An example of this might be "FarmerAndroid.ALLOW_REFUEL". I'm not entirely sure how to go about this but its intended so that the access levels have some coherence and structure. Handlers should not have to care about every single level, and only some, depending on precision. The concrete access level serves to be of the lowest precision so to speak.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/TierAccessible.java
md5sha256: Noted, I'll add the annotation if I do not add more methods to the interface.

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
md5sha256: I'll give that go.

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
TheBusyBiscuit: uuuh, hmmm. I think restricting access on a per-instruction basis is a bit out of scope for the intend behind this issue.
The original suggestion that was posted on discord only wanted a "friends list" for their androids so they could allow others to access the android's inventory or some very trusted individuals to even edit the source code of the android.
I think adding the ability for new integrations like a "Adhere to protection plugin" mode or such would be awesome.
But I'm not sure if this might be a bit too deep.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/TierAccessible.java
TheBusyBiscuit: Great :+1:

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
md5sha256: Righty, well in this case I'm not sure how best to proceed for "future proofing". I'll leave the api-centric design for this as is sans the fine-grain access control for now.

md5sha256:
Currently, I've got *some* of the implementation in place. I highly dislike the current serialization methods I've put in place and am open to suggestions. I have considered some form of a regsistry, but I'm not sure if that would be out of scope. 

I'd also like to clarify: When opening an android, does this automatically open up a menu for script editing or are there other (less sensitive) things which are also accessible? From what I saw in the code, there isn't but I'd like to make sure.

src/main/java/io/github/thebusybiscuit/slimefun4/core/accessibility/ConcreteAccessLevel.java
md5sha256: Honestly, as I've been going about implementing I *strongly* see that it would make sense for a "per-operation" access level to be in place alongside general ("concrete") access levels. For example, I'd imagine some players would friends to be able to execute scripts but not modify them. This would (therefore) require fine-grained access control.

TheBusyBiscuit:
I'll close this issue due to inactivity/staleness.
In its current state the code is still a bit convoluted, but I still see this as a great addition to the project. Perhaps we can work together on a new draft in the future.
Thanks though!

Issue #2365, opened 2020-09-30
Hacktoberfest 2020


WalshyDev:
## 📘 Why Hacktoberfest?
Slimefun is participating in the annual Hacktoberfest event this year. 
This is an awesome opportunity to contribute to open-source repostitories and get a **free limited edition t-shirt** or to **plant a tree**!
You can sign up for Hacktoberfest right here: https://hacktoberfest.digitalocean.com/

In order to complete Hacktoberfest, you need to submit four pull requests to any public repository on GitHub.
Slimefun has been open-source and home to many multinational contributors throughout several years, so this is a perfect oppurtunity for you to get involved! And you can win a cool shirt or plant a tree of course!

This year we specifically made some [Hacktoberfest issues](https://github.com/Slimefun/Slimefun4/labels/Hacktoberfest) for Slimefun4 which explain what is needed, the difficulty level and the classes you will likely need to deal with/look at. We hope this helps people not familiar with the codebase to get a more welcoming entry into the project.

## 💡 Suggested types of pull requests
* Fulfilling any of the designated [Hacktoberfest Issues](https://github.com/issues?q=is%3Aopen+is%3Aissue+user%3ASlimefun+label%3AHacktoberfest+archived%3Afalse) on the Slimefun GitHub organization
* [Contributing to our wiki](https://github.com/Slimefun/Slimefun4/wiki/Expanding-the-Wiki)
* [Bug fixes](https://github.com/issues?q=is%3Aopen+is%3Aissue+user%3ASlimefun+label%3A%22Bug+Report%22+archived%3Afalse)
* Implementing suggestions (Please check the [Hacktoberfest Issues](https://github.com/Slimefun/Slimefun4/issues?q=is%3Aopen+label%3AHacktoberfest+is%3Aissue) or, check out the [#approved](https://discord.gg/slimefun) channel on our [Discord Server](https://discord.gg/slimefun)
* Performance improvements to any of our [public projects](https://github.com/Slimefun)
* Improving code quality and adding documentation, check us out on [sonarcloud.io](https://sonarcloud.io/organizations/slimefun/projects?sort=-analysis_date) for a big and detailed list of improvements you can make

## ⚠️ What we don't want to see
* Spam Pull Requests that serve no purpose other than abusing the event in order to get a shirt, we will mark such pull requests as invalid!
* Pull Requests that break Slimefun or its addons
* Fixing one or more typos is fine, but don't make 5 Pull Requests each fixing a single typo
[...]

PR #2373, opened 2020-10-01
Update README.md

monu1315:
## Checklist


poma123:
Please do not mess up the readme. It's clean and great as it is.

Btw, you could participate in Hacktoberfest! Read more about it: #2365

PR #2385, opened 2020-10-03
Speed can affect enchanter/disenchanter times, fixed pom typo

Mooy1:
## Description
The auto enchanter and disenchanter times can be affected by speed, so if more tiers were to be created in an addon, you could @override the speed to actually make it faster. I also fixed a pom typo.

## Changes
in pom, pakages -> packages

in enchanter /disenchanter, the old time is now divided by the speed, which doesn't affect the current ones.

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -213,7 +213,7 @@
                         <link>${spigot.javadocs}</link>
                     </links>
 
-                    <!-- We can group pakages together in our Javadocs -->
+                    <!-- We can group packages together in our Javadocs -->
                     <groups>
                         <group>
                             <title>Slimefun4 - API</title>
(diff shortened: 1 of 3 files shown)


LinoxGH (commented):
The balancing aspect of this needs to be tested through.

Mooy1:
I mean it doesn't change base sf, just makes it a lot easier for addon makers to create faster versions with custom speeds + energy consumption etc.

TheBusyBiscuit:
Poor override being tagged in countless pull requests 😢

TheBusyBiscuit (approved):

Sfiguz7 (approved):
Sorry Mr. Override

PR #2394, opened 2020-10-05
Improve magical zombie pills

svr333:
## Description
Add the possibility to convert Zombified Piglins to Piglins using Magical Zombie Pills (as suggested and approved on the Discord)

## Changes
- Add the possibility to convert Zombified Piglins to Piglins
- Update the items lore

## Checklist

<br>
N/A


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -63,6 +64,20 @@ public EntityInteractHandler getItemHandler() {
...
+                if (p.getGameMode() != GameMode.CREATIVE) {
+                    ItemUtils.consumeItem(item, false);
+                }
+
+                Location loc = entity.getLocation();
+                p.playSound(loc, Sound.ENTITY_ZOMBIE_VILLAGER_CONVERTED, 1, 1);
+
+                entity.remove();
+                loc.getWorld().spawnEntity(loc, EntityType.PIGLIN);
+            }
```
TheBusyBiscuit, line 80: This is pretty redundant with the above, there's probably lots of ways to refactor this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
```diff
@@ -63,7 +63,7 @@ private SlimefunItems() {}
...
     public static final SlimefunItemStack VITAMINS = new SlimefunItemStack("VITAMINS", Material.NETHER_WART, "&cVitamins", "", "&aLevel III - Medical Supply", "", "&fRestores 4 Hearts", "&fExtinguishes Fire", "&fCures Poison/Wither/Radiation", "", LoreBuilder.RIGHT_CLICK_TO_USE);
     public static final SlimefunItemStack MEDICINE = new SlimefunItemStack("MEDICINE", Material.POTION, Color.RED, "&cMedicine", "", "&aLevel III - Medical Supply", "", "&fRestores 4 Hearts", "&fExtinguishes Fire", "&fCures Poison/Wither/Radiation");
-    public static final SlimefunItemStack MAGICAL_ZOMBIE_PILLS = new SlimefunItemStack("MAGICAL_ZOMBIE_PILLS", Material.NETHER_WART, "&6Magical Zombie Pills", "", "&eRight Click &7a Zombified Villager to", "&7instantly cure it from its curse");
+    public static final SlimefunItemStack MAGICAL_ZOMBIE_PILLS = new SlimefunItemStack("MAGICAL_ZOMBIE_PILLS", Material.NETHER_WART, "&6Magical Zombie Pills", "", "&eRight Click &7a Zombified Villager", "&eor &7a Zombified Piglin to", "&7instantly cure it from its curse");
```
LinoxGH, line 66: I'm not sure about the lore change here. It may break some items in some servers.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -63,6 +64,20 @@ public EntityInteractHandler getItemHandler() {
...
+
+                if (p.getGameMode() != GameMode.CREATIVE) {
+                    ItemUtils.consumeItem(item, false);
+                }
+
+                Location loc = entity.getLocation();
+                p.playSound(loc, Sound.ENTITY_ZOMBIE_VILLAGER_CONVERTED, 1, 1);
+
+                entity.remove();
+                loc.getWorld().spawnEntity(loc, EntityType.PIGLIN);
```
LinoxGH, line 79: This part is now a duplicate of the code above. Instead move this to a private method and call it. For both entity types. I am not sure if this is still possible. Though it's worth a try.

LinoxGH (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
TheBusyBiscuit: Other than simply not stacking with older items this should be fine.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -74,4 +76,19 @@ public EntityInteractHandler getItemHandler() {
     public ItemUseHandler onRightClick() {
         return PlayerRightClickEvent::cancel;
     }
-}
\ No newline at end of file
+
+    private void healZombieVillager(ZombieVillager zombieVillager, Player p) {
```
TheBusyBiscuit, line 80: `@Nonnull` :eyes:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -74,4 +76,19 @@ public EntityInteractHandler getItemHandler() {
...
+        zombieVillager.setConversionTime(1);
+
+        if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
+            zombieVillager.setConversionPlayer(p);
+        }
+    }
+
+    private void healZombifiedPiglin(Entity zombiePiglin) {
```
TheBusyBiscuit, line 88: `@Nonnull` :eyes:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
svr333: True, but if you dont put it in the lore, not many people will know about it

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -56,11 +57,12 @@ public EntityInteractHandler getItemHandler() {
...
-                zombieVillager.setConversionTime(1);
+                if (entity.getType() == EntityType.ZOMBIE_VILLAGER) {
+                    healZombieVillager((ZombieVillager) entity, p);
+                }
 
-                if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
-                    zombieVillager.setConversionPlayer(p);
+                else if (entity.getType() == EntityType.ZOMBIFIED_PIGLIN) {
+                    healZombifiedPiglin(entity);
                 }
```
TheBusyBiscuit, line 66: Also, while this should not cause any probems as it is, it could be a better practice to do an `instanceof` check here instead because you are doing a class cast, just in case.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -3,11 +3,7 @@
 import org.bukkit.GameMode;
 import org.bukkit.Location;
 import org.bukkit.Sound;
-import org.bukkit.entity.Entity;
-import org.bukkit.entity.EntityType;
-import org.bukkit.entity.Player;
-import org.bukkit.entity.Villager;
-import org.bukkit.entity.ZombieVillager;
+import org.bukkit.entity.*;
```
TheBusyBiscuit: No star imports!

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
svr333: Yeah my ide did that automatically, already fixed

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -74,4 +79,19 @@ public EntityInteractHandler getItemHandler() {
...
+
+    private void healZombieVillager(@Nonnull ZombieVillager zombieVillager, @Nonnull Player p) {
+        zombieVillager.setConversionTime(1);
+
+        if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
+            zombieVillager.setConversionPlayer(p);
+        }
+    }
+
+    private void healZombifiedPiglin(@Nonnull Entity zombiePiglin) {
```
LinoxGH, line 91:
```suggestion
    private void healZombifiedPiglin(@Nonnull PigZombie zombiePiglin) {
```

LinoxGH (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
svr333: Casting here isnt necessary, it saves a few milliseconds so I dont think it matters

TheBusyBiscuit: I agree with Linox, changing the parameter type restricts the type of argument this can accept.
So this is definitely a desired change in terms of maintainability.

LinoxGH: > Casting here isnt necessary, it saves a few milliseconds so I dont think it matters
> 
> In the current context, were 100% sure its a PigZombie, but we dont have to cast it since we can use the generic Entity methods
> 
> Though if this is preferred, Ill change it 

You are already checking instanceof befıre this is called. That doesn't change a thing.

LinoxGH (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -56,11 +60,12 @@ public EntityInteractHandler getItemHandler() {
 
                 p.playSound(p.getLocation(), Sound.ENTITY_ZOMBIE_VILLAGER_CONVERTED, 1, 1);
 
-                ZombieVillager zombieVillager = (ZombieVillager) entity;
-                zombieVillager.setConversionTime(1);
+                if (entity instanceof ZombieVillager) {
+                    healZombieVillager((ZombieVillager) entity, p);
+                }
```
TheBusyBiscuit, line 66: Can you please remove this empty line here?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -74,4 +78,19 @@ public EntityInteractHandler getItemHandler() {
...
+
+    private void healZombieVillager(@Nonnull ZombieVillager zombieVillager, @Nonnull Player p) {
+        zombieVillager.setConversionTime(1);
+
+        if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
+            zombieVillager.setConversionPlayer(p);
+        }
+    }
+
+    private void healZombifiedPiglin(@Nonnull PigZombie zombiePiglin) {
```
TheBusyBiscuit, line 90: Wait a second, I only just realized that you removed the 1.16 version check? O_O

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
svr333: entity.getType will never be equal to a ZombiePiglin, because they dont exist in other versions

svr333: I could add it back though, if thats preferred

LinoxGH: > entity.getType will never be equal to a ZombiePiglin, because they dont exist in other versions

Yes and cause an exception

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -47,7 +51,7 @@ public EntityInteractHandler getItemHandler() {
         return (e, item, offhand) -> {
             Entity entity = e.getRightClicked();
 
-            if (entity.getType() == EntityType.ZOMBIE_VILLAGER) {
+            if (entity instanceof ZombieVillager || entity instanceof PigZombie) {
```
poma123, line 54: Also, why are we using PigZombie instead of Piglin? Then the version check should be a bit different.
Is a piglin entity really an instance of PigZombie?

poma123 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
svr333: Because we want to turn a PigZombie into a Piglin, not vice versa

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -56,11 +60,11 @@ public EntityInteractHandler getItemHandler() {
...
 
-                ZombieVillager zombieVillager = (ZombieVillager) entity;
-                zombieVillager.setConversionTime(1);
-
-                if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
-                    zombieVillager.setConversionPlayer(p);
+                if (entity instanceof ZombieVillager) {
+                    healZombieVillager((ZombieVillager) entity, p);
+                }
+                else if (entity instanceof PigZombie) {
```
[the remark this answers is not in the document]
svr333: Im aware that it is, but if there will ever be more recipes added, theyd have to create the else if statement back

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
svr333: I dont actually think it will create an exception? But sure Ill add the check

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
poma123: But in 1.16 there is no PigZombie. Or am I wrong?

poma123: We are looking for zombified piglins, right?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
LinoxGH: > I dont actually think it will create an exception?  (im probably wrong here)
> But sure Ill add the check

Of course it will. The server will try to access an EntityType that doesn't exist and boom. Exception.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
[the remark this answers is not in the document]
poma123: It will be their problem. Currently, this `else if` is redundant since we are working with only two cases (ZombieVillager and PigZombie), and the first one is already handled by an `if` check, for the second one it could simply be an `else`.

poma123:
Please check only those checkboxes that state what you actually did.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
svr333: Oh I was assuming that it still uses the latest version of the api, and that it would just throw if anything out of that version was accessed. Good to know this!

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
svr333: Zombie pigman were renamed to Zombified Piglin, theyre the same

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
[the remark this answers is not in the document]
TheBusyBiscuit: I agree with @svr333 here, this `else if` clearly indicates what exact case this is for.
An else could lead to unwanted side effects if we ever add more entity types.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
poma123: I believe you. Did you test this ingame?

TheBusyBiscuit: Yep, Spigot still calls the Zombified Pigling PigZombie internally. That's the corrret class

poma123: Okay then

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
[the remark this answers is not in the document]
poma123: Fair enough.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -47,7 +51,8 @@ public EntityInteractHandler getItemHandler() {
         return (e, item, offhand) -> {
             Entity entity = e.getRightClicked();
 
-            if (entity.getType() == EntityType.ZOMBIE_VILLAGER) {
+            if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_16) && entity instanceof PigZombie
+                    || entity instanceof ZombieVillager) {
```
TheBusyBiscuit, line 55: Not a fan of this line here, it's really messy and redundant with the two if statements below.
I'll suggest an alternative below

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -56,11 +61,11 @@ public EntityInteractHandler getItemHandler() {
...
 
-                ZombieVillager zombieVillager = (ZombieVillager) entity;
-                zombieVillager.setConversionTime(1);
-
-                if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_15)) {
-                    zombieVillager.setConversionPlayer(p);
+                if (entity instanceof ZombieVillager) {
+                    healZombieVillager((ZombieVillager) entity, p);
+                }
+                else {
```
poma123, line 67: Please change this back to an `else if` as stated [here](https://github.com/Slimefun/Slimefun4/pull/2394#discussion_r499760581)

poma123 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
```diff
@@ -47,7 +51,8 @@ public EntityInteractHandler getItemHandler() {
         return (e, item, offhand) -> {
             Entity entity = e.getRightClicked();
 
-            if (entity.getType() == EntityType.ZOMBIE_VILLAGER) {
+            if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_16) && entity instanceof PigZombie
+                    || entity instanceof ZombieVillager) {
                 Player p = e.getPlayer();
```
TheBusyBiscuit, line 56: What if you put Line 56-62 into a seperate method like `useItem` or whatever and then just call this method before the specific heal method, this way we save redundancy even further because we no longer need two different if statements for the same check.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
TheBusyBiscuit, line 67:
```suggestion
                else if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_16) && entity instanceof PigZombie) {
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/MagicalZombiePills.java
poma123: More specifically: https://github.com/Slimefun/Slimefun4/pull/2394#discussion_r499767817

svr333: I did the minecraft version check above it, is that not good?

TheBusyBiscuit (approved):
Perfect! This is really clean now

LinoxGH (approved):

poma123 (approved):

PR #2400, opened 2020-10-06
Feature/json deserialization

md5sha256:
## Description
Adds a class which maps classes to their respective deserialization logic, a bit like Bukkit's ConfigurationSerialization class.
This addition is intended to make deserialization for objects to be less dependant on knowing the implementation detail. The rationale behind this pr stems from #2372, I needed a good way to deserialize `AccessLevel`s without knowing the deserialization implementation.
 
Although I have not marked this PR as a draft, I'm open to any suggestions, particularly the unit tests which for some reason do not run normally (`mvn test`). I've verified the code passes the test by running the test individually.

## Changes
- Add the `JsonDeserializationService` class which acts as the deserialization logic mapper.
- Add a getter for this service in the SlimefunPlugin class

## Related Issues
N/A

## Checklist
- [ ] I have fully tested the proposed changes and promise that they will not break everything into chaos. (Irrelevant)
- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them. (Irrelevant)


TheBusyBiscuit:
I'm really not seeing much of a point for adding this right now, especially because this seems like it should have been a part of your other pull request?

I mean, more API is always good and the code... well I'll still get to the code later on...
But this really raises the question about what the point of a Pull Request is if it adds just yet unused code snippets that would only clutter the codebase as of now.
So why would you not include this in the PR that actually needs it?

md5sha256:
I wasn't sure if it would be out of scope, as I've already had that happen once with #2372

TheBusyBiscuit:
I would be very much in favour of including these changes in a PR that gives them some use and context.
Every additional piece of code means additional workload and maintenance for the future, so better have a very good reason for it before we start to declutter something that we may not even need in the end.

Also I think I asked this before but why would you even need this at all?
This seems far too overengineered for a simple enum parser and frankly not quite maintainable in my opinion.
I appreciate the effort but don't overengineer too much okay?

md5sha256:
I agree with the "don't overengineer" but I still see merit in this being implemented for future use.

TheBusyBiscuit:
> but I still see merit in this being implemented for future use.

I am not denying that there wouldn't be a future use for it. I just think that this use could become more apparent if put in a context, if you know what I mean.
Like I said, more API is generally a good thing but on the other side additional code also means additional maintenance and future workload. So I think that this will be easier to evaluate once we see the full picture.

PR #2407, opened 2020-10-07
Added some code-style guidelines to CONTRIBUTING

TheBusyBiscuit:
## Description

I am proposing some changes to the `CONTRIBUTING.md`.
Currently my advice has always been "stick to the current style".
But due to the ongoing increase in contributions, it may be wise to facilitate some concrete dos and don'ts.

Anything you wanna change or add?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,15 +119,13 @@ Try to stay inline with the code that surrounds you, having an entire package or
 * Unit Test classes and methods should have no access modifier, not `public`, `protected` nor `private`.
 * Each Test should have a plain text `@DisplayName` annotation!
 #### 5. General best-practices
-* Use **Spaces**, not Tabs!
-* Do not declare multiple fields/variables on the same line! (e.g. Don't do this: `int x, y, z;`)
-* Comments should never go on the same line as code! Always above or below.
-* Use a Logger, try to avoid `System.out.println(...)` and `Throwable#printStacktrace()`!
 * Do not use `Collection#forEach(x -> ...)`, use a proper `for (...)` loop!
 * Do not create new `Random` objects, use `ThreadLocalRandom.current()` instead!
 * Always declare Maps or Collections using their base type! (e.g. `List<String> list = new ArrayList<>();`
 * When doing String operations like `String#toUppercase()`, always specify `Locale.ROOT` as an argument!
 * When reading or writing files, always specify the encoding using `StandardCharsets.UTF_8`!
+* Do not declare multiple fields/variables on the same line! (e.g. Don't do this: `int x, y, z;`)
+* Use a Logger, try to avoid `System.out.println(...)` and `Throwable#printStacktrace()`, use `Logger#log` instead!
 * Do not use Exceptions to validate data, empty catch blocks are a very bad practice, use other means like a regular expression to validate data.
 * If a parameter is annotated with `@Nonnull`, you should enforce this behaviour by doing `Validate.notNull(variable, "...");` and give a meaningful message about what went wrong
 * Any `switch/case` should always have a `default:` case at the end.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -94,3 +94,48 @@ Then you should be able build it via Maven using the goals `clean package`.
...
+## :black_nib: Code Style guidelines
+The general gist when it comes to code style: **Try to be consistent!**.<br>
+Try to stay inline with the code that surrounds you, having an entire package or even a single file that's filled with plenty of different and inconsistent code styles is just hard to read or maintain. That's why we wanna make sure everyone follows these simple principles:
+
+1. Imports
+* Don't use wildcard (`*`) imports!
+* Don't import unused classes!
+* Don't use static imports!
+2. Annotations
+* Methods and parameters should be annotated with `@Nullable` or `@Nonnull`!
```
LinoxGH, line 107: Should also tell from which api/package.

CONTRIBUTING.md
```diff
@@ -94,3 +94,48 @@ Then you should be able build it via Maven using the goals `clean package`.
...
+* Don't use static imports!
+#### 2. Annotations
+* Methods and parameters should be annotated with `@Nullable` or `@Nonnull`!
+* Methods that override a method must be annotated with `@Override`!
+* Interfaces with only one method should be annotated using `@FunctionalInterface`!
+* If you deprecate a method, add an `@deprecated` section to the javadocs explaining why you did it.
+#### 3. Documentation
+* Every class and every public method should have a Javadocs section assigned to it.
+* New packages should have a `package-info.java` file with documentation about the package.
+* Classes should have an `@author` tag.
```
LinoxGH, line 114:
```suggestion
* Classes should have an `@author` tag.
* Classes should have `@see` tags pinpointing the apis used in the class.
```

LinoxGH (changes requested):

CONTRIBUTING.md
TheBusyBiscuit: Yes!

CONTRIBUTING.md
TheBusyBiscuit: Hmm, I will include this suggestion but I'll probably rephrase it a little.

poma123 (changes requested):
Maybe something about putting `else {`, `else if {`, `catch {` etc. on a new line?

TheBusyBiscuit:
> Maybe something about putting `else {`, `else if {`, `catch {` etc. on a new line?

Thought about that but I decided against it first, many people seem to put them on the same line.
So I wouldn't have a problem with changing it to that style of brackets, we would of course need to update all the code before though.
So yeah, I'm currently leaving that open, what's your opinion on it?

LinoxGH:
> > Maybe something about putting `else {`, `else if {`, `catch {` etc. on a new line?
> 
> Thought about that but I decided against it first, many people seem to put them on the same line.
> So I wouldn't have a problem with changing it to that style of brackets, we would of course need to update all the code before though.
> So yeah, I'm currently leaving that open, what's your opinion on it?

I personally prefer it on the same line.

poma123:
> Thought about that but I decided against it first, many people seem to put them on the same line.
> So I wouldn't have a problem with changing it to that style of brackets, we would of course need to update all the code before though.
> So yeah, I'm currently leaving that open, what's your opinion on it?

Well, I've always put them on the same line so... Although putting them on a new line seems clean as well, I don't think that putting them on the same line would be better or worse.

TheBusyBiscuit:
I don't really mind which of the two we follow, as long as it is consistent. Looks like we got two people in favour of same-line, @Sfiguz7 what about you?

TheBusyBiscuit:
I'll go with the OTBS standard then, so else/else if now go on the same line, since that's what some people seem to prefer.
I'll update master in a second too

Sfiguz7 (approved):
It seems I'm always late to the party uff
I am used to OTBS as well but I didn't mind using new lines as plenty of times it was clearer. Someone is going to have fun cleaning every single damn class (or we could just PR a bunch at a time by maybe notifying others of who is taking what just so we can split the burden)

poma123 (approved):

TheBusyBiscuit:
> It seems I'm always late to the party uff
> I am used to OTBS as well but I didn't mind using new lines as plenty of times it was clearer. Someone is going to have fun cleaning every single damn class (or we could just PR a bunch at a time by maybe notifying others of who is taking what just so we can split the burden)

I already commited that >.>

LinoxGH (dismissed):

Seggan:
Don't overuse inverted if + `return` statements in listeners.

LinoxGH:
Maybe something about putting spaces in array initializers, and in if statements between brackets and etc.

TheBusyBiscuit:
@LinoxGH That is already in there under "horizontal whitespaces"

poma123 (approved):

Sfiguz7 (approved):

LinoxGH (approved):

PR #2409, opened 2020-10-08
Added optional lore to categories

Mooy1:
## Description
Addon makers can add up to 5 lines of lore to their category before the "" and "click to open" by setting the first line of their categories ItemStack to "lore", and the following 5 lines to what they want. 

image

## Changes
Added variable for the click to open string

If the Categories ItemStack has line 1 as "lore", it will take the following 5 lines and add them to the category icon as well as the click to open string. Any extra lines past 5 will be cut off. If they only add 3, it wont have 2 extra spaces.

Current categories aren't affected.
image

## Related Issues
none

## Checklist


TheBusyBiscuit:
Sorry but I am gonna reject this.
Removing the lore from the item was very much intentional because of localization and uniformity.

PR #2426, opened 2020-10-09
Hologram movement and GUI update

NCBPFluffyBear:
## Description
Added more movement options to the Hologram, fixed colors, and _added a default goal to the pom_
Video: https://streamable.com/if78pz

## Changes
- Split the Y axis movement into a horizontal and vertical button
- Changed right click to move 0.5 blocks
- Mirrored Y for the X and Z axis by giving them their own parameters and click handlers
- Added a way to convert old holograms to the new format
- Fixed colors on all the text
- Got rid of the extra 1 added to the Y display (Shouldn't have been there in the first place, doesn't make sense)
- Filled up all the empty spots in the GUI so that players can not move items in
_- Added a defaultGoal to pom so that other contributors can build easier_

## Related Issues
N/A

## Checklist


LinoxGH:
I don't think I support this change. This can easily let players to prank each other and some players to lose their projectors. Since this will let you move the hologram a lot of blocks away.

NCBPFluffyBear:
> I don't think I support this change. This can easily let players to prank each other and some players to lose their projectors. Since this will let you move the hologram a lot of blocks away from the projector.

Could always put a cap on how far they can move horizontally, the main purpose of this is to center holograms between two blocks

LinoxGH:
> > I don't think I support this change. This can easily let players to prank each other and some players to lose their projectors. Since this will let you move the hologram a lot of blocks away from the projector.
> 
> Could always put a cap on how far they can move horizontally, the main purpose of this is to center holograms between two blocks

Yeah you should add that. 5 blocks cap would be fine. Although a configurable one with 5 as default would be even better.

This cap should only apply for x and z axis. Not y axis.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+
+            BlockStorage.addBlockInfo(projector, Y_OFFSET_PARAMETER, String.valueOf(0.5));
+            BlockStorage.addBlockInfo(projector, X_OFFSET_PARAMETER, String.valueOf(0.5));
+            BlockStorage.addBlockInfo(projector, Z_OFFSET_PARAMETER, String.valueOf(0.5));
+            openEditor(pl, projector);
+            return false;
+        });
+
+        // Y Axis
+        menu.addItem(2, new CustomItem(SkullItem.fromHash("b221da4418bd3bfb42eb64d2ab429c61decb8f4bf7d4cfb77a162be3dcb0b927"),
```
LinoxGH, line 126: Use HeadTexture here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+            ArmorStand hologram = getArmorStand(projector, true);
+            Location l = new Location(projector.getWorld(), projector.getX() + xOffset, projector.getY() + yOffset, projector.getZ() + zOffset);
+            hologram.teleport(l);
+
+            BlockStorage.addBlockInfo(projector, Y_OFFSET_PARAMETER, String.valueOf(yOffset));
+            openEditor(pl, projector);
+            return false;
+        });
+
+        menu.addItem(11, new CustomItem(SkullItem.fromHash("3b83bbccf4f0c86b12f6f79989d159454bf9281955d7e2411ce98c1b8aa38d8"),
```
LinoxGH, line 142: Use HeadTexture please.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+            "&7Vertical Offset (-): &e" + DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Y_OFFSET_PARAMETER))),
+            "", "&eLeft Click: &7+0.1", "&eRight Click: &7+0.5"));
+        menu.addMenuClickHandler(11, (pl, slot, item, action) -> {
+            double yOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Y_OFFSET_PARAMETER)) + (action.isRightClicked() ? -0.5F : -0.1F));
+            double xOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), X_OFFSET_PARAMETER));
+            double zOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Z_OFFSET_PARAMETER));
+            ArmorStand hologram = getArmorStand(projector, true);
+            Location l = new Location(projector.getWorld(), projector.getX() + xOffset, projector.getY() + yOffset, projector.getZ() + zOffset);
+            hologram.teleport(l);
```
LinoxGH, line 151: Don't do this. Just use Location.add() or hologram.getLocation().getRelative()

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+
+            BlockStorage.addBlockInfo(projector, Y_OFFSET_PARAMETER, String.valueOf(yOffset));
+            openEditor(pl, projector);
+            return false;
+        });
+
+
+        // X Axis
+        menu.addItem(4, new CustomItem(SkullItem.fromHash("f4628ace7c3afc61a476dc144893aaa642ba976d952b51ece26abafb896b8"),
```
LinoxGH, line 160: Use HeadTexture pls.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+            ArmorStand hologram = getArmorStand(projector, true);
+            Location l = new Location(projector.getWorld(), projector.getX() + xOffset, projector.getY() + yOffset, projector.getZ() + zOffset);
+            hologram.teleport(l);
+
+            BlockStorage.addBlockInfo(projector, X_OFFSET_PARAMETER, String.valueOf(xOffset));
+            openEditor(pl, projector);
+            return false;
+        });
+
+        menu.addItem(13, new CustomItem(SkullItem.fromHash("2ae425c5ba9f3c2962b38178cbc23172a6c6215a11accb92774a4716e96cada"),
```
LinoxGH, line 176: HeadTexture

LinoxGH (changes requested):
Also please move the armor stand location setting to a private method accepting x y and z relative coords. And use Location.add() _not sure if this one works_ or Location.getRelative() and then teleport.

LinoxGH:
You don't need to copy the same code 6 times, just implement it into a private method which will be called.

svr333:
> I don't think I support this change. This can easily let players to prank each other and some players to lose their projectors. Since this will let you move the hologram a lot of blocks away from the projector.
Definitely agree with this

> Yeah you should add that. 5 blocks cap would be fine. Although a configurable one with 5 as default would be even better.
Same here, though I think 3 is more then plenty.

NCBPFluffyBear:
> Also please move the armor stand location setting to a private method accepting x y and z relative coords. And use Location.add() _not sure if this one works_ or Location.getRelative() and then teleport.

Where would I use Location.add()? Teleporting should be enough, since everything is calculated and stored in BlockStorage

LinoxGH:
Just use Locatiom.getRelative to get the new location to teleport to. Do not get 2 redundant values(other coords) from BlockStorage. And do not create a new Location.

NCBPFluffyBear:
> You don't need to copy the same code 6 times, just implement it into a private method which will be called.

But the lores are different, the coordinate the locations are added are different, and the offsets are different

LinoxGH:
> > You don't need to copy the same code 6 times, just implement it into a private method which will be called.
> 
> But the lores are different, the coordinate the locations are added are different, and the offsets are different. could u elaborate pls

Methods accept arguments...

NCBPFluffyBear:
> > > You don't need to copy the same code 6 times, just implement it into a private method which will be called.
> > 
> > 
> > But the lores are different, the coordinate the locations are added are different, and the offsets are different. could u elaborate pls
> 
> Methods accept arguments...

Id need like 7 different arguments tho... is it still worth it

LinoxGH:
> > > > You don't need to copy the same code 6 times, just implement it into a private method which will be called.
> > > 
> > > 
> > > But the lores are different, the coordinate the locations are added are different, and the offsets are different. could u elaborate pls
> > 
> > Methods accept arguments...
> 
> Id need like 7 different arguments tho... is it still worth it

Is it worth to not take up a lot of memory with every single click? Of course it is

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -41,8 +45,10 @@ private BlockPlaceHandler onPlace() {
             @Override
             public void onPlayerPlace(BlockPlaceEvent e) {
                 Block b = e.getBlockPlaced();
-                BlockStorage.addBlockInfo(b, "text", "Edit me via the Projector");
-                BlockStorage.addBlockInfo(b, OFFSET_PARAMETER, "0.5");
+                BlockStorage.addBlockInfo(b, "text", ChatColors.color("&fEdit me via the Projector"));
```
TheBusyBiscuit, line 48: Don't put chatcolors into the block data :NotLikeThis:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +71,30 @@ public BlockUseHandler onRightClick() {
...
     private static void openEditor(Player p, Block projector) {
         ChestMenu menu = new ChestMenu(SlimefunPlugin.getLocalization().getMessage(p, "machines.HOLOGRAM_PROJECTOR.inventory-title"));
 
-        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", "&r" + ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"))));
+        for (int s : BORDER) {
+            menu.addItem(s, new CustomItem(Material.GRAY_STAINED_GLASS_PANE, " ", " "));
+            menu.addMenuClickHandler(s, (player, slot, item, action) -> false);
+        }
+
+        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"))));
```
TheBusyBiscuit, line 97: This would appear as purple otherwise.
```suggestion
        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", ChatColor.WHITE + ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"))));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +71,30 @@ public BlockUseHandler onRightClick() {
...
         };
     }
 
     private static void openEditor(Player p, Block projector) {
         ChestMenu menu = new ChestMenu(SlimefunPlugin.getLocalization().getMessage(p, "machines.HOLOGRAM_PROJECTOR.inventory-title"));
 
-        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", "&r" + ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"))));
+        for (int s : BORDER) {
+            menu.addItem(s, new CustomItem(Material.GRAY_STAINED_GLASS_PANE, " ", " "));
+            menu.addMenuClickHandler(s, (player, slot, item, action) -> false);
```
TheBusyBiscuit, line 94: :NotLikeThis: `ChestMenuUtils.getEmptyClickHandler()`

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +71,30 @@ public BlockUseHandler onRightClick() {
...
             }
         };
     }
 
     private static void openEditor(Player p, Block projector) {
         ChestMenu menu = new ChestMenu(SlimefunPlugin.getLocalization().getMessage(p, "machines.HOLOGRAM_PROJECTOR.inventory-title"));
 
-        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", "&r" + ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"))));
+        for (int s : BORDER) {
+            menu.addItem(s, new CustomItem(Material.GRAY_STAINED_GLASS_PANE, " ", " "));
```
TheBusyBiscuit, line 93: Also why add an empty line of lore...
Also... `ChestMenuUtils.getBackground()` or so

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+            "", "&eLeft Click: &7+0.1", "&eRight Click: &7+0.5"));
+        menu.addMenuClickHandler(2, (pl, slot, item, action) -> {
+            double yOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Y_OFFSET_PARAMETER)) + (action.isRightClicked() ? 0.5F : 0.1F));
+            double xOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), X_OFFSET_PARAMETER));
+            double zOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Z_OFFSET_PARAMETER));
+            ArmorStand hologram = getArmorStand(projector, true);
+            Location l = new Location(projector.getWorld(), projector.getX() + xOffset, projector.getY() + yOffset, projector.getZ() + zOffset);
+            hologram.teleport(l);
+
+            BlockStorage.addBlockInfo(projector, Y_OFFSET_PARAMETER, String.valueOf(yOffset));
```
TheBusyBiscuit, line 137: Those are waaaaaaaaaaaayyyy too many reads/writes to BlockStorage, get your data handling sorted.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+        });
+
+        // Y Axis
+        menu.addItem(2, new CustomItem(SkullItem.fromHash("b221da4418bd3bfb42eb64d2ab429c61decb8f4bf7d4cfb77a162be3dcb0b927"),
+            "&7Vertical Offset (+): &e" + DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Y_OFFSET_PARAMETER))),
+            "", "&eLeft Click: &7+0.1", "&eRight Click: &7+0.5"));
+        menu.addMenuClickHandler(2, (pl, slot, item, action) -> {
+            double yOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Y_OFFSET_PARAMETER)) + (action.isRightClicked() ? 0.5F : 0.1F));
+            double xOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), X_OFFSET_PARAMETER));
+            double zOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Z_OFFSET_PARAMETER));
```
TheBusyBiscuit, line 132: Why isonly the first one fixed...?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit, line 142: You should NOT put texture hashes in here, place them into the HeadTexture enum to make them reusable.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit, line 126: You should NOT put texture hashes in here, place them into the HeadTexture enum to make them reusable.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit, line 160: You should NOT put texture hashes in here, place them into the HeadTexture enum to make them reusable.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit, line 176: You should NOT put texture hashes in here, place them into the HeadTexture enum to make them reusable.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+            hologram.teleport(l);
+
+            BlockStorage.addBlockInfo(projector, X_OFFSET_PARAMETER, String.valueOf(xOffset));
+            openEditor(pl, projector);
+            return false;
+        });
+
+
+        // Z Axis
+        menu.addItem(6, new CustomItem(SkullItem.fromHash("2d9287616343d833e9e7317159caa2cb3e59745113962c1379052ce478884fa"),
```
TheBusyBiscuit, line 194: You should NOT put texture hashes in here, place them into the HeadTexture enum to make them reusable.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,14 +109,116 @@ private static void openEditor(Player p, Block projector) {
...
+            ArmorStand hologram = getArmorStand(projector, true);
+            Location l = new Location(projector.getWorld(), projector.getX() + xOffset, projector.getY() + yOffset, projector.getZ() + zOffset);
+            hologram.teleport(l);
+
+            BlockStorage.addBlockInfo(projector, Z_OFFSET_PARAMETER, String.valueOf(zOffset));
+            openEditor(pl, projector);
+            return false;
+        });
+
+        menu.addItem(15, new CustomItem(SkullItem.fromHash("a3852bf616f31ed67c37de4b0baa2c5f8d8fca82e72dbcafcba66956a81c4"),
```
TheBusyBiscuit, line 210: You should NOT put texture hashes in here, place them into the HeadTexture enum to make them reusable.

TheBusyBiscuit (changes requested):
performance and code wise there is a lot to be desired, way too many calls to BlockStorage.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -27,7 +28,10 @@
 
 public class HologramProjector extends SlimefunItem {
 
-    private static final String OFFSET_PARAMETER = "offset";
+    private static final String Y_OFFSET_PARAMETER = "y_offset";
+    private static final String X_OFFSET_PARAMETER = "x_offset";
+    private static final String Z_OFFSET_PARAMETER = "z_offset";
```
TheBusyBiscuit, line 33: Also... why is this not in order...

TheBusyBiscuit (changes requested):

TheBusyBiscuit:
Also why...
> added a default goal to the pom

There is already a default goal.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: Y is likely the most often used offset, but I can change it for uniformity

NCBPFluffyBear:
> Also why...
> 
> > added a default goal to the pom
> 
> There is already a default goal.

where?

TheBusyBiscuit:
https://github.com/Slimefun/Slimefun4/blob/master/pom.xml#L80

Hence why the build failed...
`Duplicated tag: 'defaultGoal' (position: START_TAG seen ...</finalName>\n        <defaultGoal>... @83:22)  @ line 83, column 22`

NCBPFluffyBear:
> https://github.com/Slimefun/Slimefun4/blob/master/pom.xml#L80
> 
> Hence why the build failed...
> `Duplicated tag: 'defaultGoal' (position: START_TAG seen ...</finalName>\n <defaultGoal>... @83:22) @ line 83, column 22`

Yes, thats the one I added, but I cant find the existing one. You sure it isnt included with the process on github? If it existed within the file the build button wouldnt be greyed out

image

TheBusyBiscuit:
Dude, I literally sent you a link to the one that's already in the pom...

NCBPFluffyBear:
oh i didnt pull... oops sorry

NCBPFluffyBear:
```error: failed to push some refs to 'https://github.com/NCBPFluffyBear/Slimefun4.git'
To https://github.com/NCBPFluffyBear/Slimefun4.git
!	refs/heads/hologram-update:refs/heads/hologram-update	[remote rejected] (refusing to allow an OAuth App to create or update workflow `.github/workflows/auto-approve.yml` without `workflow` scope)
Done
20:11:15.537: [Slimefun4] git -c credential.helper= -c http.sslBackend=schannel -c core.quotepath=false -c log.showSignature=false add --ignore-errors -A -- .github/workflows/auto-approve.yml```

I encountered this issue when trying to push, any ideas?

File: src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java [hunk 15]
```diff
@@ -1,5 +1,9 @@
 package io.github.thebusybiscuit.slimefun4.implementation.items.blocks;
 
+import io.github.thebusybiscuit.slimefun4.api.items.ItemSetting;
+import io.github.thebusybiscuit.slimefun4.utils.ChestMenuUtils;
+import io.github.thebusybiscuit.slimefun4.utils.HeadTexture;
```
LinoxGH, line 5: Misplaced imports.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -27,12 +31,18 @@
 
 public class HologramProjector extends SlimefunItem {
 
-    private static final String OFFSET_PARAMETER = "offset";
+    private static final String X_OFFSET_PARAMETER = "x_offset";
+    private static final String Y_OFFSET_PARAMETER = "y_offset";
+    private static final String Z_OFFSET_PARAMETER = "z_offset";
+    private static final int[] BORDER = {1, 3, 5, 7, 8, 10, 12, 14, 16, 17};
+
+    private static final ItemSetting<Double> maxDistance = new ItemSetting<>("max-horizontal-offset", 5.0);
```
LinoxGH, line 39:
```suggestion
    private ItemSetting<Double> maxDistance = new ItemSetting<>("max-horizontal-offset", 5.0);
```
This should not be static let alone final.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -41,8 +51,10 @@ private BlockPlaceHandler onPlace() {
             @Override
             public void onPlayerPlace(BlockPlaceEvent e) {
                 Block b = e.getBlockPlaced();
-                BlockStorage.addBlockInfo(b, "text", "Edit me via the Projector");
-                BlockStorage.addBlockInfo(b, OFFSET_PARAMETER, "0.5");
+                BlockStorage.addBlockInfo(b, "text", ChatColors.color("Edit me via the Projector"));
```
LinoxGH, line 54:
```suggestion
                BlockStorage.addBlockInfo(b, "text", "Edit me via the Projector");
```
Redundant call, the text currently has no colors.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,25 +115,36 @@ private static void openEditor(Player p, Block projector) {
...
 
-        menu.addItem(1, new CustomItem(Material.CLOCK, "&7Offset: &e" + DoubleHandler.fixDouble(Double.valueOf(BlockStorage.getLocationInfo(projector.getLocation(), OFFSET_PARAMETER)) + 1.0D), "", "&rLeft Click: &7+0.1", "&rRight Click: &7-0.1"));
-        menu.addMenuClickHandler(1, (pl, slot, item, action) -> {
-            double offset = DoubleHandler.fixDouble(Double.valueOf(BlockStorage.getLocationInfo(projector.getLocation(), OFFSET_PARAMETER)) + (action.isRightClicked() ? -0.1F : 0.1F));
+        menu.addItem(9, new CustomItem(Material.CLOCK, "&cReset Position", "", "&eClick &7to reset hologram location"));
+        menu.addMenuClickHandler(9, (pl, slot, item, action) -> {
             ArmorStand hologram = getArmorStand(projector, true);
-            Location l = new Location(projector.getWorld(), projector.getX() + 0.5, projector.getY() + offset, projector.getZ() + 0.5);
-            hologram.teleport(l);
+            hologram.teleport(new Location(projector.getWorld(), projector.getX() + 0.5, projector.getY() + 0.5, projector.getZ() + 0.5));
```
LinoxGH, line 121:
```suggestion
            hologram.teleport(projector.getLocation().add(0.5, 0.5, 0.5));
```

I think this should work better. Well at least it looks nicer.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,25 +115,36 @@ private static void openEditor(Player p, Block projector) {
...
 
-            BlockStorage.addBlockInfo(projector, OFFSET_PARAMETER, String.valueOf(offset));
+            BlockStorage.addBlockInfo(projector, X_OFFSET_PARAMETER, String.valueOf(0.5));
+            BlockStorage.addBlockInfo(projector, Y_OFFSET_PARAMETER, String.valueOf(0.5));
+            BlockStorage.addBlockInfo(projector, Z_OFFSET_PARAMETER, String.valueOf(0.5));
```
LinoxGH, line 125:
```suggestion
            BlockStorage.addBlockInfo(projector, X_OFFSET_PARAMETER, "0.5");
            BlockStorage.addBlockInfo(projector, Y_OFFSET_PARAMETER, "0.5");
            BlockStorage.addBlockInfo(projector, Z_OFFSET_PARAMETER, "0.5");
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,25 +115,36 @@ private static void openEditor(Player p, Block projector) {
...
     }
 
     private static ArmorStand getArmorStand(Block projector, boolean createIfNoneExists) {
         String nametag = BlockStorage.getLocationInfo(projector.getLocation(), "text");
-        double offset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), OFFSET_PARAMETER));
-        Location l = new Location(projector.getWorld(), projector.getX() + 0.5, projector.getY() + offset, projector.getZ() + 0.5);
+        double xOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), X_OFFSET_PARAMETER));
+        double yOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Y_OFFSET_PARAMETER));
+        double zOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), Z_OFFSET_PARAMETER));
+        Location l = new Location(projector.getWorld(), projector.getX() + xOffset, projector.getY() + yOffset, projector.getZ() + zOffset);
```
LinoxGH, line 147:
```suggestion
        Location l = projector.getLocation(xOffset, yOffset, zOffset).add();
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +172,62 @@ private static void remove(Block b) {
             hologram.remove();
         }
     }
+
+    private static void createSelector(int menuSlot, char coordinate, char direction, String parameter, Block projector, ChestMenu menu) {
```
LinoxGH, line 176: Why don't you make a more general method? Like:
```
private void createSelector(int slot, double change, ChestMenu menu, Block projector, ItemStack button) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +172,62 @@ private static void remove(Block b) {
...
+            BlockStorage.addBlockInfo(projector, parameter,
+                String.valueOf(storedOffset + offset));
+            openEditor(pl, projector);
+            return false;
+        });
+    }
+
+    private static void moveHologram(ArmorStand hologram, double x, double y, double z) {
+        hologram.teleport(hologram.getLocation().add(x, y, z));
+    }
```
LinoxGH, line 232: This one is explicitly redundant.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
LinoxGH, line 176: Why is this even static?

LinoxGH (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: image

LinoxGH: > image
> 

So what?

TheBusyBiscuit: This should definitely be final, just absolutely not static.

NCBPFluffyBear: image

TheBusyBiscuit, line 39: I think a ranged Setting would be quite useful here, I quickly threw together one for you:
https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/settings/DoubleRangeSetting.java

(Yes, I literally just copied the file on my phone and replaced int with double, deal with it.

LinoxGH: > image
> 

And why do you even have that method static? You are overusing it...

NCBPFluffyBear: > > image
> 
> So what?

maxDistance is never modified

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -27,12 +32,18 @@
 
 public class HologramProjector extends SlimefunItem {
 
-    private static final String OFFSET_PARAMETER = "offset";
+    private static final String X_OFFSET_PARAMETER = "x_offset";
+    private static final String Y_OFFSET_PARAMETER = "y_offset";
+    private static final String Z_OFFSET_PARAMETER = "z_offset";
+    private static final int[] BORDER = {1, 3, 5, 7, 8, 10, 12, 14, 16, 17};
+
+    private final DoubleRangeSetting maxDistance = new DoubleRangeSetting("max-horizontal-offset", 0.0, 5.0, 10.0);
```
TheBusyBiscuit, line 40: Let's rather declare this with the base type.
```suggestion
    private final ItemSetting<Double> maxDistance = new DoubleRangeSetting("max-horizontal-offset", 0.0, 5.0, 10.0);
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
...
+                hologram.teleport(hologram.getLocation().add(offset, 0, 0));
+            } else if (coordinate == 'Y') {
+                hologram.teleport(hologram.getLocation().add(0, offset, 0));
+            } else if (coordinate == 'Z') {
+                hologram.teleport(hologram.getLocation().add(0, 0, offset));
+            }
+
+            BlockStorage.addBlockInfo(projector, parameter,
+                String.valueOf(storedOffset + offset));
+            openEditor(pl, projector);
```
TheBusyBiscuit, line 204:
```suggestion
            BlockStorage.addBlockInfo(projector, parameter, String.valueOf(storedOffset + offset));
            openEditor(pl, projector);
```

NCBPFluffyBear:
When moving the hologram, sometimes this happens: 
image

If a player tries to break it, they might be able to

TheBusyBiscuit:
> When moving the hologram, sometimes this happens:
> image
> 
> If a player tries to break it, they might be able to

You mean the floating point value? Yeah, you would have needed to account for it hence why I commented on you only calling fix on one of the values.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -1,5 +1,7 @@
 package io.github.thebusybiscuit.slimefun4.implementation.items.blocks;
 
+import io.github.thebusybiscuit.slimefun4.api.items.settings.DoubleRangeSetting;
```
LinoxGH, line 3: Misplaced import, again...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
             hologram.remove();
         }
     }
+
+    private void createSelector(int menuSlot, ItemStack button, char coordinate, char direction, String parameter, float smallChange, float largeChange, Block projector, ChestMenu menu) {
```
LinoxGH, line 177:
```suggestion
    private void createSelector(int menuSlot, @Nonnull ItemStack button, char coordinate, char direction, @Nonnull String parameter, float smallChange, float largeChange, @Nonnull Block projector, @Nonnull ChestMenu menu) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
...
+            }
+            ArmorStand hologram = getArmorStand(projector, true);
+
+            if (coordinate == 'X') {
+                hologram.teleport(hologram.getLocation().add(offset, 0, 0));
+            } else if (coordinate == 'Y') {
+                hologram.teleport(hologram.getLocation().add(0, offset, 0));
+            } else if (coordinate == 'Z') {
+                hologram.teleport(hologram.getLocation().add(0, 0, offset));
+            }
```
LinoxGH, line 200:
```
hologram.teleport(hologram.getLocation().add(xOffset, yOffset, zOffset);
```

Make the offsets parameters, and remove these parameters:
coordinate, direction, smallChange, largeChange

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
...
+
+            BlockStorage.addBlockInfo(projector, parameter,
+                String.valueOf(storedOffset + offset));
```
LinoxGH, line 203: Check which one of the offsets are bigger than like 0.05 and set its value.

LinoxGH (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit: Or just `@ParametersAreNonnullByDefault`

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: x and z are handled differently than y, wouldn't giving them all their own variables me less efficient?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: This just shifts the if/else statements down.. and feels jankier?

LinoxGH: This was a continuation of the change requested above.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
LinoxGH: Well not sure tbh I just thought it would look  bit better. @TheBusyBiscuit what do you think?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: I know

LinoxGH: Now that I think about it, hmm yes that might be right. A char check is checking 2 bytes, whereas double check is mucu more complicated. Okay then keep it likewise. I'll review this with a fresh mind tomorrow.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +78,30 @@ public BlockUseHandler onRightClick() {
...
             if (BlockStorage.getLocationInfo(b.getLocation(), "owner").equals(p.getUniqueId().toString())) {
                 openEditor(p, b);
+            } else {
+                SlimefunPlugin.getLocalization().sendMessage(p, "no-permission", true);
             }
         };
     }
 
-    private static void openEditor(Player p, Block projector) {
+    private void openEditor(Player p, Block projector) {
```
TheBusyBiscuit, line 96: Nonnull? :eyes:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +78,30 @@ public BlockUseHandler onRightClick() {
...
         ChestMenu menu = new ChestMenu(SlimefunPlugin.getLocalization().getMessage(p, "machines.HOLOGRAM_PROJECTOR.inventory-title"));
 
-        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", "&r" + ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"))));
+        for (int s : BORDER) {
+            menu.addItem(s, new CustomItem(ChestMenuUtils.getBackground(), " "));
+            menu.addMenuClickHandler(s, ChestMenuUtils.getEmptyClickHandler());
+        }
+
+        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", ChatColor.WHITE + ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"))));
```
TheBusyBiscuit, line 104: You could do this to make it a bit more readable.
```suggestion
        String text = ChatColor.WHITE + ChatColors.color(BlockStorage.getLocationInfo(projector.getLocation(), "text"));
        menu.addItem(0, new CustomItem(Material.NAME_TAG, "&7Text &e(Click to edit)", "", text));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit, line 177: Nonnull :eyes:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
             hologram.remove();
         }
     }
+
+    private void createSelector(int menuSlot, ItemStack button, char coordinate, char direction, String parameter, float smallChange, float largeChange, Block projector, ChestMenu menu) {
+
+        menu.addItem(menuSlot, new CustomItem(button,
+            "&7" + coordinate + " Offset (" + direction + ") &e" + DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), parameter))),
+            "", "&eLeft Click: &7+0.1", "&eRight Click: &7+0.5"));
```
TheBusyBiscuit, line 181:
```suggestion
        String currentOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), parameter)));
        menu.addItem(menuSlot, new CustomItem(button, "&7" + coordinate + " Offset (" + direction + ") &e" + currentOffset, "", "&eLeft Click: &7+0.1", "&eRight Click: &7+0.5"));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
...
+        menu.addMenuClickHandler(menuSlot, (pl, slot, item, action) -> {
+            double storedOffset = Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), parameter));
+            double offset;
+
+            offset = action.isRightClicked() ? largeChange : smallChange;
```
TheBusyBiscuit, line 186:
```suggestion
            double offset  = action.isRightClicked() ? largeChange : smallChange;
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
...
+
+            if ((coordinate == 'X' || coordinate == 'Z')
+                && offset + storedOffset > maxDistance.getValue()) {
```
TheBusyBiscuit, line 189:
```suggestion
            if ((coordinate == 'X' || coordinate == 'Z') && offset + storedOffset > maxDistance.getValue()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +173,36 @@ private static void remove(Block b) {
...
+                return false;
+            }
+            ArmorStand hologram = getArmorStand(projector, true);
```
TheBusyBiscuit, line 192:
```suggestion
            }

            ArmorStand hologram = getArmorStand(projector, true);
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit, line 203:
```suggestion
            BlockStorage.addBlockInfo(projector, parameter, String.valueOf(storedOffset + offset));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,25 +116,36 @@ private static void openEditor(Player p, Block projector) {
...
         });
 
+        createSelector(2, HeadTexture.RED_UP_ARROW.getAsItemStack(), 'X', '+', X_OFFSET_PARAMETER, 0.1F, 0.5F, projector, menu);
+        createSelector(11, HeadTexture.RED_DOWN_ARROW.getAsItemStack(), 'X', '-', X_OFFSET_PARAMETER, -0.1F, -0.5F, projector, menu);
+
+        createSelector(4, HeadTexture.LIME_UP_ARROW.getAsItemStack(), 'Y', '+', Y_OFFSET_PARAMETER, 0.1F, 0.5F, projector, menu);
+        createSelector(13, HeadTexture.LIME_DOWN_ARROW.getAsItemStack(), 'Y', '-', Y_OFFSET_PARAMETER, -0.1F, -0.5F, projector, menu);
+
+        createSelector(6, HeadTexture.BLUE_UP_ARROW.getAsItemStack(), 'Z', '+', Z_OFFSET_PARAMETER, 0.1F, 0.5F, projector, menu);
+        createSelector(15, HeadTexture.BLUE_DOWN_ARROW.getAsItemStack(), 'Z', '-', Z_OFFSET_PARAMETER, -0.1F, -0.5F, projector, menu);
```
TheBusyBiscuit, line 138: Not the biggest fan of using characters here...
I think you should create an enum for this, positive x, negative x, positive y, etc...
This could be placed into the `utils.holograms` package

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -42,7 +53,9 @@ private BlockPlaceHandler onPlace() {
             public void onPlayerPlace(BlockPlaceEvent e) {
                 Block b = e.getBlockPlaced();
                 BlockStorage.addBlockInfo(b, "text", "Edit me via the Projector");
-                BlockStorage.addBlockInfo(b, OFFSET_PARAMETER, "0.5");
+                BlockStorage.addBlockInfo(b, X_OFFSET_PARAMETER, "0.5");
+                BlockStorage.addBlockInfo(b, Y_OFFSET_PARAMETER, "0.5");
+                BlockStorage.addBlockInfo(b, Z_OFFSET_PARAMETER, "0.5");
```
TheBusyBiscuit, line 58: You are using "0.5" a freaking lot here, I think you should make a final field for the default values here and just read from that...
If we ever wanna change these, we'd need to change them in a thousand places. This is not feasible.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit: I think it would be fine like that, but I am suggesting an enum for this anyway, so you could move this into the enum as a method anyway.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: What if I grabbed the sign off of the double

TheBusyBiscuit: I mean it would be better. But an enum would still be ideal for this. Especially because passing the axis via a character is not really a good design choice.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -27,12 +34,21 @@
...
+    private static final String X_OFFSET_PARAMETER = "x_offset";
+    private static final String Y_OFFSET_PARAMETER = "y_offset";
+    private static final String Z_OFFSET_PARAMETER = "z_offset";
+    private static final int[] BORDER = {1, 3, 5, 7, 8, 10, 12, 14, 16, 17};
 
-    public HologramProjector(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, ItemStack recipeOutput) {
+    private final ItemSetting<Double> maxDistance = new DoubleRangeSetting("max-horizontal-offset", 0.0, 5.0, 10.0);
+
+    public HologramProjector(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe,
+                             ItemStack recipeOutput) {
```
LinoxGH, line 47: Why 2 lines here? Also `@Nonnull` tags.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +83,33 @@ public BlockUseHandler onRightClick() {
...
                 openEditor(p, b);
+            } else {
+                SlimefunPlugin.getLocalization().sendMessage(p, "no-permission", true);
             }
         };
     }
 
-    private static void openEditor(Player p, Block projector) {
-        ChestMenu menu = new ChestMenu(SlimefunPlugin.getLocalization().getMessage(p, "machines.HOLOGRAM_PROJECTOR.inventory-title"));
+    private void openEditor(Player p, @Nonnull Block projector) {
```
LinoxGH, line 102:
```suggestion
    private void openEditor(@Nonnull Player p, @Nonnull Block projector) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +83,33 @@ public BlockUseHandler onRightClick() {
...
+        ChestMenu menu = new ChestMenu(SlimefunPlugin.getLocalization().getMessage(p, "machines.HOLOGRAM_PROJECTOR" +
+                ".inventory-title"));
```
LinoxGH, line 104: Why 2 lines here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +179,63 @@ private static void remove(Block b) {
             hologram.remove();
         }
     }
+
+    private void createSelector(Type type, @Nonnull Block projector, @Nonnull ChestMenu menu) {
+
+        double currentOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(
+                projector.getLocation(), type.parameter)));
+
+        menu.addItem(type.slot, new CustomItem(type.button.getAsItemStack(), "&7" + type.coordinate + " Offset (" + type.direction + ") &e" + currentOffset, "", "&eLeft Click: &7" + type.direction + Math.abs(type.smallChange), "&eRight Click: &7" + type.direction + Math.abs(type.smallChange)));
```
LinoxGH, line 188: This one needs a bit refactoring.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +179,63 @@ private static void remove(Block b) {
...
+        });
+    }
+
+    private enum Type {
+        X_POSITIVE(2, HeadTexture.RED_UP_ARROW, 'X', '+', X_OFFSET_PARAMETER, 0.1F, 0.5F),
+        X_NEGATIVE(11, HeadTexture.RED_DOWN_ARROW, 'X', '-', X_OFFSET_PARAMETER, -0.1F, -0.5F),
+        Y_POSITIVE(4, HeadTexture.LIME_UP_ARROW, 'Y', '+', Y_OFFSET_PARAMETER, 0.1F, 0.5F),
+        Y_NEGATIVE(13, HeadTexture.LIME_DOWN_ARROW, 'Y', '-', Y_OFFSET_PARAMETER, -0.1F, -0.5F),
+        Z_POSITIVE(6, HeadTexture.BLUE_UP_ARROW, 'Z', '+', Z_OFFSET_PARAMETER, 0.1F, 0.5F),
+        Z_NEGATIVE(15, HeadTexture.BLUE_DOWN_ARROW, 'Z', '-', Z_OFFSET_PARAMETER, -0.1F, -0.5F);
```
LinoxGH, line 220: Put some spaces between x, y and z coords. Like:

```suggestion
        X_POSITIVE(2, HeadTexture.RED_UP_ARROW, 'X', '+', X_OFFSET_PARAMETER, 0.1F, 0.5F),
        X_NEGATIVE(11, HeadTexture.RED_DOWN_ARROW, 'X', '-', X_OFFSET_PARAMETER, -0.1F, -0.5F),

        Y_POSITIVE(4, HeadTexture.LIME_UP_ARROW, 'Y', '+', Y_OFFSET_PARAMETER, 0.1F, 0.5F),
        Y_NEGATIVE(13, HeadTexture.LIME_DOWN_ARROW, 'Y', '-', Y_OFFSET_PARAMETER, -0.1F, -0.5F),

        Z_POSITIVE(6, HeadTexture.BLUE_UP_ARROW, 'Z', '+', Z_OFFSET_PARAMETER, 0.1F, 0.5F),
        Z_NEGATIVE(15, HeadTexture.BLUE_DOWN_ARROW, 'Z', '-', Z_OFFSET_PARAMETER, -0.1F, -0.5F);
```

LinoxGH (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: Passes 100 character mark

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: Passes 100 character mark

TheBusyBiscuit: There is no 100 character mark.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
NCBPFluffyBear: How would you like it to be refactored?

LinoxGH (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -65,16 +83,32 @@ public BlockUseHandler onRightClick() {
             Player p = e.getPlayer();
             Block b = e.getClickedBlock().get();
 
+            // Updates old holograms to use new coordinate system
+            if (BlockStorage.getLocationInfo(b.getLocation(), Y_OFFSET_PARAMETER) == null) {
+                BlockStorage.addBlockInfo(b, X_OFFSET_PARAMETER, DEFAULT_OFFSET);
+                BlockStorage.addBlockInfo(b, Y_OFFSET_PARAMETER, BlockStorage.getLocationInfo(b.getLocation(), "offset")
+                );
```
TheBusyBiscuit, line 90: Formatting
```suggestion
                BlockStorage.addBlockInfo(b, Y_OFFSET_PARAMETER, BlockStorage.getLocationInfo(b.getLocation(), "offset"));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -27,12 +34,21 @@
 
 public class HologramProjector extends SlimefunItem {
 
-    private static final String OFFSET_PARAMETER = "offset";
+    private final String DEFAULT_OFFSET = "0.5";
```
TheBusyBiscuit, line 37: This should rather be static, since it is a constant anyway.

```suggestion
    private static final String DEFAULT_OFFSET = "0.5";
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -89,25 +123,34 @@ private static void openEditor(Player p, Block projector) {
...
 
-        menu.addItem(1, new CustomItem(Material.CLOCK, "&7Offset: &e" + DoubleHandler.fixDouble(Double.valueOf(BlockStorage.getLocationInfo(projector.getLocation(), OFFSET_PARAMETER)) + 1.0D), "", "&rLeft Click: &7+0.1", "&rRight Click: &7-0.1"));
-        menu.addMenuClickHandler(1, (pl, slot, item, action) -> {
-            double offset = DoubleHandler.fixDouble(Double.valueOf(BlockStorage.getLocationInfo(projector.getLocation(), OFFSET_PARAMETER)) + (action.isRightClicked() ? -0.1F : 0.1F));
+        menu.addItem(9, new CustomItem(Material.CLOCK, "&cReset Position", "", "&eClick &7to reset hologram location"));
+        menu.addMenuClickHandler(9, (pl, slot, item, action) -> {
             ArmorStand hologram = getArmorStand(projector, true);
-            Location l = new Location(projector.getWorld(), projector.getX() + 0.5, projector.getY() + offset, projector.getZ() + 0.5);
-            hologram.teleport(l);
+            hologram.teleport(projector.getLocation().add(0.5, 0.5, 0.5));
```
TheBusyBiscuit, line 129: This magic value here isn't really ideal. A static constant for it would be better. Maybe change the `DEFAULT_OFFSET` to be a `double` and then just do `String.valueOf(...)` when it is needed as a String.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +178,65 @@ private static void remove(Block b) {
             hologram.remove();
         }
     }
+
+    private void createSelector(Type type, @Nonnull Block projector, @Nonnull ChestMenu menu) {
+
+        double currentOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(
+                projector.getLocation(), type.parameter)));
```
TheBusyBiscuit, line 185: Formatting.
```suggestion
        double currentOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), type.parameter)));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +178,65 @@ private static void remove(Block b) {
...
+        Z_POSITIVE(6, HeadTexture.BLUE_UP_ARROW, 'Z', '+', Z_OFFSET_PARAMETER, 0.1F, 0.5F),
+        Z_NEGATIVE(15, HeadTexture.BLUE_DOWN_ARROW, 'Z', '-', Z_OFFSET_PARAMETER, -0.1F, -0.5F);
+
+        int slot;
+        HeadTexture button;
+        char coordinate;
+        char direction;
+        String parameter;
+        float smallChange;
+        float largeChange;
```
TheBusyBiscuit, line 229: Missing access modifiers (private)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +178,65 @@ private static void remove(Block b) {
...
+
+        Type(int slot, HeadTexture button,
+             char coordinate, char direction, String parameter, float smallChange, float largeChange) {
```
TheBusyBiscuit, line 232: Formatting.
```suggestion
        Type(int slot, HeadTexture button, char coordinate, char direction, String parameter, float smallChange, float largeChange) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +178,65 @@ private static void remove(Block b) {
...
+                hologram.teleport(hologram.getLocation().add(0, 0, offset));
+            }
+
+            BlockStorage.addBlockInfo(projector, type.parameter, String.valueOf(storedOffset + offset));
+            openEditor(pl, projector);
+            return false;
+        });
+    }
+
+    private enum Type {
```
TheBusyBiscuit, line 213: Do **NOT** declare the enum inside of a class, every class/enum should have its own file.

TheBusyBiscuit, line 213: Also the name "Type" is not very descriptive and too generic.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +178,65 @@ private static void remove(Block b) {
...
+        Y_POSITIVE(4, HeadTexture.LIME_UP_ARROW, 'Y', '+', Y_OFFSET_PARAMETER, 0.1F, 0.5F),
+        Y_NEGATIVE(13, HeadTexture.LIME_DOWN_ARROW, 'Y', '-', Y_OFFSET_PARAMETER, -0.1F, -0.5F),
+
+        Z_POSITIVE(6, HeadTexture.BLUE_UP_ARROW, 'Z', '+', Z_OFFSET_PARAMETER, 0.1F, 0.5F),
+        Z_NEGATIVE(15, HeadTexture.BLUE_DOWN_ARROW, 'Z', '-', Z_OFFSET_PARAMETER, -0.1F, -0.5F);
+
+        int slot;
+        HeadTexture button;
+        char coordinate;
+        char direction;
```
TheBusyBiscuit, line 226: Using chars for this is kinda meh...
Especially the direction, it is a binary value, so it just screams for booleans.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
TheBusyBiscuit, line 229: Also missing getters.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +178,65 @@ private static void remove(Block b) {
...
+        double currentOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(
+                projector.getLocation(), type.parameter)));
+
+        menu.addItem(type.slot, new CustomItem(type.button.getAsItemStack(), "&7" + type.coordinate + " Offset (" + type.direction + ") &e" + currentOffset, "", "&eLeft Click: &7" + type.direction + Math.abs(type.smallChange), "&eRight Click: &7" + type.direction + Math.abs(type.largeChange)));
+        menu.addMenuClickHandler(type.slot, (pl, slot, item, action) -> {
+            double storedOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(projector.getLocation(), type.parameter)));
+            double offset = action.isRightClicked() ? type.largeChange : type.smallChange;
+
+            if ((type.coordinate == 'X' || type.coordinate == 'Z') && offset + storedOffset >
+                    maxDistance.getValue()) {
```
TheBusyBiscuit, line 193: formatting
```suggestion
            if ((type.coordinate == 'X' || type.coordinate == 'Z') && offset + storedOffset > maxDistance.getValue()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -135,4 +178,65 @@ private static void remove(Block b) {
             hologram.remove();
         }
     }
+
+    private void createSelector(Type type, @Nonnull Block projector, @Nonnull ChestMenu menu) {
+
+        double currentOffset = DoubleHandler.fixDouble(Double.parseDouble(BlockStorage.getLocationInfo(
+                projector.getLocation(), type.parameter)));
+
+        menu.addItem(type.slot, new CustomItem(type.button.getAsItemStack(), "&7" + type.coordinate + " Offset (" + type.direction + ") &e" + currentOffset, "", "&eLeft Click: &7" + type.direction + Math.abs(type.smallChange), "&eRight Click: &7" + type.direction + Math.abs(type.largeChange)));
```
TheBusyBiscuit, line 187: Also do **NOT** access field values directly, always use a getter.

TheBusyBiscuit (changes requested):
Lots of room for improvement

TheBusyBiscuit:
Closing this since there hasn't been a response in over two months now.

PR #2435, opened 2020-10-10
Stop unwanted actions from StrangeNetherGoo and Heavy Cream

### LINKED ISSUE #2420: Strange Nether Goo can be used to dye sign texts

The Nether Goo is actually purple dye, it can be used to dye sign texts.

1. Give Strange Nether Goo.
2. Place sign with text.
3. Right click to sign with Strange Nether Goo

Strange Nether Goo can not used to dye sign texts

No error report.

[16:37:52 INFO]: This Server uses the following setup of Slimefun:
[16:37:52 INFO]: Paper v1_16_R2
[16:37:52 INFO]: CS-CoreLib vDEV - 91 (git e42a05d1)
[16:37:52 INFO]: Slimefun vDEV - 708 (git de5db265)
[16:37:52 INFO]: Metrics build: #17
[16:37:52 INFO]:
[16:37:52 INFO]: Installed Addons: (0)

### LINKED ISSUE #2422: Heavy Cream can be used as snowball

The Heavy Cream can be used as snowball.

1. Give Heavy Cream item.
2. Right click to air with Heavy Cream.

Heavy Cream can not used as snowball.

No error report.

[16:37:52 INFO]: This Server uses the following setup of Slimefun:
[16:37:52 INFO]: Paper v1_16_R2
[16:37:52 INFO]: CS-CoreLib vDEV - 91 (git e42a05d1)
[16:37:52 INFO]: Slimefun vDEV - 708 (git de5db265)
[16:37:52 INFO]: Metrics build: #17
[16:37:52 INFO]:
[16:37:52 INFO]: Installed Addons: (0)

TheSilentPro:
## Description
Modified StrangeNetherGoo to disable rightclick action (Prevents the dying of signs)
Added HeavyCream class to disable rightclick action (Prevents throwing of the snowball)

## Changes
Added ItemUseHandler to StrangeNetherGoo
Created HeavyCream.java with ItemUseHandler

## Related Issues
Resolves #2420
Resolves #2422

## Checklist


LinoxGH:
You should instead listen to events of these actions and check if they were done by a slimefun item. If so cancel. This is both resource intensive and less maintainable.

LinoxGH:
This way you are not only making it so signs cannot be dyed by nether goo, but you also make player unable to interact with snything using that item.

TheSilentPro:
> You should instead listen to events of these actions and check if they were done by a slimefun item. If so cancel. This is both resource intensive and less maintainable.

I mean PlayerRightClickEvent already listens for a PlayerInteractEvent, wouldn't creating a listener for this just be adding another one for no reason?

TheSilentPro:
> This way you are not only making it so signs cannot be dyed by nether goo, but you also make player unable to interact with anything using that item. Including like workbenches, which are fine to be interacted with nether goo

Yeah, i forgot about that

LinoxGH:
> > You should instead listen to events of these actions and check if they were done by a slimefun item. If so cancel. This is both resource intensive and less maintainable.
> 
> I mean PlayerRightClickEvent already listens for a PlayerInteractEvent, wouldn't creating a listener for this just be adding another one for no reason?

No ehat you should do is listen to SignChangeEvent which I believe is called.

TheSilentPro:
> > > You should instead listen to events of these actions and check if they were done by a slimefun item. If so cancel. This is both resource intensive and less maintainable.
> > 
> > 
> > I mean PlayerRightClickEvent already listens for a PlayerInteractEvent, wouldn't creating a listener for this just be adding another one for no reason?
> 
> No what you should do is listen to SignChangeEvent which I believe is called.
> 
> You should absolutely not cancel all interact events with that item in player's hand.

Wouldn't it be better to listen to a PlayerInteractListener instead of SignChangeEvent. What if there are other actions that have to be cancelled, then we would just be adding listeners for everything?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
```diff
@@ -0,0 +1,28 @@
...
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+import org.bukkit.inventory.ItemStack;
+
+/**
+ * This {@link SlimefunItem} can be obtained by crafting and, it's
+ * used for various food recipes recipes
+ *
+ * @author TheSilentPro
+ */
+public class HeavyCream extends SlimefunItem {
```
TheBusyBiscuit, line 17: You can also just make this a `SimpleSlimefunItem<ItemUseHandler>` to avoid the class casting.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
```diff
@@ -0,0 +1,28 @@
...
+
```
TheBusyBiscuit, line 18: ParamsAreNonnullByDefault

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
```diff
@@ -37,4 +39,9 @@ public int getBarteringLootChance() {
         return chance.getValue();
     }
 
+    @Override
+    public void preRegister() {
+        addItemHandler((ItemUseHandler) PlayerRightClickEvent::cancel);
```
TheBusyBiscuit, line 44: Again, SimpleSlimefunItem<ItemUseHandler> and probably only cancel it for signs

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
TheSilentPro: Add it to the class?
May you tell me what that does, haven't really seen it in other sf classes?

LinoxGH:
> > > > You should instead listen to events of these actions and check if they were done by a slimefun item. If so cancel. This is both resource intensive and less maintainable.
> > > 
> > > 
> > > I mean PlayerRightClickEvent already listens for a PlayerInteractEvent, wouldn't creating a listener for this just be adding another one for no reason?
> > 
> > No what you should do is listen to SignChangeEvent which I believe is called.
> > 
> > You should absolutely not cancel all interact events with that item in player's hand.
> 
> Wouldn't it be better to listen to a PlayerInteractListener (then do necessary checks) instead of SignChangeEvent. What if there are other actions that have to be cancelled, then we would just be adding listeners for everything?

That would work out fine too, I guess.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
TheBusyBiscuit: To the constructor, it marks the parameters to be "non-null".
It is used for most of our constructors and methods.

I see the preview being a bit inaccurate there...
I marked the line above the constructor but forgot that it doesn't show the constructor below in the preview :/

LinoxGH:
For Heavy Cream you should check if player launched a projectile, by its specific event.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
```diff
@@ -37,4 +43,18 @@ public int getBarteringLootChance() {
...
 
+    @Nonnull
+    @Override
+    public ItemUseHandler getItemHandler() {
+        return this::use;
+    }
+
+    public void use(PlayerRightClickEvent e) {
+        if (e.getClickedBlock().isPresent()) {
+            if (e.getClickedBlock().get().getType().name().contains("SIGN")) {
```
LinoxGH, line 54: Use tags.

LinoxGH (changes requested):

TheSilentPro:
> For Heavy Cream you should check if player launched a projectile, by its specific event.

But What's wrong with doing it this way?

LinoxGH:
> > For Heavy Cream you should check if player launched a projectile, by its specific event.
> 
> But What's wrong with doing it this way?

Just like we said now people cannot interact with anything if they hold a heavy cream.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
TheBusyBiscuit: Tag.SIGNS please

TheBusyBiscuit:
Can also just check if the material is interactable.

LinoxGH:
> Can also just check if the material is interactable.

That's what he does rn but the issue with heavy cream is that players can throw it. Why do we want to restrict players from opening chests and etc. while it's in their hand.

TheSilentPro:
> > Can also just check if the material is interactable.
> 
> That's what he does rn but the issue with heavy cream is that players can throw it. Why do we want to restrict players from opening chests and etc. while it's in their hand.

But i already check for that?
image
This fires when they try to interact with an interactible block like crafting table, dispenser etc..

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
TheSilentPro: Didn't know that was a thing, i will get on it sorry.

LinoxGH:
> > > Can also just check if the material is interactable.
> > 
> > That's what he does rn but the issue with heavy cream is that players can throw it. Why do we want to restrict players from opening chests and etc. while it's in their hand.
> 
> But i already check for that?
> image
> This fires when they try to interact with an interactible block like crafting table, dispenser etc..

Exactly.

TheSilentPro:
> > > > Can also just check if the material is interactable.
> > > 
> > > 
> > > That's what he does rn but the issue with heavy cream is that players can throw it. Why do we want to restrict players from opening chests and etc. while it's in their hand.
> > 
> > 
> > But i already check for that?
> > image
> > This fires when they try to interact with an interactible block like crafting table, dispenser etc..
> 
> Yes but still the same code will be needed to be done for any other future snowball slimefun item.
> 
> Whereas with listening to the projectile event and cancelling it if the projectile is ANY slimefun item will automatically fix that.

True, but wouldn't this interfere with any slimefun items that might be throwable? Cancelling that will break them.
(Though i can't think of any that are throwable, they might be added at some point?)

LinoxGH:
> > > > > Can also just check if the material is interactable.
> > > > 
> > > > 
> > > > That's what he does rn but the issue with heavy cream is that players can throw it. Why do we want to restrict players from opening chests and etc. while it's in their hand.
> > > 
> > > 
> > > But i already check for that?
> > > image
> > > This fires when they try to interact with an interactible block like crafting table, dispenser etc..
> > 
> > Yes but still the same code will be needed to be done for any other future snowball slimefun item.
> > 
> > Whereas with listening to the projectile event and cancelling it if the projectile is ANY slimefun item will automatically fix that.
> 
> True, but wouldn't this interfere with any slimefun items that might be throwable? Cancelling that will break them.
> (Though i can't think of any that are throwable, they might be added at some point?)

Making an override in a listener class is much much more easier than implementing the same thing for x amount of items.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
```diff
@@ -0,0 +1,45 @@
...
+        super(category, item, recipeType, recipe, recipeOutput);
+    }
+
+    @Nonnull
+    @Override
+    public ItemUseHandler getItemHandler() {
+        return this::use;
+    }
+
+    public void use(PlayerRightClickEvent e) {
```
TheBusyBiscuit, line 34: First off: This should NOT be public.
Also a missing nullability annotation.
But most importantly, there is no need for this method, this can all easily be done as a lambda expression.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
```diff
@@ -0,0 +1,45 @@
...
+        if (e.getClickedBlock().isPresent()) {
+            if (!e.getClickedBlock().get().getType().isInteractable()) {
+                e.cancel();
+            }
+        }
+        else {
```
TheBusyBiscuit, line 40: Formatting
```suggestion
        } else {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
```diff
@@ -37,4 +44,18 @@ public int getBarteringLootChance() {
         return chance.getValue();
     }
 
+    @Nonnull
+    @Override
+    public ItemUseHandler getItemHandler() {
+        return this::use;
+    }
+
+    public void use(PlayerRightClickEvent e) {
```
TheBusyBiscuit, line 53: First off: This should NOT be public.
Also a missing nullability annotation.
But most importantly, there is no need for this method, this can all easily be done as a lambda expression.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
```diff
@@ -37,4 +44,18 @@ public int getBarteringLootChance() {
...
+        if (e.getClickedBlock().isPresent()) {
+            if (Tag.SIGNS.isTagged(e.getClickedBlock().get().getType())) {
+                e.cancel();
+            }
+        }
```
TheBusyBiscuit, line 58: You can merge these two statements using &&

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
```diff
@@ -37,4 +44,18 @@ public int getBarteringLootChance() {
...
 
+    @Nonnull
+    @Override
+    public ItemUseHandler getItemHandler() {
+        return this::use;
+    }
+
+    public void use(PlayerRightClickEvent e) {
+        if (e.getClickedBlock().isPresent()) {
+            if (Tag.SIGNS.isTagged(e.getClickedBlock().get().getType())) {
```
TheBusyBiscuit, line 55: Calling the method twice is a bit non-optimal, you should store the Optional<Block> as a variable and use that.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Nonnull
+    @Override
+    public ItemUseHandler getItemHandler() {
+        return this::use;
+    }
+
+    public void use(PlayerRightClickEvent e) {
+        if (e.getClickedBlock().isPresent()) {
+            if (!e.getClickedBlock().get().getType().isInteractable()) {
```
TheBusyBiscuit, line 36: Calling the method twice is a bit non-optimal, you should store the Optional<Block> as a variable and use that.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
TheSilentPro: Last time i had it like that and you said to put it in the next line

TheBusyBiscuit: I know, but since many seemed to prefer the other style, we have revised our styling guidelines to now go with what the majority seems to prefer.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/HeavyCream.java
```diff
@@ -28,18 +29,16 @@ public HeavyCream(Category category, SlimefunItemStack item, RecipeType recipeTy
...
-            if (!e.getClickedBlock().get().getType().isInteractable()) {
+        return e -> {
+            Optional<Block> block = e.getClickedBlock();
+            if (block.isPresent()) {
+                if (!block.get().getType().isInteractable()) {
+                    e.cancel();
+                }
+            } else {
                 e.cancel();
             }
```
TheSilentPro: I don't think i can use && here due to that else statement, correct me if i'm wrong though

LinoxGH: Correct.

TheBusyBiscuit: > I don't think i can use && here due to that else statement, correct me if i'm wrong though

Yes buuuuuut what cou could do is the following.
```suggestion
        return e -> {
            Optional<Block> block = e.getClickedBlock();

            if (!block.isPresent() || !block.get().getType().isInteractable()) {
                e.cancel();
            }
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/misc/StrangeNetherGoo.java
```diff
@@ -47,15 +47,12 @@ public int getBarteringLootChance() {
...
     public ItemUseHandler getItemHandler() {
-        return this::use;
-    }
-
-    public void use(PlayerRightClickEvent e) {
-        if (e.getClickedBlock().isPresent()) {
-            if (Tag.SIGNS.isTagged(e.getClickedBlock().get().getType())) {
+        return e -> {
+            Optional<Block> block = e.getClickedBlock();
+            if (block.isPresent() && Tag.SIGNS.isTagged(block.get().getType())) {
```
TheBusyBiscuit, line 52: An empty line before the if always looks a bit nicer
```suggestion
            Optional<Block> block = e.getClickedBlock();

            if (block.isPresent() && Tag.SIGNS.isTagged(block.get().getType())) {
```

TheBusyBiscuit (changes requested):
Just two more things

TheSilentPro:
That should be all, sorry for all the mistakes

TheBusyBiscuit (approved):
LGTM.

The only thing I would have liked would be some comments on these changes explaining why they're here.

LinoxGH (approved):
Still I think we should have listened to projectile launch event... But anyways.

TheBusyBiscuit:
The pojectile launch event would have been more accurate, yes.
But this allows us to keep things in their respective classes.
Should there ever come up another projectile-related issue like this, we can always consider redesigning it.

PR #2440, opened 2020-10-10
Fixed #2439

### LINKED ISSUE #2439: PyroMining and Slimefun Incompatibility

Whilst using the PyroMining plugin along side Slimefun's Explosive Pickaxe, it appears that players are able to mass produce the items the plugin drops from mining the ore. I've gone ahead and talked to Pyro (developer of the PyroMining plugin), and he's mentioned that he has already set the events to the highest priority so that its more than likely Slimefun overriding something.

https://youtu.be/3wgpkvai008

Prevent the Explosive Pickaxe from being able to override the events from other plugins.

 - Server Software (Spigot/Paper):
Capture

LinoxGH:
## Description

Added a check if the event is cancelled to explosive tool.
Changed BlockBreakEvent and BlockPlaceEvent listener priorities to HIGHEST from MONITOR.

## Checklist


svr333 (dismissed):
Dont exactly know why we'd have to change from Monitor to highest, but changes look good to me

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosiveTool.java
```diff
@@ -52,6 +52,8 @@ public ExplosiveTool(Category category, SlimefunItemStack item, RecipeType recip
     @Override
     public ToolUseHandler getItemHandler() {
         return (e, tool, fortune, drops) -> {
+            if (e.isCancelled()) return;
```
TheBusyBiscuit, line 55: This is completely redundant as the Listener which calls this method already checks if it is cancelled.
Also missing brackets.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -75,7 +75,7 @@ public void onBlockPlace(BlockPlaceEvent e) {
         }
     }
 
-    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    @EventHandler(priority = EventPriority.HIGHEST, ignoreCancelled = true)
```
TheBusyBiscuit, line 78:
```suggestion
    @EventHandler(priority = EventPriority.HIGH, ignoreCancelled = true)
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
svr333: Can I ask why were even changing this?

TheBusyBiscuit: The bug was about this running AFTER HIGHEST, which should not be done anyway. By moving it TO HIGHEST we basically didn't fix the issue but rather let randomness decide.

TheBusyBiscuit: In theory events should try to run on NORMAL unless there is a valid reason not to, due to this being dependent on some things to process first, there is a point to be made to move it to HIGH but HIGHEST is definitely too late.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosiveTool.java
LinoxGH: > This is completely redundant as the Listener which calls this method already checks if it is cancelled.
> Also missing brackets.

No it doesn't

LinoxGH: It only does a material check.

TheBusyBiscuit: **ignoreCancelled = true**
```java
@EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
    public void onBlockBreak(BlockBreakEvent e) {
```

LinoxGH: > **ignoreCancelled = true**
> ```java
> @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
>     public void onBlockBreak(BlockBreakEvent e) {
> ```

Yes. So listener is fired no matter if the even is cancelled. If that worked like it ignored cancelled events, then why is there a cancelled check below?

LinoxGH: Ok just looked up, I was wrong.

TheBusyBiscuit (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosiveTool.java
svr333: ignoreCancelled is very ambiguous

svr333 (approved):

Issue #2111, opened 2020-07-22
Androids stop working and turn into CS core-lib heads


pmyers123:
## Description (Required)

When any android is given a program that includes movement off of it's initial position (not turning, but actual movement) given enough time (between 10 minutes and 4 hours on average) the androids cease to move, are no longer androids and turn back into heads.

## Steps to reproduce the Issue (Required)

When writing a script that involves android movement forward up or down, after a period of time of the android running the script, the android no longer functions. When mined, it turns into a cs core-lib head. You cannot interact with the android. I have tested this with all types of androids and the same thing happens. The video below is from a lets play. I apologize but it's the only video I have. 

https://www.youtube.com/watch?v=chHG6aJVsGw&t=74s

You should be able to see in the first few minutes the farm design i made and will see a number of androids have stopped. In the video i attempt to show you that the android interface cannot be brought up by right clicking the android. Instead it is now simply a head. 

In the video, the script that the farmer android has is as follows

Pull fuel, turn left, go forward, turn left (advanced harvest and replant, go forward) 9 times, push items into interface, turn left, go forward, turn left, (advanced harvest and replant, go forward) 9 times, then repeat.

## Expected behavior (Required)
 

The androids should continue to function as expected instead of turning into heads.

## Server Log / Error Report

https://github.com/TheBusyBiscuit/Slimefun4/files/4962340/2020-07-11-2.log.zip

The error logs in slimefun only contained information relating to a barrel bug that has also been reported.

## Environment (Required)

![image](https://user-images.githubusercontent.com/40214829/88232209-14a95580-cc3b-11ea-8102-40b97df8ee97.png)

 - Minecraft Version: Minecraft 1.15.2
 - Slimefun Version:

 - CS-CoreLib Version:

TheBusyBiscuit:
Malformed issue, and the video link seems to be wrong.

TheBusyBot:
Your issue was closed, it may fall under one or more of the following categories.
**Please wait for an admin to tick off the points that apply.**

Please respond below, if you have any questions.
Do **not** open a new Issue unless explicitly told otherwise, comment below or edit your post instead.

* [ ] You did not follow our template. Please follow the Issue template to help us identify your issue more effectively.
* [ ] You did not provide any information about your versions (We absolutely need the exact version numbers that are installed on your Server, \"latest\" is not helpful)
[...]

PR #2444, opened 2020-10-11
Android fix

### LINKED ISSUE #2443: Miner Androids cause block errors while mining

When using Dig and Move Forward, androids sometimes create a block type disagreement. Here's an error:
https://pastebin.com/iNgjXg58
Once this block is corrupt, any kind of interaction produces console errors, including right clicking, left click, and breaking the block with an android. It does not behave like a ghost block, and players can place blocks inside it perfectly fine. It does not have any BlockStorage either.

In addition, I noticed that when a block is placed in a spot right after the android moves away from it, that block is broken and a Cs CoreLib head (android) appears in the android's inventory.

In addition to that addition, this also appeared and may help diagnose the problem: https://pastebin.com/fAcF5rCz

I have grouped all these issues together because they may be caused by the same source.

This issue may be related to #2408 and #2111 .

1. Place an android
2. Create a script that uses a combination of Move and Dig Forwards and Turn Right to create a loop that moves back and forth
3. After some time, errors may start to appear in console
4. Attempting to quickly place a block where the android used to be right after it moves will cause that block to be broken.

Androids should not bug out, and the tasks it performs should be executed in order.

## :scroll: Server Log

https://pastebin.com/iNgjXg58

https://pastebin.com/fAcF5rCz

 - Minecraft Version: 1.16.3
 - Slimefun Version: DEV 709 (git 4c47a)
 - CS-CoreLib Version: DEV 91 (git e42a0)

LinoxGH:
## Description

Fixes android related bugs.

## Changes

Some refactoring.
Reordering what happens in TickerTask.

## Related Issues

Supposed to fix #2443 and #2111 

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/MinerAndroid.java
```diff
@@ -22,21 +25,33 @@
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 
+/**
+ * Represents a {@link MinerAndroid Programmable Android (Miner)}.
+ * This {@link ProgrammableAndroid} can dig blocks and move.
+ *
+ * @author Linox
```
TheBusyBiscuit, line 32: You are not the sole author of this class...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/MinerAndroid.java
```diff
@@ -45,15 +60,16 @@ protected void dig(Block b, BlockMenu menu, Block block) {
...
             if (SlimefunPlugin.getProtectionManager().hasPermission(owner, block.getLocation(), ProtectableAction.BREAK_BLOCK)) {
                 AndroidMineEvent event = new AndroidMineEvent(block, new AndroidInstance(this, b));
-                Bukkit.getPluginManager().callEvent(event);
 
-                if (event.isCancelled()) {
-                    return;
+                // We only want to break non-Slimefun blocks by default, although an addon may want it differently.
+                String blockId = BlockStorage.checkID(block);
+                if (blockId != null) {
+                    event.setCancelled(true);
```
TheBusyBiscuit, line 67: This can be simplified

```suggestion
                if (BlockStorage.hasBlockInfo(block)) {
                    event.setCancelled(true);
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/MinerAndroid.java
```diff
@@ -68,40 +84,10 @@ protected void dig(Block b, BlockMenu menu, Block block) {
...
-                    move(b, face, block);
-                }
-            } else {
-                move(b, face, block);
-            }
-        } else {
-            move(b, face, block);
-        }
+        dig(b, menu, block);
+        move(b, face, block);
```
TheBusyBiscuit, line 90: I don't think this simplification works here because it would be quite different.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/MinerAndroid.java
LinoxGH: > You are not the sole author of this class...

Well idk who else is so I just added mine.

TheBusyBiscuit: `git blame`, also me most likely.

LinoxGH: Android has no git

PR #2473, opened 2020-10-14
Fixes ArmorForge

### LINKED ISSUE #2469: Unable to craft Elytra Cap without an Output Chest
## :round_pushpin: Description
When attempting to craft Elytra Cap in an Armor Forge, it writes "Inventory full", despite the fact that the slots would clear after consuming the materials. This is fixed when you put an output chest next to it as it gives it a free slot for output.

## :bookmark_tabs: Steps to reproduce the Issue
1. Attempt to craft Elytra Cap.
2. Fail

## :bulb: Expected behavior
Materials should be consumed and the Elytra Cap should appear in the dispenser of an Armor Forge

## :scroll: Server Log
None

None

image

CURVX:
## Description

#2469 Crafted Elytra Cap in ArmorForge
https://streamable.com/p28b4f

## Changes

Made `ArmorForge` extend `BackpackCrafter` instead of `MultiBlockMachine`
Used `EnchancedCraftingTable.java` as Reference to make further changes

## Related Issues

Resolves #2469 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
@@ -2,7 +2,6 @@ package io.github.thebusybiscuit.slimefun4.implementation.items.multiblocks;
 
 import java.util.List;
 
-import org.bukkit.Bukkit;
 import org.bukkit.Material;
 import org.bukkit.Sound;
 import org.bukkit.block.Block;
@@ -15,7 +14,6 @@ import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.cscorelib2.inventory.ItemUtils;
 import io.github.thebusybiscuit.cscorelib2.item.CustomItem;
-import io.github.thebusybiscuit.slimefun4.core.multiblocks.MultiBlockMachine;
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
 import io.papermc.lib.PaperLib;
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
```diff
@@ -71,26 +66,51 @@ private boolean isCraftable(Inventory inv, ItemStack[] recipe) {
...
+
+            fakeInv.setItem(j, stack);
         }
 
-        for (int j = 0; j < 4; j++) {
-            int current = j;
+        return fakeInv;
+    }
+
+    private void craft(Player p, ItemStack output, Inventory inv, Block dispenser) {
```
TheBusyBiscuit, line 86: This could use some annotations.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
```diff
@@ -71,26 +66,51 @@ private boolean isCraftable(Inventory inv, ItemStack[] recipe) {
...
+                ItemUtils.consumeItem(stack, true);
             }
+
+            fakeInv.setItem(j, stack);
         }
 
-        for (int j = 0; j < 4; j++) {
-            int current = j;
+        return fakeInv;
+    }
```
TheBusyBiscuit, line 84: Is this method copied from anywhere?
I think it would make sense to make it a static utility method or so since we would otherwise face a lot of redundancies and duplications.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
CURVX: > Is this method copied from anywhere?

Yes, from BackpackCrafter: https://github.com/Slimefun/Slimefun4/blob/c9adba2f39068490518bafe0cb8fd9bc04a7b572/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/BackpackCrafter.java#L43

How do I tackle this? Make `ArmorForge extends BackpackCrafter` instead of MultiBlockMachine since latter is already extended in BackpackCrafter. Sorry. I know its messy but I am learning and loving it so far :D

TheBusyBiscuit: I think making it extend BackpackCrafter should be fine, even though the BackpackCrafter class might need a new name after this XD

TheBusyBiscuit (approved):

poma123 (approved):
LGTM

PR #2481, opened 2020-10-17
Fix issue #2460 Explosive tools not working with silk touch

MoKe12g:
Explosive Tools no longer ignore their SILK_TOUCH enchantment.

## Description
I've added some instructions to check whether the pickaxe or shovel is enchanted with silk touch.
If the pickaxe or shovel does not have the silk touch enchantment, the old code is used.

## Changes
Added if statement for the SILK_TOUCH enchantment.

## Related Issues
Fixes PR #2460

## Checklist


LinoxGH (dismissed):

MoKe12g:
Could you please add the label "hacktoberfest-accepted" because GitHub isn't aware, that this PR has already been merged and accepted.

LinoxGH (approved):

poma123 (approved):

MoKe12g:
Is there a method for getting silk_touch block drops?
Because there blocks that shouldn't get dropped like monster spawners and also some that drop twice (beds). (Item Duplication)
I'm thinking of creating a silk_touch drop method (in BlockListener) or is there already one that I can use?
I didn't find anything in Spigot for this use case.
https://minecraft.gamepedia.com/Silk_Touch

TheBusyBiscuit:
> Is there a method for getting silk_touch block drops?
> Because there blocks that shouldn't get dropped like monster spawners and also some that drop twice (beds). (Item Duplication)
> I'm thinking of creating a silk_touch drop method (in BlockListener) or is there already one that I can use?
> I haven't found anything in the Spigot documentation for this use case.
> https://minecraft.gamepedia.com/Silk_Touch

Sadly not, that's the main reason why I held back on merging this thus far.
However, as of 1.14 `Block#getDrops(ItemStack` now takes fortune and silk touch into account.
The whole reason why we had to apply fortune manually is because the plugin is much older than this, since Slimefun is now 1.14+ only though, we can take advantage of that and just pass the actually used item to the `getDrops(...)` call instead.
This way we will not need any special cases for fortune or silk touch.

MoKe12g:
But then there is still the problem with door and beds being dropped twice.
Also on some point i got this error (https://gist.github.com/72a3bd2914d2e7af8021726f8908a803) because it tried to drop air. Should we intercept on trying to drop air, because we then have to intercept also water, lava, cave-air etc.

TheBusyBiscuit:
That's exactly why this check is in place:
https://github.com/Slimefun/Slimefun4/pull/2481/files#diff-4d641b879f584305d510ed13fabcb6be4fbd794c59b6b294ad4b2c84aefcd2efR148

You cannot drop null or air.
And on newer versions AIR is the fallback material for a lot of methods, so you need to check for it.
But that doesn't apply to anything like water or lava since you cannot mine these with a pickaxe.
Only the check for air is necessary.

PR #2484, opened 2020-10-18
Refactored Block Tickers

TheBusyBiscuit:
## Description

I am changing Block Tickers to go from an ItemHandler to a proper interface.
This makes identifying tickers easier and reduces the need for null checks on the getTicker() method, if something is a ticker, it is a ticker.

This also reduces nesting a LOT.
The only thing I am not so fond about yet is the getTickingMethod() method as it is quite a repetitive pattern.
I thought about determining this via sub interfaces of TickingBlock but those would just get confusing to implement for addon devs. The method forces them to implement it and make a decision, even if it may encourage redundance. If you have a better idea let me know.

This currently is just a wrapper for the old system, but this layer of abstraction will allow us to handle tickers internally however we want.

## Checklist


LinoxGH:
I am not sure about the priorities of ticker tasks. I think we should focus on performance improvements more than how they're accessed by addon devs, or how it looks, or anything really.

TheBusyBiscuit:
It's not a priority, hence why this is a draft.
But it's something that should be tackled to improve the readibility, accessibility and maintainability in the future.
This is also just an alternative way at first, this still leaves the old way in tact and just deprecates it, but it will allow us to make further improvements down the line, also in regards to performance, because TickingMethod is an enum compared to the boolean before, this would allow us to offer fine-tuning in terms of concurrency.

TheBusyBiscuit:
whoops

TheBusyBiscuit:
Postponed.

PR #2488, opened 2020-10-19
Added hazmat backpack

TheSilentPro:
## Description
Added hazmat backpack for storing ONLY radioactive items

## Changes
Added HazmatBackpack.java to represent a hazmat backpack
Modified isAllowed in BackpackListener.java to check for hazmat backpack and radioactive items
Added hazmat backpack item and research

## Related Issues
N/A

## Checklist


TheSilentPro:
Oops Forgot to make it so other backpacks cant have radioactive items.
Will do that tomorrow.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/backpacks/HazmatBackpack.java
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * This implementation of {@link SlimefunBackpack} is used for holding {@link Radioactive} items.
+ *
+ * @author TheSilentPro
+ *
+ */
+public class HazmatBackpack extends SlimefunBackpack {
+
+    public HazmatBackpack(int size, Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
```
TheBusyBiscuit, line 17: `@ParamsAreNonnullByDefault`

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
```diff
@@ -80,6 +80,7 @@ private SlimefunItems() {}
     public static final SlimefunItemStack BOUND_BACKPACK = new SlimefunItemStack("BOUND_BACKPACK", HeadTexture.ENDER_BACKPACK, "&cSoulbound Backpack", "", "&7Size: &e36", BACKPACK_ID, "", LoreBuilder.RIGHT_CLICK_TO_OPEN);
     public static final SlimefunItemStack COOLER = new SlimefunItemStack("COOLER", HeadTexture.COOLER, "&bCooler", "&fAllows you to store Juices/Smoothies", "&fand automatically consumes them when you are hungry", "&fand you have this in your Inventory", "", "&7Size: &e27", BACKPACK_ID, "", LoreBuilder.RIGHT_CLICK_TO_OPEN);
     public static final SlimefunItemStack RESTORED_BACKPACK = new SlimefunItemStack("RESTORED_BACKPACK", HeadTexture.RESTORED_BACKPACK, "&eRestored Backpack", "", "&7Retrieve your lost items", BACKPACK_ID, "", LoreBuilder.RIGHT_CLICK_TO_OPEN);
+    public static final SlimefunItemStack HAZMAT_BACKPACK = new SlimefunItemStack("HAZMAT_BACKPACK", HeadTexture.BACKPACK, "&2Hazmat Backpack", "", "&7Used for storing radioactive items", "", "&7Size: &e27", BACKPACK_ID, "", LoreBuilder.RIGHT_CLICK_TO_OPEN);
```
TheBusyBiscuit, line 83: This would still need a proper texture

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BackpackListener.java
```diff
@@ -105,8 +107,15 @@ private boolean isAllowed(@Nonnull SlimefunBackpack backpack, @Nullable ItemStac
         if (item == null || item.getType() == Material.AIR) {
             return true;
         }
+        SlimefunItem sfitem = SlimefunItem.getByItem(item);
+        if (backpack instanceof HazmatBackpack && !(sfitem instanceof Radioactive)) {
+            return false;
+        }
+        if (sfitem instanceof Radioactive && !(backpack instanceof HazmatBackpack)) {
+            return false;
+        }
```
TheBusyBiscuit, line 116: This logic should absolutely NOT happen here...
It should be applied in the `isItemAllowed` method of the respective backpack, non-nuclear for normal ones and also nuclear for hazmat ones.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BackpackListener.java
TheSilentPro: Should I add both logics in SlimefunBackpack?
Or would I have to add each logic for each backpack implementation of it?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BackpackListener.java
```diff
@@ -29,6 +29,8 @@
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.implementation.items.backpacks.Cooler;
 import io.github.thebusybiscuit.slimefun4.implementation.items.backpacks.SlimefunBackpack;
+import io.github.thebusybiscuit.slimefun4.core.attributes.Radioactive;
+import io.github.thebusybiscuit.slimefun4.implementation.items.backpacks.HazmatBackpack;
```
TheBusyBiscuit, line 33: Unused imports

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/backpacks/SlimefunBackpack.java
```diff
@@ -65,6 +66,12 @@ public boolean isItemAllowed(@Nonnull ItemStack item, @Nullable SlimefunItem ite
         if (SlimefunTag.SHULKER_BOXES.isTagged(item.getType())) {
             return false;
         }
+        if (SlimefunItem.getByID(getId()) instanceof HazmatBackpack && !(itemAsSlimefunItem instanceof Radioactive)) {
+            return false;
+        }
+        if (itemAsSlimefunItem instanceof Radioactive && !(SlimefunItem.getByID(getId()) instanceof HazmatBackpack)) {
```
TheBusyBiscuit, line 72: `(SlimefunItem.getByID(getId())`

WTF...

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BackpackListener.java
TheBusyBiscuit: You should generally keep the log in the respective class.
So normal SlimefunBackpack logic -> SlimefunBackpack.java
HazmatBackpack logic -> HazmatBackpack.java

The HazmatBackpack could just fallback to the super logic if it isn't radioactive, to avoid redundancies.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/backpacks/SlimefunBackpack.java
TheSilentPro: getId returns the id of the backpack

NCBPFluffyBear: you just got the id from the item then got the item from the id... loopy loopy

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/backpacks/SlimefunBackpack.java
```diff
@@ -65,7 +66,13 @@ public boolean isItemAllowed(@Nonnull ItemStack item, @Nullable SlimefunItem ite
         if (SlimefunTag.SHULKER_BOXES.isTagged(item.getType())) {
             return false;
         }
-
+        if (this instanceof HazmatBackpack && !(itemAsSlimefunItem instanceof Radioactive)) {
+            return false;
+        }
+        if (itemAsSlimefunItem instanceof Radioactive && !(this instanceof HazmatBackpack)) {
+            return false;
+        }
```
TheBusyBiscuit, line 74: This is still super redundant and should belong into the HazmatBackpack class...

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/backpacks/SlimefunBackpack.java
TheSilentPro: Well one of them prevents the radioactive item from being entered into the normal backpacks, if i put that in the HazmatBackpack wouldn't that not trigger on normal backpacks?
Also since I'd be overriding the method from here, i will need to do both checks again anyway.
Or am I trippin?

TheBusyBiscuit: Look into polymorphism.

You can override a method in child classes to replace the behaviour or fallback to the parent behaviour.
It is a very bad design pattern to do instanceof checks for child classes inside the parent class, as it renders the child class useless and goes against the principles of polymorphism.

Think about it simply like this:

What does the normal Backpack do? Allow every item except for backpacks, shulkers and radioactive materials.
The method should reflect this.

What exception does the Hazmat Backpack provide? Radioactive substances ARE allowed.
So the method **inside** HazmatBackpack should have a special case for that and fallback to the default behaviour otherwise. Since the radioactive check will no longer pass due to the prior exclusion, the instanceof check is made obsolete.

TheBusyBiscuit:
Closing this since there hasn't been a response in over two months now.

PR #2494, opened 2020-10-20
Fixes brewing potion from Slimefun dusts using hopper

### LINKED ISSUE #2447: The potion brewing works from Slimefun dusts when using hoppers

The potion brewing works from Slimefun dusts when using hoppers.

1. /sf cheat - give Lead dust.
2. Place Awkward Potion and Blaze Powder to Brewing Stand.
3. Place hopper towards to the brewing stand.
4. Place dusts to the hopper.
5. The Brewing Stand start create the potion.

88456020-abb02080-ce7a-11ea-9cbb-47a4f7b88afb

The potion brewing not work from Slimefun dusts when using hoppers.

No error report.

[15:04:52 INFO]: This Server uses the following setup of Slimefun:
[15:04:52 INFO]: Paper v1_16_R2
[15:04:52 INFO]: CS-CoreLib vDEV - 91 (git e42a05d1)
[15:04:52 INFO]: Slimefun vDEV - 712 (git 15884c65)
[15:04:52 INFO]: Metrics build: #17
[15:04:52 INFO]:
[15:04:52 INFO]: Installed Addons: (0)

CURVX:
## Description

The potion brewing no longer works from Slimefun dusts when using hoppers
https://www.youtube.com/watch?v=7aDqnIIMmhw

## Changes

Added EventHandler `hopperOnBrew` in `BrewingStandListener.java`

## Related Issues

Resolves #2447

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/BrewingStandListener.java
```diff
@@ -52,4 +53,12 @@ public void onPreBrew(InventoryClickEvent e) {
         }
     }
 
+    @EventHandler
+    public void hopperOnBrew(InventoryMoveItemEvent e) {
+        if (e.getDestination().getType().equals(InventoryType.BREWING)) {
```
TheBusyBiscuit, line 58: Enums must always be compared using ==, not .equals()

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/BrewingStandListener.java
```diff
@@ -52,4 +53,12 @@ public void onPreBrew(InventoryClickEvent e) {
...
+            if (isUnallowed(SlimefunItem.getByItem(e.getItem()))) {
```
TheBusyBiscuit, line 59:
```suggestion
            if (isUnallowed(e.getItem())) {
```

TheBusyBiscuit, line 59: Also you can connect these two if-statements using &&

TheBusyBiscuit (changes requested):

TheBusyBiscuit:
You should also keep in mind that this event gets called on every single tick and is often disabled on servers for performance reasons.
I don't really know how beneficial it would be to even fix this or just say "Live with it".

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/BrewingStandListener.java
CURVX: image


Throws NPE, when `isUnallowed` is called without. Included a null check in method `isUnallowed` of `SlimefunCraftingListener` interface and did a little refractoring in the above `hasUnallowedItems` method. Let me know if this is ok.

CURVX:
> 
> 
> You should also keep in mind that this event gets called on every single tick and is often disabled on servers for performance reasons.
> I don't really know how beneficial it would be to even fix this or just say "Live with it".

I will leave this to you as I have little to no idea about performance as of now. :D

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/SlimefunCraftingListener.java
```diff
@@ -32,7 +27,7 @@ default boolean isUnallowed(@Nullable ItemStack item) {
         }
 
         SlimefunItem sfItem = SlimefunItem.getByItem(item);
-        return !(sfItem instanceof VanillaItem) && !sfItem.isDisabled();
+        return sfItem != null && !(sfItem instanceof VanillaItem) && !sfItem.isDisabled();
```
TheBusyBiscuit, line 30: This is the exact same as line 34.
So you can very much utilise this method:

```suggestion
        return isUnallowed(sfItem);
```

TheBusyBiscuit (changes requested):
One small improvement

TheBusyBiscuit (approved):

Sfiguz7 (approved):

PR #2501, opened 2020-10-24
Changed some things to public (from protected)

Seggan:
## Description
I changed the access modifiers in `ProgrammableAndroid.getScript()` and `ProgrammableAndroid.setScript()` from protected to public so external addons can change android scrips (I need it for precisely that reason). I also changed `Instruction` enum to public

## Changes

`ProgrammableAndroid.getScript()` and `ProgrammableAndroid.setScript()` are now public
`Instruction` is now public

## Related Issues

None

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/Instruction.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/Instruction.java
@@ -102,7 +102,7 @@ public enum Instruction {
     }),
 
     ATTACK_ANIMALS_ADULT(AndroidType.FIGHTER, HeadTexture.SCRIPT_ATTACK, (android, b, inv, face) -> {
-        Predicate<LivingEntity> predicate = e -> e instanceof Animals && e instanceof Ageable && ((Ageable) e).isAdult();
+        Predicate<LivingEntity> predicate = e -> e instanceof Animals && ((Ageable) e).isAdult();
         android.attack(b, face, predicate);
     }),
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/Instruction.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/Instruction.java
@@ -20,7 +20,7 @@ import io.github.thebusybiscuit.slimefun4.utils.HeadTexture;
 import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 
-enum Instruction {
+public enum Instruction {
 
     // Start and End Parts
     START(AndroidType.NONE, HeadTexture.SCRIPT_START),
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -534,12 +534,12 @@ protected void editInstruction(Player p, Block b, String[] script, int index) {
     }
 
     @Nonnull
-    protected String getScript(@Nonnull Location l) {
+    public String getScript(@Nonnull Location l) {
```
TheBusyBiscuit, line 537: Don't open things up that don't have sufficient parameter validation, the `@Nonnull` is not enforced here because it was an internal method, so you'd need to add a Validate call here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -534,12 +534,12 @@ protected void editInstruction(Player p, Block b, String[] script, int index) {
...
         String script = BlockStorage.getLocationInfo(l, "script");
         return script != null ? script : DEFAULT_SCRIPT;
     }
 
-    protected void setScript(@Nonnull Location l, @Nonnull String script) {
+    public void setScript(@Nonnull Location l, @Nonnull String script) {
```
TheBusyBiscuit, line 542: Don't open things up that don't have sufficient parameter validation, the `@Nonnull` is not enforced here because it was an internal method, so you'd need to add a Validate call here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/Instruction.java
```diff
@@ -20,7 +20,7 @@
 import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 
-enum Instruction {
+public enum Instruction {
```
TheBusyBiscuit, line 23: why?

TheBusyBiscuit (changes requested):
Why even are you exposing these methods and the enum...?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/Instruction.java
Seggan: As I already said, so addons can modify android code

TheBusyBiscuit: This is an enum though...

Seggan: I plan to generate an Instruction[] in my addon code, then turn it into a string that the android recognises

TheBusyBiscuit: Okay, I thought you meant adding more Instructions, because for that we would need an interface (which we should add at some point anyway)

TheBusyBiscuit (approved):

poma123 (approved):

PR #2503, opened 2020-10-25
Cargo Net Optimization: Cached Item Filters

TheBusyBiscuit:
## Description

This PR adds a little performance optimization to cargo networks.
Since `ItemStack#getItemMeta()` is a heavy call which creates a cloned instance everytime it is invoked, any way to reduce these calls to a bare minimum is a welcome performance improvement.
Here I have created an `ItemFilter` class, which is a `Predicate<ItemStack>` and it serves the point of representing a cargo node's configuration as a simple object.
This instance is cached in a `HashMap`, it will be updated once marked dirty.
It will be marked dirty when:
* The cargo node's configuration was directly modified
* The cargo node's configuration screen is closed (that way we will know when people modified the filtered items.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/AbstractItemNetwork.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/AbstractItemNetwork.java
@@ -23,6 +23,7 @@ import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
+import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.Directional;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.InventoryHolder;
@@ -47,12 +48,13 @@ import me.mrCookieSlime.Slimefun.api.inventory.UniversalBlockMenu;
 import me.mrCookieSlime.Slimefun.api.item_transport.ItemTransportFlow;
 
 /**
- * An abstract super class of {@link CargoNet} that handles interactions with ChestTerminal.
+ * An abstract super class of {@link CargoNet} that handles
+ * interactions with ChestTerminal.
  * 
  * @author TheBusyBiscuit
  *
  */
-abstract class ChestTerminalNetwork extends Network {
+abstract class AbstractItemNetwork extends Network {
 
     private static final int[] slots = { 19, 20, 21, 28, 29, 30, 37, 38, 39 };
     private static final int[] TERMINAL_SLOTS = { 0, 1, 2, 3, 4, 5, 6, 9, 10, 11, 12, 13, 14, 15, 18, 19, 20, 21, 22, 23, 24, 27, 28, 29, 30, 31, 32, 33, 36, 37, 38, 39, 40, 41, 42 };
@@ -64,14 +66,23 @@ abstract class ChestTerminalNetwork extends Network {
     protected final Set<Location> imports = new HashSet<>();
     protected final Set<Location> exports = new HashSet<>();
 
-    // This represents a Queue of requests to handle
+    /**
+     * This represents a {@link Queue} of requests to handle
+     */
     private final Queue<ItemRequest> itemRequests = new LinkedList<>();
 
-    // This is a cache for the BlockFace a node is facing, so we don't need to request the
-    // BlockData each time we visit a node
+    /**
+     * This is a cache for the {@link BlockFace} a node is facing, so we don't need to
+     * request the {@link BlockData} each time we visit a node
+     */
     protected Map<Location, BlockFace> connectorCache = new HashMap<>();
 
-    protected ChestTerminalNetwork(Location regulator) {
+    /**
+     * This is our cache for the {@link ItemFilter} for each node.
+     */
+    protected Map<Location, ItemFilter> filterCache = new HashMap<>();
+
+    protected AbstractItemNetwork(Location regulator) {
         super(SlimefunPlugin.getNetworkManager(), regulator);
     }
 
@@ -127,7 +138,7 @@ abstract class ChestTerminalNetwork extends Network {
             Optional<Block> target = getAttachedBlock(l);
 
             if (target.isPresent()) {
-                item = CargoUtils.insert(inventories, l.getBlock(), target.get(), item);
+                item = CargoUtils.insert(this, inventories, l.getBlock(), target.get(), item);
 
                 if (item == null) {
                     terminal.replaceExistingItem(request.getSlot(), null);
@@ -159,7 +170,7 @@ abstract class ChestTerminalNetwork extends Network {
             Optional<Block> target = getAttachedBlock(l);
 
             if (target.isPresent()) {
-                ItemStack is = CargoUtils.withdraw(inventories, l.getBlock(), target.get(), item);
+                ItemStack is = CargoUtils.withdraw(this, inventories, l.getBlock(), target.get(), item);
 
                 if (is != null) {
                     if (stack == null) {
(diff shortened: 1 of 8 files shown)


andris155:
Will this be included in the plugin?

TheBusyBiscuit:
> Will this be included in the plugin?

If it gets through the review process, of course.

Sfiguz7 (approved):
Saw nothing wrong or missing, yes from me

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/AbstractItemNetwork.java
```diff
@@ -326,10 +337,27 @@ protected long updateTerminals(@Nonnull Set<Location> providers) {
 
     @Override
     public void markDirty(@Nonnull Location l) {
-        connectorCache.remove(l);
+        markCargoNodeConfigurationDirty(l);
         super.markDirty(l);
     }
 
+    /**
+     * This will markt the {@link ItemFilter} of the given node dirty.
```
svr333, line 345:
```suggestion
     * This will mark the {@link ItemFilter} of the given node dirty.
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/AbstractItemNetwork.java
```diff
@@ -432,7 +460,7 @@ private void gatherItemsFromBarrel(Location l, BlockMenu blockMenu, List<ItemSta
             for (int slot : blockMenu.getPreset().getSlotsAccessedByItemTransport((DirtyChestMenu) blockMenu, ItemTransportFlow.WITHDRAW, null)) {
                 ItemStack is = blockMenu.getItemInSlot(slot);
 
-                if (is != null && CargoUtils.matchesFilter(l.getBlock(), is)) {
+                if (is != null && CargoUtils.matchesFilter(this, l.getBlock(), is)) {
```
svr333, line 463: Im not a fan of the name is because when reading it reads like:
`if is is not null`

However this is just my opinion

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/AbstractItemNetwork.java
```diff
@@ -477,4 +505,21 @@ private void filter(@Nullable ItemStack stack, List<ItemStackAndInteger> items,
...
+
+        if (filter == null) {
+            ItemFilter newFilter = new ItemFilter(node);
+            filterCache.put(loc, newFilter);
+            return newFilter;
+        } else if (filter.isDirty()) {
+            filter.update(node);
+            return filter;
+        } else {
+            return filter;
```
svr333, line 521: Not sure what the default here is
```suggestion
        } else if (filter.isDirty()) {
            filter.update(node);
        }
        return filter;
```

svr333 (changes requested):
One typo and then 2 remarks which Im unsure about

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/AbstractItemNetwork.java
TheBusyBiscuit: Just no...

svr333: lol

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
```diff
@@ -0,0 +1,175 @@
...
+
+        // This is a first check for materials to see if we might even have any match.
+        // If there is no potential match then we won't need to perform the quite
+        // intense operation .getItemMeta()
+        for (ItemStackWrapper stack : items) {
+            if (stack.getType() == item.getType()) {
+                // We found a potential match based on the Material
+                potentialMatches++;
+            }
+        }
```
WalshyDev, line 149: Not sure I'm the biggest fan of having another loop just for the sake of saving 1 #getItemMeta call on `item`. Especially since both the loops are for an Array meaning O(n). If it was a hash table and be O(1) then sure maybe but yeah. I feel like we're losing performance to try and improve performance.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
```diff
@@ -0,0 +1,175 @@
...
+        this.defaultValue = defaultValue;
+    }
+
+    /**
+     * Whether this {@link ItemFilter} is outdated and needs to be refreshed.
+     * 
+     * @return Whether the filter is outdated.
+     */
+    public boolean isDirty() {
+        return dirty;
```
WalshyDev, line 122: this. for consistency

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
```diff
@@ -0,0 +1,175 @@
...
+    private final List<ItemStackWrapper> items = new ArrayList<>();
+
+    /**
+     * Our default value for this {@link ItemFilter}.
+     * A default value of {@literal true} will mean that it returns true if no
+     * match was found. It will deny any items that match.
+     * A default value of {@literal false} means that it will return false if no
+     * match was found. Only items that match will make it past this {@link ItemFilter}.
+     */
+    private boolean defaultValue;
```
WalshyDev, line 42: I don't really like the name of this variable. `rejectItemIfMatched` or something would be a bit clearer.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
```diff
@@ -0,0 +1,175 @@
...
+        if (item == null || menu == null) {
+            // Don't filter for a non-existing item (safety check)
+            clear(false);
+        } else if (id.equals("CARGO_NODE_OUTPUT")) {
+            // Output Nodes have no filter, allow everything
+            clear(true);
+        } else {
+            this.items.clear();
+            this.checkLore = Objects.equals(blockData.getString("filter-lore"), "true");
+            this.defaultValue = !Objects.equals(blockData.getString("filter-type"), "whitelist");
```
WalshyDev, line 89: Not an issue just cool to see Objects.equals used here :+1:

WalshyDev (changes requested):
Overall I like this. It is a good optimisation!

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
TheBusyBiscuit: getItemMeta() is a very heavy function that can explode quite quickly.
This is a bit ugly but it does improve performance overall actually.
The horror of ItemMeta cloning is really worth this sacrifice here sadly.

TheBusyBiscuit: Also, keep in mind that `items` has a maximum capacity of 9. So the impact will be quite small.

Sfiguz7 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
Sfiguz7: (In the meantime I approved since I assume we're just waiting on Walshy to decide he believes you, take this as a bump for the blob)

WalshyDev (approved):
LGTM

svr333 (approved):

PR #2508, opened 2020-10-25
Smallest documentation fix ever

Sfiguz7:
## Description
Anvil Listener had doc comment copied from Brewing Stand.

## Changes
Changed one link

## Related Issues
N/A

## Checklist
N/A:

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/AnvilListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/AnvilListener.java
@@ -15,8 +15,8 @@ import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
 
 /**
- * This {@link Listener} prevents any {@link SlimefunItem} from being used in a
- * {@link BrewingStand}.
+ * This {@link Listener} prevents any {@link SlimefunItem} from being used in an
+ * anvil.
  * 
  * @author TheBusyBiscuit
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/AnvilListener.java
```diff
@@ -15,8 +15,8 @@
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
 
 /**
- * This {@link Listener} prevents any {@link SlimefunItem} from being used in a
- * {@link BrewingStand}.
+ * This {@link Listener} prevents any {@link SlimefunItem} from being used in an
+ * anvil.
```
TheBusyBiscuit, line 19: link pls

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/crafting/AnvilListener.java
Sfiguz7: No link to Anvil, just to AnvilInventory (at least that I saw). All other Listener classes have the machine just typed out like that too so I went with it

TheBusyBiscuit: Well the brewing stand has a link though, so I would argue in favour of uniformity to remove the link from the brewing stand listener then

TheBusyBiscuit (approved):

poma123 (approved):

PR #2510, opened 2020-10-26
Fixes #2509

### LINKED ISSUE #2509: Machines that run at "10x" speed require no energy network

The Electric Dust Washer (III) and Electric Gold Pan (III) are fully functional with and without any energy network supplying them power. This is most likely unintended. The other two tiers function as normal for both machines

Here is the video proof for both machines, as well as proof that this is an issue with 10x speed machines from other addons, with the example being the Electric Dust Fabricator from Fluffy Machines.
https://streamable.com/jtdud1

I was expecting to supply the required power in order for the Tier III Electric Dust Washer and Electric Gold Pan to work, but found that it nonetheless still converts Sifted Ore into dusts (and Gravel into its outputs) at the same processing speed, 1 op/ 24 gt, which is 10x processing speed, regardless of the presence of an energy network. According to the specs, the tier 3 washer requires 30 J/s and the tier 3 gold pan requires 14 J/s.

## :scroll: Server Log

There are no errors in the server log:
[latest.log](https://github.com/Slimefun/Slimefun4/files/5436127/latest.log)

There are no files to upload in that folder.

 - Minecraft Version: 1.16.3 (stable)
 - Slimefun Version: vDEV - 732
 - CS-CoreLib Version: vDEV - 91

Mooy1:
## Description
AContainers will now check and take charge when they are outputting, not just when progressing.

## Changes
Added a method to check and take the charge ,called it before outputting

## Related Issues
Resolves #2509 

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
@@ -8,6 +8,7 @@ import java.util.Map;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import org.apache.commons.lang.Validate;
 import org.bukkit.Bukkit;
 import org.bukkit.Location;
 import org.bukkit.Material;
@@ -281,6 +282,8 @@ public abstract class AContainer extends SlimefunItem implements InventoryBlock,
      * @return Whether charge was taken if its chargeable
      */
     protected boolean takeCharge(@Nonnull Location l) {
+        Validate.notNull(l, "Can't attempt to take charge from a null location!");
+
         if (isChargeable()) {
             if (getCharge(l) < getEnergyConsumption()) {
                 return false;
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
@@ -8,6 +8,7 @@ import java.util.Map;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 import org.bukkit.Bukkit;
+import org.bukkit.Location;
 import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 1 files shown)


Mooy1:
https://streamable.com/7j2z1b vid to show it AContainer still works

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
```diff
@@ -243,15 +244,16 @@ protected void tick(Block b) {
...
-                    removeCharge(b.getLocation(), getEnergyConsumption());
+                if (!takeCharge(b.getLocation())) {
+                    return;
                 }
+
                 progress.put(b, timeleft - 1);
             } else {
+                if (!takeCharge(b.getLocation())) {
+                    return;
+                }
```
TheBusyBiscuit, line 255: No need for this much nesting and unnecessary guard blocks.

```suggestion
            } else if (takeCharge(b.getLocation())) {
```

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
```diff
@@ -273,6 +275,17 @@ protected void tick(Block b) {
         }
     }
 
+    protected boolean takeCharge(Location l) {
```
TheBusyBiscuit, line 278: Missing annotation for the parameter.
Also missing javadocs

TheBusyBiscuit (changes requested):

Mooy1:
https://streamable.com/cdxu6z

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
```diff
@@ -273,6 +274,23 @@ protected void tick(Block b) {
         }
     }
 
+    /**
+     * This method will remove charge from a location if it is chargeable.
+     *
+     * @param l location to try to remove charge from
+     * @return Whether charge was taken if its chargeable
+     */
+    protected boolean takeCharge(@Nonnull Location l) {
```
TheBusyBiscuit, line 283: Since this is a "protected" method, the Non-nullability of this parameter should also be enforced via `Validate.notNull(l, message)`

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
```diff
@@ -238,31 +240,30 @@ protected void tick(Block b) {
         BlockMenu inv = BlockStorage.getInventory(b);
 
         if (isProcessing(b)) {
-            int timeleft = progress.get(b);
 
-            if (timeleft > 0) {
-                ChestMenuUtils.updateProgressbar(inv, 22, timeleft, processing.get(b).getTicks(), getProgressBar());
+            if (takeCharge(b.getLocation())) {
```
TheBusyBiscuit, line 244: Why was this now placed in front of everything?

TheBusyBiscuit (changes requested):

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
Mooy1: because it should be called in both cases, so might as well right?

TheBusyBiscuit: huh, you are right.

TheBusyBiscuit (approved):

poma123 (approved):

PR #2548, opened 2020-11-10
Adding slimefun item recipe uses to guide

Mooy1:
## Description
Right click a Slimefun item in the guide to see it's uses in other recipes. 

## Changes
Added new map in registry that keeps a list of recipe uses for each SlimefunItem, filled in PostSetup.
Added new class SlimefunItemRecipeUse which holds a slimefunItem, the uses of it, and the page of use.
Added that class to history which will open the correct menu.
Added right click handlers to many spots that will open the recipe uses if its a SliemfunItem.
Added recipe uses menu to ChestSlimefunGuide implementation.
Added some methods to history which can replace the last object in history and check for an object in history.

## Related Issues
none

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
@@ -13,6 +13,8 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import javax.annotation.Nonnull;
 
+import me.mrCookieSlime.Slimefun.api.Slimefun;
+import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.Validate;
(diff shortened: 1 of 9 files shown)


Mooy1:
quick video to demonstrate https://streamable.com/wsfm5h

TheBusyBiscuit:
I am sorry but I will have to reject this pull request for a couple of reasons.

As I already mentioned a lot of times on discord: The Recipe System is long overdue for a complete overhaul and this PR just confirms this, the code is not really that clean to be frank but that's more due to the underlying system.

I feel like it would be smarter to redo the recipe system and then implement something like this, rather than just making the rewrite even harder due to the added complexity this brings.

But I appreciate the effort and find the idea really good.

Mooy1:
Is there any plans for the new recipe system or a discussion of it somewhere? I would like to work on something like that

PR #2555, opened 2020-11-17
Fixed a bug of Duplicate items with BlockPlacer

wzpzw5:
Fixed a bug of Duplicate items with BlockPlacer's delay

## Description
on BlockPlacer place an item, just take out the item when it place,then one in hand and 
another is placed.
## Changes
when BlockPlacer place,there shouldn't be someone viewing it.
I have no better way to fix it.

## Related Issues
With BlockPlacer's Place delay,it can Duplicate almost any block,just take out the item in BlockPlacer on BlockPlacer place.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
@@ -145,10 +145,9 @@ public class BlockPlacer extends SlimefunItem {
 
     private void placeSlimefunBlock(SlimefunItem sfItem, ItemStack item, Block block, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, block);
-        Bukkit.getPluginManager().callEvent(e);
         
-        if(!dispenser.getInventory().getViewers().isEmpty()){
-        	e.setCancelled(true);
+        if(dispenser.getInventory().getViewers().isEmpty()) {
+        	Bukkit.getPluginManager().callEvent(e);
         }
 
         if (!e.isCancelled()) {
@@ -185,10 +184,9 @@ public class BlockPlacer extends SlimefunItem {
 
     private void placeBlock(ItemStack item, Block facedBlock, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, facedBlock);
-        Bukkit.getPluginManager().callEvent(e);
         
-        if(!dispenser.getInventory().getViewers().isEmpty()){
-        	e.setCancelled(true);
+        if(dispenser.getInventory().getViewers().isEmpty()) {
+        	Bukkit.getPluginManager().callEvent(e);
         }
 
         if (!e.isCancelled()) {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
@@ -93,10 +93,10 @@ public class BlockPlacer extends SlimefunItem {
                 return;
             }
 
-            if (facedBlock.isEmpty() && !isBlacklisted(material)) {
+            if (facedBlock.isEmpty() && !isBlacklisted(material) && dispenser.getInventory().getViewers().isEmpty()) {
                 SlimefunItem item = SlimefunItem.getByItem(e.getItem());
 
-                if (item != null && dispenser.getInventory().getViewers().isEmpty()) {
+                if (item != null) {
                     // Check if this Item can even be placed down
                     if (!(item instanceof NotPlaceable)) {
                         placeSlimefunBlock(item, e.getItem(), facedBlock, dispenser);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
@@ -96,7 +96,7 @@ public class BlockPlacer extends SlimefunItem {
             if (facedBlock.isEmpty() && !isBlacklisted(material)) {
                 SlimefunItem item = SlimefunItem.getByItem(e.getItem());
 
-                if (item != null) {
+                if (item != null && dispenser.getInventory().getViewers().isEmpty()) {
                     // Check if this Item can even be placed down
                     if (!(item instanceof NotPlaceable)) {
                         placeSlimefunBlock(item, e.getItem(), facedBlock, dispenser);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -146,6 +146,10 @@ private boolean isBlacklisted(Material type) {
     private void placeSlimefunBlock(SlimefunItem sfItem, ItemStack item, Block block, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, block);
         Bukkit.getPluginManager().callEvent(e);
+        
+        if(!dispenser.getInventory().getViewers().isEmpty()){
```
TheBusyBiscuit, line 150: Formatting, please adhere to our code style guidelines, there should be a space after the `if` and before the curly bracket `{`

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -182,6 +186,10 @@ private void placeSlimefunBlock(SlimefunItem sfItem, ItemStack item, Block block
     private void placeBlock(ItemStack item, Block facedBlock, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, facedBlock);
         Bukkit.getPluginManager().callEvent(e);
+        
+        if(!dispenser.getInventory().getViewers().isEmpty()){
```
TheBusyBiscuit, line 190: Formatting, please adhere to our code style guidelines, there should be a space after the `if` and before the curly bracket `{`

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -146,6 +146,10 @@ private boolean isBlacklisted(Material type) {
     private void placeSlimefunBlock(SlimefunItem sfItem, ItemStack item, Block block, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, block);
         Bukkit.getPluginManager().callEvent(e);
+        
+        if(!dispenser.getInventory().getViewers().isEmpty()){
+        	e.setCancelled(true);
```
TheBusyBiscuit, line 151: There is no point in calling an Event when it gets cancelled anyway.
The whole point of the Event contract is that **others** get a say in the outcome of the Event, not that their changes get overridden otherwise.

If no block gets placed, then there shouldn't even be an Event called.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -182,6 +186,10 @@ private void placeSlimefunBlock(SlimefunItem sfItem, ItemStack item, Block block
     private void placeBlock(ItemStack item, Block facedBlock, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, facedBlock);
         Bukkit.getPluginManager().callEvent(e);
+        
+        if(!dispenser.getInventory().getViewers().isEmpty()){
+        	e.setCancelled(true);
```
TheBusyBiscuit, line 191: There is no point in calling an Event when it gets cancelled anyway.
The whole point of the Event contract is that **others** get a say in the outcome of the Event, not that their changes get overridden otherwise.

If no block gets placed, then there shouldn't even be an Event called.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
wzpzw5: Sorry, but the event has been called at Bukkit.getPluginManager().callEvent(e).

wzpzw5:
Although when someone viewing blockplacer, it will not work, but most of the time, no one will view blockplacer when it is working.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
wzpzw5: But if add if before it ,maybe it will be slower?
Because most of the time, dispenser.getInventory().getViewers().isEmpty() is true.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -145,7 +145,10 @@ private boolean isBlacklisted(Material type) {
 
     private void placeSlimefunBlock(SlimefunItem sfItem, ItemStack item, Block block, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, block);
-        Bukkit.getPluginManager().callEvent(e);
+        
+        if(dispenser.getInventory().getViewers().isEmpty()) {
```
TheBusyBiscuit, line 149: Still formatting...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -145,7 +145,10 @@ private boolean isBlacklisted(Material type) {
...
+        	Bukkit.getPluginManager().callEvent(e);
```
TheBusyBiscuit, line 150: Putting this here makes absolutely no sense.
The only thing it would do is break all other plugins and allow for many other duplications.

TheBusyBiscuit: You cannot just make only the event calling conditional. The check must occur far before any Event is even called, so better before these methods are even invoked in the first place.

Also, you should really test your changes before-hand.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -181,7 +184,10 @@ private void placeSlimefunBlock(SlimefunItem sfItem, ItemStack item, Block block
 
     private void placeBlock(ItemStack item, Block facedBlock, Dispenser dispenser) {
         BlockPlacerPlaceEvent e = new BlockPlacerPlaceEvent(dispenser.getBlock(), item, facedBlock);
-        Bukkit.getPluginManager().callEvent(e);
+        
+        if(dispenser.getInventory().getViewers().isEmpty()) {
```
TheBusyBiscuit, line 188: Still formatting...

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
Some feedback for the user to let them know why it didn't work would be nice.
But it looks good to me as it is! Thanks :)

Sfiguz7 (approved):
Agreed on some QoL "You can't open the BlockPlacer while it's operating" or something. Other than that looks good

wzpzw5:
I have tested it on my server for a long time, and no one ask for why.
Because it has little effect to using it.

TheBusyBiscuit:
Yeah, as Sfiguz and I both said: It would be a "Quality of life" improvement, but neither of us really see a huge need for it.
It's definitely no necessity.

Anyway, thanks for the pull request! I am gonna merge it shortly.

TheBusyBiscuit:
Merging this now.
Please also pay attention to our "checklist" for future pull requests btw!

PR #2562, opened 2020-11-21
Add Rainbow beds

svr333:
## Description
Added Rainbow Beds to the Rainbow Blocks collection.
The beds are still spazzing out right now but I havent found the issue with it, though I wanted to push the code already.
Spazzing bed

## Changes
- Create Rainbow Bed SlimefunItem
- Add Rainbow Bed to SlimefunItemSetup
- Add Rainbow Bed to `wiki.json`
- Add Rainbow Bed to Rainbow Blocks Research Setup
- Add BED Material array in `ColoredMaterial.java`
- Add logic so both bed blocks update in `RainbowTickHandler.java`

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/handlers/RainbowTickHandler.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/handlers/RainbowTickHandler.java
@@ -180,11 +180,11 @@ public class RainbowTickHandler extends BlockTicker {
     }
 
     private void setBedBlock(Block b, Material material, BlockFace facing, Part part, boolean doPhysics) {
-		b.setType(material, false);
+        b.setType(material, false);
         BlockData data = b.getBlockData();
         
 		if (data instanceof Bed) {
-			((Bed) data).setFacing(facing);
+            ((Bed) data).setFacing(facing);
             ((Bed) data).setPart(part);
         }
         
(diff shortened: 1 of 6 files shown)


svr333:
If anyone knows the issue with the bed switching sides, feel free to mention, otherwise Ill take a look at it tomorrow

src/main/java/io/github/thebusybiscuit/slimefun4/core/handlers/RainbowTickHandler.java
```diff
@@ -129,4 +157,38 @@ public boolean isSynchronized() {
...
+    private void setBedBlock(Block b, Material material, BlockFace facing, Part part, boolean doPhysics) {
+		b.setType(material, false);
+        BlockData data = b.getBlockData();
+        
+		if (data instanceof Bed) {
+			((Bed) data).setFacing(facing);
+            ((Bed) data).setPart(part);
+        }
+        
+        b.setBlockData(data, doPhysics);
```
TheBusyBiscuit, line 191: Indentation hell awaits you

TheBusyBiscuit (changes requested):
Also, the block info is likely only stored on one side of the bed, so destroying the other half probably leaves a ghost block, right?
That would obviously need to be fixed...

src/main/java/io/github/thebusybiscuit/slimefun4/core/handlers/RainbowTickHandler.java
svr333: I dont know where you found that code but it looks perfectly fine on my end

svr333:
I see, I do currently lack the knowledge to do this.
Ill check the docs on beds later when I have time for it

TheBusyBiscuit:
I'm gonna close this for now.
If you find a way to address this issue, feel free to get back to us.
But I cannot really think of a "good" way to realize this concept at the moment either.

PR #2595, opened 2020-12-05
Fix ghost blocks

Patbox:
## Description
Fixes instances of ghost blocks (places where you can't place blocks, because for SF it exist and for server not).
As for now I did quick test of it and it looks like everything worked. I also plan to test in on working server soon.
## Changes
Added some logic to BlockListener.onBlockPlaceExisting for checking if targeted block is air (and if it is, removing it and dropping item).

## Related Issues
https://github.com/Slimefun/Slimefun4/wiki/Common-Issues#unplaceable-blocks
https://youtu.be/HYlNRKud5kI

## Checklist

^ Leaving this empty, as I prefer someone else to rate it (1st and 2nd one I will check after tests).

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -66,14 +66,14 @@ public class BlockListener implements Listener {
 
             if (sfItem != null) {
                 for (ItemStack item : sfItem.getDrops()) {
-                    if (!item.getType().isAir()) {
+                    if (item != null && !item.getType().isAir()) {
                         block.getWorld().dropItemNaturally(block.getLocation(), item);
                     }
                 }
                 BlockStorage.clearBlockInfo(block);
-            } else {
-                e.setCancelled(true);
             }
+        } else if (BlockStorage.hasBlockInfo(e.getBlock())) {
+            e.setCancelled(true);
         }
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -4,16 +4,13 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 import java.util.Random;
-import java.util.Iterator;
 import java.util.concurrent.ThreadLocalRandom;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import org.bukkit.Location;
 import org.bukkit.Material;
-import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.enchantments.Enchantment;
@@ -41,10 +38,10 @@ import me.mrCookieSlime.Slimefun.api.Slimefun;
 /**
  * The {@link BlockListener} is responsible for listening to the {@link BlockPlaceEvent}
  * and {@link BlockBreakEvent}.
- * 
+ *
  * @author TheBusyBiscuit
  * @author Linox
- * 
+ *
  * @see BlockPlaceHandler
  * @see BlockBreakHandler
  * @see ToolUseHandler
@@ -62,18 +59,17 @@ public class BlockListener implements Listener {
         // While this can cause ghost blocks it also prevents them from replacing grass
         // or saplings etc...
         Block block = e.getBlock();
-        if (BlockStorage.hasBlockInfo(block)) {
+        SlimefunItem sfItem = BlockStorage.check(block);
+
+        if (sfItem != null) {
             if (e.getBlockReplacedState().getType().isAir()) {
-                Iterator<ItemStack> item = BlockStorage.check(block).getDrops().iterator();
-                World world = block.getWorld();
-                Location loc = block.getLocation();
-                while (item.hasNext()) {
-                    world.dropItem(loc, item.next());
+                for (ItemStack item : sfItem.getDrops()) {
+                    if (item != null && !item.getType().isAir()) { block.getWorld().dropItemNaturally(block.getLocation(), item); }
                 }
                 BlockStorage.clearBlockInfo(block);
-                return;
+            } else {
+                e.setCancelled(true);
             }
-            e.setCancelled(true);
         }
     }
 
@@ -184,7 +180,7 @@ public class BlockListener implements Listener {
      * This method checks for a sensitive {@link Block}.
      * Sensitive {@link Block Blocks} are pressure plates or saplings, which should be broken
      * when the block beneath is broken as well.
-     * 
+     *
      * @param p
      *            The {@link Player} who broke this {@link Block}
      * @param b
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -41,6 +41,7 @@ import me.mrCookieSlime.Slimefun.api.Slimefun;
  *
  * @author TheBusyBiscuit
  * @author Linox
+ * @author Patbox
  *
  * @see BlockPlaceHandler
  * @see BlockBreakHandler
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,7 +61,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
...
+        if (BlockStorage.hasBlockInfo(block)) {
+            if (e.getBlockReplacedState().getType().isAir()) {
+                Iterator<ItemStack> item = BlockStorage.check(block).getDrops().iterator();
+                World world = block.getWorld();
+                Location loc = block.getLocation();
+                while (item.hasNext()) {
+                    world.dropItem(loc, item.next());
+                }
+                BlockStorage.clearBlockInfo(block);
+                return;
```
TheBusyBiscuit, line 74: Why would you `return;` here...
If this is to prevent the Event from being cancelled, use an `if/else` structure.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,7 +61,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
         // This prevents Players from placing a block where another block already exists
         // While this can cause ghost blocks it also prevents them from replacing grass
         // or saplings etc...
-        if (BlockStorage.hasBlockInfo(e.getBlock())) {
+        Block block = e.getBlock();
+        if (BlockStorage.hasBlockInfo(block)) {
+            if (e.getBlockReplacedState().getType().isAir()) {
+                Iterator<ItemStack> item = BlockStorage.check(block).getDrops().iterator();
```
TheBusyBiscuit, line 67: Why the iterator...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,7 +61,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
...
+                World world = block.getWorld();
+                Location loc = block.getLocation();
+                while (item.hasNext()) {
+                    world.dropItem(loc, item.next());
```
TheBusyBiscuit, line 71: You cannot drop null or air items, so you would need to check for that.

TheBusyBiscuit, line 71: Also *dropItemNaturally

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,7 +61,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
         // This prevents Players from placing a block where another block already exists
         // While this can cause ghost blocks it also prevents them from replacing grass
         // or saplings etc...
-        if (BlockStorage.hasBlockInfo(e.getBlock())) {
+        Block block = e.getBlock();
+        if (BlockStorage.hasBlockInfo(block)) {
+            if (e.getBlockReplacedState().getType().isAir()) {
```
TheBusyBiscuit, line 66: These two statements can be merged.

TheBusyBiscuit, line 66: The order of these checks should be swapped for performance reasons.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
TheBusyBiscuit, line 67: Calling `BlockStorage.check()` here is completely redundant.
You are already doing a check by doing `hasBlockInfo()` above, this is essentially doing two concurrent operations redundantly, a null check would be more sufficient here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,7 +61,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
...
+                Iterator<ItemStack> item = BlockStorage.check(block).getDrops().iterator();
+                World world = block.getWorld();
+                Location loc = block.getLocation();
```
TheBusyBiscuit, line 69: Not really much of a point in making these variables here.

TheBusyBiscuit (changes requested):
I appreciate the effort, but the code quality leaves a lot of room for improvement.

Also, this would essentially erase/override existing blocks, so that is a thing that should be kept in mind. 
This is a compromise and not an ideal solution.

Patbox:
Thanks for feedback! I'm still learning java, so it helps me to get better with it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -38,10 +38,10 @@
 /**
  * The {@link BlockListener} is responsible for listening to the {@link BlockPlaceEvent}
  * and {@link BlockBreakEvent}.
- * 
+ *
  * @author TheBusyBiscuit
  * @author Linox
- *
```
TheBusyBiscuit, line 44: Speaking of which, since you modified this class, feel free to add yourself here.
So we know who to contact XD

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,8 +58,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
         // This prevents Players from placing a block where another block already exists
         // While this can cause ghost blocks it also prevents them from replacing grass
         // or saplings etc...
-        if (BlockStorage.hasBlockInfo(e.getBlock())) {
-            e.setCancelled(true);
+        Block block = e.getBlock();
+        SlimefunItem sfItem = BlockStorage.check(block);
+
+        if (sfItem != null) {
+            if (e.getBlockReplacedState().getType().isAir()) {
```
TheBusyBiscuit, line 65: As stated in a prior review, these two checks should be swapped for performance reason.
Any `BlockStorage` operations might be heavy depending on their concurrent behaviour and take a relatively long time to compute, whereas the `.isAir()` check is very inexpensive.
Swapping will reduce those heavy lookups to a bare minimum.

```suggestion
        if (e.getBlockReplacedState().getType().isAir()) {
            Block block = e.getBlock();
            SlimefunItem sfItem = BlockStorage.check(block);

            if (sfItem != null) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,8 +58,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
...
+                for (ItemStack item : sfItem.getDrops()) {
+                    if (item != null && !item.getType().isAir()) { block.getWorld().dropItemNaturally(block.getLocation(), item); }
```
TheBusyBiscuit, line 67: Formatting, also for non-blocks a simple AIR check is sufficient.
```suggestion
                    if (item != null && item.getType() != Material.AIR) {
                        block.getWorld().dropItemNaturally(block.getLocation(), item);
                    }
```

TheBusyBiscuit (changes requested):
Looking much better, still two things left that should be improved though.

Thanks for the pull request!

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
WalshyDev: Well.... There's multiple different airs now (1.16 api and 1.17 implementation).

isAir is definitely best to use unless we know 100% that it would be AIR and not something like CAVE_AIR

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,8 +59,21 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
...
-        if (BlockStorage.hasBlockInfo(e.getBlock())) {
-            e.setCancelled(true);
+        Block block = e.getBlock();
+
+        if (e.getBlockReplacedState().getType().isAir()) {
+            SlimefunItem sfItem = BlockStorage.check(block);
+
+            if (sfItem != null) {
+                for (ItemStack item : sfItem.getDrops()) {
+                    if (!item.getType().isAir()) {
```
TheBusyBiscuit, line 69: I probably worded that wrongly, the null check is still required here, I simply meant that you could replace `.isAir()` with a `Material.AIR` check.
But either works tbh.

```suggestion
                    if (item != null && !item.getType().isAir()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -58,8 +59,21 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
...
+                        block.getWorld().dropItemNaturally(block.getLocation(), item);
+                    }
+                }
+                BlockStorage.clearBlockInfo(block);
+            } else {
+                e.setCancelled(true);
+            }
         }
```
TheBusyBiscuit, line 77: I would suggest to add a fallback here, because the `BlockPlaceEvent` will also be called when a Player tries to place the block in the same space as tall grass or other passable materials.
For these, we want to cancel it nonetheless as they would need to destroy the block first.

```suggestion
        } else if (BlockStorage.hasBlockInfo(e.getBlock())) {
            e.setCancelled(true);
        }
```

TheBusyBiscuit (changes requested):
One small mistake from me and another suggestion.
Otherwise it is looking good!

TheBusyBiscuit (approved):

svr333 (approved):

WalshyDev (approved):
LGTM

PR #2597, opened 2020-12-05
Refactoring + Fixes to Block Handlers

### LINKED ISSUE #2560: Breaking Infused Hoppers with a hopper limitation applied doesn't subtract them from the hopper limit.

Placing an infused hopper down with a hopper limitation applied and then breaking said hopper will still count towards the hopper limitation. So if you for example have 1 hopper as a limit and you place down 1 infused hopper, then break the hopper, you can no longer place any other hoppers down. This only seems to happen with the infused hopper, regular hoppers and hoppers from other plugins work fine.

Set a hopper limit (Preferably 1 for easier testing).
Get an infused hopper and place it down.
Break infused hopper.
Try to place another hopper down (any) and it will tell you it's not possible.

Breaking the infused hopper should no longer count towards the limit.

## :scroll: Server Log

No errors but still got a link for you anyway:
https://pastebin.com/sFFGHfv3

The latest one was october 25th (Nearly a month ago) I figure you don't want those error reports
https://i.imgur.com/RPjcwgc.png

I can't include it all in one screenshot so I split it in two with the last line being the first in the second:
https://i.imgur.com/v3IXQ2z.png
https://i.imgur.com/ODLFSO9.png

 - Server Software (Spigot/Paper): Yatopia
 - Minecraft Version: 1.16.4
 - Slimefun Version: vDEV - 744
 - CS-CoreLib Version: vDEV - 90

TheBusyBiscuit:
## Description

This Pull Request is meant to replace the old `SlimefunBlockHandler` in the future.
Slimefun has been using an `ItemHandler`-based system for quite some time now and block-breaking should finally be transferred over.
This will also fix two bugs.

## Related Issues

* Resolves #2560 

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiningAndroidListener.java
```diff
@@ -0,0 +1,32 @@
...
+import me.mrCookieSlime.Slimefun.api.BlockStorage;
+
+public class MiningAndroidListener implements Listener {
+
+    public MiningAndroidListener(@Nonnull SlimefunPlugin plugin) {
+        plugin.getServer().getPluginManager().registerEvents(this, plugin);
+    }
+
+    @EventHandler(ignoreCancelled = true)
+    public void onEntityExplode(AndroidMineEvent e) {
```
Sfiguz7, line 21: Miners using dynamite :p

Sfiguz7 (approved):
Loots of annotation to go through, had to double check as I was getting deja vu's ahaha
This will need changes in places for most Addons but it does make sense to move on; couldn't spot anything much (other than the method name for androids) so I'll approve this

EDIT: if you want to send me a Jar with the changes I can test it on our Dev server, it has a bunch of Addons and one of the skyblock hopper limits so it could be a decent test, definitely better than no test at all

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiningAndroidListener.java
TheBusyBiscuit: whooooops XD

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -51,13 +60,28 @@ public void onPlayerPlace(BlockPlaceEvent e) {
...
+            @Override
+            public void onExplode(Block b, List<ItemStack> drops) {
+                remove(b);
+            }
+
+            @Override
+            public void onAndroidBreak(AndroidMineEvent e) {
+                remove(e.getBlock());
+            }
         };
```
WalshyDev, line 81:
```suggestion
        return new SimpleBlockBreakHandler() {

            @Override
            public void onBlockBreaking(@Nonnull Block b) {
                remove(b);
            }
        };
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/handlers/SimpleBlockBreakHandler.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * This method is called when a {@link Block} of this type is broken by a {@link Player},
+     * by a {@link MinerAndroid} or through an explosion.
+     * 
+     * @param b
+     *            The broken {@link Block}
+     */
+    public abstract void onBlockBreaking(@Nonnull Block b);
```
WalshyDev, line 42: should be called onBlockBreak

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -51,13 +60,28 @@ public void onPlayerPlace(BlockPlaceEvent e) {
...
     }
 
+    @Nonnull
     private BlockBreakHandler onBreak() {
-        return (e, item, fortune, drops) -> {
-            remove(e.getBlock());
-            return true;
+        return new BlockBreakHandler(true, true) {
```
WalshyDev, line 65: Can just use SimpleBlockBreakHandler here

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ExplosionsListener.java
```diff
@@ -41,6 +64,24 @@ public void onEntityExplode(EntityExplodeEvent e) {
...
+                        item.callItemHandler(BlockBreakHandler.class, handler -> {
+                            if (handler.isExplosionAllowed()) {
+                                BlockStorage.clearBlockInfo(block);
+                                block.setType(Material.AIR);
+
+                                List<ItemStack> drops = new ArrayList<>();
+                                handler.onExplode(block, drops);
+
+                                for (ItemStack drop : drops) {
+                                    if (drop != null && drop.getType() != Material.AIR) {
```
WalshyDev, line 77: should use isAir, if we're doing an air check may as well make it future proof. Who knows what MC, Spigot or other plugins will do

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ExplosionsListener.java
```diff
@@ -41,6 +64,24 @@ public void onEntityExplode(EntityExplodeEvent e) {
...
                     if (blockHandler != null) {
                         success = blockHandler.onBreak(null, block, item, UnregisterReason.EXPLODE);
+                    } else {
+                        item.callItemHandler(BlockBreakHandler.class, handler -> {
+                            if (handler.isExplosionAllowed()) {
+                                BlockStorage.clearBlockInfo(block);
+                                block.setType(Material.AIR);
+
+                                List<ItemStack> drops = new ArrayList<>();
+                                handler.onExplode(block, drops);
```
WalshyDev, line 74: would be great if we could cancel without disabling all explosions. If we had the event we could cancel and then check if cancelled above.

Would be nice for cases like, block durability (for like factions), allows explosion if out of electricity, etc.

WalshyDev (changes requested):
Looking good, a few things I've seen

Sfiguz7 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ExplosionsListener.java
TheBusyBiscuit: I added a Block param to the `isExplosionAllowed` method, it will be ignored by default but if you were to implement sth like that, you could override the method and handle durability there.

WalshyDev (approved):
LGTM

WalshyDev (approved):
LGTM

Sorry for the long review time on this 😅

WalshyDev (approved):
LGTM

Phoenix-Starlight:
> 
> 
> Loots of annotation to go through, had to double check as I was getting deja vu's ahaha
> This will need changes in places for most Addons but it does make sense to move on; couldn't spot anything much (other than the method name for androids) so I'll approve this
> 
> EDIT: if you want to send me a Jar with the changes I can test it on our Dev server, it has a bunch of Addons and one of the skyblock hopper limits so it could be a decent test, definitely better than no test at all

Kinda late for me to say something, but here is a jar if you want to test. (If you already have one, no big deal, just ignore this.)
[Slimefun v4.9-UNOFFICIAL.zip](https://github.com/Slimefun/Slimefun4/files/6012585/Slimefun.v4.9-UNOFFICIAL.zip)

Sfiguz7 (approved):
Sorry for the long wait, I'm back! LGTM

PR #2598, opened 2020-12-05
Added Bee Armor

TheBusyBiscuit:
## Description

This is a modified version of the original Pull Request #1909 .
It has been optimized to use a task-based approach instead of listening to the `PlayerMoveEvent` which will have a smaller impact on overall performance.
Lore should probably be changed too at some point.

I am only re-submitting this pull request since the original author hadn't continued it.
I merged it into a new branch and improved the performance a bit but all credit goes to @beSn0w and @LinoxGH .

Oh yeah, this adds bee armor by the way.

## Checklist


Sfiguz7 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/BeeWings.java
```diff
@@ -0,0 +1,19 @@
...
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+
+public class BeeWings extends SlimefunItem {
+
+    @ParametersAreNonnullByDefault
+    public BeeWings(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        super(category, item, recipeType, recipe);
+    }
+
```
svr333, line 18: Unnecessary whiteline (unless its a style thing)

svr333 (approved):

PR #2599, opened 2020-12-06
Add Tier 3 Electric Ore Grinder

svr333:
## Description
Add a tier 3 Electric Ore Grinder (reinforced plates, blistering ingots, T2 electric ore grinders and heating coils)
image

## Changes
- Add tier 3 Electric Ore Grinder to ResearchSetup of endgame grinding
- Add new SlimefunItemStack to `SlimefunItems.java`
- Add new ElectricOreGrinder to `SlimefunItemSetup.java`

## Feedback
I was very unsure about the recipe, seeing as the recipe for the Tier 2 Ore Grinder is also very pricy.
I was thinking of reducing their cost but instead I increased this cost.
Very open for discussion though!

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
```diff
@@ -1658,6 +1658,13 @@ public static void setup(@Nonnull SlimefunPlugin plugin) {
         .setProcessingSpeed(4)
         .register(plugin);
 
+        new ElectricOreGrinder(categories.electricity, SlimefunItems.ELECTRIC_ORE_GRINDER_3, RecipeType.ENHANCED_CRAFTING_TABLE,
+        new ItemStack[] {SlimefunItems.REINFORCED_PLATE, SlimefunItems.HEATING_COIL, SlimefunItems.REINFORCED_PLATE, SlimefunItems.ELECTRIC_ORE_GRINDER_2, SlimefunItems.BLISTERING_INGOT_3, SlimefunItems.ELECTRIC_ORE_GRINDER_2, SlimefunItems.BLISTERING_INGOT_3, SlimefunItems.HEATING_COIL, SlimefunItems.BLISTERING_INGOT_3})
+        .setCapacity(1024)
+        .setEnergyConsumption(90)
```
TheBusyBiscuit, line 1664: This should be 45, otherwise it wouldn't be inline with the lore you set.
This is per tick, not per second.

TheBusyBiscuit (changes requested):

hex8F:
Just 2 cents about recipe/power cost - for me, there is no reason to ever craft this machine.
2 Electric Ore Grinders IIs process items the same speed as proposed Grinder III while being cheaper AND requiring less power (2x30 J vs 90 J).

svr333:
> 
> 
> Just 2 cents about recipe/power cost - for me, there is no reason to ever craft this machine.
> 2 Electric Ore Grinders IIs process items the same speed as proposed Grinder III while being cheaper AND requiring less power (2x30 J vs 90 J).

Well this is the same with all the other tier 3 machines, they use more power than 2 tier 2s combined and they start using blistering ingots in their recipes, seeing as the tier 2 of this machine already used blistering ingots, I made this one even more expensive.
I did think about making tier 2 cheaper and change it from endgame to advanced machine but I cant make that decision alone

Sfiguz7 (dismissed):
I mean other than the recipe discussion (which I agree is less efficient but as all Tier III are, I've always assumed their purpose was mostly dust farm compactness in the late game when you have enough production already rather than efficiency so I don't mind the one you used) it looks good to me.

hex8F:
Re: recipe discussion

Sure, compactness is a good reason, but I'm pretty sure every Tier III is more efficient per item (important bit that I've of course forgot to type :P ).
Also, grand total of 5 Blistering Ingots is really rough to create on servers that don't run adminshops/electric spawners, especially  since you can craft 5 Grinders II using them.

svr333:
Well I would suggest the following then:

RP HC RP
EO BI EO
RP HC RP

with RP being Reinforced Plates, HC being Heating Coils, EO being Electric Ore Grinder tier 2 and BI being a Blistering ingot

svr333:
Seeing as no one objected, I pushed my proposal.

Sfiguz7 (approved):
Continues to be fine by me, I've already expressed my thoughts above

WalshyDev (approved):
LGTM

TheBusyBiscuit (approved):

PR #2627, opened 2020-12-13
Added PR labels

WalshyDev:
## Description
Made this action forever ago, was meant to add it in Hacktoberfest but... yeah.

Anyway, this adds labels to PRs based on the branch prefix. This is good for categorisation and helpful for people like myself for which PRs to look at first. We should add more labels but I thought these were some basic starting ones.

## Changes
* new workflow file to add labels

## Related Issues
N/A

## Checklist

### DIFF
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -1,0 +1,18 @@
+name: PR Labeler
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  pr-labeler:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: WalshyDev/pr-labels@v1
+        with:
+          feature: 'New Feature'
+          fix: '🐞 Bug Fix'
+          chore: 'Chore'
+          perf: '🛠️ Performance Improvement'
+          performance: '🛠️ Performance Improvement'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - uses: WalshyDev/pr-labels@v1
         with:
-          token: "${{ secrets.GITHUB_TOKEN }}"
+          token: "${{ secrets.ACCESS_TOKEN }}"
           feature: 'New Feature'
           fix: '🐞 Bug Fix'
           chore: 'Chore'
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -1,17 +1,21 @@
-name: PR Labeler
+name: Pull Request Labels
+
 on:
   pull_request:
     types: [opened]
 
 jobs:
   pr-labeler:
+
+    name: Pull Request Labels
     runs-on: ubuntu-latest
+
     steps:
       - uses: WalshyDev/pr-labels@v1
+        name: Apply labels based on branch
         with:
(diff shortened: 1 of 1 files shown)


svr333 (dismissed):
Nice addition, wont trigger many if any false positives
Im all for it

Sfiguz7 (approved):
Mr Blob forced me at gunpoint, please he has my children




LGTM

.github/workflows/pr-labels.yml
```diff
@@ -0,0 +1,17 @@
+name: PR Labeler
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  pr-labeler:
+    runs-on: ubuntu-latest
```
TheBusyBiscuit, line 8: Missing job name.

.github/workflows/pr-labels.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
+      - uses: WalshyDev/pr-labels@v1
```
TheBusyBiscuit, line 10: Missing name for this step

.github/workflows/pr-labels.yml
```diff
@@ -0,0 +1,17 @@
...
+        with:
+          token: "${{ secrets.GITHUB_TOKEN }}"
```
TheBusyBiscuit, line 12: We can use the token for @TheBusyBot here.

```suggestion
          token: "${{ secrets.ACCESS_TOKEN }}"
```

.github/workflows/pr-labels.yml
```diff
@@ -0,0 +1,17 @@
...
+          feature: 'New Feature'
```
TheBusyBiscuit, line 13: emoji? :eyes:

.github/workflows/pr-labels.yml
```diff
@@ -0,0 +1,17 @@
...
+          fix: '🐞 Bug Fix'
+          chore: 'Chore'
```
TheBusyBiscuit, line 15: emoji? :eyes:

.github/workflows/pr-labels.yml
```diff
@@ -0,0 +1,17 @@
...
+          perf: '🛠️ Performance Improvement'
+          performance: '🛠️ Performance Improvement'
```
TheBusyBiscuit, line 17: Redundant. I think we should remove one of those, or we could approach it with a regular expression.

TheBusyBiscuit (changes requested):
I'll be honest and say I am not really a big fan of this, seems quite unnecessary and I definitely would veto against this being a requirement for PRs.

I am not a big fan of having too many labels, but if you guys find this helpful, sure, we can always manually label PRs, they come in rare enough for that to be not too much of a stretch.

TheBusyBiscuit:
Would be awesome if the changes will be addressed, I took the liberty to add labels for it already.

* "🧹 Chores"
* "💡 Performance Optimization"
* "🎈 Feature"
* "✨ Fix"

TheBusyBiscuit (commented):
I fixed the given issues myself now.
One suggestion I still have would be have an output for this action, something like "applied-label".
Then we could do a null check on this and perhaps make TheBusyBot leave a comment like "Hey, we recommend you to use this branch naming convention..."

WalshyDev:
Sure, looking into that now

WalshyDev:
image

Yeah :sunglasses:

WalshyDev:
Pushed v1.1 and updated the workflow file

Example of usage is on the release page: https://github.com/WalshyDev/pr-labels/releases/tag/v1.1

TheBusyBiscuit (approved):

PR #2643, opened 2020-12-22
Refactoring for hasUnlocked() checks

TheBusyBiscuit:
## Description

Some refactoring, changing the old `Slimefun.hasUnlocked()` to be a bit cleaner, better documented and object-oriented.

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/core/services/localization/SlimefunLocalization.java
```diff
@@ -215,76 +215,76 @@ public ItemStack getRecipeTypeItem(Player p, RecipeType recipeType) {
     }
 
     @Override
-    public void sendMessage(CommandSender sender, String key, boolean addPrefix) {
+    public void sendMessage(CommandSender recipient, String key, boolean addPrefix) {
```
svr333, line 218: Its weird because the variable name `recipient` for some reason sounds wrong in this context (same for all the other uses in this file), and I would then suggest something like "endUser" but it literally means the same thing, guess my mind is just acting weird xD

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/runes/SoulboundRune.java
```diff
@@ -59,7 +60,7 @@ public ItemDropHandler getItemHandler() {
         };
     }
 
-    private void activate(Player p, Item rune) {
+    private void activate(@Nonnull Player p, @Nonnull Item rune) {
```
svr333, line 63: No need for Validate.notNull here?
Or is it checked upon in the methods which it calls later on?

svr333 (commented):
Not gonna approve yet because Im not sure if the Validate#notNull is a requirement or not, besides that its looking really good though, I like the changes

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/localization/SlimefunLocalization.java
TheBusyBiscuit: Yeah, it sounds weird and it is more accurate. `sender` had different implications, so despite this sounding weird it is more accurate.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/runes/SoulboundRune.java
TheBusyBiscuit: It's a private method anyway.
Only we can call it, so there isn't really any need for param validation since we have full control over what goes into this method.

svr333 (approved):

Sfiguz7 (dismissed):
LGTM

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -1012,6 +1015,90 @@ public void error(@Nonnull String message, @Nonnull Throwable throwable) {
...
+     * @return Whether this {@link Player} is able to use this {@link SlimefunItem}.
+     */
+    public boolean canUse(@Nonnull Player p, boolean sendMessage) {
+        Validate.notNull(p, "The Player cannot be null!");
+
+        if (getState() == ItemState.VANILLA_FALLBACK) {
+            // Vanilla items (which fell back) can always be used.
+            return true;
+        } else if (isDisabled()) {
+            // The Item has been disabled in the config
```
svr333, line 1048: Are these comments under the if statements necessary everywhere? They seem pretty self-explanatory to me

svr333 (dismissed):
Besides the comment remark it looks good, so Ill approve

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
TheBusyBiscuit: Better too many comments than too few tbh

WalshyDev (dismissed):
LGTM

svr333 (approved):

svr333 (approved):

WalshyDev (approved):
LGTM

PR #2637, opened 2020-12-19
Temporary fix for #2636

Patbox:
## Description
Quick, temporary fix for #2636. I try fixing it better when I will have more time.## Changes
Items won't drop from ghost blocks (for now)

## Related Issues
Resolves #2636

PR #2646, opened 2020-12-23
Readd drops from ghost blocks

Patbox:
## Description
This PR readds drops with ghost blocks (with additional checks, so it doesn't cause duplication)
## Changes
- Added `isPendingDeletion` to `TickerTask`
- Enabled drops from ghost blocks
- Event is now canceled after ghost block is removed (so data of new block isn't deleted)

## Related Issues
Replaces tempfix for #2636 (#2637)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -63,18 +63,16 @@ public class BlockListener implements Listener {
          */
         Block block = e.getBlock();
 
-        if (e.getBlockReplacedState().getType().isAir()) {
+        if (e.getBlockReplacedState().getType().isAir() && !SlimefunPlugin.getTickerTask().isPendingDeletion(block.getLocation())) {
             SlimefunItem sfItem = BlockStorage.check(block);
-
             if (sfItem != null) {
-                /* Temp fix for #2636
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -63,18 +63,16 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
...
             if (sfItem != null) {
-                /* Temp fix for #2636
                 for (ItemStack item : sfItem.getDrops()) {
                     if (item != null && !item.getType().isAir()) {
                         block.getWorld().dropItemNaturally(block.getLocation(), item);
                     }
                 }
-                 */
                 BlockStorage.clearBlockInfo(block);
+                e.setCancelled(true);
```
TheBusyBiscuit, line 75: Why is it cancelled here?

TheBusyBiscuit (commented):
This fix was already addressed in #2643 btw.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
Patbox: While testing, I found out, that sometimes placed blocks lost their data (like Charging Bench becoming just crafting table)

Patbox:
Ok, I'm closing it then.

PR #2648, opened 2020-12-23
Fixes #2647

### LINKED ISSUE #2647: Boots of Stomper work on NPCs

The boots of stomper can move NPC's in claimed area like spawn.

Fly up land and it moves the NPC's
I tested with op and no op and both move them
https://imgur.com/a/snIJYOw

Not to move the NPC's when landing

## :scroll: Server Log

N/a

N/a

 - Server Software: 
 - Minecraft Version: 
 - Slimefun Version: 
 - CS-CoreLib Version: 
https://imgur.com/a/snIJYOw

WalshyDev:
## Description
Add some more checks in order to make this more valid. This will resolve a lot of conflicts with other plugins. This has been confirmed fix with Citizens

## Changes
Check #isCollidable and #hasGravity

## Related Issues
Resolves #2647 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
@@ -49,7 +49,7 @@ public class StomperBoots extends SlimefunItem {
         player.setVelocity(new Vector(0, 0.7, 0));
 
         for (Entity entity : player.getNearbyEntities(4, 4, 4)) {
-            if (entity instanceof LivingEntity && canMoveEntity((LivingEntity) entity, player)) {
+            if (entity instanceof LivingEntity && canPush(player, (LivingEntity) entity)) {
                 Vector velocity = getShockwave(player.getLocation(), entity.getLocation());
                 entity.setVelocity(velocity);
 
@@ -93,9 +93,10 @@ public class StomperBoots extends SlimefunItem {
 
     /**
      * Checks if the stomper boots can move an entity and is not the player who is using the boots.
-     * <br><b>For developers:</b> If you're spawning an immovable NPC, you should be denying
-     * collision with {@code LivingEntity#setCollidable(false)} or
-     * gravity with {@code LivingEntity#setGravity(false)}.
+     * <p>
+     * <b>For developers:</b> If you're spawning an immovable NPC, you should be denying
+     * collision with {@link LivingEntity#setCollidable(boolean)} or
+     * gravity with {@link LivingEntity#setGravity(boolean)}.
      *
      * @param entity
      *            The {@link LivingEntity} to check.
@@ -103,7 +104,7 @@ public class StomperBoots extends SlimefunItem {
      *            The {@link Player} using the {@link StomperBoots}.
      * @return If the entity can move.
      */
-    private boolean canMoveEntity(@Nonnull LivingEntity entity, @Nonnull Player player) {
+    protected boolean canPush(@Nonnull Player player, @Nonnull LivingEntity entity) {
         return entity.isValid() && !entity.getUniqueId().equals(player.getUniqueId())
             && entity.isCollidable() && entity.hasGravity();
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
@@ -66,8 +66,8 @@ public class StomperBoots extends SlimefunItem {
         }
 
         for (BlockFace face : BlockFace.values()) {
-            Block b = player.getLocation().getBlock().getRelative(BlockFace.DOWN).getRelative(face);
-            player.getWorld().playEffect(b.getLocation(), Effect.STEP_SOUND, b.getType());
+            Block block = player.getLocation().getBlock().getRelative(BlockFace.DOWN).getRelative(face);
+            player.getWorld().playEffect(block.getLocation(), Effect.STEP_SOUND, block.getType());
         }
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
@@ -1,5 +1,7 @@
 package io.github.thebusybiscuit.slimefun4.implementation.items.armor;
 
+import javax.annotation.Nonnull;
+
 import org.bukkit.Bukkit;
 import org.bukkit.Effect;
 import org.bukkit.Location;
@@ -25,7 +27,7 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 /**
  * The Boots of the Stomper are boots which damage nearby enemies whenever the {@link Player}
  * takes fall damage.
- * 
+ *
  * @author TheBusyBiscuit
  *
  */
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
```diff
@@ -47,7 +47,12 @@ public void stomp(EntityDamageEvent fallDamageEvent) {
         p.setVelocity(new Vector(0, 0.7, 0));
 
         for (Entity n : p.getNearbyEntities(4, 4, 4)) {
-            if (n instanceof LivingEntity && n.isValid() && !n.getUniqueId().equals(p.getUniqueId())) {
+            if (n instanceof LivingEntity
+                && n.isValid()
+                && !n.getUniqueId().equals(p.getUniqueId())
+                && ((LivingEntity) n).isCollidable()
+                && n.hasGravity()
+            ) {
```
TheBusyBiscuit, line 55: I think this unholy abomination of new lines justifies to have its own dedicated method.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
WalshyDev: "abomination"

you seen a few lines down? 181 chars on a line monkaS

TheBusyBiscuit: Still, putting this statement into its own method would also allow for flexibility in case a sub implementation wants to add additional checks.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
```diff
@@ -37,54 +39,49 @@ public StomperBoots(Category category, SlimefunItemStack item, RecipeType recipe
...
+                        ((LivingEntity) entity).damage(event.getDamage());
                     }
                 }
             }
         }
 
         for (BlockFace face : BlockFace.values()) {
-            Block b = p.getLocation().getBlock().getRelative(BlockFace.DOWN).getRelative(face);
-            p.getWorld().playEffect(b.getLocation(), Effect.STEP_SOUND, b.getType());
+            Block b = player.getLocation().getBlock().getRelative(BlockFace.DOWN).getRelative(face);
```
J3fftw1, line 69: changes p to player but b not to block?

WalshyDev: reeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee leave me alone. Was just changing the shit associated with my method. Also `b` is a bit clearer than `n` or `p`

J3fftw1: i mean b is as clear as p in this case i agree n is weird but b and p are both really clear in this case so i would say make them all full name or 1 letter be consistent

svr333 (approved):
Looks good

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
```diff
@@ -89,4 +91,20 @@ private Vector getShockwave(Location origin, Location target) {
         }
     }
 
+    /**
+     * Checks if the stomper boots can move an entity and is not the player who is using the boots.
+     * <br><b>For developers:</b> If you're spawning an immovable NPC, you should be denying
+     * collision with {@code LivingEntity#setCollidable(false)} or
+     * gravity with {@code LivingEntity#setGravity(false)}.
```
TheBusyBiscuit, line 98: Why are you using `@code` instead of `@link` here?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
```diff
@@ -89,4 +91,20 @@ private Vector getShockwave(Location origin, Location target) {
...
+     *
+     * @param entity
+     *            The {@link LivingEntity} to check.
+     * @param player
+     *            The {@link Player} using the {@link StomperBoots}.
+     * @return If the entity can move.
+     */
+    private boolean canMoveEntity(@Nonnull LivingEntity entity, @Nonnull Player player) {
```
TheBusyBiscuit, line 106: I would suggest to swap the order of parameters here and rename the method.
The current order is unintuitive in my opinion or at least in a grammatical sense. If the method was called "canBePushedBy" then the current order would make more sense. So I am in favour of swapping these here.

Also if you want this to be overridden, then it should be at least `protected`.

```suggestion
    protected boolean canPush(@Nonnull Player player, @Nonnull LivingEntity entity) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
```diff
@@ -89,4 +91,20 @@ private Vector getShockwave(Location origin, Location target) {
         }
     }
 
+    /**
+     * Checks if the stomper boots can move an entity and is not the player who is using the boots.
+     * <br><b>For developers:</b> If you're spawning an immovable NPC, you should be denying
```
TheBusyBiscuit, line 96: According to the javadocs convention `<br>` should not be used. Instead `<p>` should be used to start a new paragraph.
Although you will see the line break quite often in this codebase already because I didn't know that until recently either.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
```diff
@@ -37,49 +39,49 @@ public StomperBoots(Category category, SlimefunItemStack item, RecipeType recipe
...
                 }
             }
         }
 
         for (BlockFace face : BlockFace.values()) {
-            Block b = p.getLocation().getBlock().getRelative(BlockFace.DOWN).getRelative(face);
-            p.getWorld().playEffect(b.getLocation(), Effect.STEP_SOUND, b.getType());
+            Block block = player.getLocation().getBlock().getRelative(BlockFace.DOWN).getRelative(face);
+            player.getWorld().playEffect(block.getLocation(), Effect.STEP_SOUND, block.getType());
         }
```
TheBusyBiscuit, line 71: I don't really see much of a problem with the naming of p, n and b but sure.

TheBusyBiscuit (changes requested):
A few nitpicky convention type of things and a mistake in the access modifier. Otherwise looking good.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
svr333: p and b could be used fine like that but n for entity is a bit odd here, and besides I think using full names is cleaner and more easily readable

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
WalshyDev: Just for the sake of adding `false` to be clear what people should do. I can change it

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/StomperBoots.java
WalshyDev: Was doing it for consistency but sute

svr333 (approved):

TheBusyBiscuit (approved):

PR #2657, opened 2020-12-27
Added a new issues template

TheBusyBiscuit:
This proposes a new "Performance Problem" issues template.
The template is very long and time-consuming.
But I am seriously tired of seeing so many people misusing the bug report template to post their incomplete or misinterpreted timings reports.

We have a full wiki article on how to investigate lag and have to point people there on a regular basis.
This template also ensures that we get an appropriate Spark report to properly investigate bottlenecks past the usual "yep, there is lag" level of insight that regular timings provide.


Sfiguz7 (approved):
Spark :D It is a bit long but if it doesn't contain all the above we ask for it anyway so it is perfect as is IMO

WalshyDev (approved):
LGTM just need to sort merge conflict

I feel like we're gonna close a lot of these right away though as people won't do Spark.

I wonder if we could have a command which runs timings and spark (could shade spark I guess) then let's them know to send the info it's generated. Would make it easier since then we know it's ran for at least X time and we have what we want.
I don't like it though as I wish people would just do as we ask and provide what we need *sigh*

J3fftw1:
> 
> 
> LGTM just need to sort merge conflict
> 
> I feel like we're gonna close a lot of these right away though as people won't do Spark.
> 
> I wonder if we could have a command which runs timings and spark (could shade spark I guess) then let's them know to send the info it's generated. Would make it easier since then we know it's ran for at least X time and we have what we want.
> I don't like it though as I wish people would just do as we ask and provide what we need _sigh_

I find it a little weird to force another plugin to servers. We want them to provide more info then we should provide them with a way to do it. Shading it in would be a good idea imo. Maybe for now make the spark section optional. Since we can get a lot of info from Aikars anyways and most of the time that's enough anyways

Also i agree on that we should have maybe a website that shows all timings 
- Aikars
- Sparks
- Slimefuns

.github/ISSUE_TEMPLATE/---performance-problem.md
```diff
@@ -0,0 +1,58 @@
...
+<!-- Make sure to install Spark and follow the steps carefully to ensure that we can give you the best response possible. -->
+
+# :warning: Prerequisites
+1. Make sure you have read the wiki article on Server-Optimization: 
+   https://github.com/Slimefun/Slimefun4/wiki/Server-Optimizations
+2. Please install Spark: 
+   https://www.spigotmc.org/resources/spark.57242/)
+
+## :timer_clock: 1. Generic Timings Report (REQUIRED)
+<!-- Run `/timings on` and wait at least 5 minutes! -->
```
J3fftw1, line 20: Just wanted to comment on it 
Timings are on by default if you just let them run /timings paste it will get it from when the server started till they do the comment
In paper atleast

TheBusyBiscuit: Not in Spigot though afaik.

TheBusyBiscuit:
@J3fftw1
The Spark portion will definitely not be optional.
Normal timings rarely gets us anywhere, it only boils down per-task which is useless in our case.
The only reason normal timings is sometimes useful is because it can determine whether there **is** lag and which plugin it's coming from. But beyond that it is not sufficient enough to isolate the issue itself.

This template shouldn't serve as a dumping ground for people to post any timings reports they get.
The reason for it would be to properly investigate performance problems with Slimefun itself, and that comes with the aforementioned prerequisites. Otherwise we would be in the same position as we are today...

I'll close this for now though, thanks for your feedback @WalshyDev @Sfiguz7.
Shading spark would of course be a bit stupid - for obvious reasons. I wish it had a proper API we can utilise but I had the same idea as Walshy, making an all-in-one command for that. Perhaps zipping the result (aikar + spark + slimefun) simply and prompting them to include the required zip in their report is a better approach.
It would definitely be more user-friendly, so I'll close this off for now until we have a command like that.

J3fftw1:
> @J3fftw1
> The Spark portion will definitely not be optional.
> Normal timings rarely gets us anywhere, it only boils down per-task which is useless in our case.
> The only reason normal timings is sometimes useful is because it can determine whether there **is** lag and which plugin it's coming from. But beyond that it is not sufficient enough to isolate the issue itself.
> 
> This template shouldn't serve as a dumping ground for people to post any timings reports they get.
> The reason for it would be to properly investigate performance problems with Slimefun itself, and that comes with the aforementioned prerequisites. Otherwise we would be in the same position as we are today...
> 
> I'll close this for now though, thanks for your feedback @WalshyDev @Sfiguz7.
> Shading spark would of course be a bit stupid - for obvious reasons. I wish it had a proper API we can utilise but I had the same idea as Walshy, making an all-in-one command for that. Perhaps zipping the result (aikar + spark + slimefun) simply and prompting them to include the required zip in their report is a better approach.
> It would definitely be more user-friendly, so I'll close this off for now until we have a command like that.

May i ask what the obvious reason is?

TheBusyBiscuit:
> May i ask what the obvious reason is?

@J3fftw1 
Shading has very high costs in terms of memory, disk space, redundancies and especially coherency.
When you shade a project then you are required to also shade all of these projects dependencies to make it run.
So shading spark, will also shade all of these:

```gradle
dependencies {
    compile 'org.ow2.asm:asm:7.1'
    compile 'com.google.protobuf:protobuf-java:3.0.0'
    compile 'com.squareup.okhttp3:okhttp:3.14.1'
    compile 'com.squareup.okio:okio:1.17.3'
    compile 'org.tukaani:xz:1.8'
    compile('net.kyori:adventure-api:4.1.1') {
        exclude(module: 'checker-qual')
    }
    compile('net.kyori:adventure-text-serializer-gson:4.1.1') {
        exclude(module: 'adventure-api')
        exclude(module: 'gson')
    }
    compile('net.kyori:adventure-text-serializer-legacy:4.1.1') {
        exclude(module: 'adventure-api')
    }
    compile('net.kyori:adventure-text-feature-pagination:4.0.0-SNAPSHOT') {
        exclude(module: 'adventure-api')
    }
    compile('net.kyori:adventure-platform-bukkit:4.0.0-SNAPSHOT') {
        exclude(module: 'adventure-api')
        exclude(module: 'adventure-text-serializer-gson')
    }
}
```

As you can see, this while skyrocket our file size suddenly which means higher cost in terms of memory and disk space.
While this in of itself is not a huge deal (because honestly... what's a few MB on disks nowadays...), it will mean that we have two options.
Either we relocate or not relocate the dependencies, relocating them means that we woill waste quite a lot of RAM through redundant resource management and will also make us worry about updating this module frequently.
If we don't relocate, then that means we will run into classpath errors as redundant modules battle for memory locations.
This will eventually lead to many more problems where we either disallow people from using the standalone plugin or constantly fix stuff and tell people to update etc...
In most cases, relocation is a much safer option at the cost of wasting memory.

If Spark was an API, then all of these issues would be worth the trouble, but it's a plugin and Bukkit doesn't like to have the same plugin loaded twice, so we would need a lot of workarounds for this, especially because there is no way to intercept a timings report. So we would eventually have to rewrite the entire Bukkit module for our needs and still carry all the luggage as mentioned before.
This is in no way feasible as it is simply not worth all this trouble. You can either spend weeks trying to make this work and run into thousands of problems or... the much more reasonable approach: "Please use Spark, thank you".
I'm all for a Spark integration, but it should stay an external dependency, shading it is just not a good idea.

J3fftw1:
> > May i ask what the obvious reason is?
> 
> @J3fftw1
> Shading has very high costs in terms of memory, disk space, redundancies and especially coherency.
> When you shade a project then you are required to also shade all of these projects dependencies to make it run.
> So shading spark, will also shade all of these:
> 
> ```groovy-gradle
> dependencies {
>     compile 'org.ow2.asm:asm:7.1'
>     compile 'com.google.protobuf:protobuf-java:3.0.0'
>     compile 'com.squareup.okhttp3:okhttp:3.14.1'
>     compile 'com.squareup.okio:okio:1.17.3'
>     compile 'org.tukaani:xz:1.8'
>     compile('net.kyori:adventure-api:4.1.1') {
>         exclude(module: 'checker-qual')
>     }
>     compile('net.kyori:adventure-text-serializer-gson:4.1.1') {
>         exclude(module: 'adventure-api')
>         exclude(module: 'gson')
>     }
>     compile('net.kyori:adventure-text-serializer-legacy:4.1.1') {
>         exclude(module: 'adventure-api')
>     }
>     compile('net.kyori:adventure-text-feature-pagination:4.0.0-SNAPSHOT') {
>         exclude(module: 'adventure-api')
>     }
>     compile('net.kyori:adventure-platform-bukkit:4.0.0-SNAPSHOT') {
>         exclude(module: 'adventure-api')
>         exclude(module: 'adventure-text-serializer-gson')
>     }
> }
> ```
> 
> As you can see, this while skyrocket our file size suddenly which means higher cost in terms of memory and disk space.
> While this in of itself is not a huge deal (because honestly... what's a few MB on disks nowadays...), it will mean that we have two options.
> Either we relocate or not relocate the dependencies, relocating them means that we woill waste quite a lot of RAM through redundant resource management and will also make us worry about updating this module frequently.
> If we don't relocate, then that means we will run into classpath errors as redundant modules battle for memory locations.
> This will eventually lead to many more problems where we either disallow people from using the standalone plugin or constantly fix stuff and tell people to update etc...
> In most cases, relocation is a much safer option at the cost of wasting memory.
> 
> If Spark was an API, then all of these issues would be worth the trouble, but it's a plugin and Bukkit doesn't like to have the same plugin loaded twice, so we would need a lot of workarounds for this, especially because there is no way to intercept a timings report. So we would eventually have to rewrite the entire Bukkit module for our needs and still carry all the luggage as mentioned before.
> This is in no way feasible as it is simply not worth all this trouble. You can either spend weeks trying to make this work and run into thousands of problems or... the much more reasonable approach: "Please use Spark, thank you".
> I'm all for a Spark integration, but it should stay an external dependency, shading it is just not a good idea.

I still find it really weird that we require them to use given software. We also dont force people to use paper if they use spigot. We always just tell them to use it. But now in this closed pr we are forcing them to use it. I think if this ever gets added we should have our own way of getting the timings. maybe how you suggested with the zip file or with a whole different method. But i really think this needs to be done within slimefun or cscorelib

TheBusyBiscuit:
We force them to use Spigot or Paper, otherwise the plugin won't work. Simple as that. CraftBukkit for example is no longer supported either.

Spark is obviously not required to run Slimefun.
But if they want a Spark performance report, they need Spark, simple as that.

PR #2680, opened 2021-01-03
Added Book Binder

ProfElements:
## Description
Added the Book Binder, an electric machine that takes enchanted books and combines them, without regards to maximum level.

## Changes
Added the book binder, put the SlimefunItemStack and SlimefunItem where I thought they would go

## Checklist

- [] I have fully tested the proposed changes and promise that they will not break everything into chaos.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
@@ -1,0 +1,111 @@
+package io.github.thebusybiscuit.slimefun4.implementation.items.electric.machines;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+
+import org.bukkit.Material;
+import org.bukkit.enchantments.Enchantment;
+import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.EnchantmentStorageMeta;
+
+import io.github.thebusybiscuit.cscorelib2.inventory.InvUtils;
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.AContainer;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineRecipe;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
+
+public class BookBinder extends AContainer {
+
+    @ParametersAreNonnullByDefault
+    public BookBinder(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        super(category, item, recipeType, recipe);
+    }
+
+    @Override
+    protected MachineRecipe findNextRecipe(BlockMenu menu) {
+        for (int slot : getInputSlots()) {
+            ItemStack target = menu.getItemInSlot(slot == getInputSlots()[0] ? getInputSlots()[1] : getInputSlots()[0]);
+
+            if (!isCompatible(target)) {
+                return null;
+            }
+
+            ItemStack item = menu.getItemInSlot(slot);
+            if  (item != null && item.getType() == Material.ENCHANTED_BOOK && target != null) {
+                Map<Enchantment, Integer> enchantments = new HashMap<>();
+                int amount = 0;
+                EnchantmentStorageMeta itemMeta = (EnchantmentStorageMeta) item.getItemMeta();
+                EnchantmentStorageMeta targetMeta = (EnchantmentStorageMeta) target.getItemMeta();
+
+                for(Map.Entry<Enchantment, Integer> e : itemMeta.getStoredEnchants().entrySet()) {
+                    for (Map.Entry<Enchantment, Integer> e2 : targetMeta.getStoredEnchants().entrySet()) {
+                        if (e.getKey() == e2.getKey()) {
+                            if (e.getValue() == e2.getValue()) {
+                                enchantments.put(e.getKey(), e.getValue()+1);
+                            } else {
+                                enchantments.put(e.getKey(), e.getValue() > e2.getValue() ? e.getValue() : e2.getValue());
+                            }
+                        }
+                        enchantments.put(e2.getKey(), e2.getValue());
+                    }
+
+                    amount++;
+                    enchantments.put(e.getKey(), e.getValue());
+                }
+
+                if (amount > 0) {
+                    ItemStack book = new ItemStack(Material.ENCHANTED_BOOK);
+                    book.setAmount(1);
+
+                    EnchantmentStorageMeta enchantMeta = (EnchantmentStorageMeta) book.getItemMeta();
+                    for (Map.Entry<Enchantment, Integer> entry : enchantments.entrySet()) {
+                        enchantMeta.addStoredEnchant(entry.getKey(), entry.getValue(), true);
+                    }
+
+                    book.setItemMeta(enchantMeta);
+                    
+                    MachineRecipe recipe = new MachineRecipe(50 * amount / this.getSpeed(), new ItemStack[] {target, item}, new ItemStack[] {book});
+
+                    if(!InvUtils.fitAll(menu.toInventory(), recipe.getOutput(), getOutputSlots())) {
+                        return null;
+                    }
+
+                    for (int inputSlot : getInputSlots()) {
+                        menu.consumeItem(inputSlot);
+                    }
+
+                    return recipe;
+                }
+
+                return null;
+            }
+
+        }
+
+        return null;
+    }
+
+    private boolean isCompatible(ItemStack item) {
+        if (item != null && item.getType() == Material.ENCHANTED_BOOK) {
+            return true;
+
+            //Figure out what to do with incompatible enchantments if we should allow it at all ex. Protection + Fire Protection
+        }
+
+        return false;
+    }
+
+    @Override
+    public ItemStack getProgressBar() {
+        return new ItemStack(Material.IRON_CHESTPLATE);
+    }
+
+    @Override
+    public String getMachineIdentifier() {
+        return "BOOK_BINDER";
+    }
+}
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
@@ -301,56 +301,70 @@ public final class SlimefunItems {
 
     public static final SlimefunItemStack BOOTS_OF_THE_STOMPER = new SlimefunItemStack("BOOTS_OF_THE_STOMPER", Material.LEATHER_BOOTS, Color.AQUA, "&bBoots of the Stomper", "", "&9All Fall Damage you receive", "&9will be applied to nearby Mobs/Players", "", "&9+ No Fall Damage");
 
+    public static final SlimefunItemStack BEE_HELMET = new SlimefunItemStack("BEE_HELMET", Material.GOLDEN_HELMET, "&e&lBee Helmet", " ", "&fBzzzzzzz");
+    public static final SlimefunItemStack BEE_WINGS = new SlimefunItemStack("BEE_WINGS", Material.ELYTRA, "&e&lBee Wings", " ", "&fBzzzzzzz", " ", "&9Activates Slow falling", "&9when approaching the ground");
+    public static final SlimefunItemStack BEE_LEGGINGS = new SlimefunItemStack("BEE_LEGGINGS", Material.GOLDEN_LEGGINGS, "&e&lBee Leggings", " ", "&fBzzzzzzz");
+    public static final SlimefunItemStack BEE_BOOTS = new SlimefunItemStack("BEE_BOOTS", Material.GOLDEN_BOOTS, "&e&lBee Boots", "", "&fBzzzzzzz", "", "&9+ Jump Boost", "&9+ No Fall Damage");
+
     static {
-        Map<Enchantment, Integer> cactus = new HashMap<>();
-        cactus.put(Enchantment.THORNS, 3);
-        cactus.put(Enchantment.DURABILITY, 6);
+        Map<Enchantment, Integer> cactusEnchs = new HashMap<>();
+        cactusEnchs.put(Enchantment.THORNS, 3);
+        cactusEnchs.put(Enchantment.DURABILITY, 6);
 
-        CACTUS_HELMET.addUnsafeEnchantments(cactus);
-        CACTUS_CHESTPLATE.addUnsafeEnchantments(cactus);
-        CACTUS_LEGGINGS.addUnsafeEnchantments(cactus);
-        CACTUS_BOOTS.addUnsafeEnchantments(cactus);
+        CACTUS_HELMET.addUnsafeEnchantments(cactusEnchs);
+        CACTUS_CHESTPLATE.addUnsafeEnchantments(cactusEnchs);
+        CACTUS_LEGGINGS.addUnsafeEnchantments(cactusEnchs);
+        CACTUS_BOOTS.addUnsafeEnchantments(cactusEnchs);
 
-        Map<Enchantment, Integer> damascus = new HashMap<>();
-        damascus.put(Enchantment.DURABILITY, 5);
-        damascus.put(Enchantment.PROTECTION_ENVIRONMENTAL, 5);
+        Map<Enchantment, Integer> damascusEnchs = new HashMap<>();
+        damascusEnchs.put(Enchantment.DURABILITY, 5);
+        damascusEnchs.put(Enchantment.PROTECTION_ENVIRONMENTAL, 5);
 
-        DAMASCUS_STEEL_HELMET.addUnsafeEnchantments(damascus);
-        DAMASCUS_STEEL_CHESTPLATE.addUnsafeEnchantments(damascus);
-        DAMASCUS_STEEL_LEGGINGS.addUnsafeEnchantments(damascus);
-        DAMASCUS_STEEL_BOOTS.addUnsafeEnchantments(damascus);
+        DAMASCUS_STEEL_HELMET.addUnsafeEnchantments(damascusEnchs);
+        DAMASCUS_STEEL_CHESTPLATE.addUnsafeEnchantments(damascusEnchs);
+        DAMASCUS_STEEL_LEGGINGS.addUnsafeEnchantments(damascusEnchs);
+        DAMASCUS_STEEL_BOOTS.addUnsafeEnchantments(damascusEnchs);
 
-        Map<Enchantment, Integer> reinforced = new HashMap<>();
-        reinforced.put(Enchantment.DURABILITY, 9);
-        reinforced.put(Enchantment.PROTECTION_ENVIRONMENTAL, 9);
+        Map<Enchantment, Integer> reinforcedEnchs = new HashMap<>();
+        reinforcedEnchs.put(Enchantment.DURABILITY, 9);
+        reinforcedEnchs.put(Enchantment.PROTECTION_ENVIRONMENTAL, 9);
 
-        REINFORCED_ALLOY_HELMET.addUnsafeEnchantments(reinforced);
-        REINFORCED_ALLOY_CHESTPLATE.addUnsafeEnchantments(reinforced);
-        REINFORCED_ALLOY_LEGGINGS.addUnsafeEnchantments(reinforced);
-        REINFORCED_ALLOY_BOOTS.addUnsafeEnchantments(reinforced);
+        REINFORCED_ALLOY_HELMET.addUnsafeEnchantments(reinforcedEnchs);
+        REINFORCED_ALLOY_CHESTPLATE.addUnsafeEnchantments(reinforcedEnchs);
+        REINFORCED_ALLOY_LEGGINGS.addUnsafeEnchantments(reinforcedEnchs);
+        REINFORCED_ALLOY_BOOTS.addUnsafeEnchantments(reinforcedEnchs);
 
-        Map<Enchantment, Integer> gilded = new HashMap<>();
-        gilded.put(Enchantment.DURABILITY, 6);
-        gilded.put(Enchantment.PROTECTION_ENVIRONMENTAL, 8);
+        Map<Enchantment, Integer> gildedEnchs = new HashMap<>();
+        gildedEnchs.put(Enchantment.DURABILITY, 6);
+        gildedEnchs.put(Enchantment.PROTECTION_ENVIRONMENTAL, 8);
 
-        GILDED_IRON_HELMET.addUnsafeEnchantments(gilded);
-        GILDED_IRON_CHESTPLATE.addUnsafeEnchantments(gilded);
-        GILDED_IRON_LEGGINGS.addUnsafeEnchantments(gilded);
-        GILDED_IRON_BOOTS.addUnsafeEnchantments(gilded);
+        GILDED_IRON_HELMET.addUnsafeEnchantments(gildedEnchs);
+        GILDED_IRON_CHESTPLATE.addUnsafeEnchantments(gildedEnchs);
+        GILDED_IRON_LEGGINGS.addUnsafeEnchantments(gildedEnchs);
+        GILDED_IRON_BOOTS.addUnsafeEnchantments(gildedEnchs);
 
         GOLDEN_HELMET_12K.addUnsafeEnchantment(Enchantment.DURABILITY, 10);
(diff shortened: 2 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+                            } else {
+                                enchantments.put(e.getKey(), e.getValue() > e2.getValue() ? e.getValue() : e2.getValue());
+                            }
+                        }
+                        enchantments.put(e2.getKey(), e2.getValue());
+                    }
+
+                    amount++;
+                    enchantments.put(e.getKey(), e.getValue());
+                }
```
TheBusyBiscuit, line 63: This section could be replaced with a more elegant solution.

```suggestion
                Map<Enchantment, Integer> storedEnchantments = itemMeta.getStoredEnchants();
                enchantments.putAll(storedEnchantments);
                for (Map.Entry<Enchantment, Integer> entry : storedEnchantments.entrySet()) {
                    enchantments.merge(entry.getKey(), entry.getValue(), (a, b) -> {
                            if (a == b) {
                                return a + 1;
                            } else {
                                return Math.max(a, b);
                            }
                    });
                }
```

I haven't tested this, so I might be missing a semicolon or similar somewhere. But i think this would be much cleaner.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+            if (!isCompatible(target)) {
+                return null;
+            }
+
+            ItemStack item = menu.getItemInSlot(slot);
+            if  (item != null && item.getType() == Material.ENCHANTED_BOOK && target != null) {
+                Map<Enchantment, Integer> enchantments = new HashMap<>();
+                int amount = 0;
+                EnchantmentStorageMeta itemMeta = (EnchantmentStorageMeta) item.getItemMeta();
+                EnchantmentStorageMeta targetMeta = (EnchantmentStorageMeta) target.getItemMeta();
```
TheBusyBiscuit, line 47: You are casting an unknown ItemMeta to EnchantmentStorageMeta here.
You should also check the Material of `target`, even if it is already done in `isCompatible`.
Otherwise one might get the impression that it is missing.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+import io.github.thebusybiscuit.cscorelib2.inventory.InvUtils;
+import io.github.thebusybiscuit.slimefun4.api.items.ItemSetting;
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.AContainer;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineRecipe;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
+
+public class BookBinder extends AContainer {
```
TheBusyBiscuit, line 22: Missing documentation, please add a javadoc description to this class with an author tag.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private final ItemSetting<Boolean> deniesVanillaRules = new ItemSetting<>("denies-vanilla-rules", true); 
+
+    @ParametersAreNonnullByDefault
+    public BookBinder(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        super(category, item, recipeType, recipe);
+        
+       addItemSetting(deniesVanillaRules);
```
TheBusyBiscuit, line 30: indentation

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+                    enchantments.put(e.getKey(), e.getValue());
+                }
+
+                if (amount > 0) {
+                    ItemStack book = new ItemStack(Material.ENCHANTED_BOOK);
+                    book.setAmount(1);
+
+                    EnchantmentStorageMeta enchantMeta = (EnchantmentStorageMeta) book.getItemMeta();
+                    for (Map.Entry<Enchantment, Integer> entry : enchantments.entrySet()) {
+                        if(deniesVanillaRules.getValue()) {
```
TheBusyBiscuit, line 71: Space after `if`. You could also store the expression `deniesVanillaRules.getValue()` inside a variable btw.
```suggestion
                        if (deniesVanillaRules.getValue()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+                            enchantMeta.addStoredEnchant(entry.getKey(), entry.getValue(), false);
+                        }
+                        
+                    }
+
+                    book.setItemMeta(enchantMeta);
+                    
+                    MachineRecipe recipe = new MachineRecipe(50 * amount / this.getSpeed(), new ItemStack[] {target, item}, new ItemStack[] {book});
+
+                    if(!InvUtils.fitAll(menu.toInventory(), recipe.getOutput(), getOutputSlots())) {
```
TheBusyBiscuit, line 83: Space after `if`
```suggestion
                    if (!InvUtils.fitAll(menu.toInventory(), recipe.getOutput(), getOutputSlots())) {
```

TheBusyBiscuit (changes requested):
Looking pretty good, a bit inconsistent code styling though but that's fine and not hard to fix.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.AContainer;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineRecipe;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
+
+public class BookBinder extends AContainer {
+
+    private final ItemSetting<Boolean> deniesVanillaRules = new ItemSetting<>("denies-vanilla-rules", true);
```
WalshyDev, line 24: I would recommend renaming this. Something like "bypass-max-level".

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+            }
+
+            ItemStack item = menu.getItemInSlot(slot);
+            if  (item != null && item.getType() == Material.ENCHANTED_BOOK && target != null) {
+                Map<Enchantment, Integer> enchantments = new HashMap<>();
+                int amount = 0;
+                EnchantmentStorageMeta itemMeta = (EnchantmentStorageMeta) item.getItemMeta();
+                EnchantmentStorageMeta targetMeta = (EnchantmentStorageMeta) target.getItemMeta();
+
+                for(Map.Entry<Enchantment, Integer> e : itemMeta.getStoredEnchants().entrySet()) {
```
WalshyDev, line 49: Space after for
```suggestion
                for (Map.Entry<Enchantment, Integer> e : itemMeta.getStoredEnchants().entrySet()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+                        enchantments.put(e2.getKey(), e2.getValue());
+                    }
+
+                    amount++;
+                    enchantments.put(e.getKey(), e.getValue());
+                }
+
+                if (amount > 0) {
+                    ItemStack book = new ItemStack(Material.ENCHANTED_BOOK);
+                    book.setAmount(1);
```
WalshyDev, line 67: This line isn't needed

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,121 @@
...
+
+                    EnchantmentStorageMeta enchantMeta = (EnchantmentStorageMeta) book.getItemMeta();
+                    for (Map.Entry<Enchantment, Integer> entry : enchantments.entrySet()) {
+                        if(deniesVanillaRules.getValue()) {
+                            enchantMeta.addStoredEnchant(entry.getKey(), entry.getValue(), true);
+                        } else {
+                            enchantMeta.addStoredEnchant(entry.getKey(), entry.getValue(), false);
+                        }
+
```
WalshyDev, line 76: Remove empty line

WalshyDev (changes requested):

Mooy1:
Dont think unlimited combining is a good idea, maybe have a config option for it to enable.

ProfElements:
This is accounted for in the book#addStoredEnchant which when true adheres to vanilla max levels

ProfElements:
Unless you mean you dont want to adhere to vanilla max levels and have a custom max level?

ProfElements:
There you go, custom max level if you want it

Mooy1:
id make bypass false by default, otherwise looks gud

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,132 @@
...
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
+
+/**
+* Represents Book Binder, a machine that binds multiple enchantments books into one.
+* @author ProfElements
+*/
+public class BookBinder extends AContainer {
+
+    private final ItemSetting<Boolean> bypassVanillaMaxLevel = new ItemSetting<>("bypass-vanilla-max-level", false);
+    private final ItemSetting<Boolean> hasCustomMaxLevel = new ItemSetting<>("has-custom-max-level", true);
```
WalshyDev, line 30: You have bypass vanilla limit as false but then also have a custom max which is above MC limit which is true by default. These should either both be true or both be false by default. If it shouldn't bypass vanilla levels then it shouldn't have a custom one too.

If you set bypass vanilla as false but custom as true, what is the behaviour?

ProfElements:
Yeah I forgot about that because I was messing around with how it felt and didnt change the default value. Thank you :bow:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,132 @@
...
+
+                return null;
+            }
+
+        }
+
+        return null;
+    }
+
+    private boolean isCompatible(ItemStack item) {
```
TheBusyBiscuit, line 115: You should add an annotation here
```suggestion
    private boolean isCompatible(@Nullable ItemStack item) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,132 @@
...
+                                
+                            } else {
+                                if (hasCustomMaxLevel.getValue()) {
+                                    return Math.max(a, b) > customMaxLevel.getValue() ? customMaxLevel.getValue() : Math.max(a, b);
+                                } else {
+                                    return Math.max(a, b);
+                                }
+                               
+                            }
+                    });
```
TheBusyBiscuit, line 76: This is getting a bit out of hand in terms of nesting :/
You could make a seperate private `combineEnchantments` method which takes two inputs and then just reference it using `this::combineEnchantments`. That would reduce the nesting here and make it a bit cleaner to read.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,132 @@
...
+import me.mrCookieSlime.Slimefun.Lists.RecipeType;
+import me.mrCookieSlime.Slimefun.Objects.Category;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.AContainer;
+import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineRecipe;
+import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
+
+/**
+* Represents Book Binder, a machine that binds multiple enchantments books into one.
+* @author ProfElements
```
TheBusyBiscuit, line 25: space :)
```suggestion
* Represents Book Binder, a machine that binds multiple enchantments books into one.
*
* @author ProfElements
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,142 @@
...
+        return null;
+    }
+
+    private boolean isCompatible(@Nullable ItemStack item) {
+        if (item != null && item.getType() == Material.ENCHANTED_BOOK) {
+            return true;
+        }
+
+        return false;
+    }
```
J3fftw1, line 103: shouldnt this be else return false;

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,142 @@
...
+    protected MachineRecipe findNextRecipe(BlockMenu menu) {
+        for (int slot : getInputSlots()) {
+            ItemStack target = menu.getItemInSlot(slot == getInputSlots()[0] ? getInputSlots()[1] : getInputSlots()[0]);
+
+            if (!isCompatible(target)) {
+                return null;
+            }
+
+            ItemStack item = menu.getItemInSlot(slot);
+            if  (isCompatible(item) && isCompatible(target)) {
```
J3fftw1, line 52: this check isCompatible(target) looks redundant

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,138 @@
...
+    public ItemStack getProgressBar() {
+        return new ItemStack(Material.IRON_CHESTPLATE);
+    }
+
+    @Override
+    public String getMachineIdentifier() {
+        return "BOOK_BINDER";
+    }
+
+    private Map<Enchantment, Integer> combineEnchantments(Map<Enchantment, Integer> ech1, Map<Enchantment, Integer> ech2) {
```
TheBusyBiscuit, line 111:
```suggestion
    @Nonnull
    @ParametersAreNonnullByDefault
    private Map<Enchantment, Integer> combineEnchantments(Map<Enchantment, Integer> ech1, Map<Enchantment, Integer> ech2) {
```

(Nonnull would need to be imported too)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,138 @@
...
+                        } else {
+                            return a + 1;
+                        }
+                        
+                    } else {
+                        if (hasCustomMaxLevel.getValue()) {
+                            return Math.max(a, b) > customMaxLevel.getValue() ? customMaxLevel.getValue() : Math.max(a, b);
+                        } else {
+                            return Math.max(a, b);
+                        }
```
TheBusyBiscuit, line 129: You are doing Math.max(a,b) here regardless of the state, in one branch even twice.
So I think it would be wise to create a variable here.

```suggestion
                        int highestLevel = Math.max(a, b);
                        if (hasCustomMaxLevel.getValue()) {
                            return highestLevel > customMaxLevel.getValue() ? customMaxLevel.getValue() : highestLevel;
                        } else {
                            return highestLevel;
                        }
```

TheBusyBiscuit (changes requested):
Looking good thus far.
I spotted two more things that could be improved. But other than these it looks good as is.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,143 @@
...
+                EnchantmentStorageMeta itemMeta = (EnchantmentStorageMeta) item.getItemMeta();
+                EnchantmentStorageMeta targetMeta = (EnchantmentStorageMeta) target.getItemMeta();
+
+                Map<Enchantment, Integer> storedItemEnchantments = itemMeta.getStoredEnchants();
+                Map<Enchantment, Integer> storedTargetEnchantments = targetMeta.getStoredEnchants();
+                Map<Enchantment, Integer> enchantments = combineEnchantments(storedItemEnchantments, storedTargetEnchantments);
+                
+                if (enchantments.size() > 0) {
+                    ItemStack book = new ItemStack(Material.ENCHANTED_BOOK);
+                    book.setAmount(1);
```
WalshyDev, line 60: this line is still redundant

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,143 @@
...
+
+                    EnchantmentStorageMeta enchantMeta = (EnchantmentStorageMeta) book.getItemMeta();
+                    for (Map.Entry<Enchantment, Integer> entry : enchantments.entrySet()) {
+                        if (bypassVanillaMaxLevel.getValue()) {
+                            enchantMeta.addStoredEnchant(entry.getKey(), entry.getValue(), true);
+                        } else {
+                            enchantMeta.addStoredEnchant(entry.getKey(), entry.getValue(), false);
+                        }
```
WalshyDev, line 68: This can be simplified
```suggestion
                        enchantMeta.addStoredEnchant(entry.getKey(), entry.getValue(), bypassVanillaMaxLevel.getValue());
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,143 @@
...
+                    }
+
+                    book.setItemMeta(enchantMeta);
+                    
+                    MachineRecipe recipe = new MachineRecipe(25 * enchantments.size() / this.getSpeed(), new ItemStack[] {target, item}, new ItemStack[] {book});
```
WalshyDev, line 73: Slight change to make this just a bit more readable
```suggestion
                    MachineRecipe recipe = new MachineRecipe(25 * (enchantments.size() / this.getSpeed()), new ItemStack[] {target, item}, new ItemStack[] {book});
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,143 @@
...
+
+        return null;
+    }
+
+    private boolean isCompatible(@Nullable ItemStack item) {
+        if (item != null && item.getType() == Material.ENCHANTED_BOOK) {
+            return true;
+        } else {
+            return false;
+        }
```
WalshyDev, line 99: Can be simplified to
```suggestion
        return item != null && item.getType() == Material.ENCHANTED_BOOK;
```

WalshyDev (changes requested):

WalshyDev (approved):
LGTM

TheBusyBiscuit (changes requested):
Looks perfect!

Only one thing missing before I will approve it.
I noticed that there is no research for this item, meaning it will be unlocked by default.
I think this should definitely be gated behind a research!

Also did you test this ingame?

J3fftw1:
> Looks perfect!
> 
> Only one thing missing before I will approve it.
> I noticed that there is no research for this item, meaning it will be unlocked by default.
> I think this should definitely be gated behind a research!
> 
> Also did you test this ingame?

Havent checked the code but does it also make the apply xp higher
So when you combine 20 books it will require more xp to be used on normal anvil?

TheBusyBiscuit:
> Havent checked the code but does it also make the apply xp higher
> So when you combine 20 books it will require more xp to be used on normal anvil?

Doesn't Minecraft do that automatically?
At least there is no sign or reference for any method in `EnchantmentStorageMeta` which would imply the opposite.

ProfElements:
I have been messing around with the book binder on both my private server and the dev server I use to test and develop my branches of slimefun and dynatech, I checked the the max vanilla bypass works, the custom max levels works. I however did not test to check how the book binding affects enchantment books in a vanilla anvil

J3fftw1:
> > Havent checked the code but does it also make the apply xp higher
> > So when you combine 20 books it will require more xp to be used on normal anvil?
> 
> Doesn't Minecraft do that automatically?
> At least there is no sign or reference for any method in `EnchantmentStorageMeta` which would imply the opposite.

Dont we also do something with this in the auto enchanter i can remember seeing something like this somewhere cant remember wherr but i think the auto enchanter

TheBusyBiscuit:
> Dont we also do something with this in the auto enchanter i can remember seeing something like this somewhere cant remember wherr but i think the auto enchanter

damageables have a repairCost attribute, yes. But Enchantment books do not have such a thing, so I assume it should be automatically inferred.

J3fftw1:
> > Dont we also do something with this in the auto enchanter i can remember seeing something like this somewhere cant remember wherr but i think the auto enchanter
> 
> damageables have a repairCost attribute, yes. But Enchantment books do not have such a thing, so I assume it should be automatically inferred.

Mmh then we should somehow cap it iguess since the prices can go sky rocket then when you apply a lot of books?

TheBusyBiscuit:
> Mmh then we should somehow cap it iguess since the prices can go sky rocket then when you apply a lot of books?

Did you even read any of my messages... Again, Enchantment books have no defined cost.

TheBusyBiscuit:
I think we might have forgotten to check if enchantments conflict with one another :NotLikeThis:
Like Fortune and Silk Touch.

ProfElements:
Checking rn.

ProfElements:
It can infact conflicting enchantments (tested with Fire Protection IV and Protection IV)

Mooy1:
Enchanter doesnt care so I think conflicting ones are fine, maybe limit it to like 10 enchants per book tho.

J3fftw1:
> 
> 
> Enchanter doesnt care so I think conflicting ones are fine, maybe limit it to like 10 enchants per book tho.

normal anvils care though

ProfElements:
> Enchanter doesnt care so I think conflicting ones are fine, maybe limit it to like 10 enchants per book tho.
Normal enchantments are not limited to number, so Im going to leave it as is, maybe make bypassVanillaMaxLevel > bypassVanillaRules and use that is a good compromise?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/BookBinder.java
```diff
@@ -0,0 +1,143 @@
...
+
+                        if (hasCustomMaxLevel.getValue()) {
+                            return highestLevel > customMaxLevel.getValue() ? customMaxLevel.getValue() : highestLevel;
+                        } else {
+                            return highestLevel;
+                        }
+                    
+                    }
+                });                
+            }
```
TheBusyBiscuit, line 137: This could probably be optimized a lot.
But I think we were holding this back for long enough and can always improve it later.

TheBusyBiscuit (dismissed):

TheBusyBiscuit:
@WalshyDev :eyes:

TheBusyBiscuit (approved):

WalshyDev (approved):
LGTM

PR #2715, opened 2021-01-13
Experimental: Slimefun but without CS-CoreLib

TheBusyBiscuit:
## Description

For ~3 years we have been slowly but surely getting rid of the old CS-CoreLib code (CSCL 1 that is).
Sure, a lot of addons may still use methods from CSCL 1 but I made sure to transfer the three main classes that are still in use.

Note here that CSCL 1 has basically remained untouched since 2018 and bStats shows that literally **noone** is running any outdated versions. So the classes should all match, in case someone still uses CSCL1 (though, our classes should take priority when loaded...), the class signatures should still be compatible.

This may need much more testing though before this is ready to be merged.
We may announce this as a public beta.

## Changes

* Removed CS-CoreLib dependency from the `pom.xml`
* Removed CS-CoreLib as a `softDepend` from the `plugin.yml`
* Removed check for CS-CoreLib on startup
* Transferred any "still-in-use" classes from CS-CoreLib to the Slimefun src

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -302,12 +302,6 @@
             <version>${spigot.version}-R0.1-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
-        <dependency>
-            <groupId>com.github.TheBusyBiscuit</groupId>
-            <artifactId>CS-CoreLib</artifactId>
-            <version>1.7</version>
-            <scope>provided</scope>
-        </dependency>
 
         <!-- Development dependencies -->
         <dependency>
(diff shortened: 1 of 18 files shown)


TheBusyBiscuit:
Note that the old CS-CoreLib code is not up for review here, it purely exists for legacy purposes and has already been replaced by newer rewritten versions anyway.

There's no point in refactoring it, it will get removed anyway.
But this will make the transition easier.

WalshyDev (approved):
Thank god it's finally going, only a little bit to go before the world can be happier.

LGTM

svr333 (approved):
*Corona virus enters*
Walshy: 
Still Walshy:
*Biscuit removes cs corelib dependency*
Walshy: WOOOOHOOOOO Finally gone! The world is almost a better place now

Also looks good by the way

TheBusyBiscuit:
Side note: Don't merge this yet. I want to release a public beta build before merging.

WalshyDev:
OH SHITTTTTTTTTTTTTTTT

PR #2717, opened 2021-01-13
Enable batch mode and disable transfer progress for Maven

WalshyDev:
## Description
Enable batch mode (If any prompts come up just use default value) and disable transfer progress (since this spams lots of lines). This will make the log cleaner and prevent any issue where a stall could happen.

Will slightly increase build time but mainly just shorten the log.

## Changes
* Change maven command to `-B --no-transfer-progress`

## Related Issues
N/A

## Checklist

### DIFF
--- a/.github/workflows/maven-compiler.yml
+++ b/.github/workflows/maven-compiler.yml
@@ -28,4 +28,4 @@ jobs:
       with:
         java-version: 1.8
     - name: Build with Maven
-      run: mvn package --file pom.xml
+      run: mvn package -B --no-transfer-progress
--- a/.github/workflows/maven-compiler.yml
+++ b/.github/workflows/maven-compiler.yml
@@ -13,6 +13,7 @@ on:
     paths:
     - 'src/**'
     - 'pom.xml'
+    - '.github/workflows/**'
 
 jobs:
   build:


TheBusyBiscuit (changes requested):
Could you add this workflow to the list of paths that trigger it?
Then we could see the results immediately.

WalshyDev:
Hmm. I thought that flag would just remove the x/x bytes
Rather than also removing "downloading" and "downloaded" messages. I feel it's not verbose enough now
:Thonk:

.github/workflows/maven-compiler.yml
```diff
@@ -13,6 +13,7 @@ on:
     paths:
     - 'src/**'
     - 'pom.xml'
+    - '.github/workflows/**'
```
TheBusyBiscuit, line 16: I mean... I think that only this particular action is necessary here, not all actions...

.github/workflows/maven-compiler.yml
```diff
@@ -28,4 +29,4 @@ jobs:
       with:
         java-version: 1.8
     - name: Build with Maven
-      run: mvn package --file pom.xml
+      run: mvn package -B --no-transfer-progress
```
TheBusyBiscuit, line 32: So is this a desirable change now or not?

PR #2724, opened 2021-01-15
Massive performance improvements and fixes to holograms

TheBusyBiscuit:
## Description

This pull request completely redesigns the way holograms work.
It adds a HologramsService to Slimefun which holds a cache where we can store the corresponding ArmorStand UUID for our Block.
We also cache the nametag now, as the operation `Entity#setCustomName(String)` is VERY expensive as it uses a regular expression internally. Lots of holograms will mostly remain constant, like Cargo's "Status: Online", the Hologram Projector or the GEO Miner. Only the Energy Regulator or Reactors may vary depending on the fluctuations of energy/cooling. So this will be an overall good optimization.

By caching the UUID, we effectively also solve any duplications that might occur.
And I also set persistent data on the entity to make sure that no holograms which are too close to each other will conflict.

I ran a few Spark samples to back this up.
Both samples were taken with a duration of **500s**.

Before: `CraftEntity#setCustomName(String)` took **1,796ms** and `SimpleHologram` was at the very top of the task list.
d7730ebb65343414faf1510d8e0246ce

After: The entire `HologramsService` took **16ms** to update all holograms continously for 500s.
6b907ce0d607bf0c390f823b1cdb316e

This was my testing rig:
a49b56143aec7c60180d0fb816c3112e

## Changes

* Added HologramsService
* Added HologramOwner
* Implemented caching for ArmorStands (UUIDs, don't worry) to reduce lookup time
* Implemented efficient caching for custom names

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/Hologram.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import org.bukkit.Bukkit;
+import org.bukkit.entity.ArmorStand;
+import org.bukkit.entity.Entity;
+
+import com.google.common.base.Objects;
```
WalshyDev, line 13: The one in Java does the same, this has been deprecated since J7

J3fftw1:
You should make all armor stands small

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
```diff
@@ -0,0 +1,296 @@
...
+     * This removes the {@link Hologram} at that given {@link Location}.
+     * <p>
+     * <strong>This method must be executed on the main {@link Server} {@link Thread}.</strong>
+     * 
+     * @param loc
+     *            The {@link Location}
+     * 
+     * @return Whether the {@link Hologram} could be removed, false if the {@link Hologram} does not exist or was
+     *         already
+     *         removed
```
WalshyDev, line 257: Formatting messed up a bit

TheBusyBiscuit:
> You should make all armor stands small

That would be more than pointless...
Besides, they are markers so they already have a hitbox of 0x0x0 and don't render as they are invisible too.

J3fftw1:
> > You should make all armor stands small
> 
> That would be more than pointless...
> Besides, they are markers so they already have a hitbox of 0x0x0 and don't render as they are invisible too.

Not completely pointless they are a pain in the ass to see when flying in spectator mode they are so huge they obstructing your view when someone has a big base of with a lot of holograms

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/HologramOwner.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * This returns the offset of the hologram as a {@link Vector}.
+     * This offset is applied to {@link Block#getLocation()} when spawning
+     * the hologram.
+     * 
+     * @return The hologram offset
+     */
+    @Nonnull
+    default Vector getHologramOffset() {
+        return new Vector(0.5, 0.75, 0.5);
```
WalshyDev, line 59: You can do a static final variable in interfaces since Java 8. Would save re-creating this every time. There isn't a single Slimefun server running Java 7 so I call this safe.

TheBusyBiscuit:
If that bothers you, feel free to make a pull request in the future.

WalshyDev (changes requested):
I like this, also like the fact there's a read access expiration. I wish Caffeine was in Spigot then this could be done with that. Either way, this is a much nicer approach than before 👍

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/Hologram.java
TheBusyBiscuit: whoops, wrong import

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/HologramOwner.java
TheBusyBiscuit: The point of this was to make it configurable for the interface implementer. But I guess a static DEFAULT_OFFSET variable would do.

WalshyDev (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
```diff
@@ -0,0 +1,311 @@
...
+
+        Runnable runnable = () -> {
+            try {
+                Hologram hologram = getHologram(loc, true);
+
+                if (hologram != null) {
+                    consumer.accept(hologram);
+                }
+            } catch (Exception | LinkageError x) {
+                SlimefunPlugin.logger().log(Level.SEVERE, "Something went wrong while trying to update a hologram", x);
```
Sfiguz7, line 252: Should we add more info here? Like the location where this happened. I am not sure whether the error already has it and if it is even useful at all but if it's nowhere more information is never bad

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
```diff
@@ -0,0 +1,311 @@
...
+
+                if (hologram != null) {
+                    cache.remove(new BlockPosition(loc));
+                    hologram.remove();
+                    return true;
+                } else {
+                    return false;
+                }
+            } catch (Exception | LinkageError x) {
+                SlimefunPlugin.logger().log(Level.SEVERE, "Something went wrong while trying to remove a hologram", x);
```
Sfiguz7, line 289: As before

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
TheBusyBiscuit: In theory, but the catch-block shouldn't ever be reached really, it's just an in-case-of-emergency scenario anyway.
And if it is, then a threading issue or a NPE is probably the most likeliest event.
The Location would not be needed in this case, as any exception would happen on a higher level anywhere.
But as you said, more context is never a bad thing.

TheBusyBiscuit: Added the location.

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
TheBusyBiscuit: Added the location.

Sfiguz7 (approved):
LGTM

WalshyDev (approved):
Still LGTM

PR #2729, opened 2021-01-16
Fix the infused hopper picking up items with pickup delay Short.MAX_VALUE

WalshyDev:
## Description
The Infused Hopper would think that an item with pickup delay of 32,767 is valid. Items with a delay of 32,767 are marked as non-pickupable by the MC client and server. We should treat them the same, there are many plugins which use this in order to say their items shouldn't be picked up.

MC Wiki Ref: https://minecraft.gamepedia.com/Item_(entity)#Entity_data

## Changes
* Added a pickup delay check to InfusedHopper#isValidItem
* Fixed a typo in the class Javadoc

## Related Issues
Skizzles on Discord: https://discord.com/channels/565557184348422174/565569196218384385/799931483635580959

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/InfusedHopper.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/InfusedHopper.java
@@ -28,13 +28,11 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 
 /**
  * The {@link InfusedHopper} is a special kind of {@link Hopper} which teleports any
- * neaby {@link Item} to itself.
+ * nearby {@link Item} to itself.
  * The radius can be configured in the config.
- * 
- * @author TheBusyBiscuit
- * 
- * @see InfusedMagnet
  *
+ * @author TheBusyBiscuit
+ * @see InfusedMagnet
  */
 public class InfusedHopper extends SimpleSlimefunItem<BlockTicker> {
 
@@ -65,7 +63,7 @@ public class InfusedHopper extends SimpleSlimefunItem<BlockTicker> {
                 if (toggleable.getValue()) {
                     Hopper hopper = (Hopper) b.getBlockData();
 
-                    /**
+                    /*
                      * If the Hopper was disabled by a redstone signal,
                      * we just don't do anything.
                      */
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/InfusedHopper.java
```diff
@@ -28,13 +28,11 @@
 
 /**
  * The {@link InfusedHopper} is a special kind of {@link Hopper} which teleports any
- * neaby {@link Item} to itself.
+ * nearby {@link Item} to itself.
  * The radius can be configured in the config.
- * 
+ *
  * @author TheBusyBiscuit
```
TheBusyBiscuit, line 34:
```suggestion
 * @author TheBusyBiscuit
 * @author Walshy
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/InfusedHopper.java
```diff
@@ -104,7 +102,10 @@ public boolean isSynchronized() {
     private boolean isValidItem(@Nonnull Location l, @Nonnull Entity entity) {
         if (entity instanceof Item && entity.isValid()) {
             Item item = (Item) entity;
-            return !SlimefunUtils.hasNoPickupFlag(item) && item.getLocation().distanceSquared(l) > 0.25;
+            // Check if the item cannot be picked up or has the "no pickup" metadata
+            return item.getPickupDelay() != Short.MAX_VALUE
```
TheBusyBiscuit, line 106:
```suggestion
            return item.getPickupDelay() <= 0
```

getPickupDelay() will count down, so we should respect the pickup delay fully here tbh.

TheBusyBiscuit (changes requested):
Did you test this?
Also, why was there no issue for it?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/InfusedHopper.java
WalshyDev: No, pickup delay does not count down if it's at MAX_VALUE. This is exactly why it can never be picked up

I can change this though if you wish

TheBusyBiscuit: I know, but for any other value it will.
Tbh I thought we checked for the pickup delay anyway (we do this with the Infused Magnet too)

TheBusyBiscuit (approved):
👍🏻

Sfiguz7 (approved):

PR #2733, opened 2021-01-16
Added script length checking

Seggan:
## Description
I added a script length check right before opening the edit script page. The reason for this is that in an addon I'm making, I can set much longer scripts than 54 instructions.

## Changes
Added a script length check

## Related Issues
N/A

## Checklist

- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos. (I don't think this needs testing)

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
@@ -474,7 +474,7 @@ public class ProgrammableAndroid extends SlimefunItem implements InventoryBlock,
                 openScript(pl, b, getScript(b.getLocation()));
             } else {
                 pl.closeInventory();
-                pl.sendMessage("&cThe script is too large to edit!");
+                pl.sendMessage(ChatColor.RED + "The script is too large to edit!");
             }
             return false;
         });
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
@@ -556,6 +556,7 @@ public class ProgrammableAndroid extends SlimefunItem implements InventoryBlock,
         Validate.notNull(script, "No script given");
         Validate.isTrue(script.startsWith(Instruction.START.name() + '-'), "A script must begin with a 'START' token.");
         Validate.isTrue(script.endsWith('-' + Instruction.REPEAT.name()), "A script must end with a 'REPEAT' token.");
+        Validate.isTrue(PatternUtils.DASH.split(script).length <= MAX_SCRIPT_LENGTH, "Scripts may not have more than " + MAX_SCRIPT_LENGTH + " segments");
 
         BlockStorage.addBlockInfo(l, "script", script);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -470,7 +470,12 @@ public void openScriptEditor(Player p, Block b) {
 
         menu.addItem(1, new CustomItem(HeadTexture.SCRIPT_FORWARD.getAsItemStack(), "&2> Edit Script", "", "&aEdits your current Script"));
         menu.addMenuClickHandler(1, (pl, slot, item, action) -> {
-            openScript(pl, b, getScript(b.getLocation()));
+            if (PatternUtils.DASH.split(BlockStorage.getLocationInfo(b.getLocation()).getString("script")).length <= MAX_SCRIPT_LENGTH) {
+                openScript(pl, b, getScript(b.getLocation()));
+            } else {
+                pl.closeInventory();
+                pl.sendMessage(ChatColor.RED + "The script is too large to edit!");
```
TheBusyBiscuit, line 477: Message isn't localized

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -470,7 +470,12 @@ public void openScriptEditor(Player p, Block b) {
 
         menu.addItem(1, new CustomItem(HeadTexture.SCRIPT_FORWARD.getAsItemStack(), "&2> Edit Script", "", "&aEdits your current Script"));
         menu.addMenuClickHandler(1, (pl, slot, item, action) -> {
-            openScript(pl, b, getScript(b.getLocation()));
+            if (PatternUtils.DASH.split(BlockStorage.getLocationInfo(b.getLocation()).getString("script")).length <= MAX_SCRIPT_LENGTH) {
+                openScript(pl, b, getScript(b.getLocation()));
+            } else {
```
TheBusyBiscuit, line 475: This check is already performed in #setScript(...).
So no real need for that.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
Seggan: That's the entire reason I made the pr. I have the need for scripts of 100 instructions or more. setscript won't allow me to do that but I can use blockstorage to circumvent that. However if a player edits a script of such a big size they will get an inventory error. That's why I added the second check

TheBusyBiscuit: :NotLikeThis:

Seggan: So...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -551,7 +556,6 @@ public void setScript(@Nonnull Location l, @Nonnull String script) {
         Validate.notNull(script, "No script given");
         Validate.isTrue(script.startsWith(Instruction.START.name() + '-'), "A script must begin with a 'START' token.");
         Validate.isTrue(script.endsWith('-' + Instruction.REPEAT.name()), "A script must end with a 'REPEAT' token.");
-        Validate.isTrue(PatternUtils.DASH.split(script).length <= MAX_SCRIPT_LENGTH, "Scripts may not have more than " + MAX_SCRIPT_LENGTH + " segments");
```
TheBusyBiscuit, line 554: Dear lord. Do not remove this :NotLikeThis:

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
The reason as to *why* this was added, I don't agree so much with.
But error-prevention and input validation is always good to have, so... approved

WalshyDev (approved):

PR #2749, opened 2021-01-21
Make RechargeableHelper public

WalshyDev:
## Description
Rechargeable takes an ItemStack, does some validation, gets the ItemMeta and passes it over to RechargeableHelper. Making RechargeableHelper public means that if people already have the ItemMeta they can save some execution time by using the helper themselves.

## Changes
* Move RechargeableHelper to the utils package
* Make RechargeableHelper class and methods public
* Added persistent data if it didn't exist and it had to fetch from the lore
* Added a unit test for the RechargeableHelper class

## Related Issues
My sanity

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
@@ -39,7 +39,7 @@ public final class ChargeUtils {
     public static void setCharge(@Nonnull ItemMeta meta, float charge, float capacity) {
         Validate.notNull(meta, "Meta cannot be null!");
         Validate.isTrue(charge >= 0, "Charge has to be equal to or greater than 0!");
-        Validate.isTrue(capacity >= 0, "Capacity has to be equal to or greater than 0!");
+        Validate.isTrue(capacity > 0, "Capacity has to be greater than 0!");
         Validate.isTrue(charge <= capacity, "Charge may not be bigger than the capacity!");
 
         BigDecimal decimal = BigDecimal.valueOf(charge).setScale(2, RoundingMode.HALF_UP);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
@@ -12,6 +12,7 @@ import io.github.thebusybiscuit.slimefun4.core.attributes.Rechargeable;
 import org.apache.commons.lang.Validate;
 import org.bukkit.NamespacedKey;
 import org.bukkit.inventory.meta.ItemMeta;
+import org.bukkit.persistence.PersistentDataContainer;
 import org.bukkit.persistence.PersistentDataType;
 
 import io.github.thebusybiscuit.cscorelib2.chat.ChatColors;
@@ -28,17 +29,17 @@ import net.md_5.bungee.api.ChatColor;
  * @see Rechargeable
  *
  */
-public final class RechargeableHelper {
+public final class ChargeUtils {
 
     private static final String LORE_PREFIX = ChatColors.color("&8\u21E8 &e\u26A1 &7");
     private static final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "[0-9.]+ / [0-9.]+ J");
 
-    private RechargeableHelper() {}
+    private ChargeUtils() {}
 
     public static void setCharge(@Nonnull ItemMeta meta, float charge, float capacity) {
         Validate.notNull(meta, "Meta cannot be null!");
         Validate.isTrue(charge >= 0, "Charge has to be equal to or greater than 0!");
-        Validate.isTrue(charge <= capacity, "Charge must be less than the capacity!");
+        Validate.isTrue(charge <= capacity, "Charge may not be bigger than the capacity!");
 
         BigDecimal decimal = BigDecimal.valueOf(charge).setScale(2, RoundingMode.HALF_UP);
         float value = decimal.floatValue();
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/RechargeableHelper.java
```diff
@@ -8,32 +8,38 @@
...
+ *
  * @author TheBusyBiscuit
- * 
+ * @author WalshyDev
+ *
  * @see Rechargeable
  *
  */
-final class RechargeableHelper {
+public final class RechargeableHelper {
```
TheBusyBiscuit, line 31: We could rename this from RechargeableHelper to `ChargeUtils` perhaps? Or have a better suggestion?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/RechargeableHelper.java
```diff
@@ -8,32 +8,38 @@
...
     private static final String LORE_PREFIX = ChatColors.color("&8\u21E8 &e\u26A1 &7");
     private static final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "[0-9.]+ / [0-9.]+ J");
 
     private RechargeableHelper() {}
 
-    static void setCharge(@Nonnull ItemMeta meta, float charge, float capacity) {
+    public static void setCharge(@Nonnull ItemMeta meta, float charge, float capacity) {
+        Validate.notNull(meta, "Meta cannot be null!");
+        Validate.isTrue(charge >= 0, "Charge has to be equal to or greater than 0!");
+        Validate.isTrue(charge <= capacity, "Charge must be less than the capacity!");
```
TheBusyBiscuit, line 41: * or equal to
Better just say "may not be bigger than the capacity"

src/main/java/io/github/thebusybiscuit/slimefun4/utils/RechargeableHelper.java
```diff
@@ -69,12 +77,14 @@ static float getCharge(@Nonnull ItemMeta meta) {
             for (String line : meta.getLore()) {
                 if (REGEX.matcher(line).matches()) {
                     String data = ChatColor.stripColor(PatternUtils.SLASH_SEPARATOR.split(line)[0].replace(LORE_PREFIX, ""));
-                    return Float.parseFloat(data);
+
+                    float loreValue = Float.parseFloat(data);
+                    meta.getPersistentDataContainer().set(key, PersistentDataType.FLOAT, loreValue);
```
TheBusyBiscuit, line 82: `meta.getPersistentDataContainer()` is called twice here, you could have it as a variable further up.

src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/utils/TestRechargeableHelper.java
```diff
@@ -0,0 +1,81 @@
...
+        RechargeableHelper.setCharge(meta, 1, 10);
+        Assertions.assertTrue(meta.hasLore());
+        Assertions.assertEquals(1, meta.getLore().size());
+
+        // Make sure the lore is correct
+        RechargeableHelper.setCharge(meta, 10.1f, 100.5f);
+        Assertions.assertEquals("&8\u21E8 &e\u26A1 &710.1 / 100.5 J".replace('&', ChatColor.COLOR_CHAR), meta.getLore().get(0));
+
+        // Make sure the persistent data was set
+        Assertions.assertEquals(10.1, PersistentDataAPI.getFloat(meta, SlimefunPlugin.getRegistry().getItemChargeDataKey()), 0.001);
```
TheBusyBiscuit, line 47: SlimefunPlugin is not loaded in the before() method, so this will return null.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
```diff
@@ -8,32 +8,39 @@
...
 
     private static final String LORE_PREFIX = ChatColors.color("&8\u21E8 &e\u26A1 &7");
     private static final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "[0-9.]+ / [0-9.]+ J");
 
-    private RechargeableHelper() {}
+    private ChargeUtils() {}
+
+    public static void setCharge(@Nonnull ItemMeta meta, float charge, float capacity) {
+        Validate.notNull(meta, "Meta cannot be null!");
+        Validate.isTrue(charge >= 0, "Charge has to be equal to or greater than 0!");
```
TheBusyBiscuit, line 41: I guess, capacity could also be checked to be > 0

TheBusyBiscuit (commented):
Looks good, one more validate check and perhaps javadocs

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
WalshyDev: true

Sfiguz7 (dismissed):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
```diff
@@ -8,32 +8,40 @@
...
+        Validate.isTrue(capacity >= 0, "Capacity has to be equal to or greater than 0!");
```
TheBusyBiscuit, line 42:
```suggestion
        Validate.isTrue(capacity > 0, "Capacity has to be greater than 0!");
```
A capacity of zero would not make any sense.
Then it would not be able to hold any charge.

TheBusyBiscuit (changes requested):
One small mistake (and I'd still love javadocs but I'm not gonna reject it only because of that)

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ChargeUtils.java
WalshyDev: *facepalm* I'm smart

TheBusyBiscuit (approved):
Javadocs would have been the cherry ontop but LGTM

Sfiguz7 (approved):
Cookie REALLY wants JavaDocs, poor guy you won't give him any :( LGTM though

PR #2755, opened 2021-01-22
Check inv maxStackSize

kii-chan-reloaded:
## Description
As mentioned in CS-CoreLib2's numbers 141 and 142 PR's, previous iterations of Slimefun AContainers would not consider the underlying Inventory's max stack size when determining sizing or pushing items. This commit changes DirtyChestMenus to act similarly by storing a variable defined as the lower of the Inventory's max stack size and the ItemStack's max stack size in a slot.

## Changes
* Changed `else if` statement in line 108 to an `else`
* Defined variable `maxStackSize`
* Resumed original conditional statement
* Changed all references of `stack.getMaxStackSize()` to `maxStackSize`

## Related Issues
N/A

## Checklist


TheBusyBiscuit (approved):
Would love to see this tested first but looks good as is.

WalshyDev (approved):
LGTM

PR #2774, opened 2021-01-26
Add a way to retrieve the raw storage

WalshyDev:
## Description

Provide a way to access BlockStorage's `storage`. This adds a way to access all the blocks loaded in the server. It may be worth making this return an immutable map so that people don't modify this. Let me know your opinion.

This is currently untested (though if it compiles, can't see why it wouldn't work) and I made this on my phone sooo compilation may fail. Not added tests or docs either due to phone (also this class really needs to be redone anyway)

## Changes

* Added BlockStorage#getStorage
* Added BlockStorage.getStorage(World)

## Related Issues

N/A

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
@@ -387,7 +387,7 @@ public class BlockStorage {
 
         BlockStorage storage = getStorage(world);
         if (storage != null) {
-            return (storage.getRawStorage();
+            return storage.getRawStorage();
         } else {
             return null;
         }
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
@@ -27,6 +27,8 @@ import org.bukkit.configuration.file.YamlConfiguration;
 import org.bukkit.entity.HumanEntity;
 import org.bukkit.inventory.ItemStack;
 
+import org.apache.commons.lang.Validate;
+
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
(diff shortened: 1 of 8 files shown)


Sfiguz7 (dismissed):
No tests and phone coding >:( But LGTM

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
```diff
@@ -360,6 +362,23 @@ public static void saveChunks() {
         }
     }
 
+    @Nonnull
+    public Map<Location, Config> getRawStorage() {
+        return this.storage;
```
TheBusyBiscuit, line 367: Like you presumed, I'd recommend making this immutable.

TheBusyBiscuit (changes requested):
I don't really see much of a point in this, especially since we will be tearing down this class at some point anyway but lgtm, just... immutable pls

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
```diff
@@ -360,6 +363,36 @@ public static void saveChunks() {
...
+     * @param world The world of which to fetch the data from.
+     * @return An {@link ImmutableMap} of the raw data or null if the world isn't registered.
+     */
+    @Nullable
+    public static Map<Location, Config> getRawStorage(@Nonnull World world) {
+        Validate.notNull(world, "World cannot be null!");
+
+        BlockStorage storage = getStorage(world);
+        if (storage != null) {
+            return (storage.getRawStorage();
```
Sfiguz7:
```suggestion
            return storage.getRawStorage();
```

Sfiguz7 (approved):

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
Sfiguz7: Didn't mean to approve since I suggested the above, eh

PR #2782, opened 2021-01-29
Allow Talismans to use the Actionbar

TheBusyBiscuit:
## Description

Long-requested feature; Talismans can now send their messages using the actionbar.
This is configurable but reduces spam :)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 * Added Talisman of the Wise
 * Added Book Binder
 * Added Tier 3 Electric Ore Grinder
+* Added an option to allow Talismans to send their notifications via the Actionbar
 
 #### Changes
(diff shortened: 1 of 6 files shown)


svr333 (approved):
Looks good, Im assuming the talismans and soulbound items arent toggleable anymore since its not as laggy?

TheBusyBiscuit:
> Looks good, Im assuming the talismans and soulbound items arent toggleable anymore since its not as laggy?

I doubt anyone really disabled these in the first place.
Besides.. that only disables the listeners, the items would still exist but be essentially useless.
So if we add an option for it, then perhaps it should be much better.
Or we just check whether the corresponding Talisman is enabled in each EventHandler or talisman check instead.

WalshyDev (approved):
LGTM

PR #2792, opened 2021-02-01
WIP - BlockStorage Rewrite

TheBusyBiscuit:
## Description

This is a milestone pull request (still work in progress though).
It proposes new thread-safe system to handle block data in a much safer and more efficient manner.

**Important**: This is a very huge milestone and a big change. We could use all the help we can get here. Both by testers and contributors. Lots of this is very hard and time-consuming to implement. So feel free to make your own pull request towards the `api/blockstorage-rewrite` branch to contribute to this draft PR.


J3fftw1:
A quick question might be completely out kf the scope of this pr/draft but will this mean db support for like sql or maria or what ever?

WalshyDev:
> A quick question might be completely out kf the scope of this pr/draft but will this mean db support for like sql or maria or what ever?

It's a step in that direction. This PR will not be adding that though. This is a general focus on just rewriting and abstracting the data. Then binary data, MySQL, etc can be added as new storage methods.

J3fftw1:
> > A quick question might be completely out kf the scope of this pr/draft but will this mean db support for like sql or maria or what ever?
> 
> It's a step in that direction. This PR will not be adding that though. This is a general focus on just rewriting and abstracting the data. Then binary data, MySQL, etc can be added as new storage methods.

This will make it possible right?

TheBusyBiscuit:
Well this system abstracts away from everything, so new storage methods would be very easy to add.

src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/AbstractDataObject.java
```diff
@@ -0,0 +1,138 @@
...
+            lock.writeLock().lock();
+
+            try {
+                data.put(key, (String) value);
+                markDirty(true);
+            } finally {
+                lock.readLock().unlock();
+            }
+        } else {
+            throw new UnsupportedOperationException("Can't set \"" + key + "\" to \"" + value + "\" (type: " + value.getClass().getSimpleName() + ") because SlimefunBlockData only supports Strings");
```
Mooy1, line 89: Just wondering, why not just accept type String instead of Object?

Seggan:
what is left to be done?

TheBusyBiscuit:
> what is left to be done?

Pretty much everything.
This is still WIP and right now this is just a shallow API skeleton. There is no implementation for anything at the moment, so it's only like 10% done.

JasperChaseTOQ:
> > what is left to be done?
> 
> Pretty much everything. This is still WIP and right now this is just a shallow API skeleton. There is no implementation for anything at the moment, so it's only like 10% done.

so 90% left is there anything i or anyone else can find a list of whats left to do or what can be contributed to it if were interested in helping out?

J3fftw1:
Closing as stale, keeping the branch alive. might get superceded because of other rewrites.

PR #2816, opened 2021-02-09
Fix broken enchant plugins stalling plugin start

WalshyDev:
## Description
There have been a few reports of servers locking up after `loadItems()` is called. Today I managed to get one of the person's servers and reproduce the issue. Thanks to `Cahsew#8395` for providing his server.

Custom enchantments... *sigh* there was a plugin (Enchantments+-) which was adding enchants which were: 1. Prefixed with `minecraft:` in their namespace... and 2: Above the MC max enchant value

This plugin was basically creating about 12 enchants all from what I can tell using Integer.MAX_VALUE as the max level. This means creating 14,400,000,000 item settings.... yeah, this caused issues...
So, this removes custom enchants in general showing up (which is good) and also fixes really stupid ones using an above Short.MAX_VALUE max level.

## Changes
* Fix broken custom enchant plugins freezing up the server.

## Related Issues
N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
@@ -6,6 +6,7 @@ import io.github.thebusybiscuit.slimefun4.implementation.settings.TalismanEnchan
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 import org.apache.commons.lang.Validate;
 import org.bukkit.Material;
+import org.bukkit.NamespacedKey;
 import org.bukkit.enchantments.Enchantment;
 import org.bukkit.entity.Player;
 import org.bukkit.inventory.ItemStack;
@@ -40,12 +41,17 @@ public class MagicianTalisman extends Talisman {
         addItemSetting(allowEnchantmentBooks);
 
         for (Enchantment enchantment : Enchantment.values()) {
-            try {
-                for (int i = 1; i <= enchantment.getMaxLevel(); i++) {
-                    enchantments.add(new TalismanEnchantment(enchantment, i));
+            // Stop custom enchant plugins appearing
+            // We also want to stop some stupid plugins which register as Minecraft enchants
+            // but also go above the max enchant level.
+            if (enchantment.getKey().getNamespace().equals(NamespacedKey.MINECRAFT) && enchantment.getMaxLevel() < Short.MAX_VALUE) {
+                try {
+                    for (int i = 1; i <= enchantment.getMaxLevel(); i++) {
+                        enchantments.add(new TalismanEnchantment(enchantment, i));
+                    }
+                } catch (Exception x) {
+                    SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "The following Exception occurred while trying to register the following Enchantment: " + enchantment);
                 }
-            } catch (Exception x) {
-                SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "The following Exception occurred while trying to register the following Enchantment: " + enchantment);
             }
         }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
@@ -124,7 +124,7 @@ public class MagicianTalisman extends Talisman {
         if (enchantment.getMaxLevel() > Short.MAX_VALUE) {
             // Enchantment levels are shorts, even if Bukkit allows integers.
             return false;
-        } else if (!SlimefunPlugin.getCfg().getBoolean("allow-custom-enchants") {
+        } else if (!SlimefunPlugin.getCfg().getBoolean("allow-custom-enchants")) {
             // Only allow Enchantments from the "minecraft:" namespace.
             return enchantment.getKey().getNamespace().equals(NamespacedKey.MINECRAFT);
         } else {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
@@ -41,9 +41,11 @@ public class MagicianTalisman extends Talisman {
         addItemSetting(allowEnchantmentBooks);
 
         for (Enchantment enchantment : Enchantment.values()) {
-            // Stop custom enchant plugins appearing
-            // We also want to stop some stupid plugins which register as Minecraft enchants
-            // but also go above the max enchant level.
+            /* 
+            * Stop custom enchant plugins appearing
+            * We also want to stop some stupid plugins which register as Minecraft enchants
+            * but also go above the max enchant level.
+            */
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
```diff
@@ -40,12 +41,17 @@ public MagicianTalisman(SlimefunItemStack item, ItemStack[] recipe) {
         addItemSetting(allowEnchantmentBooks);
 
         for (Enchantment enchantment : Enchantment.values()) {
-            try {
-                for (int i = 1; i <= enchantment.getMaxLevel(); i++) {
-                    enchantments.add(new TalismanEnchantment(enchantment, i));
+            // Stop custom enchant plugins appearing
+            // We also want to stop some stupid plugins which register as Minecraft enchants
+            // but also go above the max enchant level.
```
TheBusyBiscuit, line 46: multi-line comments :eyes:
```suggestion
            /* 
            * Stop custom enchant plugins appearing
            * We also want to stop some stupid plugins which register as Minecraft enchants
            * but also go above the max enchant level.
            */
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
```diff
@@ -40,12 +41,17 @@ public MagicianTalisman(SlimefunItemStack item, ItemStack[] recipe) {
...
+            if (enchantment.getKey().getNamespace().equals(NamespacedKey.MINECRAFT) && enchantment.getMaxLevel() < Short.MAX_VALUE) {
```
TheBusyBiscuit, line 47: I think you mixed this one up.
This will exclude any enchantment registering under their own namespace though.
So only `minecraft:**` enchantments are supported now, which doesn't fix the issue of faulty plugins registering theirs under the wrong namespace.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
WalshyDev: It's probably best to not support custom enchants either way honestly. They can be buggy but yeah, the second check fixes if they register under MC and messed up max level.

I can remove the first check and allow custom enchants I guess but it can still have issues.
For example, none of these custom enchants will actually show up on books or items. So, it can look like a book has no enchant when there's actually a custom one there.

I think for the sake of simplicity it's best to just make it only vanilla.

TheBusyBiscuit: Or perhaps a config option for it.
And then have it disabled by default and if someone asks about their favourite enchantment plugin, we can tell them to enable this setting but warn of the consequences.

WalshyDev: Yeah that can work. 

I hate custom enchant plugins though. Also hate Bukkit for having getMaxLevel as an int rather than short.

TheBusyBiscuit: same and same.
But then again... very typical thing for Bukkit :LUL:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
```diff
@@ -40,12 +42,20 @@ public MagicianTalisman(SlimefunItemStack item, ItemStack[] recipe) {
         addItemSetting(allowEnchantmentBooks);
```
TheBusyBiscuit, line 42: You forgot to add it to the item
```suggestion
        addItemSetting(allowEnchantmentBooks, allowCustomEnchantments);
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
```diff
@@ -30,22 +31,31 @@
...
-                    enchantments.add(new TalismanEnchantment(enchantment, i));
+            /* 
+             * Stop custom enchant plugins appearing
+             * We also want to stop some stupid plugins which register as Minecraft enchants
+             * but also go above the max enchant level.
+             */
+            if ((allowCustomEnchantments.getValue() || enchantment.getKey().getNamespace().equals(NamespacedKey.MINECRAFT))) {
+                try {
+                    // Make sure we cap this at max level or if set it incorrectly, use Short.MAX_VALUE
+                    for (int i = 1; i <= Math.max(enchantment.getMaxLevel(), Short.MAX_VALUE); i++) {
```
TheBusyBiscuit, line 53: You probably meant `min` here :monkaS:
```suggestion
                    for (int i = 1; i <= Math.min(enchantment.getMaxLevel(), Short.MAX_VALUE); i++) {
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
WalshyDev: *facepalm* I swear, I do know how to code

WalshyDev:
Hmm, ItemSetting isn't loaded till register but I use it in the constructor... I can hackily this by loading it right after adding.

Errr, not sure what to do here

TheBusyBiscuit:
Oh right...
Simply add it to the `config.yml` then?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/talismans/MagicianTalisman.java
```diff
@@ -31,23 +31,24 @@
...
+        addItemSetting(allowEnchantmentBooks);
 
         for (Enchantment enchantment : Enchantment.values()) {
             /* 
              * Stop custom enchant plugins appearing
              * We also want to stop some stupid plugins which register as Minecraft enchants
              * but also go above the max enchant level.
              */
-            if ((allowCustomEnchantments.getValue() || enchantment.getKey().getNamespace().equals(NamespacedKey.MINECRAFT))) {
+            if ((SlimefunPlugin.getCfg().getBoolean("allow-custom-enchants")
```
TheBusyBiscuit, line 49: Should probably cache it, like all other config settings.
`SlimefunRegistry` is currently used for this

TheBusyBiscuit (changes requested):

Sfiguz7 (approved):

TheBusyBiscuit (commented):
I refactored this a bit, so it becomes easier to extend upon or override in child classes.
However, I still think we should cache the config setting for performance reasons.

TheBusyBiscuit:
REEEEE why the build fail...

TheBusyBiscuit:
Oh great... Github actions servers are having problems again

TheBusyBiscuit:
Closing this as "stale" since there has been no response since over a month and requested changes were still left unaddressed.

PR #2831, opened 2021-02-20
Fixes #2677

### LINKED ISSUE #2677: Cargo Filters treat all potions as equal

Cargo filters do not know the difference between potions. Water bottle is the same for cargo as Mundane Potion, Potion of Healing, Potion of Regeneration, etc. the same problem for throwers and I assume that for lingering

1. Place the chest and fill it with several different potions of the same type (normal, throwing or lingering)
2. Connect the Cargo Node (Input) to the chest and enable the blacklist in it.
3. Place the second chest and connect the Advanced Cargo Node (Output) to it
4. Insert any potion of the selected type in the first chest into the whitelist filter.
Video: https://youtu.be/euOi2rSkB5M

The cargo filter will distinguish potions.
## :scroll: Server Log

N/A

N/A

image

Thehrz:
## Description
Fixes #2677 

## Changes
Added judgment on potion for SlimefunUtils#isItemSimilar()

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -296,7 +296,7 @@ public final class SlimefunUtils {
             return false;
         } else if (itemMeta instanceof PotionMeta && sfitemMeta instanceof PotionMeta) {
             return ((PotionMeta) itemMeta).getBasePotionData().equals(((PotionMeta) sfitemMeta).getBasePotionData());
-        }else if (!checkLore) {
+        } else if (!checkLore) {
             return true;
         } else if (itemMeta.hasLore() && sfitemMeta.hasLore()) {
             return equalsLore(itemMeta.getLore(), sfitemMeta.getLore());
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -295,7 +295,7 @@ public final class SlimefunUtils {
         } else if (itemMeta.hasDisplayName() && sfitemMeta.hasDisplayName() && !itemMeta.getDisplayName().equals(sfitemMeta.getDisplayName())) {
             return false;
         } else if (itemMeta instanceof PotionMeta && sfitemMeta instanceof PotionMeta) {
-            return ((PotionMeta) itemMeta).getBasePotionData().equals(((PotionMeta) sfitemMeta).getBasePotionData());
+            return ((PotionMeta) itemMeta).getBasePotionData().hashCode() == ((PotionMeta) sfitemMeta).getBasePotionData().hashCode();
         } else if (!checkLore) {
             return true;
         } else if (itemMeta.hasLore() && sfitemMeta.hasLore()) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -293,7 +294,9 @@ private static boolean equalsItemMeta(@Nonnull ItemMeta itemMeta, @Nonnull ItemM
             return false;
         } else if (itemMeta.hasDisplayName() && sfitemMeta.hasDisplayName() && !itemMeta.getDisplayName().equals(sfitemMeta.getDisplayName())) {
             return false;
-        } else if (!checkLore) {
+        } else if (itemMeta instanceof PotionMeta && sfitemMeta instanceof PotionMeta) {
+            return ((PotionMeta) itemMeta).getBasePotionData().equals(((PotionMeta) sfitemMeta).getBasePotionData());
+        }else if (!checkLore) {
```
WalshyDev, line 299: space missing
```suggestion
        } else if (!checkLore) {
```

WalshyDev (changes requested):
CraftMetaPotion does not override #equals (however they do have #hashCode...). This means it will fallback to CraftMetaItem's equals may result in it being incorrect. Not sure if their `return CraftItemFactory.instance().equals(this, (ItemMeta) object);` will do some weirdness to make it compare correctly.

Just something to note

svr333 (dismissed):
Looks good, keep walshys remark in mind though

Sfiguz7 (approved):
LGTM

LinoxGH (approved):

svr333 (approved):

TheBusyBiscuit (changes requested):
I am not actually requesting any changes yet.
I just wanna prevent this from being merged until I took a look at it since this modifies one of the most frequently used methods within Slimefun.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -293,6 +294,8 @@ private static boolean equalsItemMeta(@Nonnull ItemMeta itemMeta, @Nonnull ItemM
             return false;
         } else if (itemMeta.hasDisplayName() && sfitemMeta.hasDisplayName() && !itemMeta.getDisplayName().equals(sfitemMeta.getDisplayName())) {
             return false;
+        } else if (itemMeta instanceof PotionMeta && sfitemMeta instanceof PotionMeta) {
+            return ((PotionMeta) itemMeta).getBasePotionData().hashCode() == ((PotionMeta) sfitemMeta).getBasePotionData().hashCode();
```
TheBusyBiscuit, line 298: This overrides any lore checks, so if we have two potions with the same potion data but different lore, they will still pass as "equal", even when the `checkLore` parameter is set to true.

Furthermore `.hashCode()` should not really be used for comparisons like this, there is nothing wrong with using `.equals()` here instead.
Note that  `.hashCode()` and `.equals()` always co-exist with each other and are never overridden by themselves only.
Or as Yoda would put it:

> Always two there are, no more, no less.

TheBusyBiscuit (changes requested):

TheBusyBiscuit:
Closing this since there hasn't been any response in over three weeks.

Issue #2835, opened 2021-02-23
Default SF items showing power per second incorrectly


Jian-Wuyou:
## :round_pushpin: Description (REQUIRED)

Default SF items show power per second as though tick duration is 0.5s and doesn't scale into different tick durations, i.e. it's incorrect for any tick durations aside from 10 ticks.

## :bookmark_tabs: Steps to reproduce the Issue (REQUIRED)

1. Set tick duration to anything aside from 10 ticks

## :bulb: Expected behavior (REQUIRED)

If the machine produces 250 J per SF tick, then it should show 500 J/s if the tick duration was 0.5s, and 416.67 J/s if the tick duration was 0.6s

## :compass: Environment (REQUIRED)

 - Server Software: 
 - Minecraft Version: Spigot 1.16.4-R0.1-SNAPSHOT
 - Slimefun Version: 
![image](https://user-images.githubusercontent.com/11474057/108799323-fdfd1f80-75ca-11eb-843d-a59d7dc4195e.png)

J3fftw1:
Your litexpansion is outdated

PR #2836, opened 2021-02-23
Scaled powerPerSecond with tick duration

Jian-Wuyou:
## Description
Fixed incorrect power per second for tick durations that are not 0.5s

## Changes
Changed `powerPerSecond(int power)` to divide `power` with the tick duration (i.e. if power given to LoreBuilder is 240 and tick duration is 12 ticks, it shows 400 J/s as the final output). This will require that current LoreBuilder values be divided by 2 (i.e. if currently it says `LoreBuilder.powerPerSecond(500)`, it should be changed to `LoreBuilder.powerPerSecond(250)`).

## Related Issues

Resolves: #2835

## Checklist


WalshyDev:
How come you closed this?

Jian-Wuyou:
Just remembered that I still have to rewrite the current LoreBuilder.powerPerSecond calls so that it displays the correct values (i.e. replace the current values with the per-tick usage) otherwise it'll display two times the actual values

Jian-Wuyou:
Example of how it looks like with Lava Generator 2 (20 J/t, tick duration is 0.6s)
image

All energy values are rounded off to 2 decimal places.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
```diff
@@ -55,12 +57,12 @@ public static String powerBuffer(int power) {
 
     @Nonnull
     public static String powerPerSecond(int power) {
-        return power(power, "/s");
+        return power((float)power * 20f / SlimefunPlugin.getCfg().getInt("URID.custom-ticker-delay"), "/s");
```
Mooy1, line 60: should add a static field for 20 / configValue

TheBusyBiscuit:
I am sorry but I have to reject this.
Changing the lore of items will have a ton of unforeseen consequences.
Stacking problems will just be one of them.
This might be possible in the future, but at the moment these changes would cause more problems than they solve.

Your changes would for example result in **every** existing item before these changes no longer being usable in some crafting recipes. So this will likely spawn even more bugs and bug reports than for what it's worth.
Sorry for the effort you put into this, but for now you need to accept these semi-accurate times.

PR #2867, opened 2021-03-06
Minor improvements to new block storage classes

md5sha256:
## Description

Minor improvements and api additions to the new blockstorage classes.
None of the changes are "tested" since the branch is WIP but there shouldn't
be anything which is broken. No javadocs were added for similar reasons as most
if not all additions are to undocumented WIP api. 

## Changes

* Fixed unlocking the wrong lock in `AbstractDataObject`
* Add `AbstractDataObject#toJson(Gson)`
* Add `SlimefunDataObject#empty`
* Add `SlimefunWorldData#getBlockData(BlockPosition)`

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/EmptyBlockData.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/EmptyBlockData.java
@@ -12,7 +12,9 @@ import me.mrCookieSlime.Slimefun.api.BlockStorage;
  * @author TheBusyBiscuit
  *
  */
-class EmptyBlockData extends SlimefunBlockData {
+final class EmptyBlockData extends SlimefunBlockData {
+
+    static EmptyBlockData INSTANCE = new EmptyBlockData();
 
     EmptyBlockData() {
         super();
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/EmptyBlockData.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/EmptyBlockData.java
@@ -14,8 +14,6 @@ import me.mrCookieSlime.Slimefun.api.BlockStorage;
  */
 final class EmptyBlockData extends SlimefunBlockData {
 
-    static EmptyBlockData INSTANCE = new EmptyBlockData();
-
     EmptyBlockData() {
         super();
     }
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/EmptyBlockData.java
```diff
@@ -12,7 +12,9 @@
  * @author TheBusyBiscuit
  *
  */
-class EmptyBlockData extends SlimefunBlockData {
+final class EmptyBlockData extends SlimefunBlockData {
+
+    static EmptyBlockData INSTANCE = new EmptyBlockData();
```
TheBusyBiscuit, line 17: Having this static singleton here is unnecessary in my opinion.
We don't yet know whether we will even need this class.
And if we do, then the singleton should be a private field in the BlockDatabase rather than a static field to avoid unnecessary memory leaks when disabling the plugin.

src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/SlimefunBlockData.java
```diff
@@ -25,6 +25,11 @@
  */
 public class SlimefunBlockData extends AbstractDataObject {
 
+    @Nonnull
+    public static SlimefunBlockData empty() {
+        return EmptyBlockData.INSTANCE;
+    }
```
TheBusyBiscuit, line 31: Again, not entirely certain whether we will be needing this or if the class will be removed.
So let's rather remove it for now.

src/main/java/io/github/thebusybiscuit/slimefun4/api/blocks/SlimefunWorldData.java
```diff
@@ -169,6 +169,21 @@ public SlimefunBlockData getBlockData(int x, int y, int z) {
         }
     }
 
+    @Nullable
+    public SlimefunBlockData getBlockData(@Nonnull BlockPosition blockPosition) {
```
TheBusyBiscuit, line 173: Should also add a `Validate.notNull()` check here for the block position

TheBusyBiscuit (changes requested):
Thanks for your pull request, I requested some changes in regard to the Empty block data.

Also since I just never seem to remember when to add it and when not to do so: What exactly does the `synchronized` modifier do here?

md5sha256:
`synchronized` just means only one thread can access a given method at one time. I've placed it there because there really shouldn't be two threads marking the object as dirty at the same time.

TheBusyBiscuit:
Any other Thread accessing this method will wait for the first Thread to process the method then I assume?

md5sha256:
> Any other Thread accessing this method will wait for the first Thread to process the method then I assume?
Yes

TheBusyBiscuit (approved):
LGTM

PR #2885, opened 2021-03-15
Switch the last two items in Magical Items

svr333:
## Description
Switch the two last items in the Magical Category around so that all the runes appear next to each other.

Example of how it looks now (before the last two items were switched):
image

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
@@ -2569,13 +2569,13 @@ public final class SlimefunItemSetup {
         }
 
         if (minecraftVersion.isAtLeast(MinecraftVersion.MINECRAFT_1_16)) {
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
🙄

WalshyDev:
Now this is the PR we were all waiting for! Thank you for fixing this game-breaking bug 😆

svr333:
I know, game changing PR
but you know what they say, the devil is in the details :)

WalshyDev (approved):

TheBusyBiscuit:
I can understand that this may satisfy some people's OCD.
But something inside me somehow feels it is wrong for the rune to come first, then the item required to craft the rune.
But I don't really care that much about the order of items to be frank.

@WalshyDev Keep in mind that the next build will break a ton of addons because people did not seem to read through the announcement for them to update to RC-21 ASAP.
Then again, we gave them deprecation warnings and a notice upfront, so not really out fault in the end when things will break.

LinoxGH (approved):

TheBusyBiscuit (approved):

PR #2888, opened 2021-03-17
Allow onBlockBreak to be cancelled by BlockBreakHandler

NCBPFluffyBear:
## Description
Added a check for if the BlockBreakHandler was cancelled and returned accordingly. This prevents Slimefun block data from being cleared when the BlockBreakEvent gets cancelled within the handler.

## Changes
Modified BlockListener.java to return if the BlockBreakEvent is cancelled.

## Related Issues
N/A

## Checklist


TheBusyBiscuit (approved):
Makes sense.

The drops get cancelled anyway already, the block stays too.
So cancelling the data clearing as well is intended behaviour.

LinoxGH (approved):
As long as it is tested. LGTM

TheBusyBiscuit:
Still waiting for this to be tested then.

NCBPFluffyBear:
Tested with FluffyMachines, LiteXpansion, and InfinityExpansion. However, I think FluffyMachines is the only addon that currently cancels the event.

PR #2893, opened 2021-03-21
Show generating in debug fish

WalshyDev:
## Description
There's no real way to see how much some things are generating, some addons obviously have a GUI for this but for the most part there isn't one. This is a quick and easy way for staff to verify the generator is producing what is expected.

Now, there are a few points to mention:
* Obviously this wont always be the same value - this does often change depending on circumstances
  * For me, this makes sense and it's what I would want this to show. But, do we add any disclaimer/info about this?
* This is per SF tick, do we add a `/t` into this? Maybe could do a in brackets seconds one too. For example: `Generating: 2/t (3.6/s)`
  * If we did add a per second, should it be correct (so, use configured ticks) or not? For normal SF items I can see bug reports from this being correct. "The lore says 4/s but it's doing 3.6/s"

I believe this is a good, informational change however there due to the weirdness around this area it could cause problems. Is it a net positive or a net negative?

## Changes
Added a new message from the debug fish listener which shows how much it is generating

## Related Issues
N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/DebugFishListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/DebugFishListener.java
@@ -155,6 +155,12 @@ public class DebugFishListener implements Listener {
             } else {
                 p.sendMessage(ChatColors.color("&dChargeable: " + redCross));
             }
+
+            if (item instanceof EnergyNetProvider) {
+                EnergyNetProvider provider = (EnergyNetProvider) item;
+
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
That's a hard reject.
The method `getGeneratedOutput()` will consume fuel etc... and does not support querying.

J3fftw1:
Doesnt the multimeter already do this?

TheBusyBiscuit:
No it doesn't.
But adding that feature to the Multimeter was a Hacktoberfest issue at one point.

J3fftw1:
Ah okay, @WalshyDev you should look at that hacktoberfest issue 😘

WalshyDev:
getGeneratedOutput really should not be processing fuel NotLikeThis

PR #2897, opened 2021-03-21
Made MultiCategory able to hold any Category

Mooy1:
## Description
Made it so multi categories can hold any category, so that flex categories or even more nested multi categories could be added.

## Changes
Changed a few params from SubCategory -> Category

## Related Issues
x

## Checklist


TheBusyBiscuit:
That would break a lot of stuff and even allow for things as Infinite recursive loops.
Also would break with any FlexCategory which is not intended for such a purpose.

This really goes against the heart of what the MultiCategory does.

PR #2900, opened 2021-03-22
Fixed a Message

Seggan:
## Description
I removed the part mentioning cscl in the `/sf` usage message. This is probably the smallest pr I ever did.

## Changes
Removed a part mentioning CS-CoreLib from a message.

## Related Issues

N/A

## Checklist

### DIFF
--- a/src/main/resources/plugin.yml
+++ b/src/main/resources/plugin.yml
@@ -28,7 +28,7 @@ commands:
   slimefun:
     aliases: sf
     description: Slimefun base command
-    usage: You either forgot to install CS-CoreLib or you installed an unsupported version.
+    usage: You probably installed an unsupported version.
 
 # Default permissions
 permissions:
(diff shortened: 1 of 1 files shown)


src/main/resources/plugin.yml
```diff
@@ -28,7 +28,7 @@ commands:
   slimefun:
     aliases: sf
     description: Slimefun base command
-    usage: You either forgot to install CS-CoreLib or you installed an unsupported version.
+    usage: You probably installed an unsupported version.
```
TheBusyBiscuit, line 31:
```suggestion
    usage: The Slimefun version you have installed does not seem to be compatible with your server setup. Check your console for more detailed info.
```

TheBusyBiscuit (changes requested):
I think that message itself is very much misleading. If we change it, we might as well just do it better.

TheBusyBiscuit (approved):

svr333 (approved):

PR #2909, opened 2021-03-26
Fixes #2903

### LINKED ISSUE #2903: Crucibles when next to each other don't make lava

When you activate multiple crucibles beside each other they try to make lava but it cancels out and you get nothing

1. Get 2 or more crucibles
2. Place them directly beside each other
3. Activate them with cobblestone to begin the lava making process
https://youtu.be/JoOVaRbo5lw

Continue to make lava until it's finished, then the lava remains there flowing out if there's nothing to contain it.

## :scroll: Server Log

N/A

N/A

 - Server Software: Purpur
 - Minecraft Version: 1.16.5
 - Slimefun Version: DEV 850

RobotHanzo:
## Description
Fixed the bug in #2903 where crucibles next to each other doesnt work due to the output updating each other causing the liquid to be updated.

## Changes
Made applyPhysics false when running block.setType

## Related Issues
Fixes #2903 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/Crucible.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/Crucible.java
@@ -188,6 +188,7 @@ public class Crucible extends SimpleSlimefunItem<BlockUseHandler> implements Rec
 
     private void placeLiquid(@Nonnull Block block, boolean water) {
         if (block.getType() == Material.AIR || block.getType() == Material.CAVE_AIR || block.getType() == Material.VOID_AIR) {
+            // Fixes #2903 - Cancel physics update to resolve weird overlapping
             block.setType(water ? Material.WATER : Material.LAVA, false);
         } else {
             if (water && block.getBlockData() instanceof Waterlogged) {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/Crucible.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/Crucible.java
@@ -14,6 +14,7 @@ import org.bukkit.Tag;
 import org.bukkit.World.Environment;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.Levelled;
 import org.bukkit.block.data.Waterlogged;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/Crucible.java
```diff
@@ -14,6 +14,7 @@
 import org.bukkit.World.Environment;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.block.data.BlockData;
```
WalshyDev, line 17: Unused import

WalshyDev (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/Crucible.java
RobotHanzo: Fixed, forgot to remove after experimentations

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/Crucible.java
```diff
@@ -188,7 +188,7 @@ private void addLiquidLevel(@Nonnull Block block, boolean water) {
 
     private void placeLiquid(@Nonnull Block block, boolean water) {
         if (block.getType() == Material.AIR || block.getType() == Material.CAVE_AIR || block.getType() == Material.VOID_AIR) {
-            block.setType(water ? Material.WATER : Material.LAVA);
+            block.setType(water ? Material.WATER : Material.LAVA, false);
```
TheBusyBiscuit, line 191: Changes like these should usually be preceded by a comment explaining them with a reference to the issue.
Example:

```suggestion
            // Fixes #2903 - Cancel physics update to resolve weird overlapping
            block.setType(water ? Material.WATER : Material.LAVA, false);
```

TheBusyBiscuit (changes requested):
One minor addition I would like to suggest to make it future-proof and transparent, otherwise looking good.

Did you ensure that this does indeed fix the corresponding issue?

TheBusyBiscuit (approved):
Lgtm if this was well tested.

RobotHanzo:
It is tested, forgot to include a demonstration video since it was a small PR
[YouTube](https://youtu.be/Ib1LvNlp_N4)

TheBusyBiscuit:
Perfect! Now we just need to wait for a second reviewer.

WalshyDev (approved):

PR #2911, opened 2021-03-26
Check lore to determine whether (Dis)Enchantable

Rothes:
## Description
Support using lore to prevent enchanting or disenchanting.
The existing config is only for Slimefun items, but this one is mainly for RPG items.
Allow to modify lores in Items.yml to facilitate administrator customization.

## Changes
Check lore to determine whether the item is Enchantable/ Disenchantable.
List is supported for config.

## Related Issues
Nothing.

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
```diff
@@ -1,7 +1,6 @@
 package io.github.thebusybiscuit.slimefun4.implementation.items.electric.machines.enchanting;
 
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
```
WalshyDev, line 3: No star import

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
```diff
@@ -29,6 +28,8 @@
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 
+import static java.util.Arrays.asList;
```
WalshyDev, line 31: Don't do static imports unless you're in unit tests

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
```diff
@@ -50,13 +51,15 @@
 
     private final ItemSetting<Boolean> useEnchantLevelLimit = new ItemSetting<>(this, "use-enchant-level-limit", false);
     private final IntRangeSetting enchantLevelLimit = new IntRangeSetting(this, "enchant-level-limit", 0, 10, Short.MAX_VALUE);
+    private final ItemSetting<List<String>> cantDisenchantLores = new ItemSetting<>(this, "cant-disenchant-lores", asList("§7- §cCan't be Auto-Disenchanted"));
```
WalshyDev, line 54: Don't use `§`. It's a deprecated char and has been for years. It isn't used in any text and hasn't been for a long time.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
```diff
@@ -147,7 +150,14 @@ private void transferEnchantments(ItemStack item, ItemStack book, Map<Enchantmen
     private boolean isDisenchantable(@Nullable ItemStack item) {
         if (item == null) {
             return false;
-        } else if (item.getType() != Material.BOOK) {
+        } else if (item.hasItemMeta() && item.getItemMeta().getLore() != null) {
```
WalshyDev, line 153:
```suggestion
        } else if (item.hasItemMeta() && item.getItemMeta().hasLore()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -24,6 +25,8 @@
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 
+import static java.util.Arrays.asList;
```
WalshyDev, line 28: Don't use static imports

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -41,13 +44,15 @@
 
     private final ItemSetting<Boolean> useEnchantLevelLimit = new ItemSetting<>(this, "use-enchant-level-limit", false);
     private final IntRangeSetting enchantLevelLimit = new IntRangeSetting(this, "enchant-level-limit", 0, 10, Short.MAX_VALUE);
+    private final ItemSetting<List<String>> cantEnchantLores = new ItemSetting<>(this, "cant-enchant-lores", asList("§7- §cCan't be Auto-Enchanted"));
```
WalshyDev, line 47: Again, old and unused char. Bad char

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -117,7 +122,13 @@ protected MachineRecipe findNextRecipe(BlockMenu menu) {
 
     private boolean isEnchantable(ItemStack item) {
         SlimefunItem sfItem = null;
-
+        if (item != null && item.hasItemMeta() && item.getItemMeta().getLore() != null) {
```
WalshyDev, line 125:
```suggestion
        if (item != null && item.hasItemMeta() && item.getItemMeta().hasLore()) {
```

WalshyDev (changes requested):
I feel like it'd just be better though for proper support to be added than workarounds.

Rothes (commented):
Thanks for the review. Is it ok now?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
```diff
@@ -147,7 +152,14 @@ private void transferEnchantments(ItemStack item, ItemStack book, Map<Enchantmen
     private boolean isDisenchantable(@Nullable ItemStack item) {
         if (item == null) {
             return false;
-        } else if (item.getType() != Material.BOOK) {
+        } else if (item.hasItemMeta() && item.getItemMeta().hasLore()) {
+            for (String lore : cantDisenchantLores.getValue()) {
+                if (item.getItemMeta().getLore().contains(ChatColors.color(lore))) {
```
Mooy1, line 157: You are calling getItemMeta() twice, assign it to a variable before you check hasLore()

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -117,7 +122,13 @@ protected MachineRecipe findNextRecipe(BlockMenu menu) {
 
     private boolean isEnchantable(ItemStack item) {
         SlimefunItem sfItem = null;
-
+        if (item != null && item.hasItemMeta() && item.getItemMeta().hasLore()) {
+            for (String lore : cantEnchantLores.getValue()) {
```
Mooy1, line 126: Same here, assign getItemMeta() to a variable as well as getLore()

Mooy1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
Rothes: edited

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
Rothes: edited

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -130,17 +130,21 @@ protected MachineRecipe findNextRecipe(BlockMenu menu) {
...
-                if (item.getItemMeta().getLore().contains(ChatColors.color(lore))) {
-                    return false;
+    private boolean isEnchantable(@Nullable ItemStack item) {
+        if (item == null) {
+            return false;
+        } else {
+            ItemMeta itemMeta = item.getItemMeta();
+            if (itemMeta != null && itemMeta.hasLore()) {
+                for (String lore : cantEnchantLores.getValue()) {
+                    if (itemMeta.getLore().contains(ChatColors.color(lore))) {
```
Mooy1, line 140: same as above

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
```diff
@@ -152,10 +151,13 @@ private void transferEnchantments(ItemStack item, ItemStack book, Map<Enchantmen
...
             return false;
-        } else if (item.hasItemMeta() && item.getItemMeta().hasLore()) {
-            for (String lore : cantDisenchantLores.getValue()) {
-                if (item.getItemMeta().getLore().contains(ChatColors.color(lore))) {
-                    return false;
+        } else {
+            ItemMeta itemMeta = item.getItemMeta();
+            if (itemMeta != null && itemMeta.hasLore()) {
+                for (String lore : cantDisenchantLores.getValue()) {
+                    if (itemMeta.getLore().contains(ChatColors.color(lore))) {
```
Mooy1, line 158: getLore() should be assigned to a variable to, it makes a copy each time its called

Mooy1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
Rothes: I thought it was easier to read... Edited now.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
Rothes: Edited

Mooy1 (changes requested):
Figured out how to do a suggestion, so i think use that for both. The issue with calling getItemMeta() without checking hasItemMeta() is it will create a new ItemMeta and then a copy for you. getLore() will return null if there isn't any so its fine.

TheBusyBiscuit:
Sorry for causing these merge conflicts but the AutoEnchanter and AutoDisenchanter classes were getting a bit out of hand and quite redundant.
I created a new super class called `AbstractEnchantmentMachine`:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -126,8 +131,23 @@ protected MachineRecipe findNextRecipe(BlockMenu menu) {
...
+                List<String> lores = itemMeta.getLore();
+                if (lores != null) {
+                    for (String lore : cantEnchantLores.getValue()) {
+                        if (lore.contains(ChatColors.color(lore))) {
+                            return false;
+                        }
+                    }
+                }
+            }
+        }
```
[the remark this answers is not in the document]
Rothes: Looks good. I haven't decompile this because I was busy at the time.

Rothes:
> Sorry for causing these merge conflicts but the AutoEnchanter and AutoDisenchanter classes were getting a bit out of hand and quite redundant.
> I created a new super class called `AbstractEnchantmentMachine`:
> [c0058a3#diff-0aecd876d75b3710e71b8613c33bd4f790c1bb0d16c5c67bf3b33e640d4d7438](https://github.com/Slimefun/Slimefun4/commit/c0058a39945e75fee5427f4ec718acfbdf1204b8#diff-0aecd876d75b3710e71b8613c33bd4f790c1bb0d16c5c67bf3b33e640d4d7438)
> 
> It would be nice if you could add this feature to that class instead as a simple `ignored-lore` setting which applies to both as well as a `hasIgnoredLore(...)` method.
> This would make it a lot easier to maintain and reduces redundancy a lot.

I think separate ignored-enchant-lores and ignored-disenchant-lores is better for some server

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -0,0 +1,84 @@
...
+ * @see AutoEnchanter
+ * @see AutoDisenchanter
+ *
+ */
+abstract class AbstractEnchantmentMachine extends AContainer {
+
+    private final ItemSetting<Boolean> useLevelLimit = new ItemSetting<>(this, "use-enchant-level-limit", false);
+    private final IntRangeSetting levelLimit = new IntRangeSetting(this, "enchant-level-limit", 0, 10, Short.MAX_VALUE);
+    private final ItemSetting<List<String>> ignoredEnchantLores = new ItemSetting<>(this, "ignored-enchant-lores", Arrays.asList("&7- &cCan't be Auto-Enchanted"));
+    private final ItemSetting<List<String>> ignoredDisenchantLores = new ItemSetting<>(this, "ignored-disenchant-lores", Arrays.asList("&7- &cCan't be Auto-Disenchanted"));
```
Mooy1, line 38: these item settings should be in the sub classes since enchanter doesnt use disenchant settings and vice versa

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -0,0 +1,84 @@
...
+            throw new IllegalStateException("Enchantment level limit not enabled, cannot display a warning.");
+        }
+
+        String notice = ChatColors.color(SlimefunPlugin.getLocalization().getMessage("messages.above-limit-level"));
+        notice = notice.replace("%level%", String.valueOf(levelLimit.getValue()));
+        ItemStack progressBar = new CustomItem(Material.BARRIER, " ", notice);
+        menu.replaceExistingItem(22, progressBar);
+    }
+
+    protected boolean hasIgnoredLore(ItemStack itemStack, AbstractEnchantmentMachine enchantmentMachine) {
```
Mooy1, line 65: i think change 2nd param to an ItemSetting<List<String>> that way each implementation can pass its own item setting.

Mooy1 (changes requested):

TheBusyBiscuit:
> > Sorry for causing these merge conflicts but the AutoEnchanter and AutoDisenchanter classes were getting a bit out of hand and quite redundant.
> > I created a new super class called `AbstractEnchantmentMachine`:
> > [c0058a3#diff-0aecd876d75b3710e71b8613c33bd4f790c1bb0d16c5c67bf3b33e640d4d7438](https://github.com/Slimefun/Slimefun4/commit/c0058a39945e75fee5427f4ec718acfbdf1204b8#diff-0aecd876d75b3710e71b8613c33bd4f790c1bb0d16c5c67bf3b33e640d4d7438)
> > It would be nice if you could add this feature to that class instead as a simple `ignored-lore` setting which applies to both as well as a `hasIgnoredLore(...)` method.
> > This would make it a lot easier to maintain and reduces redundancy a lot.
> 
> I think separate ignored-enchant-lores and ignored-disenchant-lores is better for some server

Item Settings are always per-SlimefunItem.
This refactoring has nothing to do that, people will still configure it the same way.
These are just code restructurings/reorganizing.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
TheBusyBiscuit, line 38: This should be one single ItemSetting to reduce redundancy.

```suggestion
    private final ItemSetting<List<String>> ignoredEnchantLores = new ItemSetting<>(this, "ignored-enchant-lores", Arrays.asList("&7- &cCan't be Auto-Enchanted"));
    private final ItemSetting<List<String>> ignoredDisenchantLores = new ItemSetting<>(this, "ignored-disenchant-lores", Arrays.asList("&7- &cCan't be Auto-Disenchanted"));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
TheBusyBiscuit, line 65: There is no point in passing itself here.

```suggestion
    protected boolean hasIgnoredLore(ItemStack item) {
```

TheBusyBiscuit (changes requested):

Rothes:
Sorry that I was unconsciousness at the time and gone to sleep immediately when updated it lol..

TheBusyBiscuit:
> Sorry that I was unconsciousness at the time and gone to sleep immediately when updated it lol..

All good.
By the way, I updated your working tree to sanitize all the leftovers from unmerged commits. Now the diff is easier to read as it should only show the actual changes that were made.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +60,18 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
         menu.replaceExistingItem(22, progressBar);
     }
 
+    protected boolean hasIgnoredLore(ItemStack itemStack) {
```
TheBusyBiscuit, line 63:
```suggestion
    protected boolean hasIgnoredLore(@Nonnull ItemStack itemStack) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoDisenchanter.java
```diff
@@ -142,6 +142,8 @@ private void transferEnchantments(ItemStack item, ItemStack book, Map<Enchantmen
     private boolean isDisenchantable(@Nullable ItemStack item) {
         if (item == null) {
             return false;
+        } else if (hasIgnoredLore(item)) {
+            return false;
```
TheBusyBiscuit, line 146: This is a rather heavy method call as it invokes `.getItemMeta()` so it should only happen further down.
The book check was put in place to aid performance, changing this order will render this performance improvement useless.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -120,8 +119,12 @@ protected MachineRecipe enchant(BlockMenu menu, ItemStack target, ItemStack ench
     }
 
     private boolean isEnchantable(@Nullable ItemStack item) {
-        // stops endless checks of getByItem for enchanted book stacks.
-        if (item != null && item.getType() != Material.ENCHANTED_BOOK) {
+        if (item == null) {
+            return false;
+        } else if (hasIgnoredLore(item)) {
+            return false;
```
TheBusyBiscuit, line 125: This is a rather heavy method call as it invokes `.getItemMeta()` so it should only happen further down.
The book check was put in place to aid performance, changing this order will render this performance improvement useless.

Material checks should always come before ItemMeta checks.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +60,18 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
...
+        List<String> ignoredLore = ignoredLores.getValue();
+        if (itemStack.hasItemMeta()) {
+            List<String> itemLore = itemStack.getItemMeta().getLore();
+            if (itemLore != null && ignoredLore != null) {
```
TheBusyBiscuit, line 67: `ignoredLore` is not nullable, so this null check is redundant.
Same goes for `itemLore`.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +60,18 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
...
+                for (String lore : ignoredLores.getValue()) {
```
TheBusyBiscuit, line 68:
```suggestion
                for (String lore : ignoredLore) {
```

TheBusyBiscuit (changes requested):

Rothes:
All is finished, but not tested yet.

Rothes:
I have some free time tormorrow and gonna have a test.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -17,12 +17,15 @@
...
+import java.util.List;
+
 /**
  * This is a super class of the {@link AutoEnchanter} and {@link AutoDisenchanter} which is
  * used to streamline some methods and combine common attributes to reduce redundancy.
- * 
+ *
  * @author TheBusyBiscuit
- * 
+ *
```
TheBusyBiscuit, line 28:
```suggestion
 * @author Rothes
 *
```

TheBusyBiscuit (approved):
Looks pretty much fine to me right now.
@WalshyDev should still have a second look at this eventually.

Only thing I would like to see would be an option to disable this check, similar to how the level limit has a boolean for it.
But that can always be added later on.

TheBusyBiscuit (dismissed):
same as before

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +61,16 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
         menu.replaceExistingItem(22, progressBar);
     }
 
+    protected boolean hasIgnoredLore(@Nonnull ItemStack itemStack) {
+        List<String> ignoredLore = ignoredLores.getValue();
+        if (itemStack.hasItemMeta()) {
+            List<String> itemLore = itemStack.getItemMeta().getLore();
```
WalshyDev, line 67: getLore can be null so your contains check can throw an NPE. You will want to do a != null check here or do hasLore (which does != null && !isEmpty)

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
TheBusyBiscuit: hasLore() preferabbly.

Mooy1:
> Looks pretty much fine to me right now.
> @WalshyDev should still have a second look at this eventually.
> 
> Only thing I would like to see would be an option to disable this check, similar to how the level limit has a boolean for it.
> But that can always be added later on.

Could return immediatly in hasIgnoredLore if the item setting's list is size 0

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
Rothes: Once we wandered between whether to do hasLore. Fixed now

Rothes:
> > Looks pretty much fine to me right now.
> > @WalshyDev should still have a second look at this eventually.
> > Only thing I would like to see would be an option to disable this check, similar to how the level limit has a boolean for it.
> > But that can always be added later on.
> 
> Could return immediatly in hasIgnoredLore if the item setting's list is size 0

Yes that was what I thought.

I have tested it and added skip the check if setting is empty.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +61,23 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
         menu.replaceExistingItem(22, progressBar);
     }
 
+    protected boolean hasIgnoredLore(@Nonnull ItemStack itemStack) {
+        List<String> ignoredLore = ignoredLores.getValue();
+        // Skip the check if not set any ignored lore.
+        if (ignoredLore.isEmpty()) {
+            return false;
+        }
```
TheBusyBiscuit, line 69: Not a fan of this, the default should be to not ignore any lores, yet we have a default value.
This will have a negative impact on performance for people who don't even want to use this setting.
That's why I would rather suggest a seperate toggle setting, similar to how the level limit has one.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +61,23 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
...
+        if (itemStack.hasItemMeta()) {
+            List<String> itemLore = itemStack.getItemMeta().getLore();
+            if (itemLore == null) {
+                return false;
+            }
```
TheBusyBiscuit, line 74: You should do `ItemMeta#hasLore()` instead of relying on a null check.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -31,13 +35,15 @@
 
     private final ItemSetting<Boolean> useLevelLimit = new ItemSetting<>(this, "use-enchant-level-limit", false);
     private final IntRangeSetting levelLimit = new IntRangeSetting(this, "enchant-level-limit", 0, 10, Short.MAX_VALUE);
+    private final ItemSetting<List<String>> ignoredLores = new ItemSetting<>(this, "ignored-lores", Arrays.asList("&7- &cCan't be used in " + this.getItemName()));
```
WalshyDev, line 38: I feel like this should be an empty list not gonna lie... I don't see a reason to do any checks when likely this will not be changed by the vast majority of servers.

WalshyDev: If we do want to keep a value here though, this should really be using Collections.singletonList

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +61,23 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
...
 
+    protected boolean hasIgnoredLore(@Nonnull ItemStack itemStack) {
+        List<String> ignoredLore = ignoredLores.getValue();
+        // Skip the check if not set any ignored lore.
+        if (ignoredLore.isEmpty()) {
+            return false;
+        }
+        if (itemStack.hasItemMeta()) {
+            List<String> itemLore = itemStack.getItemMeta().getLore();
+            if (itemLore == null) {
```
WalshyDev, line 72: as Cookie mentioned before, probably best to do hasLore, quick circuit-break if empty too (and if they change stuff in the future it'll be supported)

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
TheBusyBiscuit: I realized that I forgot to press "submit" on my review

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
Rothes: Collections.singletonList can cause error in ItemSetting process before.. Let's keep it empty simply.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
Rothes: OK. I thought empty list are rare to see so didn't check.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
Rothes: Yes.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
Rothes: Or maybe I will add a toggle setting, as BusyBiscuit suggested.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +63,23 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
...
     }
 
+    protected boolean hasIgnoredLore(@Nonnull ItemStack itemStack) {
+        List<String> ignoredLore = ignoredLores.getValue();
+        // Skip the check if not enable this feature or set any ignored lore.
+        if (!useIgnoredLores.getValue() || ignoredLore.isEmpty()) {
+            return false;
+        }
+        if (itemStack.hasItemMeta() && itemStack.getItemMeta().hasLore()) {
+            List<String> itemLore = itemStack.getItemMeta().getLore();
```
WalshyDev, line 73: don't call getItemMeta twice, it's intensive.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +63,23 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
...
+            if (itemLore.isEmpty()) {
```
WalshyDev, line 74: this is redundant with hasLore

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -117,7 +116,7 @@ protected MachineRecipe enchant(BlockMenu menu, ItemStack target, ItemStack ench
 
     private boolean isEnchantable(@Nullable ItemStack item) {
         // stops endless checks of getByItem for enchanted book stacks.
-        if (item != null && !item.getType().isAir() && item.getType() != Material.ENCHANTED_BOOK) {
+        if (item != null && !item.getType().isAir() && item.getType() != Material.ENCHANTED_BOOK && !hasIgnoredLore(item)) {
```
WalshyDev, line 119: I'd flip the order of these material checks to just make this a little quicker
```suggestion
        if (item != null && item.getType() != Material.ENCHANTED_BOOK && !item.getType().isAir() && !hasIgnoredLore(item)) {
```

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
```diff
@@ -55,4 +64,23 @@ protected void showEnchantmentLevelWarning(@Nonnull BlockMenu menu) {
...
+            }
+            List<String> itemLore = itemMeta.getLore();
+            for (String lore : ignoredLore) {
+                if (itemLore.contains(ChatColors.color(lore))) {
+                    return true;
+                }
+            }
+        }
+        return false;
+    }
```
TheBusyBiscuit, line 85: This can be tidied up a bit, `ignoredLore` is not used until much further down, so we dont need to declare it this early on.
Also replaced the guard block with a simple if (AND)
```suggestion
        if (useIgnoredLores.getValue() && itemStack.hasItemMeta()) {
            ItemMeta itemMeta = itemStack.getItemMeta();

            if (!itemMeta.hasLore()) {
                return false;
            }

            List<String> itemLore = itemMeta.getLore();
            List<String> ignoredLore = ignoredLores.getValue();

            for (String lore : ignoredLore) {
                if (itemLore.contains(ChatColors.color(lore))) {
                    return true;
                }
            }
        }

        return false;
    }
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -36,7 +36,6 @@
  *
  */
 public class AutoEnchanter extends AbstractEnchantmentMachine {
-
```
TheBusyBiscuit, line 39: Why was this line removed? :(

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AbstractEnchantmentMachine.java
Rothes: Should be edited after I added some check earlier... But I forgot.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
Rothes: I am sorry about that!?!?

TheBusyBiscuit (approved):
Looks good to me!

WalshyDev (approved):
lgtm

PR #2917, opened 2021-03-28
The update that fixed world hunger (or how I made a cake factory)

TheBusyBiscuit:
## Description

Added the Produce Collector.
This machine can fill up buckets with milk and bowls with mushrrom stew if a nearby Cow/Mooshroom is found.

Also fixed a bug in the Auto Crafter which consumed the milk instead of leaving a bucket behind.
And allowed Block Placers to place cake (special case that I really wanted)

All of this makes for a nice completely automated cake factory.
When you eat the cake, it is immediately replenished by the Block Placer below.
Infinite Cake

## Changes

* Added Produce Collector
* Fixed Auto Crafters not handling milk buckets properly
* Allowed Block Placers to place down cakes
* Fixed Multimeter not working on Auto-Crafters

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/ResearchSetup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/ResearchSetup.java
@@ -282,6 +282,7 @@ public final class ResearchSetup {
         register("wise_talisman", 271, "Talisman of the Wise", 20, SlimefunItems.TALISMAN_WISE);
         register("book_binder", 272, "Enchantment Book Binding", 26, SlimefunItems.BOOK_BINDER);
         register("auto_crafting", 273, "Automatic Crafting", 30, SlimefunItems.VANILLA_AUTO_CRAFTER, SlimefunItems.ENHANCED_AUTO_CRAFTER, SlimefunItems.ARMOR_AUTO_CRAFTER);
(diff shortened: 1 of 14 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/ResearchSetup.java
```diff
@@ -282,6 +282,7 @@ public static void setupResearches() {
         register("wise_talisman", 271, "Talisman of the Wise", 20, SlimefunItems.TALISMAN_WISE);
         register("book_binder", 272, "Enchantment Book Binding", 26, SlimefunItems.BOOK_BINDER);
         register("auto_crafting", 273, "Automatic Crafting", 30, SlimefunItems.VANILLA_AUTO_CRAFTER, SlimefunItems.ENHANCED_AUTO_CRAFTER, SlimefunItems.ARMOR_AUTO_CRAFTER);
+        register("produce_collector", 274, "Automatic Milking", 20, SlimefunItems.PRODUCE_COLLECTOR);
```
Sfiguz7, line 285: Now I'll have to change my researches cause Blob is slow and hasn't merged my PR yet ffs

Sfiguz7 (approved):

WalshyDev (approved):
LGTM

PR #2919, opened 2021-03-28
Fixed damageable items not respecting `unbreakable` tag

GallowsDove:
## Description

Add unbreakable check to damageItem, this should be probably default behaviour, and I specifically require it for my Unbreakable rune to work properly.

## Checklist


WalshyDev:
You should be checking for hasItemMeta first

GallowsDove:
Yop that makes sense, however there wasn't any hasItemMeta check to begin with, so I wasn't sure whether it would be required

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/DamageableItem.java
```diff
@@ -47,22 +47,25 @@
...
-        if (isDamageable() && item != null && item.getType() != Material.AIR && item.getAmount() > 0) {
-            int unbreakingLevel = item.getEnchantmentLevel(Enchantment.DURABILITY);
+        if (isDamageable() && item != null && item.getType() != Material.AIR && item.getAmount() > 0 && item.hasItemMeta()) {
+            ItemMeta meta = item.getItemMeta();
 
-            if (unbreakingLevel > 0 && Math.random() * 100 <= (60 + Math.floorDiv(40, (unbreakingLevel + 1)))) {
-                return;
-            }
+            if (!meta.isUnbreakable()) {
+                Damageable damageable = (Damageable) meta;
```
TheBusyBiscuit, line 54: `getItemMeta()` is an expensive call, so please revert this swap around because checking for the unbreaking level first is a rather important performance optimization.

TheBusyBiscuit (changes requested):

GallowsDove:
I'm sorry I'm just getting mixed signals there, will fix

TheBusyBiscuit (approved):
LGTM

svr333 (approved):

PR #2950, opened 2021-04-08
Optimizations to cargo

md5sha256:
## Description

This commit introduces two optimizations to cargo handling. 

## Changes

- Utilization of regex to validate integers in CargoNet#getFrequency,
- Smart shallow-cloning in CargoNetworkTask: using an ArrayList when round-robin is disabled

The commit used to test these optimizations can be found [here.](https://github.com/md5sha256/Slimefun4/commit/6d6e45f93193bee6e25f3dd52edbbc358f082680)

## Test Environment
- Cargo System: Electric Smelter (Gold Ore --> Gold Ingot) connected to a chest. 
- 1x Input Node, 1x Output node
- Input node and Output node are whitelisted to accept Gold Ingots + 4 other random items
- Input node has round robin enabled
- Network visualisation is enabled

## Results 

**Item Routing**
image

**Get Frequency**
image

Brief concluding notes: 
Whilst neither were tested using a proper benchmark such as JMH, the rough results do show a clear trend. 
- `distributeItems` is on average 78% faster when using an ArrayDeque for round robin and ArrayList otherwise
- `getFrequency` is 55% faster. 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
@@ -251,7 +251,6 @@ public class CargoNet extends AbstractItemNetwork implements HologramOwner {
      * @return The frequency of the given node
      */
     private static int getFrequency(@Nonnull Location node) {
-
         String str = BlockStorage.getLocationInfo(node,"frequency");
         if (str == null) {
             return 0;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
@@ -255,7 +255,7 @@ public class CargoNet extends AbstractItemNetwork implements HologramOwner {
         if (str == null) {
             return 0;
         } else if (!PatternUtils.NUMERIC.matcher(str).matches()) {
-            SlimefunPlugin.logger().log(Level.SEVERE, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
+            SlimefunPlugin.logger().log(Level.SEVERE, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + ',' + node.getBlockY() + ',' + node.getBlockZ() + ')');
             return 0;
         }
         return Integer.parseInt(str);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
@@ -242,7 +242,7 @@ public class CargoNet extends AbstractItemNetwork implements HologramOwner {
 
     /**
      * This method returns the frequency a given node is set to.
-     * Should there be an {@link Exception} to this method it will fall back to zero in
+     * Should there be invalid data this method it will fall back to zero in
      * order to preserve the integrity of the {@link CargoNet}.
      * 
      * @param node
@@ -250,13 +250,12 @@ public class CargoNet extends AbstractItemNetwork implements HologramOwner {
      * 
      * @return The frequency of the given node
      */
-    private static int getFrequency(Location node) {
+    private static int getFrequency(@Nonnull Location node) {
 
-        String str = BlockStorage.getLocationInfo(node).getString("frequency");
+        String str = BlockStorage.getLocationInfo(node,"frequency");
         if (str == null) {
             return 0;
-        }
-        if (!PatternUtils.NUMERIC.matcher(str).matches()) {
+        } else if (!PatternUtils.NUMERIC.matcher(str).matches()) {
             SlimefunPlugin.logger().log(Level.SEVERE, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
             return 0;
         }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
@@ -11,6 +11,7 @@ import java.util.logging.Level;
 
 import javax.annotation.Nonnull;
 
+import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
 import org.bukkit.Location;
 import org.bukkit.block.Block;
 import org.bukkit.inventory.ItemStack;
@@ -250,12 +251,16 @@ public class CargoNet extends AbstractItemNetwork implements HologramOwner {
      * @return The frequency of the given node
      */
     private static int getFrequency(Location node) {
-        try {
-            String str = BlockStorage.getLocationInfo(node).getString("frequency");
-            return str == null ? 0 : Integer.parseInt(str);
-        } catch (Exception x) {
-            SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
+
+        String str = BlockStorage.getLocationInfo(node).getString("frequency");
+        if (str == null) {
             return 0;
         }
+        if (!PatternUtils.NUMERIC.matcher(str).matches()) {
+            SlimefunPlugin.logger().log(Level.SEVERE, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
+            return 0;
+        }
+        return Integer.parseInt(str);
+
     }
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
@@ -1,10 +1,10 @@
 package io.github.thebusybiscuit.slimefun4.core.networks.cargo;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Deque;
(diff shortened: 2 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
```diff
@@ -1,13 +1,6 @@
...
 
-import java.util.Deque;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
J3fftw1, line 3: No star imports

md5sha256: ah, missed that

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -250,12 +251,16 @@ public void tick(Block b) {
...
-            return str == null ? 0 : Integer.parseInt(str);
-        } catch (Exception x) {
-            SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
+
+        String str = BlockStorage.getLocationInfo(node).getString("frequency");
+        if (str == null) {
+            return 0;
+        }
+        if (!PatternUtils.NUMERIC.matcher(str).matches()) {
+            SlimefunPlugin.logger().log(Level.SEVERE, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
```
WalshyDev, line 260:
```suggestion
            SlimefunPlugin.logger().log(Level.SEVERE, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + ',' + node.getBlockY() + ',' + node.getBlockZ() + ')');
```

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
md5sha256: Is there a reason why chars shpuld be used instead? I was just keeping the old error logging.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
```diff
@@ -158,13 +160,17 @@ private void insertItem(Block inputTarget, int previousSlot, ItemStack item) {
...
 
+        Collection<Location> destinations;
         if (roundrobin) {
-            roundRobinSort(inputNode, destinations);
+            Deque<Location> tempDestinations = new LinkedList<>(outputNodes);
+            roundRobinSort(inputNode, tempDestinations);
+            destinations = tempDestinations;
+        } else {
+            destinations = new ArrayList<>(outputNodes);
         }
```
TheBusyBiscuit, line 174: Some comments here would be nice.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -250,12 +251,16 @@ public void tick(Block b) {
...
-            String str = BlockStorage.getLocationInfo(node).getString("frequency");
-            return str == null ? 0 : Integer.parseInt(str);
-        } catch (Exception x) {
-            SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
+
+        String str = BlockStorage.getLocationInfo(node).getString("frequency");
+        if (str == null) {
+            return 0;
+        }
+        if (!PatternUtils.NUMERIC.matcher(str).matches()) {
```
TheBusyBiscuit, line 259: I don't really see the point of putting a regular expression here.

Yes, it validates the data but we can already be certain that it is gonna match 99.9999% of the time since we control what data goes in there.
Only reason why this would not match is if someone manually edited their data files or if a file got corrupted.

So this to my eyes would rather be hurting performance if anything. How on earth does the extra check make it faster though...?

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -250,12 +251,16 @@ public void tick(Block b) {
      * @return The frequency of the given node
      */
     private static int getFrequency(Location node) {
-        try {
-            String str = BlockStorage.getLocationInfo(node).getString("frequency");
-            return str == null ? 0 : Integer.parseInt(str);
-        } catch (Exception x) {
-            SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
+
+        String str = BlockStorage.getLocationInfo(node).getString("frequency");
```
TheBusyBiscuit, line 255:
```suggestion
        String str = BlockStorage.getLocationInfo(node, "frequency");
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -250,12 +251,16 @@ public void tick(Block b) {
      * @return The frequency of the given node
      */
     private static int getFrequency(Location node) {
```
TheBusyBiscuit, line 253:
```suggestion
    private static int getFrequency(@Nonnull Location node) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
TheBusyBiscuit, line 259: Why not an else if?
```suggestion
        } else if (!PatternUtils.NUMERIC.matcher(str).matches()) {
```

TheBusyBiscuit (changes requested):
Generally not too sure why the frequency one would even be an improvement in the first place.
Also some comments would be great

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
md5sha256: Well, we had a try-catch before for the number-format-exception this is why the performance improved.

Sfiguz7 (approved):
Agree with the couple of review comments already put in, nothing else needing changes that I can see. Approved in the meantime since their requested changes block merge

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
TheBusyBiscuit: But it should almost never reach the catch block.
Is the existence of the try/catch itself really affecting performance this much??

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
md5sha256: I just noticed an ArrayDeque might be faster; I'll redo the test and add comments

TheBusyBiscuit: ooh, nice

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
md5sha256: yes, try-catch is expensive.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
md5sha256: it was indeed faster!

md5sha256:
Okay, I've fixed most of the nitpicks (bar Walshy's as I'm waiting for the reasoning behind the use of chars for ','). Otherwise, should be good to go.

TheBusyBiscuit (dismissed):
Looks good to me! Thanks

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
WalshyDev: Haha! Yes! Thanks for asking. Chars are faster and use slightly less memory when doing this concatination. When you do "," you're making a new immutable String object which is heavier than just a single char. You're also maintaining an array (byte array or char array depending on version) and hash. Whereas char just holds it's value.

There is a good StackOverflow answer about this too: https://stackoverflow.com/a/24860958

Both are still very quick but if we can use a char, we should.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -241,21 +242,24 @@ public void tick(Block b) {
...
      * @return The frequency of the given node
      */
-    private static int getFrequency(Location node) {
-        try {
-            String str = BlockStorage.getLocationInfo(node).getString("frequency");
-            return str == null ? 0 : Integer.parseInt(str);
-        } catch (Exception x) {
-            SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "An Error occurred while parsing a Cargo Node Frequency (" + node.getWorld().getName() + " - " + node.getBlockX() + "," + node.getBlockY() + "," + +node.getBlockZ() + ")");
+    private static int getFrequency(@Nonnull Location node) {
+
```
WalshyDev, line 254:
```suggestion
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
md5sha256: Fair enough, thanks for the insight!

WalshyDev (approved):
LGTM

TheBusyBiscuit (approved):
LGTM too

PR #2982, opened 2021-04-23
Fixes #2977

### LINKED ISSUE #2977: Auto-Enchanter has strict slot order

The author enchanter does not work if the item is on the left (first) slot

1. Place auto enchanter
2. Insert an item (such as a pickaxe) into the left (first) slot
3. Insert an enchanted book (eg unbreaking) into the right (second) slot
4. Nothing happens
If the enchanted book is placed in the left (first) slot and the item in the right (second), the auto enchanter works.
Video: https://youtu.be/NPX6k2nsg-Q

The auto enchanter will work if there is an item in the right (first) slot and an enchanted book in the left (second) slot

## :scroll: Server Log

N/A

N/A

image

martinbrom:
## Description

Auto Enchanter now considers both combinations of enchanted book and item (item in first or second slot).

## Proposed changes

The machine now tries to check if the second combination could possibly be valid instead of exiting immediately as before.
Uses `addUnsafeEnchantments` instead of `addUnsafeEnchantment` to skip the loop over map entries.
Uses `isEmpty` and `map.size()` instead of a separate amount variable.

Resolves #2977 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
@@ -53,7 +53,7 @@ public class AutoEnchanter extends AbstractEnchantmentMachine {
 
             // Check if the item is enchantable
             if (!isEnchantable(item)) {
-                return null;
+                continue;
             }
 
             // Call an event so other Plugins can modify it.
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -53,7 +53,7 @@ protected MachineRecipe findNextRecipe(BlockMenu menu) {
 
             // Check if the item is enchantable
             if (!isEnchantable(item)) {
```
TheBusyBiscuit, line 55: I just noticed that `isEnchantable` is missing an air check but that's just a side note.

TheBusyBiscuit (approved):
LGTM

Don't even know why we had a seperate `amount` variable there. Parts of this class are over 5 years old...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
martinbrom: Do you want me to add that or can it be ignored?

TheBusyBiscuit: You can add it if you want but can also ignore it and we will add it later.
I approved it, so no explicit actions are expected nor required :)

WalshyDev (approved):

PR #2822, opened 2021-02-13
WIP - Refactored config handling + Added /sf reload

TheBusyBiscuit:
## Description

This pull requests moves the config-related stuff from SlimefunRegistry to a dedicated Config Manager.
It also allows some basic reload functionality,, accessible via `/sf reload` (Many people requested this for a long time).
Note that this is not fully finished though, so not all changes are supported and a lot will still require a restart.

## Changes

* Added /sf reload
* Added Config Manager to handle our configs
* All relevant config files have been moved to this Config Manager class
* All relevant cached data was moved to the Config Manager too

## Checklist

- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [x] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [x] I followed the existing code standards and didn't mess up the formatting.
- [x] I did my best to add documentation to any public classes or methods I added.
- [x] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

PR #2983, opened 2021-04-23
Add the option to disable learning animations server & client side

martinbrom:
## Description

Discord suggestion 4620
It takes forever to unlock researches even with all of the required XP.

## Proposed changes

Add a new config key and guide option to allow both server owners and players to disable/enable the learning animation.
The default value is 'enabled' to keep the behavior the same it currently is.
If the animations are enabled server-side, the player can choose to disable them. If the animations are disabled server-side, the player cannot choose.

Discord suggestion 4620

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
@@ -240,6 +240,11 @@ public final class SlimefunRegistry {
         return researchFireworks;
     }
 
+    /**
+     * Returns whether the research learning animations is disabled
+     *
+     * @return Whether the research learning animations is disabled
+     */
     public boolean isLearningAnimationDisabled() {
         return disableLearningAnimation;
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
@@ -69,6 +69,7 @@ public final class SlimefunRegistry {
     private boolean enableResearches;
     private boolean freeCreativeResearches;
     private boolean researchFireworks;
+    private boolean disableLearningAnimation;
     private boolean logDuplicateBlockEntries;
     private boolean talismanActionBarMessages;
 
@@ -109,6 +110,7 @@ public final class SlimefunRegistry {
         backwardsCompatibility = cfg.getBoolean("options.backwards-compatibility");
         freeCreativeResearches = cfg.getBoolean("researches.free-in-creative-mode");
         researchFireworks = cfg.getBoolean("researches.enable-fireworks");
+        disableLearningAnimation = cfg.getBoolean("researches.disable-learning-animation");
         logDuplicateBlockEntries = cfg.getBoolean("options.log-duplicate-block-entries");
         talismanActionBarMessages = cfg.getBoolean("talismans.use-actionbar");
     }
@@ -236,6 +238,10 @@ public final class SlimefunRegistry {
 
     public boolean isResearchFireworkEnabled() {
         return researchFireworks;
+    }
+
+    public boolean isLearningAnimationDisabled() {
+        return disableLearningAnimation;
     }
 
     /**
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
@@ -9,6 +9,7 @@ import org.bukkit.entity.Player;
 import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.api.player.PlayerProfile;
+import io.github.thebusybiscuit.slimefun4.core.guide.options.SlimefunGuideSettings;
 import io.github.thebusybiscuit.slimefun4.core.researching.Research;
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.implementation.guide.SurvivalSlimefunGuide;
@@ -18,7 +19,7 @@ import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.SlimefunItem;
 /**
  * This interface is used for the different implementations that add behaviour
  * to the {@link SlimefunGuide}.
- * 
+ *
  * @author TheBusyBiscuit
  * 
  * @see SlimefunGuideMode
@@ -30,7 +31,7 @@ public interface SlimefunGuideImplementation {
     /**
      * Every {@link SlimefunGuideImplementation} can be associated with a
      * {@link SlimefunGuideMode}.
-     * 
+     *
      * @return The mode this {@link SlimefunGuideImplementation} represents
      */
     @Nonnull
@@ -40,7 +41,7 @@ public interface SlimefunGuideImplementation {
      * Returns the {@link ItemStack} representation for this {@link SlimefunGuideImplementation}.
      * In other words: The {@link ItemStack} you hold in your hand and that you use to
      * open your {@link SlimefunGuide}
-     * 
+     *
      * @return The {@link ItemStack} representation for this {@link SlimefunGuideImplementation}
      */
     @Nonnull
@@ -63,7 +64,9 @@ public interface SlimefunGuideImplementation {
             research.unlock(p, true, callback);
         } else {
             p.setLevel(p.getLevel() - research.getCost());
-            research.unlock(p, false, callback);
+            boolean skipLearningAnimation = SlimefunPlugin.getRegistry().isLearningAnimationDisabled()
+                    || !SlimefunGuideSettings.hasLearningAnimationEnabled(p);
+            research.unlock(p, skipLearningAnimation, callback);
         }
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/FireworksOption.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/FireworksOption.java
@@ -4,11 +4,8 @@ import java.util.Optional;
 
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
-import org.bukkit.enchantments.Enchantment;
 import org.bukkit.entity.Player;
-import org.bukkit.inventory.ItemFlag;
 import org.bukkit.inventory.ItemStack;
-import org.bukkit.inventory.meta.ItemMeta;
 
 import io.github.thebusybiscuit.cscorelib2.data.PersistentDataAPI;
 import io.github.thebusybiscuit.cscorelib2.item.CustomItem;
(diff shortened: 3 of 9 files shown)


src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/researches/TestResearchUnlocking.java
```diff
@@ -47,7 +47,7 @@ private Player awaitUnlock(Player player, Research research, boolean instant) th
             latch.countDown();
         });
 
-        latch.await(10, TimeUnit.SECONDS);
+        latch.await(instant ? 1 : 10, TimeUnit.SECONDS);
```
WalshyDev, line 50: this doesn't really matter, this is just a max wait time. As soon as the latch hits zero this will no longer be blocking

WalshyDev (changes requested):

src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/researches/TestResearchUnlocking.java
martinbrom: I added this because the non-instant unlocking process takes longer than 1 second so the test would fail.

WalshyDev: Exactly why it was 10 seconds, you made it 1 second if instant 🤔

if it was just as it was, this should be fine.

martinbrom: I made it 1 second so it 'would fail' if the instant research took longer, is that wrong?

WalshyDev: Oh I see what you mean... definitely not a great way to do this. It would work but not exactly ideal

martinbrom: Reverted to the previous version

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/SlimefunGuideSettings.java
```diff
@@ -249,4 +250,25 @@ public static boolean hasFireworksEnabled(@Nonnull Player p) {
...
+    public static boolean hasLearningAnimationEnabled(@Nonnull Player p) {
+        for (SlimefunGuideOption<?> option : options) {
+            if (option instanceof LearningAnimationOption) {
+                LearningAnimationOption learningAnimation = (LearningAnimationOption) option;
+                ItemStack guide = SlimefunGuide.getItem(SlimefunGuideMode.SURVIVAL_MODE);
+                return learningAnimation.getSelectedOption(p, guide).orElse(true);
+            }
+        }
+
+        return true;
```
TheBusyBiscuit, line 271: This method is very similar to the method above, therefore I think a private static helper method might be appropriate here to reduce this redundancy.

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
```diff
@@ -63,7 +64,9 @@ default void unlockItem(Player p, SlimefunItem sfitem, Consumer<Player> callback
             research.unlock(p, true, callback);
         } else {
             p.setLevel(p.getLevel() - research.getCost());
-            research.unlock(p, false, callback);
+            boolean skipLearningAnimation = SlimefunPlugin.getRegistry().isLearningAnimationDisabled()
+                    || !SlimefunGuideSettings.hasLearningAnimationEnabled(p);
```
TheBusyBiscuit, line 68: Using an OR connector here will render the Player setting useless if it has been disabled server-wide.
Just something to keep in mind.

src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
```diff
@@ -238,6 +240,15 @@ public boolean isResearchFireworkEnabled() {
...
     }
 
+    /**
+     * Returns whether the research learning animations is disabled
+     *
+     * @return Whether the research learning animations is disabled
+     */
+    public boolean isLearningAnimationDisabled() {
+        return disableLearningAnimation;
+    }
```
TheBusyBiscuit, line 250: This is fine for now, but we wanna refactor this class quite soon, see #2822

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/LearningAnimationOption.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Nonnull
+    @Override
+    public Optional<ItemStack> getDisplayItem(Player p, ItemStack guide) {
+        if (SlimefunPlugin.getRegistry().isLearningAnimationDisabled()) {
+            return Optional.empty();
+        } else {
+            boolean enabled = getSelectedOption(p, guide).orElse(true);
+            ItemStack item = new CustomItem(Material.PAPER, "&bLearning Animation: &" + (enabled ? "aYes" : "4No"), "", "&7You can now toggle whether you", "&7will see information about your pondering in chat", "&7upon researching an item.", "", "&7\u21E8 &eClick to " + (enabled ? "disable" : "enable") + " your learning animation");
```
TheBusyBiscuit, line 38: Could you please add a locale string for the description here?
So it can be translated.

TheBusyBiscuit (changes requested):
Looks pretty good to me.
I think one part could be improved to be less redundant and another snippet needs localization in my opinion.

Other than that I only added some comments to keep in mind but don't require any further action for now.

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/LearningAnimationOption.java
martinbrom: I added it but the Unicode character gets displayed as \u.... instead of the actual character

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
martinbrom: Yeah, that was the way I intended

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/SlimefunGuideSettings.java
martinbrom: I added a generic one which might seem a little overengineered but I think it will might be useful in the future when there is an option which has a different value than boolean

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/LearningAnimationOption.java
TheBusyBiscuit: Yeah I only meant the actual description.
Though our locale should support unicodes in the format of [u1234] or something like that.

martinbrom: The square bracket placeholder did not work for me and even though I tried googling a way to do it in the end I just went with a simple way and split the message into two parts (and included the Unicode straight in java).

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/LearningAnimationOption.java
```diff
@@ -0,0 +1,67 @@
...
+import org.bukkit.NamespacedKey;
+import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+
+import io.github.thebusybiscuit.cscorelib2.data.PersistentDataAPI;
+import io.github.thebusybiscuit.cscorelib2.item.CustomItem;
+import io.github.thebusybiscuit.slimefun4.api.SlimefunAddon;
+import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
+
+public class LearningAnimationOption implements SlimefunGuideOption<Boolean> {
```
TheBusyBiscuit, line 18: This class should be package-private.
It's also missing javadocs.

```suggestion
class LearningAnimationOption implements SlimefunGuideOption<Boolean> {
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/LearningAnimationOption.java
```diff
@@ -0,0 +1,67 @@
...
+        if (SlimefunPlugin.getRegistry().isLearningAnimationDisabled()) {
+            return Optional.empty();
+        } else {
+            String enabled = getSelectedOption(p, guide).orElse(true) ? "enabled" : "disabled";
+            List<String> lore = SlimefunPlugin.getLocalization().getMessages(
+                    p, "guide.options.learning-animation." + enabled + ".text");
+            lore.add("");
+            lore.add("&7\u21E8 " + SlimefunPlugin.getLocalization().getMessage(
+                    p, "guide.options.learning-animation." + enabled + ".click"));
+            ItemStack item = new CustomItem(Material.PAPER, lore);
```
TheBusyBiscuit, line 44: If we wanna be fancy, we could switch the item here if it is enabled or disabled :o

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/SlimefunGuideSettings.java
```diff
@@ -232,21 +233,48 @@ private static void addConfigurableOptions(Player p, ChestMenu menu, ItemStack g
...
+                OPTION o = (OPTION) option;
                 ItemStack guide = SlimefunGuide.getItem(SlimefunGuideMode.SURVIVAL_MODE);
-                return fireworks.getSelectedOption(p, guide).orElse(true);
+                return o.getSelectedOption(p, guide).orElse(defaultValue);
             }
         }
 
-        return true;
+        return defaultValue;
     }
```
TheBusyBiscuit, line 278: The use of T and OPTION here as generic types is kinda unsettling to me.

I would rather suggest to use T and V or K and V (for type/key and value) and putting the options type first.
This seems confusing right now.

There should also be an annotation on the method to indicate whether this is nullable or not.

TheBusyBiscuit (changes requested):
Thanks for addressing our requested changes and sorry that the review is taking a bit longer.
As you can see, there are 17 open pull requests and we are having a hard time to keep up.

I addressed two more things I would probably like to be changed before merging this.
But otherwise it is looking good to me.

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/FireworksOption.java
```diff
@@ -29,6 +32,12 @@ public NamespacedKey getKey() {
         if (SlimefunPlugin.getRegistry().isResearchFireworkEnabled()) {
             boolean enabled = getSelectedOption(p, guide).orElse(true);
             ItemStack item = new CustomItem(Material.FIREWORK_ROCKET, "&bFireworks: &" + (enabled ? "aYes" : "4No"), "", "&7You can now toggle whether you", "&7will be presented with a big firework", "&7upon researching an item.", "", "&7\u21E8 &eClick to " + (enabled ? "disable" : "enable") + " your fireworks");
+            if (enabled) {
+                item.addUnsafeEnchantment(Enchantment.MENDING, 1);
+                ItemMeta meta = item.getItemMeta();
+                meta.addItemFlags(ItemFlag.HIDE_ENCHANTS);
+                item.setItemMeta(meta);
+            }
```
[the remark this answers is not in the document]
J3fftw1: iirc we dont have 1, but would maybe be better to fit in cscorelib 2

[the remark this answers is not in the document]
TheBusyBiscuit: Yeah, don't use a fake enchantment...
I personally think using a seperate item type would be best though.

TheBusyBiscuit (approved):
LGTM I think

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/options/LearningAnimationOption.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Nonnull
+    @Override
+    public NamespacedKey getKey() {
+        return new NamespacedKey(SlimefunPlugin.instance(), "research_learning_animation");
+    }
+
+    @Nonnull
+    @Override
+    public Optional<ItemStack> getDisplayItem(Player p, ItemStack guide) {
```
WalshyDev, line 41:
```suggestion
    public Optional<ItemStack> getDisplayItem(@Nonnull Player p, @Nonnull ItemStack guide) {
```

WalshyDev (changes requested):

WalshyDev (approved):
LGTM

TheBusyBiscuit (approved):

PR #2985, opened 2021-04-23
Added the option to disable the Per World Configuration file generation + a fast way to disabled the worlds

xSavior-of-God:
## Description
I have implemented an option to disable the creation of files (therefore the management of items) of each world, in order to avoid large quantities in case you have more worlds ... I have also implemented a blacklist of the worlds (in order to fill the gap of these files)

```Yaml
disable-per-world-config: false
disabled-worlds:
  - 'THIS_WORLD_IS_DISABLED'
```

## Proposed changes
The idea is to avoid the increase of files in the folder in case of server based on multi world (SkyBlock type with SlimeWorldManager)

## Checklist


TheBusyBiscuit (changes requested):
Just a heads up:

We are currently quite busy and there are 12 other Pull Requests in this repository alone.
This isn't high up on our priority list at the moment, so it will still take a while until we get to reviewing this pull request.

We will also still need to debate about this internally about potential side effects that may come from such a change.

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/PerWorldSettingsService.java
```diff
@@ -196,20 +196,27 @@ public void save(@Nonnull World world) {
             return optional.get();
         } else {
             Set<String> items = new LinkedHashSet<>();
-            Config config = getConfig(world);
 
-            config.getConfiguration().options().header("This file is used to disable certain items in a particular world.\nYou can set any item to 'false' to disable it in the world '" + name + "'.\nYou can also disable an entire addon from Slimefun by setting the respective\nvalue of 'enabled' for that Addon.\n\nItems which are disabled in this world will not show up in the Slimefun Guide.\nYou won't be able to use these items either. Using them will result in a warning message.");
-            config.getConfiguration().options().copyHeader(true);
-            config.setDefaultValue("enabled", true);
+            if (!SlimefunPlugin.getCfg().getBoolean("options.disable-per-world-config")) {
```
TheBusyBiscuit, line 200: There is no need to read this boolean everytime a World is loaded.
The value should probably be stored as a field and read via the `load()` method.

src/main/resources/config.yml
```diff
@@ -18,6 +18,9 @@ options:
   log-duplicate-block-entries: true
   burn-players-when-radioactive: true
   drop-excess-sf-give-items: false
+  disable-per-world-config: false
+  disabled-worlds:
+    - 'THIS_WORLD_IS_DISABLED'
```
TheBusyBiscuit, line 23: I am not a big fan of the `disabled-worlds` list right here as it is kinda redundant with what World Settings try to achieve which is not really comfortable.

TheBusyBiscuit (changes requested):
Sorry that it has taken us this long to get back to you.
We have had some busy weeks and lots of higher priority pull requests.

But I finally found some time to review this!

src/main/resources/config.yml
xSavior-of-God: I put it there in case some one decides not to use the base system, where the world file is no longer present

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/PerWorldSettingsService.java
xSavior-of-God: ok

TheBusyBiscuit:
Closing this as stale

PR #2992, opened 2021-04-26
Wood Cutter Android fix

### LINKED ISSUE #2964: Android Woodcutter will stuck on chopping tree if the inventory is full until the fuel ran out.

Android Woodcutter will stuck on chopping tree if the inventory is full until the fuel ran out.

1. Place Woodcutter Android
2. Plant a tree
3. Wait until the inventory is full (or just 'force' fill it)
4. See that loop on breaking the tree until the fuel ran out

<details>
<summary>Video</summary>

https://user-images.githubusercontent.com/34732414/115132937-b032ef80-a02e-11eb-9af4-01b17804d2c0.mp4

</details>

I guess like a Farmer Android just skip past that block or kinda like Android Miner just mine the blocks but the items go missing.

## :scroll: Server Log

There is no relevant server logs.

There is no error logs.

image

Mooy1:
## Description
The wood cutter android won't stop at the chopTree instruction forever, it will keep looping.

## Proposed changes
Changed WoodcutterAndroid#chopTree to return void

Fixes #2964 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
@@ -706,10 +706,9 @@ public class ProgrammableAndroid extends SlimefunItem implements InventoryBlock,
                     BlockStorage.addBlockInfo(b, "index", String.valueOf(0));
                     break;
                 case CHOP_TREE:
-                    // We only move to the next step if we finished chopping wood
-                    if (chopTree(b, inv, face)) {
-                        BlockStorage.addBlockInfo(b, "index", String.valueOf(index));
-                    }
+                    // Move to the next step and try to chop the tree
+                    BlockStorage.addBlockInfo(b, "index", String.valueOf(index));
+                    chopTree(b, inv, face);
                     break;
                 default:
                     // We set the index here in advance to fix moving android issues
@@ -894,7 +893,7 @@ public class ProgrammableAndroid extends SlimefunItem implements InventoryBlock,
         throw new UnsupportedOperationException("Non-mining Android tried to mine!");
     }
 
-    protected boolean chopTree(Block b, BlockMenu menu, BlockFace face) {
+    protected void chopTree(Block b, BlockMenu menu, BlockFace face) {
         throw new UnsupportedOperationException("Non-woodcutter Android tried to chop a Tree!");
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -894,7 +893,7 @@ protected void moveAndDig(Block b, BlockMenu menu, BlockFace face, Block block)
         throw new UnsupportedOperationException("Non-mining Android tried to mine!");
     }
 
-    protected boolean chopTree(Block b, BlockMenu menu, BlockFace face) {
+    protected void chopTree(Block b, BlockMenu menu, BlockFace face) {
```
TheBusyBiscuit, line 896: Do not change the return type on this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -706,10 +706,9 @@ private void executeInstruction(Instruction instruction, Block b, BlockMenu inv,
                     BlockStorage.addBlockInfo(b, "index", String.valueOf(0));
                     break;
                 case CHOP_TREE:
-                    // We only move to the next step if we finished chopping wood
-                    if (chopTree(b, inv, face)) {
-                        BlockStorage.addBlockInfo(b, "index", String.valueOf(index));
-                    }
+                    // Move to the next step and try to chop the tree
+                    BlockStorage.addBlockInfo(b, "index", String.valueOf(index));
+                    chopTree(b, inv, face);
```
TheBusyBiscuit, line 711: This will make the android only mine a single block...

TheBusyBiscuit (changes requested):
Changes the behaviour drastically and not really what was reported.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
Mooy1: It previously only called it once

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
Mooy1: The return value is useless though, if the issue is to be solved.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
TheBusyBiscuit: You are advancing the `index` on every iteration now.
Previously the `index` was only incremented if the woodcutting operation has finished.

TheBusyBiscuit (changes requested):
Now the boolean value isn't even used.
And it will return true everytime, making this entirely pointless.

I feel like this is just a troll PR at this point as it completely breaks android behaviour.

PR #2997, opened 2021-04-27
Trivial fixes/improvements to TickerTask

md5sha256:
## Description

This pr addresses some of the issues in the TickerTask implementation

## Proposed changes

- Fix some incorrect acquiring of the read lock when performing writes
- changes the task execution logic to handle tasks using a shallow-cloned copy of the queue.
-  synchronization blocks have been added to a few methods. 

## Checklist


TheBusyBiscuit (approved):
Seems sensible to me!

md5sha256:
There's a few issues that I'd like this PR to resolve in one go, specifically regarding the excessive try-catch and the policy regarding catching `throwable`s when executing ("foreign") scheduled tasks. Unless there are objections, I'll leave this as a draft for now.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -300,12 +300,13 @@ private void tick() {
...
-            // Only continue if the Chunk is actually loaded
-            if (chunk.isLoaded()) {
-                for (Location l : locations) {
-                    tickLocation(tickers, l);
-                }
+            for (Location l : locations) {
+                tickLocation(tickers, l);
             }
         } catch (ArrayIndexOutOfBoundsException | NumberFormatException x) {
             SlimefunPlugin.logger().log(Level.SEVERE, x, () -> "An Exception has occurred while trying to resolve Chunk: " + chunk);
```
WalshyDev, line 312: this try-catch still seems to be redundant and just unnecessary overhead. Did you find a reason to keep it?

TheBusyBiscuit:
I am looking into that now.
The `ArrayIndexOutOfBoundsException | NumberFormatException` try/catch should definitely be removed.
We don't rely on Number parsing or String splitting anymore.

You can remove that if you want or I do it inside #2996 myself (which might result in a merge conflict here)

md5sha256:
I'll just rebase, seems like that should be taken care of elsewhere.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -246,39 +246,39 @@ private void tick() {
...
-                Iterator<Map.Entry<Location, Boolean>> removals = deletionQueue.entrySet().iterator();
-
-                while (removals.hasNext()) {
-                    Map.Entry<Location, Boolean> entry = removals.next();
-                    BlockStorage.deleteLocationInfoUnsafely(entry.getKey(), entry.getValue());
-                    removals.remove();
-                }
+                // Copy the queue, this way we don't have keep the queue locked whilst removing
+                deletionQueueCopy = new HashMap<>(deletionQueue);
+                deletionQueue.clear();
```
WalshyDev, line 253: This is a good change, we shouldn't keep this locked for the (heavy as fuck) BlockStorage modifications.

md5sha256:
I'm not entirely convinced that we need the lock in a lot of places. There seems to a "better" way of doing things. Additionally, there is the issue of the following not being properly read-only. (The values in the map are mutable!)
```java
/**
     * This method returns a <strong>read-only</strong> {@link Map}
     * representation of every {@link ChunkPosition} and its corresponding
     * {@link Set} of ticking {@link Location Locations}.
     * 
     * This does include any {@link Location} from an unloaded {@link Chunk} too!
     * 
     * @return A {@link Map} representation of all ticking {@link Location Locations}
     */
    @Nonnull
    public Map<ChunkPosition, Set<Location>> getLocations() {
        return Collections.unmodifiableMap(tickingLocations);
    }
    ```

File: src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java [hunk 3]
```diff
@@ -246,39 +246,39 @@ private void tick() {
...
             } finally {
                 queueLock.writeLock().unlock();
             }
 
+            for (Map.Entry<Location, Boolean> entry : deletionQueueCopy.entrySet()) {
+                BlockStorage.deleteLocationInfoUnsafely(entry.getKey(), entry.getValue());
```
WalshyDev, line 259: I feel like we need a different method for bulk deletions... I'll need to check the code on `deleteLocationInfoUnsafely` to see if that's a good idea

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -408,12 +409,12 @@ public void queueMove(Location from, Location to) {
...
 
         // This collection is iterated over in a different thread. Need to lock it.
-        queueLock.readLock().lock();
+        queueLock.writeLock().lock();
 
         try {
             movingQueue.put(from, to);
         } finally {
-            queueLock.readLock().unlock();
+            queueLock.writeLock().unlock();
```
WalshyDev, line 417: I feel like we should make a LockableCollection so that we can make sure things are locked properly at all times.

`map.put(key, value)`
will actually run like

```java
lock.writeLock().lock();
try {
    realCollection.put(key, value);
} finally {
    lock.writeLock().unlock();
}
```

md5sha256: Isn't that what `Collections#synchronizedX` does?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
TheBusyBiscuit: There is no reason to keep it, see my comment.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
md5sha256: fair enough

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
md5sha256: As mentioned above, I'm leaving that to be done by Biscuit in #2996

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
WalshyDev: Yeah Collections does have synchronized things. I do prefer synchronized here since we're locking in a very simple way but I didn't want to change everything right away 😆

md5sha256: ah fair enough

md5sha256:
I'm concerned that `start` is allowed to happen multiple times, is this behaviour correct?

TheBusyBiscuit:
No. It should only be called once.

md5sha256:
I guess I'll add a check inside start to ensure the task can't be started multiple times.

md5sha256:
Although that being said, why is the ticker task a `Runnable`, it feels much more like an executor than a runnable to me.

TheBusyBiscuit:
I actually already changed that myself about two minutes ago XD

TheBusyBiscuit:
Alright, removed the unnecessary `try/catch`, removed the implementation of `Runnable` and refactored some other parts within #2996

TheBusyBiscuit (approved):
Still looking good to me.

svr333 (approved):
Looks fine

PR #2998, opened 2021-04-28
Make energy net use long

Mooy1:
## Description
This will allow energy networks to have a total production/capactity of more than 2.1 billion. Previously it would reset to 0 or go negative.

image

## Proposed changes
Changed a fews ints to longs

x

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
@@ -226,7 +226,7 @@ public class EnergyNet extends Network implements HologramOwner {
                         loc.getBlock().setType(Material.LAVA);
                         loc.getWorld().createExplosion(loc, 0F, false);
                     });
-                } else {
+                } else if (supply != MAX_ENERGY) {
                     supply += provider.getGeneratedOutput(loc, data);
 
                     if (provider.isChargeable()) {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
@@ -131,9 +131,9 @@ public class EnergyNet extends Network implements HologramOwner {
         if (connectorNodes.isEmpty() && terminusNodes.isEmpty()) {
             updateHologram(b, "&4No Energy Network found");
         } else {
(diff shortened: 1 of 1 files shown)


WalshyDev:
This is delaying the problem rather than solving it. It's also causing new issues since you cast to int in other places.

This should instead just be Math.max'd.

TheBusyBiscuit:
Agreed, this just delays the underlying issue.
Also I think the energy net limit should maybe become a config option.

Mooy1:
Why a config option? No other energy related stuff is in configs.

WalshyDev:
Configuration is good and it's very easy to just add here. Also allows for users to make it harder and limit it to like 10k.

Mooy1:
> This is delaying the problem rather than solving it. It's also causing new issues since you cast to int in other places.
> 
> This should instead just be Math.max'd.

Casting to int is fine since anywhere that its cast is for sure under Integer.MAX_VALUE, and what should i Math.max?

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
```diff
@@ -231,7 +227,15 @@ private int tickAllGenerators(@Nonnull LongConsumer timings) {
...
-                    supply += energy;
+                    supply += provider.getGeneratedOutput(loc, data);
+
+                    if (provider.isChargeable()) {
+                        supply += provider.getCharge(loc, data);
+                    }
+                    
+                    if (supply < 0 || supply > MAX_ENERGY) {
+                        supply = MAX_ENERGY;
+                    }
```
TheBusyBiscuit, line 238: Congratulations, you broke reactors.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
Mooy1: forgot that getGeneratedOutput ran other code xD

TheBusyBiscuit (changes requested):
I don't know what @WalshyDev might think about this.
But I would be in favour of discarding this pull request for now.

This is no simple change that has massive consequences for the entire codebase and merging something like this without giving it proper thought would have severe side effects for the whole project.
Thanks for bringing this up but I would be against merging it as of now, but we may revisit this at a later time, though it should probably be done properly on the entire energy code then.

TheBusyBiscuit:
Closed, as discussed.

PR #3002, opened 2021-04-30
New form-based Issue templates

WalshyDev:
I have no idea if you can still have the old markdown format alongside this, I would assume so. This just brings over the current structure without any real changes.

Full documentation can be found here: https://gh-community.github.io/issue-template-feedback/structured/

Changelogs are also on that same site

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,4 +1,4 @@
-name: Bug Report (NEW)
+name: Bug Report (Beta)
 about: Report a Bug or an Issue with Slimefun 4.
 title: ''
 labels: ['\U0001F3AF Needs testing', '\U0001F41E Bug Report']
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -42,6 +42,18 @@ body:
   description: |
     Check the folder /plugins/Slimefun/error-reports/ and upload any files inside that folder.
     You can also post these files via https://pastebin.com/
+- type: dropdown
+  validations:
+    required: true
+  description: 'Please select the Minecraft version of the server'
+  attributes:
+    label: ':video_game: Minecraft Version'
+    options:
+      #- 1.17.x
+      - 1.16.x
+      - 1.15.x
+      - 1.14.x
+      - Older
 - type: textarea
   label: ':compass: Environment'
   validations:
@@ -54,5 +66,4 @@ body:
     If your issue is related to other plugins, make sure to include the versions of these plugins too!
   value: |
     - Server software:
-    - Minecraft version:
     - Slimefun version:
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,4 +1,4 @@
-name: Bug Report
+name: Bug Report (NEW)
 about: Report a Bug or an Issue with Slimefun 4.
 title: ''
 labels: ['\U0001F3AF Needs testing', '\U0001F41E Bug Report']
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+  validations:
+    required: true
+  description: |
+    Any issue without the exact version numbers will be closed!
+    "latest" IS NOT A VERSION NUMBER. -->
+    We recommend running "/sf versions" and showing us a screenshot of that.
+    Make sure that the screenshot covers the entire output of that command.
+    If your issue is related to other plugins, make sure to include the versions of these plugins too!
+  value: |
+    - Server software:
```
TheBusyBiscuit, line 56: I think we should use a dropdown for this. Mainly we are interested in whether they use Spigot, Paper or some other fork.

I would suggest changing the Server Software to a dedicated dropdown with the following options:
* Spigot
* Paper
* Tuinity
* Yatopia
* Purpur
* Airplane
* Other

(or just Spigot, Paper and other.)

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+  description: |
+    Check the folder /plugins/Slimefun/error-reports/ and upload any files inside that folder.
+    You can also post these files via https://pastebin.com/
+- type: textarea
+  label: ':compass: Environment'
+  validations:
+    required: true
+  description: |
+    Any issue without the exact version numbers will be closed!
+    "latest" IS NOT A VERSION NUMBER. -->
```
TheBusyBiscuit, line 51:
```suggestion
    "latest" IS NOT A VERSION NUMBER.
```

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+    We recommend running "/sf versions" and showing us a screenshot of that.
+    Make sure that the screenshot covers the entire output of that command.
+    If your issue is related to other plugins, make sure to include the versions of these plugins too!
+  value: |
+    - Server software:
+    - Minecraft version:
```
TheBusyBiscuit, line 57: Same here, *could* do a dropdown for Minecraft version too.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,58 @@
...
+    required: true
+  description: |
+    What were you expecting to happen?
+    What do you think would have been the correct behaviour?
+- type: textarea
+  label: ':scroll: Server Log'
+  description: |
+    Take a look at your Server Log and post any errors you can find via https://pastebin.com/
+    If you are unsure about it, post your full log, you can find it under /logs/latest.log
+- type: textarea
```
TheBusyBiscuit, line 40: Is there a way to do a file upload here?

TheBusyBiscuit (changes requested):

.github/ISSUE_TEMPLATE/bug-report.yml
WalshyDev: All textarea support file upload by default

TheBusyBiscuit: Alright, I would really need to see this in action first I guess

.github/ISSUE_TEMPLATE/bug-report.yml
WalshyDev: Yeah, the only thing is how much do we care about the version of it.

TheBusyBiscuit: Ultimately I would still prefer a screenshot of `/sf versions` I guess

WalshyDev: I wish we could accept only a file/link. I may have to send that as a suggestion

TheBusyBiscuit: Also is there any way the `validations` field can be used for... proper input validation? :eyes:

WalshyDev: One nice thing is that we can accept input which is automatically highlighted. So for the logs we can make the text there highlighted with like java or something

TheBusyBiscuit: what?

TheBusyBiscuit (approved):
Let's try it out I guess

.github/ISSUE_TEMPLATE/bug-report.yml
WalshyDev: So, we have an input for log file and if they decide to send text they can and it'll turn into a code block

"Post your log file"
```java
enabling slimefun
exception NullPointerException
  at some.class.Slimefun:10
```

would be nice if we ever wanted configs too, we could highlight with YAML

TheBusyBiscuit:
It doesn't seem to work.

WalshyDev:
I can see, let me shoot an email off

PR #3015, opened 2021-05-03
Improved memory usage of cargo/energy networks and fixed #3013

### LINKED ISSUE #3013: Energy Network isn't 'updating' after some generator/capasitor/machine being broken with explosive pickaxe.
### ❗ Checklist

### 📍 Description

Energy Network isn't 'updating' after some generator/capasitor/machine being broken with explosive pickaxe.
It will 'refresh' to 0J and regenerate again.

### 📑 Reproduction Steps

1. Place an Energy Regulator 
2. Place a generator (Solar Generator on video)
3. Place a machine or capasitor (Capasitor on video)
4. Break them indirectly with explosive pickaxe
5. It will 'refresh' and still works

### 💡 Expected Behavior

The energy network should be updated and remove the removed generator/capasitor/machine that are broken.

### 📷 Screenshots / Videos

https://cdn.discordapp.com/attachments/565569196218384385/838736977624563722/2021-05-03_18-19-54.mp4

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.16.x

### ⭐ Slimefun version

image

### 🧭 Other plugins

_No response_

TheBusyBiscuit:
## Description

This pull request not only fixes #3013 by listening to the `ExplosiveToolBreakBlocksEvent`.
But due to the increase in `Network#connectsTo` calls, I decided to speed up the network architecture and use a less memory-heavy way of handling the locations.
This should also speed up  `Network#connectsTo` as the `.hashCode()` and `.equals()` methods for longs are much faster than the ones for `Location`.

## Proposed changes

* Use `BlockPosition`s long to reduce memory usage
* Enforce world equality. A network cannot be transdimensional

Resolves #3013 

## Checklist


Sfiguz7 (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/NetworkListener.java
```diff
@@ -39,4 +45,12 @@ public void onBlockBreak(BlockBreakEvent e) {
     public void onBlockPlace(BlockPlaceEvent e) {
         manager.updateAllNetworks(e.getBlock().getLocation());
```
WalshyDev, line 46: unrelated to this PR but surely this can be optimised. Placing will only ever affect the network if it's a SF block. Looking at the method it looks pretty intense.

Why does the manager even have a List rather than a Set ughh, I really hate the way this is done currently.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/NetworkListener.java
```diff
@@ -39,4 +45,12 @@ public void onBlockBreak(BlockBreakEvent e) {
...
     }
+
+    @EventHandler(priority = EventPriority.MONITOR, ignoreCancelled = true)
+    public void onExplosiveToolUse(ExplosiveToolBreakBlocksEvent e) {
+        // Fixes #3013 - Also update networks when using an explosive tool
+        for (Block b : e.getAdditionalBlocks()) {
+            manager.updateAllNetworks(b.getLocation());
```
WalshyDev, line 53: is it quicker to run this for all blocks rather than checking if it's an SF block first... I can't see checking if it's an SF block being more intensive honestly, that's a Map lookup whereas this loops a list and does all sorts of crap

TheBusyBiscuit:
@WalshyDev I implemented the performance improvement you suggested

WalshyDev (approved):
LGTM - This should definitely help quite a bit

PR #3042, opened 2021-05-11
Android and coal generators to burn bamboo to generate electricity

3093FengMing:
## Description

I want Android and coal generators to burn bamboo to generate electricity
## Checklist


variananora:
I think you are making a suggestion....
Suggestion goes on to the Discord server... (Links in the https://github.com/Slimefun/Slimefun4/)

PR #3043, opened 2021-05-11
Change the way the TickerTask tracks Blocks

md5sha256:
## Description

As per a discussion in the SF discord, this PR removes the use of explicit locks used in TickerTask. 
I'm not a massive fan of the implementation of `TickingBlocks` as it is right now. Currently its just been wrapped (mutex) so it isn't that much better for performance than it was before (might even be worse on that end). However, we now map the world to a long as opposed to `ChunkPosition --> Set<Location>` which means the memory footprint is considerably lower. 

## Proposed changes

- Add `TickingBlocks`, a class that serves as an accessor/wrapper for a `Map<World, Set<Long>>`. In other words, a `BlockPositionSet`. 
- Removes the public getters for the `tickingLocations`
- Swap out the `Lock` + `Map<ChunkPosition, Set<Location>>` for `TickingBlocks`
- Added relevant docs

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
@@ -121,10 +121,8 @@ final class TickerCollection {
     private static @Nonnull Set<Location> convertToLocation(@Nonnull World world, @Nonnull Collection<Long> raw) {
         Set<Location> locations = new HashSet<>(raw.size());
         for (long compressed : raw) {
-            int x = (int) (compressed >> 38);
-            int y = (int) (compressed & 0XFFF);
-            int z = (int) (compressed << 26 >> 38);
-            locations.add(new Location(world, x, y, z));
+            int[] decompressed = BlockPosition.decompress(compressed);
+            locations.add(new Location(world, decompressed[0], decompressed[1], decompressed[2]));
         }
         return locations;
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
@@ -1,8 +1,8 @@
 package io.github.thebusybiscuit.slimefun4.implementation.tasks;
 
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
+
 import io.github.thebusybiscuit.cscorelib2.blocks.BlockPosition;
+import org.bukkit.Bukkit;
 import org.bukkit.Location;
 import org.bukkit.World;
 
@@ -13,7 +13,8 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Memory efficient, synchronized collection of {@link BlockPosition BlockPositions} used by the
@@ -24,7 +25,7 @@ import java.util.WeakHashMap;
 @ThreadSafe
 final class TickerCollection {
 
-    private final Cache<World, Set<Long>> blocks = CacheBuilder.newBuilder().weakKeys().build();
+    private final Map<UUID, Set<Long>> blocks = new ConcurrentHashMap<>();
 
     /**
      * Get the compressed {@link BlockPosition BlockPositions} mapped to a given {@link World}.
@@ -35,7 +36,7 @@ final class TickerCollection {
      *
      */
     public @Nonnull Set<Long> getBlocks(@Nonnull World world) {
-        return blocks.asMap().computeIfAbsent(world, unused -> Collections.synchronizedSet(new HashSet<>()));
+        return blocks.computeIfAbsent(world.getUID(), unused -> Collections.synchronizedSet(new HashSet<>()));
     }
 
     /**
@@ -46,8 +47,10 @@ final class TickerCollection {
      */
     public @Nonnull Set<Location> getBlocksAsLocations() {
         Set<Location> locations = new HashSet<>();
-        for (Map.Entry<World, Set<Long>> entry : blocks.asMap().entrySet()) {
-            World world = entry.getKey();
+        blocks.keySet().removeIf(uid -> Bukkit.getWorld(uid) == null);
+        for (Map.Entry<UUID, Set<Long>> entry : blocks.entrySet()) {
+            World world = Bukkit.getWorld(entry.getKey());
+            // assert world != null
             Set<Location> converted = convertToLocation(world, entry.getValue());
             locations.addAll(converted);
         }
@@ -82,13 +85,13 @@ final class TickerCollection {
      */
     public void removeBlock(@Nonnull BlockPosition block) {
         World world = block.getWorld();
-        Collection<Long> positions = blocks.getIfPresent(world);
+        Collection<Long> positions = blocks.get(world.getUID());
         if (positions == null) {
             return;
         }
         positions.remove(block.getPosition());
         if (positions.isEmpty()) {
-            blocks.invalidate(world);
+            blocks.remove(world.getUID());
         }
     }
 
@@ -107,14 +110,14 @@ final class TickerCollection {
      * @param key The world to clear the key for
      */
     public void clear(@Nonnull World key) {
-        blocks.invalidate(key);
+        blocks.remove(key.getUID());
     }
 
     /**
      * Clear all entries from this instance
      */
     public void clear() {
-        blocks.invalidateAll();
+        blocks.clear();
     }
 
     private static @Nonnull Set<Location> convertToLocation(@Nonnull World world, @Nonnull Collection<Long> raw) {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
@@ -1,6 +1,5 @@
 package io.github.thebusybiscuit.slimefun4.implementation.tasks;
 
-
 import io.github.thebusybiscuit.cscorelib2.blocks.BlockPosition;
 import org.bukkit.Bukkit;
 import org.bukkit.Location;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
@@ -49,17 +49,15 @@ final class TickerCollection {
         blocks.keySet().removeIf(uid -> Bukkit.getWorld(uid) == null);
         for (Map.Entry<UUID, Set<Long>> entry : blocks.entrySet()) {
             World world = Bukkit.getWorld(entry.getKey());
-            // assert world != null
-            Set<Location> converted = convertToLocation(world, entry.getValue());
-            locations.addAll(converted);
+            locations.addAll(convertToLocation(world, entry.getValue()));
         }
         return locations;
     }
 
     /**
-     * Add a location to this ticker collection, duplicates will be ignored.
+     * Add a {@link BlockPosition} to this ticker collection, duplicates will be ignored.
      *
-     * @param block The block to add
+     * @param block The {@link BlockPosition} to add
      * @see #addBlock(Location)
      */
     public void addBlock(@Nonnull BlockPosition block) {
@@ -67,9 +65,9 @@ final class TickerCollection {
     }
 
     /**
-     * Add a location to this ticker collection, duplicates will be ignored.
+     * Add a {@link Location} to this ticker collection, duplicates will be ignored.
      *
-     * @param location The location of the block to add
+     * @param location The {@link Location} of the block to add
      * @see #addBlock(BlockPosition)
      */
     public void addBlock(@Nonnull Location location) {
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickingBlocks.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.Set;
+import java.util.WeakHashMap;
+
+/**
+ * Memory efficient, synchronized collection of {@link BlockPosition BlockPositions} used by the
+ * {@link TickerTask}. All methods are thread-safe unless explicitly stated otherwise.
+ *
+ * @author md5sha256
+ */
+final class TickingBlocks {
```
TheBusyBiscuit, line 21: Not a fan of this name tbh.
Perhaps `TickerCollection`?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickingBlocks.java
```diff
@@ -0,0 +1,165 @@
...
+     * Get the compressed {@link BlockPosition BlockPositions} mapped to a given {@link World}.
+     * The returned {@link Set} is thread-safe. Changes to set will be reflected in this instance.
+     *
+     * @param world The world
+     * @return Returns a {@link Set} of {@link Long Longs} which are compressed {@link BlockPosition BlockPositions}
+     *
+     * @see #getBlocksCopy()
+     */
+    public @Nonnull Set<Long> getBlocks(@Nonnull World world) {
+        return blocks.computeIfAbsent(world, (unused) -> Collections.synchronizedSet(new HashSet<>()));
```
TheBusyBiscuit, line 35: Don't put braces around a single variable in a lambda expression
```suggestion
        return blocks.computeIfAbsent(world, x -> Collections.synchronizedSet(new HashSet<>()));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickingBlocks.java
```diff
@@ -0,0 +1,165 @@
...
+        }
+        positions.remove(block.getPosition());
+        if (positions.isEmpty()) {
+            blocks.remove(world);
+        }
+    }
+
+    public void removeBlock(@Nonnull Location location) {
+        removeBlock(new BlockPosition(location));
+    }
```
TheBusyBiscuit, line 125: sum javadocs? :D

(No necessity, just nice to have)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickingBlocks.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    public @Nonnull Set<BlockPosition> getBlocksCopy() {
+        Set<BlockPosition> positions = new HashSet<>();
+        for (Map.Entry<World, Set<Long>> entry : blocks.entrySet()) {
+            World world = entry.getKey();
+            Set<BlockPosition> converted = convertToBlockPosition(world, entry.getValue());
+            positions.addAll(converted);
+        }
+        return positions;
+    }
```
TheBusyBiscuit, line 69: These methods seem to be unused and are public, yet this class isn't even public and the TickerTask provides no getter for this object anyway.

TheBusyBiscuit (changes requested):

md5sha256:
I've resolved the comments. That being said I've also changed the implementation of `TickingCollection` to use a guava `Cache` as that offers better concurrent performance whilst also supporting weak keys. Alongside biscuit, could @WalshyDev also have a look through?

TheBusyBiscuit:
> Alongside biscuit, could @WalshyDev also have a look through?

I am not enough :FeelsBadMan:

md5sha256:
> > Alongside biscuit, could @WalshyDev also have a look through?
> 
> I am not enough :FeelsBadMan:

Sorry :P Just thought Walshy's experience with the guava Cache might make a difference since we've discussed how to implement this for a bit back in April and early May

md5sha256:
👀

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Set;
+import java.util.WeakHashMap;
+
+/**
+ * Memory efficient, synchronized collection of {@link BlockPosition BlockPositions} used by the
+ * {@link TickerTask}. All methods are thread-safe unless explicitly stated otherwise.
+ *
+ * @author md5sha256
+ */
+final class TickerCollection {
```
WalshyDev, line 23: We should annotate with @ThreadSafe

WalshyDev (commented):
Looking good. Has this been stress tested?

(also apologies, didn't realise this was waiting on me)

md5sha256:
No, it has not. I can start working on a JMH bench. I just needed an opinion on the use of the Guava Cache as I feel like I'm abusing it here.

WalshyDev:
Hmmm,Guava's Cache is made and optimised for automatic eviction (usually size or time). We don't really use any of the Cache's features at all nor use it for the intended purpose. I feel like we're better off just using ConcurrentHashMap.

The only feature/benefit we get from Cache is the weakKeys. Considering the compute and retrieval of things isn't expensive and memory is more valuable to MC servers than usual production software I think we stick to native Java ConcurrentHashMap.

md5sha256:
Noted. So should I wrap all keys with a weak reference?

WalshyDev:
Probably best yeah

md5sha256:
I guess I have to stick with holding the UUID instance of the world instead?

WalshyDev:
Yeah UUID would be better.

md5sha256:
Also, by stress test do you mean a unit test that stresses the collection to check for concurrency errors or a benchmark?

WalshyDev:
> Also, by stress test do you mean a unit test that stresses the collection to check for concurrency errors or a benchmark?

Making sure it can handle concurrency correctly.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
```diff
@@ -0,0 +1,134 @@
+package io.github.thebusybiscuit.slimefun4.implementation.tasks;
+
+
```
WalshyDev, line 3: remove double new line

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -75,10 +66,11 @@
      */
     private static final long MAX_WAIT_TIME = 1_500_000L;
 
+
```
WalshyDev, line 69: double new line

md5sha256:
Should be good EXCEPT for https://discord.com/channels/565557184348422174/573872826327629844/856183240736899104. 
The tests currently pass as I've modified the test to not fail (I changed it to assert against a fresh `Set` from `#getBlocks` instead of `compressedPositions`. The stress test passes; however, I'm all eyes and ears for whether the stress test needs more work. 
Tested on an Intel i7-7700 (4 cores, 8 threads).

md5sha256:
Also, I was forced to merge in `master` because I couldn't compile the new spelling tests as the localisation stuff changed. Oh well.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
```diff
@@ -0,0 +1,133 @@
...
+            return;
+        }
+        positions.remove(block.getPosition());
+        if (positions.isEmpty()) {
+            blocks.remove(world.getUID());
+        }
+    }
+
+    /**
+     * Remove a location from this ticker collection.
```
WalshyDev, line 98:
```suggestion
     * Remove a {@link Location} from this ticker collection.
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @param location The location of the block to remove
```
WalshyDev, line 100:
```suggestion
     * @param location The {@link Location} of the block to remove
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
```diff
@@ -0,0 +1,133 @@
...
+     * @see #removeBlock(BlockPosition)
+     */
+    public void removeBlock(@Nonnull Location location) {
+        removeBlock(new BlockPosition(location));
+    }
+
+    /**
+     * Clear all the blocks for a given {@link World}
+     * @param key The world to clear the key for
```
WalshyDev, line 109:
```suggestion
     * Clear all the blocks for a given {@link World}
     *
     * @param key The {@link World} to remove blocks from
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -101,17 +92,19 @@
...
     private final ExecutorService asyncExecutor = Executors.newSingleThreadExecutor(threadFactory);
 
+
     /**
-     * Making collections concurrent doesn't make their iterators thread safe.
+     *  These are "Queues" of blocks that need to be removed or moved.
+     *  <p>
+     *  Whilst the {@link ConcurrentHashMap} implementation is thread-safe,
+     *  the {@link Location Locations} are not. Therefore, the internal copy of the locations
+     *  should be treated as <strong>read only</strong> and never exposed externally.
```
WalshyDev, line 106: 2 spaces at the start here
```suggestion
     * These are "Queues" of blocks that need to be removed or moved.
     * <p>
     * Whilst the {@link ConcurrentHashMap} implementation is thread-safe,
     * the {@link Location Locations} are not. Therefore, the internal copy of the locations
     * should be treated as <strong>read only</strong> and never exposed externally.
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -338,22 +320,21 @@ private void processAsyncTasks() {
...
-                if (item != null && item.getBlockTicker() != null) {
-                    try {
-                        runTask(l, data, item);
-                        tickers.add(item.getBlockTicker());
-                    } catch (Exception x) {
-                        // Catch and report any errors caused by this ticker.
-                        reportErrors(l, item, x);
-                    }
+    /**
+     * Executes all tasks in a given {@link Collection} of locations
```
WalshyDev, line 324:
```suggestion
     * Executes all tasks in a given {@link Collection} of {@link Location locations}
```

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
```diff
@@ -0,0 +1,133 @@
...
+     * Get a copy of all the blocks stored in the form of {@link BlockPosition BlockPositions}
+     * The returned {@link Set} is not guaranteed to be thread-safe.
+     *
+     * @return Returns a {@link Set} of {@link BlockPosition BlockPositions}
+     */
+    public @Nonnull Set<Location> getBlocksAsLocations() {
+        Set<Location> locations = new HashSet<>();
+        blocks.keySet().removeIf(uid -> Bukkit.getWorld(uid) == null);
+        for (Map.Entry<UUID, Set<Long>> entry : blocks.entrySet()) {
+            World world = Bukkit.getWorld(entry.getKey());
```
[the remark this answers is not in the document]
md5sha256: Won't that throw a concurrent modification exception, when removing keys whilst iterating over the entryset?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private static @Nonnull Set<Location> convertToLocation(@Nonnull World world, @Nonnull Collection<Long> raw) {
+        Set<Location> locations = new HashSet<>(raw.size());
+        for (long compressed : raw) {
+            int x = (int) (compressed >> 38);
+            int y = (int) (compressed & 0XFFF);
+            int z = (int) (compressed << 26 >> 38);
+            locations.add(new Location(world, x, y, z));
+        }
```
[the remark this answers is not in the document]
md5sha256: I'll PR that to cscl2 and wait for a bump.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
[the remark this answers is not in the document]
WalshyDev: Nope, it's thread-safe.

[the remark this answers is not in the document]
md5sha256: I guess, but won't that also cause the original iterator the for-loop is using to have undertiministic behaviour?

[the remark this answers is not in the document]
WalshyDev: Hmmm yeah fair enough.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerCollection.java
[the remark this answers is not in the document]
md5sha256: Righty then, now waiting on https://github.com/TheBusyBiscuit/CS-CoreLib2/pull/194

WalshyDev (approved):
LGTM

TheBusyBiscuit (approved):
What Walshy said 🤷🏻

PR #3044, opened 2021-05-11
Add PortableTeleporter

martinbrom:
## Description

I got a request for this item from a few members of our Minecraft community and when I saw the suggestion on SF discord I decided to implement it. Some people may argue that this is too overpowered, but I think that the cost of the item coupled with the fact that most servers use another way of teleportation anyway makes this item okay to add.

## Proposed changes

Add new rechargeable item - PortableTeleporter
Add TeleportationManager::openTeleporterGUI overload without network complexity parameter, which then gets calculated from the given ownerUUID

Discord suggestion 1229

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/teleporter/PortableTeleporter.java
```diff
@@ -0,0 +1,54 @@
...
+    @Nonnull
+    @Override
+    public ItemUseHandler getItemHandler() {
+        return e -> {
+            ItemStack item = e.getItem();
+            e.cancel();
+
+            if (removeItemCharge(item, COST)) {
+                SlimefunPlugin.getGPSNetwork().getTeleportationManager().openTeleporterGUI(
+                        e.getPlayer(), e.getPlayer().getUniqueId(), e.getPlayer().getLocation().getBlock().getRelative(BlockFace.DOWN));
```
TheBusyBiscuit, line 45: Please make a Player variable here. This calls `e.getPlayer()` waayyyyy too often.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/teleporter/PortableTeleporter.java
```diff
@@ -0,0 +1,54 @@
...
+ * This item allows a {@link Player} to access and teleport to his waypoints
+ * from anywhere.
+ *
+ * @author martinbrom
+ *
+ * @see Teleporter
+ */
+public class PortableTeleporter extends SimpleSlimefunItem<ItemUseHandler> implements Rechargeable {
+
+    private static final float COST = 2F;
```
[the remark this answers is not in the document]
martinbrom: I added a new `FloatRangeSetting` but there are two more options with their potential drawbacks:
- Use `DoubleRangeSetting` instead but we could get overflow if the max value is specified wrongly
- Use some sort of generic numerical ItemSetting but that would maybe introduce errors in precision when comparing the boxed types and also would get rid of the implicit null-safety because there are no primitive generics yet

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/settings/FloatRangeSetting.java
```diff
@@ -0,0 +1,63 @@
...
+ * This variation of {@link ItemSetting} allows you to define an {@link Float} range
+ * and enforces this range using the {@link #validateInput(Float)} method.
+ * 
+ * @author TheBusyBiscuit
+ *
+ * @see ItemSetting
+ * @see IntRangeSetting
+ * @see DoubleRangeSetting
+ */
+public class FloatRangeSetting extends ItemSetting<Float> {
```
TheBusyBiscuit, line 20: I do not think that Floats are properly supported here as the Config tends to auto-interpret them as doubles most of the time.
Either way, this is a potential room for error. So I suggested you to simply use an integer here.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/settings/FloatRangeSetting.java
martinbrom: Sure, that is fine as well

WalshyDev (dismissed):
I think the default cost is a bit too low still but LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/teleporter/PortableTeleporter.java
```diff
@@ -0,0 +1,61 @@
...
+ * from anywhere.
+ *
+ * @author martinbrom
+ *
+ * @see Teleporter
+ */
+public class PortableTeleporter extends SimpleSlimefunItem<ItemUseHandler> implements Rechargeable {
+
+    private static final int CAPACITY = 30;
+    private static final int DEFAULT_COST = 5;
```
TheBusyBiscuit, line 31: I think something like cost: 10J, capacity 50J sounds more appropriate.

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
LGTM

svr333 (approved):
Looks good

PR #3048, opened 2021-05-12
Add the item to the research message

WalshyDev:
## Description
Very simple change, it adds the item to the not researched message. This way people know what they actually need to research. This should significantly reduce the number of people coming to Discord asking why they have the message.

**Note:** I had to delete `messages.yml` in `plugins/Slimefun/messages.yml` - Do we have a good way around this currently?

## Proposed changes
`&4You do not have enough knowledge to understand this` -> `&4You do not have enough knowledge to understand this. You need to research %item%`

N/A

## Checklist

### DIFF
--- a/src/main/resources/languages/en/messages.yml
+++ b/src/main/resources/languages/en/messages.yml
@@ -139,7 +139,7 @@ guide:
       translator: '&9Translator'
 
 messages:
-  not-researched: '&4You do not have enough knowledge to understand this'
+  not-researched: '&4You do not have enough knowledge to understand this. You need to research %item%'
   not-enough-xp: '&4You do not have enough XP to unlock this'
   unlocked: '&bYou have unlocked &7"%research%"'
   only-players: '&4This command is only for Players'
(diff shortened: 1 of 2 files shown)


src/main/resources/languages/en/messages.yml
```diff
@@ -139,7 +139,7 @@ guide:
       translator: '&9Translator'
 
 messages:
-  not-researched: '&4You do not have enough knowledge to understand this'
+  not-researched: '&4You do not have enough knowledge to understand this. You need to research %item%'
```
TheBusyBiscuit, line 142:
```suggestion
  not-researched: '&4You do not have enough knowledge to understand this. &cYou will need to unlock &7"%item%&7"'
```

TheBusyBiscuit (changes requested):
Quotation marks and coloring improvements

src/main/resources/languages/en/messages.yml
WalshyDev: Do we maybe want to reset before %item% just so if the item has no colour code it is white?

TheBusyBiscuit: It will be gray, that's why I added &7 there

WalshyDev: exactly why I'm saying do we want to reset. So quotes are gray but item is true color no matter what (even if white)

TheBusyBiscuit: You mean reset as in `ChatColor.stripColor` and not `ChatColor.RESET` then?

TheBusyBiscuit (approved):

TheBusyBiscuit (approved):

Sfiguz7 (approved):

PR #3052, opened 2021-05-13
Fix WoodcutterAndroid getting stuck while chopping a tree with full inventory

### LINKED ISSUE #2964: Android Woodcutter will stuck on chopping tree if the inventory is full until the fuel ran out.

martinbrom:
## Description

Currently, players need to come up with clever ways to handle the overflow of logs from the Woodcutter Android otherwise the android will inevitably get stuck in a loop and will need to be restarted manually by adding fuel to it when it runs out.

## Proposed changes

If the android's inventory is full while in the middle of chopping a tree, the logs will 'get lost', meaning chopped successfully but not added to the android's inventory. The miner android does kind of the same thing.

Resolves #2964 

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,7 +67,6 @@
 * Fixed #3041
 * Fixed #3036
 * Possibly fixed #2927
-* Fixed #2964
 
 ## Release Candidate 22 (18 Apr 2021)
 https://thebusybiscuit.github.io/builds/TheBusyBiscuit/Slimefun4/stable/#22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@
 * Fixed #3041
 * Fixed #3036
 * Possibly fixed #2927
+* Fixed #2964
 
 ## Release Candidate 22 (18 Apr 2021)
 https://thebusybiscuit.github.io/builds/TheBusyBiscuit/Slimefun4/stable/#22
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/WoodcutterAndroid.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/WoodcutterAndroid.java
@@ -69,13 +69,14 @@ public class WoodcutterAndroid extends ProgrammableAndroid {
 
         if (menu.fits(drop, getOutputSlots())) {
             menu.pushItem(drop, getOutputSlots());
-            log.getWorld().playEffect(log.getLocation(), Effect.STEP_SOUND, log.getType());
+        }
 
-            if (log.getY() == android.getRelative(face).getY()) {
-                replant(log);
-            } else {
-                log.setType(Material.AIR);
-            }
+        log.getWorld().playEffect(log.getLocation(), Effect.STEP_SOUND, log.getType());
+
+        if (log.getY() == android.getRelative(face).getY()) {
+            replant(log);
+        } else {
+            log.setType(Material.AIR);
         }
     }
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -67,6 +67,7 @@
 * Fixed #3041
 * Fixed #3036
 * Possibly fixed #2927
+* Fixed #2964
```
TheBusyBiscuit, line 70: For future pull requests: Rather leave the changelog maintenance up to us as it can lead to very much avoidable and unnecessary merge conflicts between pull requests if everyone updates it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/WoodcutterAndroid.java
```diff
@@ -69,13 +69,14 @@ private void breakLog(Block log, Block android, BlockMenu menu, BlockFace face)
 
         if (menu.fits(drop, getOutputSlots())) {
             menu.pushItem(drop, getOutputSlots());
-            log.getWorld().playEffect(log.getLocation(), Effect.STEP_SOUND, log.getType());
+        }
```
TheBusyBiscuit, line 72: I think the `fits` check can be completely removed here, `pushItem()` will just not do anything if it cannot be fit. So this may be a bit redundant.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/WoodcutterAndroid.java
```diff
@@ -69,13 +69,14 @@ private void breakLog(Block log, Block android, BlockMenu menu, BlockFace face)
...
-            } else {
-                log.setType(Material.AIR);
-            }
+        log.getWorld().playEffect(log.getLocation(), Effect.STEP_SOUND, log.getType());
+
+        if (log.getY() == android.getRelative(face).getY()) {
+            replant(log);
+        } else {
+            log.setType(Material.AIR);
         }
```
TheBusyBiscuit, line 80: This will void a bit of logs sadly but better than the alternative I guess.

Some comments here would be nice.

TheBusyBiscuit (changes requested):

CHANGELOG.md
martinbrom: I removed it, I just assumed I always forgot to add it and you had to do it afterwards :smile:

Boomer-1:
Just a quick note on this, I don't have this problem because the android is programmed correctly. If you only select once to chop down the tree, then pivot to the fuel and item interfaces, then go back to the tree, it works perfectly. That setup is best utilized on a 2x2 tree. If they are using smaller trees they have to program the android correctly to maximize the time the android spends mining, but could still pivot after each one. It's no different than every android. You simply need to program it right. Just my thoughts.

martinbrom:
> Just a quick note on this, I don't have this problem because the android is programmed correctly. If you only select once to chop down the tree, then pivot to the fuel and item interfaces, then go back to the tree, it works perfectly. That setup is best utilized on a 2x2 tree. If they are using smaller trees they have to program the android correctly to maximize the time the android spends mining, but could still pivot after each one. It's no different than every android. You simply need to program it right. Just my thoughts.

In the case the system gets filled with logs there is no "correctly programmed" way to handle that because even if you only put 1 chopping instruction the android still has nowhere to push its items if it gets filled in the middle of the chopping process.

variananora:
> Just a quick note on this, I don't have this problem because the android is programmed correctly. If you only select once to chop down the tree, then pivot to the fuel and item interfaces, then go back to the tree, it works perfectly. That setup is best utilized on a 2x2 tree. If they are using smaller trees they have to program the android correctly to maximize the time the android spends mining, but could still pivot after each one. It's no different than every android. You simply need to program it right. Just my thoughts.

i mean if you have a bottleneck somewhere like lagging cargo or smth it can make a woodcutter android stuck.

TheBusyBiscuit (approved):
LGTM

Sfiguz7 (approved):
LGTM

PR #3054, opened 2021-05-13
Cleanup SlimefunLocalization

### LINKED ISSUE #3053: Recipe type has 'null' as a display name
### ❗ Checklist

### 📍 Description

When using any other language than English, some of the recipes in item detail (in the survival guide) have `null` as the item display name. The problem is probably caused by the recipe not looking for a fallback if the translation does not exist.

### 📑 Reproduction Steps

1. Open Slimefun guide
2. Change language to a different language than English (reproducible with Czech)
3. Open the recipe for Damascus Steel Helmet (for example)
4. See the display name of the anvil is `null`

### 💡 Expected Behavior

The display name is not `null`. It is either the correct translation (unavailable at the time of writing) or the fallback English name.

### 📷 Screenshots / Videos

Screenshot from 2021-05-13 17-35-10

### 📜 Server Log

No errors

No errors

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.16.x

### ⭐ Slimefun version

Screenshot from 2021-05-13 17-37-06

### 🧭 Other plugins

_No response_

martinbrom:
Use fallbacks for every method in SlimefunLocalization
Resolves #3053

## Description

The SlimefunLocatization has quite a bit of duplicated code and some of the methods do not use fallbacks at all

## Proposed changes

Use common methods to load translations (with fallback to English)

Resolves #3053

## Personal Checklist (todo)
- [ ] Nullability in the getString method
- [ ] Unit tests?

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/localization/SlimefunLocalization.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/localization/SlimefunLocalization.java
@@ -1,11 +1,11 @@
 package io.github.thebusybiscuit.slimefun4.core.services.localization;
 
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit (changes requested):
For me this is a hard pass.
- overuse of ternary operators
- slow performance via use of Streams (and that in such a vital and frequently called part of the code)
- Unnecessary `namespacedKeyToString` method

I appreciate that you took the time to fix this but the refactoring part does not seem like an improvement to me.

PR #3058, opened 2021-05-14
Fixed menu loading issues and probably improved BlockStorage load times

### LINKED ISSUE #2979: Multiple Block Menus are created for the same Location in some worlds
On Server startup, BlockMenus created in custom/other plugins's worlds are created twice. The first one is the one that actually gets stored, the second one can cause problems for code which overrides newInstance() in BlockMenuPreset. This doesn't cause any problems for Slimefun from what i've tested, but can and has affected Addons which use newInstance().

I found it by adding this method to InventoryBlock to see when it occurs

@Override
public void newInstance(@Nonnull BlockMenu menu, @Nonnull Block b) {
    System.out.println("NEW INSTANCE " + menu + menu.getLocation());
}

This results in the following with 1 Auto Enchanter in world, and 1 Auto Enchanter in bskyblock_world_nether

[00:20:24 INFO]: NEW INSTANCE me.mrCookieSlime.Slimefun.api.inventory.BlockMenu@5dbf1608Location{world=CraftWorld{name=world},x=-63989.0,y=68.0,z=70355.0,pitch=0.0,yaw=0.0}
[00:20:24 INFO]: NEW INSTANCE me.mrCookieSlime.Slimefun.api.inventory.BlockMenu@78c30b07Location{world=CraftWorld{name=bskyblock_world_nether},x=-7997.0,y=69.0,z=8802.0,pitch=0.0,yaw=0.0}
[00:20:24 INFO]: NEW INSTANCE me.mrCookieSlime.Slimefun.api.inventory.BlockMenu@39877c72Location{world=CraftWorld{name=bskyblock_world_nether},x=-7997.0,y=69.0,z=8802.0,pitch=0.0,yaw=0.0}

The second and third logs are the same location, but different BlockMenus.

Somewhere something in Slimefun is creating a BlockMenu for a location that already had one.

1. Add a custom world, for example with a skyblock plugin
2. Place a block with an inventory
3. Restart server

There should only be 1 BlockMenu created per location on startup

## :scroll: Server Log
https://pastebin.com/ML10Td6h

none

 - Server software: Purpur
 - Minecraft version: 1.16.5
 - Slimefun version: Dev 881 + 1 debug method

Mooy1:
## Description
Stored inventories were getting loaded multiple times, overwriting the previous ones each time. Universal menus and stored chunks aswell. This is intended to be a temporary fix while BlockStorage is hopefully deleted and rewritten lol.

## Proposed changes
Added 2 checks to see if chunks and UniversalBlockMenus have already been loaded, and a check for world when loading BlockMenus.

Fixes #2979

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
@@ -242,7 +242,7 @@ public class BlockStorage {
 
     private void loadInventories() {
         for (File file : new File("data-storage/Slimefun/stored-inventories").listFiles()) {
-            if (file.getName().endsWith(".sfi")) {
+            if (file.getName().startsWith(this.world.getName()) && file.getName().endsWith(".sfi")) {
                 try {
                     Location l = deserializeLocation(file.getName().replace(".sfi", ""));
                     
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
@@ -59,6 +59,8 @@ public class BlockStorage {
     private final Map<String, Config> blocksCache = new ConcurrentHashMap<>();
 
     private static int chunkChanges = 0;
+    private static boolean chunksLoaded = false;
+    private static boolean universalInventoriesLoaded = false;
 
     private int changes = 0;
     private AtomicBoolean isMarkedForRemoval = new AtomicBoolean(false);
@@ -214,6 +216,12 @@ public class BlockStorage {
     }
 
     private void loadChunks() {
+        if (chunksLoaded) {
+            return;
+        }
+
+        chunksLoaded = true;
+        
         File chunks = new File(PATH_CHUNKS + "chunks.sfc");
 
         if (chunks.exists()) {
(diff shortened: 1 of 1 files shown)


src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
```diff
@@ -234,9 +242,15 @@ private void loadChunks() {
 
     private void loadInventories() {
         for (File file : new File("data-storage/Slimefun/stored-inventories").listFiles()) {
-            if (file.getName().startsWith(world.getName()) && file.getName().endsWith(".sfi")) {
+            if (file.getName().endsWith(".sfi")) {
```
WalshyDev, line 245: Why not just change this to a .equals rather than removing the startsWith and slowing loading down?

WalshyDev (dismissed):

src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
Mooy1: because its in the format world;x;y;z.sfi

WalshyDev: God this stuff is horrible...

Hmm. I need to look into how this stuff is working a bit

TheBusyBiscuit (approved):
Not perfect but that entire class needs to be rewritten anyway...
So we can make this better in the future.

Sfiguz7 (approved):
Agreed with both of you BlockStorage is supposed to be redone but hey small improvements for now work

PR #3059, opened 2021-05-14
Add Rainbow Armor

martinbrom:
## Description

We already have rainbow glass and other rainbow blocks. Now you can not only place the rainbow and *taste the rainbow*™, but you can also **BECOME** the rainbow!

Discord suggestion 0806

## Proposed changes

Add `RainbowArmorTask` because the existing one does not need to tick as often
Add `RainbowLeather` (as an ancient altar recipe) to craft the rainbow armor
Extract common functionality from `ArmorTask` and `RainbowArmorTask` to `AbstractArmorTask`

Discord suggestion 0806

## Personal Checklist (todo)
- [x] Rearrange the colors
- [x] Add Rainbow Leather recipe and change the crafting material for the armor recipes
- [x] When missing the research you get an error message for missing knowledge when shifting ??

## Checklist


J3fftw1:
Ohno rainbow puke

martinbrom:
> rainbow puke
Your discord suggestion was so nicely written it practically forced me to add this :smile:

J3fftw1:
> > rainbow puke
> 
> Your discord suggestion was so nicely written it practically forced me to add this 😄

Yea we were talking about it a long time ago in general and then i wrote up a little meme suggestion didnt expect it to become reality

martinbrom:
I feel like it is ready for review at this point, I poked around and fixed a few issues that popped up even though the code is a little overengineered at the moment.
I can also see one way to improve performance a bit, and that would be to check firstmost if the armor is a piece of leather armor in the `RainbowArmorTask` case, but that would need to be added in the `AbstractArmorTask` and would lead to the code being even more overengineered.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
```diff
@@ -2586,6 +2603,11 @@ public int getCapacity() {
             .register(plugin);
         }
 
+        new SlimefunItem(categories.magicalResources, SlimefunItems.RAINBOW_LEATHER, RecipeType.ANCIENT_ALTAR,
+        new ItemStack[] {new ItemStack(Material.RABBIT_HIDE), new ItemStack(Material.LEATHER), new ItemStack(Material.RABBIT_HIDE), new ItemStack(Material.LEATHER), SlimefunItems.RAINBOW_RUNE, new ItemStack(Material.LEATHER), new ItemStack(Material.RABBIT_HIDE), new ItemStack(Material.LEATHER), new ItemStack(Material.RABBIT_HIDE)},
+        new SlimefunItemStack(SlimefunItems.RAINBOW_LEATHER, 8))
```
TheBusyBiscuit, line 2608: I think this is way too cheap.

```suggestion
        new ItemStack[] {new ItemStack(Material.RABBIT_HIDE), new ItemStack(Material.EMERALD), new ItemStack(Material.RABBIT_HIDE), new ItemStack(Material.LEATHER), SlimefunItems.RAINBOW_RUNE, new ItemStack(Material.LEATHER), new ItemStack(Material.RABBIT_HIDE), new ItemStack(Material.LEATHER), new ItemStack(Material.RABBIT_HIDE)},
        new SlimefunItemStack(SlimefunItems.RAINBOW_LEATHER, 4))
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/RainbowArmorTask.java
```diff
@@ -0,0 +1,57 @@
...
+        iterator = new LoopIterator<>(Arrays.stream(DYE_COLORS)
+                .map(DyeColor::getColor)
+                .collect(Collectors.toList()));
+    }
+
+    @Override
+    @ParametersAreNonnullByDefault
+    protected void handleArmorPiece(Player p, SlimefunArmorPiece sfArmorPiece, ItemStack armorPiece) {
+        if (sfArmorPiece instanceof RainbowArmorPiece) {
+            LeatherArmorMeta meta = (LeatherArmorMeta) armorPiece.getItemMeta();
```
TheBusyBiscuit, line 42: Should always check `hasItemMeta()` before accessing this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -99,14 +99,7 @@ public void run() {
...
             // Fixes #2576 - Remove any deleted instances of BlockStorage
-            Iterator<BlockStorage> worlds = SlimefunPlugin.getRegistry().getWorlds().values().iterator();
-            while (worlds.hasNext()) {
-                BlockStorage storage = worlds.next();
-
-                if (storage.isMarkedForRemoval()) {
-                    worlds.remove();
-                }
-            }
+            SlimefunPlugin.getRegistry().getWorlds().values().removeIf(BlockStorage::isMarkedForRemoval);
```
TheBusyBiscuit, line 102: This is already part of another PR but I don't mind it being here already.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/RainbowArmorTask.java
```diff
@@ -0,0 +1,57 @@
...
+import io.github.thebusybiscuit.slimefun4.implementation.items.armor.SlimefunArmorPiece;
+
+/**
+ * The {@link RainbowArmorTask} is responsible for handling the change in color of any Rainbow Armor piece.
+ *
+ * @author martinbrom
+ */
+public class RainbowArmorTask extends AbstractArmorTask {
+
+    private static final DyeColor[] DYE_COLORS = { DyeColor.RED, DyeColor.ORANGE, DyeColor.YELLOW, DyeColor.LIME, DyeColor.LIGHT_BLUE, DyeColor.PURPLE, DyeColor.MAGENTA };
```
TheBusyBiscuit, line 26: This line is way too long and this does not need to be static.

TheBusyBiscuit (changes requested):
Personally, I am not the biggest fan of Rainbow Armor.

I removed these four items from Slimefun for good reasons as they were quite a performance drag in the end.
But ehh, if people really want them to be added back into the game...

(Though I do like your idea of Rainbow leather)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/ResearchSetup.java
```diff
@@ -287,6 +287,7 @@ public static void setupResearches() {
         register("improved_generators", 275, "Improved Generators", 24, SlimefunItems.COAL_GENERATOR_2, SlimefunItems.LAVA_GENERATOR_2);
         register("ingredients_and_cheese", 276, "Slimefun Cuisine", 5, SlimefunItems.SALT, SlimefunItems.WHEAT_FLOUR, SlimefunItems.HEAVY_CREAM, SlimefunItems.CHEESE, SlimefunItems.BUTTER);
         register("medium_tier_auto_enchanting", 277, "Fast Automatic Enchanting and Disenchanting", 48, SlimefunItems.AUTO_ENCHANTER_2, SlimefunItems.AUTO_DISENCHANTER_2);
+        register("rainbow_armor", 279, "Wear the Rainbow", 22, SlimefunItems.RAINBOW_HELMET, SlimefunItems.RAINBOW_CHESTPLATE, SlimefunItems.RAINBOW_LEGGINGS, SlimefunItems.RAINBOW_BOOTS);
```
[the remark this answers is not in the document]
martinbrom: > Even if you reserved that id for a different PR

Yes, I did (the portable teleporter which is close to being merged).
How else are we supposed to number these?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/RainbowArmorTask.java
martinbrom: > does not need to be static

What do you mean by that?
Do you think the colors should be configurable or? I might be missing what's wrong with the constant being static.

martinbrom:
This is not forgotten, I'm just waiting for a reply from Cookie

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/ResearchSetup.java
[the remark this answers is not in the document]
TheBusyBiscuit: Next available.
Because if we somehow decided against the merging of the Portable Teleporter PR for example, there would have been a gap.
And when we did merge it, this results in a merge conflict either way. So no matter what, there will be a secondary action required if you skip a number here.
By choosing the next available id at the time, you'd only have to take actions in the event of a successful merge, otherwise not.
This is more efficient.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/RainbowArmorTask.java
TheBusyBiscuit: Static variables can cause memory leaks due to the way Bukkit is written.
This is completely harmless for primitives but not so much applies for complex types, like this array.

While, this is not the end of the world and there are many places where this trade off is a-okay...
I think here it serves no real benefit.
There is no concrete reasoning as to why the potential memory leak is outweighed by any benefits here.
And more importantly, this is not exactly future-proof, if we intend to add different kinds of Rainbow armor (perhaps a christmas armor for example), we would need to make this non-constant yet again.

Furthermore, this is a singleton class right now.
There won't be another instance of this in the JVM at runtime, which makes the static-ness of this field not provide any of the memory improvements that a static variable would yield.

All in all, there is really no need for this to be static.

WalshyDev:
Resolve conflicts please ❤️

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/RainbowArmorTask.java
TheBusyBiscuit: You can have multiple constructors though.
The colors could be passed via the constructor and then we could have a default constructor which calls `this(...);`

Mooy1: howabout a RainbowArmorPiece#getColors then

martinbrom: Check latest commit 😉

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/RainbowArmorPiece.java
```diff
@@ -0,0 +1,64 @@
...
+            throw new IllegalArgumentException("Rainbow armor needs to be a leather armor piece!");
+        }
+
+        int length = dyeColors.length;
+        Color[] colors = new Color[length];
+        for (int i = 0; i < length; i++) {
+            colors[i] = dyeColors[i].getColor();
+        }
+
+        this.colors = colors;
```
TheBusyBiscuit, line 51: This should probably go inside a private method and not directly into the constructor.
Can also use `Arrays.stream().map()` here, it is only run on startup so that is fine.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
```diff
@@ -430,6 +433,35 @@ public static void setup(@Nonnull SlimefunPlugin plugin) {
...
+        DyeColor[] RAINBOW_ARMOR_COLORS = {
+                DyeColor.RED,
+                DyeColor.ORANGE,
+                DyeColor.YELLOW,
+                DyeColor.LIME,
+                DyeColor.LIGHT_BLUE,
+                DyeColor.PURPLE,
+                DyeColor.MAGENTA
+        };
+        new RainbowArmorPiece(categories.magicalArmor, SlimefunItems.RAINBOW_HELMET, RecipeType.ARMOR_FORGE,
```
TheBusyBiscuit, line 445:
```suggestion
        };

        new RainbowArmorPiece(categories.magicalArmor, SlimefunItems.RAINBOW_HELMET, RecipeType.ARMOR_FORGE,
```

TheBusyBiscuit (changes requested):

martinbrom:
I will profile it with Spark once I figure out how to do it :smile:

martinbrom:
https://spark.lucko.me/uS2xR1BMTB not even visible - 1 player, once per 3 seconds
https://spark.lucko.me/yqmdKDkYMp 10 players, once per second

for comparison, the cargo network has 1 input and output node + 1 access terminal (and tick impact is almost the same in the second test)
I'd say the performance is just fine

md5sha256:
> Personally, I am not the biggest fan of Rainbow Armor.
> 
> I removed these four items from Slimefun for good reasons as they were quite a performance drag in the end.
> But ehh, if people really want them to be added back into the game...
> 
> (Though I do like your idea of Rainbow leather)

Just curious, where was the original bottleneck/performance hit?

TheBusyBiscuit:
> Just curious, where was the original bottleneck/performance hit?

:LUL: Believe me, you better not wanna know.
https://github.com/Slimefun/Slimefun2/blob/master/src/me/mrCookieSlime/Slimefun/RunningTimer.java#L193-L236

TheBusyBiscuit (approved):
Similarities to #3131 hence why I will merge the two pull requests into a staging branch to work out compatibility inbetween.

PR #3062, opened 2021-05-15
Add a configurable step delay to the Ancient Altar

martinbrom:
## Description

Mass crafting rune in base Slimefun is currently very tedious. This gives the server owners the option to configure the speed themselves.

## Proposed changes

Add `IntRangeSetting` to the `AncientAltar`
Remove check for zero-tick step delay, now it is possible to craft instantly

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/altar/AncientAltar.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/altar/AncientAltar.java
@@ -3,6 +3,9 @@ package io.github.thebusybiscuit.slimefun4.implementation.items.altar;
 import java.util.ArrayList;
 import java.util.List;
 
+import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNonnullByDefault;
+
 import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.api.events.AncientAltarCraftEvent;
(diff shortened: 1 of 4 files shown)


WalshyDev:
What's the behaviour if this is 0? Also, if this was let's say 50 would there be any indication it is still doing it's thing?

martinbrom:
> What's the behaviour if this is 0?

https://youtu.be/VIfwxpv6piU

> Also, if this was let's say 50 would there be any indication it is still doing it's thing?

https://youtu.be/86PFF-uuRxY

The particles are shown 4 times (at equal intervals) per item consumed. I guess a better way to show them would be sth like once per item consumed and then once per 2 seconds, that way there wouldn't be so many when the delay is zero and so little when the delay is large. The question then would be if it's worth it and how to do it :smile:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/altar/AncientAltar.java
```diff
@@ -28,28 +30,27 @@
  */
 public class AncientAltar extends SlimefunItem {
 
-    private final int speed;
+    private static final int DEFAULT_STEP_DELAY = 8;
```
TheBusyBiscuit, line 33: javadocs on this would be nice.
Can also add yourself as an `@author` to this class.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/altar/AncientAltar.java
```diff
@@ -28,28 +30,27 @@
...
     private final List<AltarRecipe> recipes = new ArrayList<>();
 
-    public AncientAltar(Category category, int speed, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
-        super(category, item, recipeType, recipe);
+    private final ItemSetting<Integer> stepDelay = new IntRangeSetting(this, "step-delay", 0, DEFAULT_STEP_DELAY, Integer.MAX_VALUE);
 
-        if (speed < 1) {
-            throw new IllegalArgumentException("The speed must be at least 1.");
-        }
+    public AncientAltar(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
```
TheBusyBiscuit, line 39: Please add a "parameters are nonnull by default" annotation to this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/altar/AncientAltar.java
```diff
@@ -28,28 +30,27 @@
...
-     * The default is 8 ticks.
+     * This number determines how many ticks happen in between a step in the ritual animation.
+     * The default is {@value #DEFAULT_STEP_DELAY} ticks.
      * 
      * @return The speed of this {@link AncientAltar}
      */
     public int getSpeed() {
-        return speed;
+        return stepDelay.getValue();
     }
```
TheBusyBiscuit, line 54: The name "speed" here is no longer accurate.

TheBusyBiscuit (changes requested):
The original plan was to have different tiers of altars.
This kinda ruins that of course but I can live with it for now, more configuration is always good.

TheBusyBiscuit (approved):
LGTM

Sfiguz7 (approved):
LGTM

PR #3068, opened 2021-05-18
Optimized use of ItemStackWrapper in the CargoNetworkTask

md5sha256:
## Description

Following some investigation in the SF discord after someone raised the question of CargoNet being slow, I have 
identified a few things: 
- The use of `ItemStackWrapper` is sometimes misused accidentally. (I.e creating an `ItemStackWrapper` of an `ItemStackWrapper`
- The `ItemStackWrapper` is not used efficiently (I.e creating multiple copies of the wrapper when performing `CargoNetTask#routeItems`: methods discard their own wrappers immediately after use instead of passing it on where plausible. 

This PR should not yet be merged; however, it can be reviewed. 

I am unsure how to proceed in benchmarking the proposed changes. There has been a ~ 9-10% performance improvement just from using `ItemStackWrapper#ofItem` instead of the constructor; however, this PR covers more changes than just that. The other changes (namely the extra parameters for `CargoUtils` methods) have not been benchmarked whatsoever. I am open to ideas as to how those changes should be tested although my inclination is that the changes are better served tested on a production environment with real-world setups.

## Proposed changes

- Deprecate the public constructor `ItemStackWrapper(ItemStack)` in favour of static initialization methods (`ItemstackWrapper#forceWrap(ItemStack)` and `ItemStackWrapper#ofItem(ItemStack)`. The latter will only create a new wrapper if the target `ItemStack` is not already a wrapper. 
- Add an `ItemStackWrapper` parameter to some methods in `CargoUtils` which will allow for `ItemStackWrapper`s to be reused between method calls. 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
@@ -6,20 +6,13 @@ import io.github.thebusybiscuit.slimefun4.utils.itemstack.ItemStackWrapper;
 
 class ItemStackAndInteger {
 
-    private final ItemStack item;
+    private ItemStack item;
     private ItemStackWrapper wrapper;
     private int number;
 
     ItemStackAndInteger(ItemStack item, int amount) {
         this.number = amount;
-        if (item instanceof ItemStackWrapper) {
-            this.item = new ItemStack(item.getType(), item.getAmount());
-            if (item.hasItemMeta()) {
-                this.item.setItemMeta(item.getItemMeta());
-            }
-        } else {
-            this.item = item;
-        }
+        this.item = item;
     }
 
     public int getInt() {
@@ -27,12 +20,13 @@ class ItemStackAndInteger {
     }
 
     public ItemStack getItem() {
+        initializeItem();
         return item;
     }
 
     public ItemStackWrapper getItemStackWrapper() {
         if (wrapper == null && item != null) {
-            wrapper = ItemStackWrapper.ofItem(item);
+            wrapper = ItemStackWrapper.wrap(item);
         }
 
         return wrapper;
@@ -40,6 +34,16 @@ class ItemStackAndInteger {
 
     public void add(int amount) {
         number += amount;
+    }
+
+    private void initializeItem() {
+        if (this.item instanceof ItemStackWrapper) {
+            ItemStack copy = new ItemStack(item.getType(), item.getAmount());
+            if (this.item.hasItemMeta()) {
+                copy.setItemMeta(this.item.getItemMeta());
+            }
+            this.item = copy;
+        }
     }
 
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
@@ -25,7 +25,7 @@ class ItemStackAndInteger {
 
     public ItemStackWrapper getItemStackWrapper() {
         if (wrapper == null && item != null) {
-            wrapper = new ItemStackWrapper(item);
+            wrapper = ItemStackWrapper.ofItem(item);
         }
 
         return wrapper;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
@@ -4,13 +4,15 @@ import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.utils.itemstack.ItemStackWrapper;
 
+import javax.annotation.Nonnull;
+
 class ItemStackAndInteger {
 
     private ItemStack item;
     private ItemStackWrapper wrapper;
     private int number;
 
-    ItemStackAndInteger(ItemStack item, int amount) {
+    ItemStackAndInteger(@Nonnull ItemStack item, int amount) {
         this.number = amount;
         this.item = item;
     }
@@ -19,13 +21,13 @@ class ItemStackAndInteger {
         return number;
     }
 
-    public ItemStack getItem() {
+    public @Nonnull ItemStack getItem() {
         initializeItem();
         return item;
     }
 
-    public ItemStackWrapper getItemStackWrapper() {
-        if (wrapper == null && item != null) {
+    public @Nonnull ItemStackWrapper getItemStackWrapper() {
+        if (wrapper == null) {
             wrapper = ItemStackWrapper.wrap(item);
         }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
@@ -1,5 +1,6 @@
 package io.github.thebusybiscuit.slimefun4.core.networks.cargo;
 
+import org.apache.commons.lang.Validate;
 import org.bukkit.inventory.ItemStack;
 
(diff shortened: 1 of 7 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/itemstack/ItemStackWrapper.java
```diff
@@ -46,6 +59,17 @@ public ItemStackWrapper(@Nonnull Material material) {
...
+    public static @Nonnull ItemStackWrapper forceWrapItem(@Nonnull ItemStack itemStack) {
+        return new ItemStackWrapper(itemStack);
+    }
+
+    public static @Nonnull ItemStackWrapper ofItem(@Nonnull ItemStack itemStack) {
+        if (itemStack instanceof ItemStackWrapper) {
+            return (ItemStackWrapper) itemStack;
+        }
+        return new ItemStackWrapper(itemStack);
+    }
```
TheBusyBiscuit, line 71: First off, this method is super useful and much needed!
Thanks!

Though I have a few remarks on this in its current state:
1. There are similar static methods for arrays and collections, so I believe these methods should all be grouped together instead of having this one so far up.
2. `ItemStackWrapper.wrap(ItemStack)` seems more adequate to me
3. The nonnull attribute here is not enforced, I also think we should allow null values (just return null)
4. documentation :)

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
```diff
@@ -12,7 +12,14 @@
 
     ItemStackAndInteger(ItemStack item, int amount) {
         this.number = amount;
-        this.item = item;
+        if (item instanceof ItemStackWrapper) {
+            this.item = new ItemStack(item.getType(), item.getAmount());
+            if (item.hasItemMeta()) {
+                this.item.setItemMeta(item.getItemMeta());
+            }
+        } else {
```
TheBusyBiscuit, line 20: Not quite sure why this is done here...

TheBusyBiscuit (changes requested):
I haven't looked at the entirety of this yet

src/main/java/io/github/thebusybiscuit/slimefun4/utils/itemstack/ItemStackWrapper.java
md5sha256: Ah, I though the `wrapList` etc was non static for some reason :P. Noted.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
TheBusyBiscuit: Depends, is `getItem()` guaranteed to be called? I would assume so, in which case a lazy computation would be a bit nonsencial. Otherwise I would be completely for the lazy route.

md5sha256: I can't say on the top of my head if it's guaranteed; however, intuitively it seems so given that the wrapper is lazy. I'll have a look at the current usages and make changes if necessary.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/itemstack/ItemStackWrapper.java
md5sha256: > The nonnull attribute here is not enforced, I also think we should allow null values (just return null)
Understood. I've been having an issue where there's a lack of a way to tell IDE/static code analysis of the proper nullability contracts. I.e if I were to do this
```java
@Nullable 
public static ItemStackWrapper#wrap(`@Nullable` ItemStack)`
```
My IDE (intellij) tells me that this: 
```java
// the wrapper variable is marked as nullable
ItemStackWrapper wrapper = ItemStackWrapper#wrap(new ItemStack(Material)
```
Jetbrains annotations has an annotation specifically for this called `@Contract`
which would allow me to annotate something like this to denote that the returned
value is null if, and only if, the parameter is null. Is there an equivalent I could use for Slimefun?
```java
@Nullable
@Contract("null -> null") 
public static ItemStackWrapper#wrap(`@Nullable` ItemStack)`
```

md5sha256: Should be resolved; however, I'm leaving this comment unresolved so that the annotations stuff can be addressed if necessary.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
```diff
@@ -20,12 +20,13 @@ public int getInt() {
     }
 
     public ItemStack getItem() {
+        initializeItem();
         return item;
     }
 
     public ItemStackWrapper getItemStackWrapper() {
```
TheBusyBiscuit, line 27: Both of these methods should probably be marked as `@Nullable`

src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/utils/TestItemStackWrapper.java
```diff
@@ -69,6 +69,22 @@ void testImmutability() {
         Assertions.assertThrows(UnsupportedOperationException.class, () -> wrapper.equals(wrapper));
     }
 
+    @Test
+    @DisplayName("Test if the ItemStackWrapper static method constructors are checking for nested wrapping properly")
+    void testWrapperChecking() {
+        Assertions.assertNull(ItemStackWrapper.wrap(null));
+        Assertions.assertNull(ItemStackWrapper.forceWrap(null));
```
TheBusyBiscuit, line 76: These two will fail right?

TheBusyBiscuit (changes requested):
Looks good to me thus far.
Only two remarks

TheBusyBiscuit:
Also this is becoming a pretty useful class, might be worth moving that to `dough` eventually :eyes:

src/test/java/io/github/thebusybiscuit/slimefun4/testing/tests/utils/TestItemStackWrapper.java
md5sha256: Now they will yes, I forgot to fix this :P

md5sha256:
> Also this is becoming a pretty useful class, might be worth moving that to `dough` eventually 👀

Aye

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
md5sha256: They weren't originally. Additionally apparently `getItemStackWrapper` is actually nullable if `item` is null... Although it shouldn't so :P

md5sha256: Actually put it this way. If the `ItemStack` is null then the wrapper will be null so no they aren't both nullable.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemStackAndInteger.java
```diff
@@ -4,13 +4,15 @@
...
+
 class ItemStackAndInteger {
 
-    private final ItemStack item;
+    private ItemStack item;
     private ItemStackWrapper wrapper;
     private int number;
 
-    ItemStackAndInteger(ItemStack item, int amount) {
+    ItemStackAndInteger(@Nonnull ItemStack item, int amount) {
```
TheBusyBiscuit, line 15: This should be enforced with a `Validate.notNull` statement

TheBusyBiscuit (changes requested):
One change left

TheBusyBiscuit (approved):

Sfiguz7 (approved):
LGTM

PR #3080, opened 2021-05-26
Easier SlimefunItem creation with small recipe

Mooy1:
## Description

This will allow slimefun items created with recipes of length less than 9 to be filled with nulls until reaching length 9 rather than being 9 nulls.

## Proposed changes

Rather than setting the recipe to a `new ItemStack[9]` when `length < 9`, set it to `Arrays.copyOf(recipe, 9)`

## Checklist


TheBusyBiscuit (approved):

TheBusyBiscuit (approved):
LGTM

Sfiguz7 (approved):
LGTM

PR #3083, opened 2021-05-28
added time and date to error report

J3fftw1:
## Description

added time and date to the error report, this makes it easier for us to see when this error has been thrown. instead of just blindly trusted the user

## Proposed changes

Added time and date to the error report

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
@@ -89,7 +89,7 @@ public class ErrorReport<T extends Throwable> {
             Date date = new Date();
             SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
 
-            stream.println(formatter.format(date));
+            stream.println("Error Generated: " + formatter.format(new Date()));
             stream.println("Block Info:");
             stream.println("  World: " + l.getWorld().getName());
             stream.println("  X: " + l.getBlockX());
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
@@ -3,9 +3,11 @@ package io.github.thebusybiscuit.slimefun4.api;
 import java.io.File;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
+import java.text.SimpleDateFormat;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Date;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
```diff
@@ -84,6 +86,10 @@ public ErrorReport(T throwable, SlimefunAddon addon, Consumer<PrintStream> print
     @ParametersAreNonnullByDefault
     public ErrorReport(T throwable, Location l, SlimefunItem item) {
         this(throwable, item.getAddon(), stream -> {
+            Date date = new Date();
+            SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
WalshyDev, line 90: This should not be initialised every single error report. Instead make this a global static variable

src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
```diff
@@ -84,6 +86,10 @@ public ErrorReport(T throwable, SlimefunAddon addon, Consumer<PrintStream> print
...
+
+            stream.println(formatter.format(date));
```
WalshyDev, line 92:
```suggestion
            stream.println(formatter.format(new Date()));
```

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
```diff
@@ -3,9 +3,11 @@
 import java.io.File;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
+import java.text.SimpleDateFormat;
 import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Date;
```
TheBusyBiscuit, line 10: unused import

src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
```diff
@@ -84,6 +86,8 @@ public ErrorReport(T throwable, SlimefunAddon addon, Consumer<PrintStream> print
     @ParametersAreNonnullByDefault
     public ErrorReport(T throwable, Location l, SlimefunItem item) {
         this(throwable, item.getAddon(), stream -> {
+
+            stream.println("Error Generated: " + dateFormat.format(LocalDateTime.now()));
```
TheBusyBiscuit, line 90: This should not be location-based but rather added to **every** error-report in my opinion. So please move that down to the part where Java version etc... are added as well

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
The wording could be improved but looks fine to me

WalshyDev (approved):

PR #3097, opened 2021-05-31
Change config for consistency and improvement

WalshyDev:
## Description
This PR aims to just cleanup the config a little to fix inconsistency and a case where issues can easily occur.

### Tick Delay
Slimefun is so much more optimised than it used to be. Insanely so. The 12 ticks has never made sense to me. Regardless though, this 12 tick delay has always caused issues for people due to the fact that all Slimefun lores assume this is 10 ticks. They have production * 2 in the lore.

This changed should cause no real issue due to the fact it's always been changeable anyway. No addon or core Slimefun functionality should be based on this being 12.

### Backwards Compatibility
It has now been 2 years since this was introduced. All servers which didn't have this config option definitely do now. If we look at bStats 80% of servers have this disabled. A lot of those which have it enabled do not actually know about it.

There's a common bug which has occurred from this being true by default. People often copy the data-storage folder when resetting or moving but do NOT copy config.yml due to them not changing it. This results in backward compatibility being enabled. We've seen multiple cases of this in the Discord server which makes me believe a good chunk of that 20% is due to this. 

## Proposed changes
* custom-ticker-delay: 12 -> 10
* backwards-compatibility: true -> false

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
@@ -256,11 +256,8 @@ public final class SlimefunPlugin extends JavaPlugin implements SlimefunAddon {
             PaperLib.suggestPaper(this);
         }
 
-        // Disabling backwards-compatibility for fresh Slimefun installs
+        // If the server has no "data-storage" folder, it's _probably_ a new install. So mark it for metrics.
         if (!new File("data-storage/Slimefun").exists()) {
-            config.setValue("options.backwards-compatibility", false);
-            config.save();
-
             isNewlyInstalled = true;
         }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
@@ -257,9 +257,7 @@ public final class SlimefunPlugin extends JavaPlugin implements SlimefunAddon {
         }
 
         // If the server has no "data-storage" folder, it's _probably_ a new install. So mark it for metrics.
-        if (!new File("data-storage/Slimefun").exists()) {
-            isNewlyInstalled = true;
-        }
+        isNewlyInstalled = !new File("data-storage/Slimefun").exists();
 
         // Creating all necessary Folders
         getLogger().log(Level.INFO, "Creating directories...");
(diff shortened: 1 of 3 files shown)


Sfiguz7 (dismissed):
I agree with both changes, nothing should really happen to already existing servers but it will save both us and less experienced owners / owners who do not look as carefully into Slimefun's config before installing it from small headaches about stopping renaming dupes and it will indeed be more consistent with the 2* for energy consumption rates.

TheBusyBiscuit (changes requested):
This should also be removed then

https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java#L261-L262

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunPlugin.java
```diff
@@ -256,11 +256,8 @@ private void onPluginStart() {
...
         }
 
-        // Disabling backwards-compatibility for fresh Slimefun installs
+        // If the server has no "data-storage" folder, it's _probably_ a new install. So mark it for metrics.
         if (!new File("data-storage/Slimefun").exists()) {
-            config.setValue("options.backwards-compatibility", false);
-            config.save();
-
             isNewlyInstalled = true;
         }
```
TheBusyBiscuit, line 262: Can be simplified to `isNewlyInstalled` = !file.exists();

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
```diff
@@ -305,7 +305,7 @@ public final boolean hasResearch() {
      * @return An {@link Optional} describing the result
      */
     @SuppressWarnings("unchecked")
-    public <T> @Nonnull Optional<ItemSetting<T>> getItemSetting(@Nonnull String key, @Nonnull Class<T> c) {
+    public @Nonnull <T> Optional<ItemSetting<T>> getItemSetting(@Nonnull String key, @Nonnull Class<T> c) {
```
TheBusyBiscuit, line 308: why?

TheBusyBiscuit (changes requested):

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/SlimefunItem.java
WalshyDev: IntelliJ bitches

TheBusyBiscuit (approved):

Sfiguz7 (approved):

PR #3102, opened 2021-06-03
Add collectAverageTimings to SlimefunProfiler

WalshyDev:
## Description
Added SlimefunProfiler#collectAverageTimings which will return the average ms for the given profiler. The supporting values get reset after each call.

I also fixed some validations.

This was done so that I can add this to our metrics.

## Proposed changes
* Created SlimefunProfiler#collectAverageTimings
* Have an AtomicLong and AtomicInteger which log the summed ms timings and ticks passed
* Fixed validations

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
@@ -9,6 +9,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.logging.Level;
 
 import javax.annotation.Nonnull;
@@ -82,6 +83,9 @@ public class SlimefunProfiler {
 
     private final Map<ProfiledBlock, Long> timings = new ConcurrentHashMap<>();
     private final Queue<PerformanceInspector> requests = new ConcurrentLinkedQueue<>();
+
+    private final AtomicLong totalMsTicked = new AtomicLong();
+    private final AtomicInteger ticksPassed = new AtomicInteger();
 
     /**
      * This method terminates the {@link SlimefunProfiler}.
@@ -186,7 +190,7 @@ public class SlimefunProfiler {
         // Wait for all timing results to come in
         while (!isProfiling && queued.get() > 0) {
             try {
-                /**
+                /*
                  * Since we got more than one Thread in our pool,
                  * blocking this one is (hopefully) completely fine
                  */
@@ -216,6 +220,11 @@ public class SlimefunProfiler {
         }
 
         totalElapsedTime = timings.values().stream().mapToLong(Long::longValue).sum();
+
+        // We log how much ms has been ticked and how many ticks have passed.
+        // This is so when bStats requests the average timings, they're super quick to figure out
+        totalMsTicked.addAndGet((long) (totalElapsedTime / 1e6));
+        ticksPassed.incrementAndGet();
 
         if (!requests.isEmpty()) {
             PerformanceSummary summary = new PerformanceSummary(this, totalElapsedTime, timings.size());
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
```diff
@@ -217,6 +221,11 @@ private void finishReport() {
 
         totalElapsedTime = timings.values().stream().mapToLong(Long::longValue).sum();
 
+        // We log how much ms has been ticked and how many ticks have passed.
```
Sfiguz7, line 224:
```suggestion
        // We log how many ms have been ticked and how many ticks have passed.
```

Sfiguz7: Not even sure this would be the best way to express it but eh, I'm not the English one 😛

Sfiguz7 (changes requested):
Already given my opinions and changes in DMs before this was PRed, other than the comment it looks good to me

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
WalshyDev: Improved it further

martinbrom: 'Have been passed' means something different :smile: I think 'have passed' is correct but I don't wanna annoy you too much :smile:

Sfiguz7 (approved):
Looks good now

WalshyDev:
*glares* i hate my life

Sfiguz7 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
```diff
@@ -217,6 +221,11 @@ private void finishReport() {
 
         totalElapsedTime = timings.values().stream().mapToLong(Long::longValue).sum();
 
+        // We log how many milliseconds have been ticked, and how many ticks have passed
+        // This is so when bStats requests the average timings, they're super quick to figure out
```
TheBusyBiscuit, line 225: Should be a multiline comment
/*

*/

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
```diff
@@ -217,6 +221,11 @@ private void finishReport() {
...
+        totalMsTicked.addAndGet((long) (totalElapsedTime / 1e6));
```
TheBusyBiscuit, line 226: `1e6`. God I hate scientific notation.

This is a conversion from nanos to millis, right?
Can't we please use something like `TimeUnit.NANOSECONDS.convert(totalElapsedTime, TimeUnit.MILLISECONDS)` then?

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
```diff
@@ -366,30 +375,37 @@ public int getTickRate() {
...
 
     public String getTime(@Nonnull SlimefunItem item) {
-        Validate.notNull("Cannot get timings for a null SlimefunItem");
+        Validate.notNull(item, "Cannot get timings for a null SlimefunItem");
 
         long time = getByItem().getOrDefault(item.getId(), 0L);
         return NumberUtils.getAsMillis(time);
     }
 
+    public long collectAverageTimings() {
```
TheBusyBiscuit, line 404: 1. javadocs
2. `collect` seems misleading, perhaps `getAndReset`?

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
```diff
@@ -366,30 +375,37 @@ public int getTickRate() {
...
+        double d = (double) totalMsTicked.get() / ticksPassed.get();
```
TheBusyBiscuit, line 405: Why cast this to a double and then to a long??

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/profiler/SlimefunProfiler.java
WalshyDev: javadocs really needed? This is just needed for metrics module

TheBusyBiscuit: It's a public method. And there is no such thing as "too much docs" after all

TheBusyBiscuit (approved):

Sfiguz7 (approved):

PR #3117, opened 2021-06-13
Add LimitedUseItem - base class for items with limited uses

martinbrom:
## Description

This change makes it easier for Slimefun or Slimefun addon devs to create items with a limited number of uses.

## Proposed changes

Add LimitedUseItem - base class for items with limited uses
Update StormStaff to extend LimitedUseItem
Add 'usesLeft' function to LoreBuilder

none

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -19,11 +19,11 @@ import org.bukkit.persistence.PersistentDataType;
 
 import io.github.thebusybiscuit.cscorelib2.chat.ChatColors;
 import io.github.thebusybiscuit.slimefun4.api.SlimefunAddon;
-import io.github.thebusybiscuit.slimefun4.api.items.ItemState;
 import io.github.thebusybiscuit.slimefun4.core.handlers.ItemUseHandler;
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.implementation.items.magical.staves.StormStaff;
 import io.github.thebusybiscuit.slimefun4.utils.LoreBuilder;
+import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
 import me.mrCookieSlime.Slimefun.Lists.RecipeType;
 import me.mrCookieSlime.Slimefun.Objects.Category;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
@@ -41,8 +41,7 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
  */
 public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
 
-    private static final NamespacedKey usageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
-    private static final Pattern REGEX = Pattern.compile(ChatColors.color("&e[0-9]+ Uses &7left"));
+    private final NamespacedKey defaultUsageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
 
     private int maxUseCount = -1;
 
@@ -72,12 +71,8 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
     public final @Nonnull LimitedUseItem setMaxUseCount(int count) {
         Validate.isTrue(count > 0, "The maximum use count must be greater than zero!");
 
-        if (getState() == ItemState.UNREGISTERED) {
-            maxUseCount = count;
-            return this;
-        } else {
-            throw new IllegalStateException("You cannot modify the maximum use count after the Item was registered.");
-        }
+        maxUseCount = count;
+        return this;
     }
 
     /**
@@ -86,7 +81,7 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
      * @return The {@link NamespacedKey} to store/load the amount of uses
      */
     protected @Nonnull NamespacedKey getStorageKey() {
-        return usageKey;
+        return defaultUsageKey;
     }
 
     @Override
@@ -140,7 +135,7 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
         if (lore != null && !lore.isEmpty()) {
             // find the correct line
             for (int i = 0; i < lore.size(); i++) {
-                if (REGEX.matcher(lore.get(i)).matches()) {
+                if (PatternUtils.USES_LEFT_LORE.matcher(lore.get(i)).matches()) {
                     lore.set(i, newLine);
                     meta.setLore(lore);
                     item.setItemMeta(meta);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -2,7 +2,6 @@ package io.github.thebusybiscuit.slimefun4.implementation.items;
 
 import java.util.Collections;
 import java.util.List;
-import java.util.regex.Pattern;
 
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
@@ -23,7 +22,6 @@ import io.github.thebusybiscuit.slimefun4.core.handlers.ItemUseHandler;
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.implementation.items.magical.staves.StormStaff;
 import io.github.thebusybiscuit.slimefun4.utils.LoreBuilder;
-import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
 import me.mrCookieSlime.Slimefun.Lists.RecipeType;
 import me.mrCookieSlime.Slimefun.Objects.Category;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -22,6 +22,7 @@ import io.github.thebusybiscuit.slimefun4.core.handlers.ItemUseHandler;
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.implementation.items.magical.staves.StormStaff;
 import io.github.thebusybiscuit.slimefun4.utils.LoreBuilder;
+import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
 import me.mrCookieSlime.Slimefun.Lists.RecipeType;
 import me.mrCookieSlime.Slimefun.Objects.Category;
 import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
@@ -38,8 +39,6 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
  * @see StormStaff
  */
 public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
-
-    public static final String USES_LEFT_SUFFIX = ChatColors.color("Uses &7left");
 
     private final NamespacedKey defaultUsageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
     private int maxUseCount = -1;
@@ -134,7 +133,7 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
         if (lore != null && !lore.isEmpty()) {
             // find the correct line
             for (int i = 0; i < lore.size(); i++) {
-                if (lore.get(i).endsWith(USES_LEFT_SUFFIX)) {
+                if (PatternUtils.USES_LEFT_LORE.matcher(lore.get(i)).matches()) {
                     lore.set(i, newLine);
                     meta.setLore(lore);
                     item.setItemMeta(meta);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -39,8 +39,9 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
  */
 public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
 
+    public static final String USES_LEFT_SUFFIX = ChatColors.color("Uses &7left");
+
     private final NamespacedKey defaultUsageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
-    private final String usesLeftSuffix = ChatColors.color("Uses &7left");
 
     private int maxUseCount = -1;
 
@@ -134,7 +135,7 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
         if (lore != null && !lore.isEmpty()) {
             // find the correct line
             for (int i = 0; i < lore.size(); i++) {
-                if (lore.get(i).endsWith(usesLeftSuffix)) {
+                if (lore.get(i).endsWith(USES_LEFT_SUFFIX)) {
                     lore.set(i, newLine);
                     meta.setLore(lore);
                     item.setItemMeta(meta);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -42,7 +42,6 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
     public static final String USES_LEFT_SUFFIX = ChatColors.color("Uses &7left");
 
     private final NamespacedKey defaultUsageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
-
     private int maxUseCount = -1;
 
     @ParametersAreNonnullByDefault
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -1,11 +1,14 @@
 package io.github.thebusybiscuit.slimefun4.implementation.items;
 
+import java.util.Collections;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 import org.apache.commons.lang.Validate;
+import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
 import org.bukkit.Sound;
 import org.bukkit.entity.Player;
@@ -39,6 +42,7 @@ import me.mrCookieSlime.Slimefun.api.SlimefunItemStack;
 public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
 
     private static final NamespacedKey usageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
+    private static final Pattern REGEX = Pattern.compile(ChatColors.color("&e[0-9]+ Uses &7left"));
 
     private int maxUseCount = -1;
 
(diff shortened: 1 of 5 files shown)


martinbrom:
offtopic - what's going on with the auto-label action :smile:

TheBusyBiscuit:
> offtopic - what's going on with the auto-label action 😄

Yeah, it only works for PRs from within the organization due to GitHub's security measures...
I will eventually fix this by making it just run on a `cron` job. But yeah... fun times.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,117 @@
...
+        addItemHandler(getItemHandler());
+    }
+
+    /**
+     * Returns the number of times this item can be used.
+     * Must be greater than zero.
+     *
+     * @return The number of times this item can be used.
+     */
+    protected abstract int getMaxUseCount();
```
TheBusyBiscuit, line 52: Really no point in making this method abstract, especially because it would allow you to return anything, such as `-1`, `Infinity` or `NaN`.

It would be better to control this via a setter and getter so we can validate parameters.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,117 @@
...
+
+    /**
+     * Returns the {@link NamespacedKey} under which will the amount of uses left stored.
+     *
+     * @return The {@link NamespacedKey} to store/load the amount of uses
+     */
+    protected @Nonnull NamespacedKey getStorageKey() {
+        return new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
```
TheBusyBiscuit, line 60: No need to instantiate a new `NamespacedKey` here. Just make a `private final` field for this and return that.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,117 @@
...
+            } else {
+                usesLeft--;
+                pdc.set(key, PersistentDataType.INTEGER, usesLeft);
+
+                updateItemLore(item, meta, usesLeft);
+            }
+        }
+    }
+
+    private void updateItemLore(ItemStack item, ItemMeta meta, int usesLeft) {
```
TheBusyBiscuit, line 105: ParamsAreNonnullByDefault?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
```diff
@@ -83,4 +83,9 @@ public static String range(int blocks) {
         return "&7Range: &c" + blocks + " blocks";
     }
 
+    @Nonnull
+    public static String usesLeft(int usesLeft) {
```
TheBusyBiscuit, line 87:
```suggestion
    public static @Nonnull String usesLeft(int usesLeft) {
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
martinbrom: Well, I did it the same way as `AContainer` does...the item is not registered if the getter returns an invalid value

TheBusyBiscuit: `AContainer` uses setters and getters though, like I suggested...

Also `AContainer` is not the go-to example for best practices :LUL:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,141 @@
...
+        return maxUseCount;
+    }
+
+    /**
+     * Sets the maximum number of times this item can be used.
+     * The number must be greater than zero.
+     *
+     * @param count The maximum number of times this item can be used.
+     */
+    public final void setMaxUseCount(int count) {
```
WalshyDev, line 67: Would be nice to have this return LimitedUseItem then we can do

```java
new LimitedUseItem(...)
	.setMaxUseCount(100)
	.register(this);
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+        } else {
+            ItemMeta meta = item.getItemMeta();
+            NamespacedKey key = getStorageKey();
+            PersistentDataContainer pdc = meta.getPersistentDataContainer();
+            int usesLeft = pdc.getOrDefault(key, PersistentDataType.INTEGER, getMaxUseCount());
+
+            if (usesLeft == 1) {
+                p.playSound(p.getLocation(), Sound.ENTITY_ITEM_BREAK, 1, 1);
+                item.setAmount(0);
```
WalshyDev, line 118: Note: This doesn't *always* work as you expect. It can just set the amount to well 0. Which is technically a valid number in some versions of MC. Some server software on some versions have this behaviour rather than removing the item.

It's best to set the type to AIR.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    @ParametersAreNonnullByDefault
+    private void updateItemLore(ItemStack item, ItemMeta meta, int usesLeft) {
+        List<String> lore = meta.getLore();
+
+        if (lore != null && !lore.isEmpty()) {
+            // replace the last line
+            lore.set(lore.size() - 1, ChatColors.color(LoreBuilder.usesLeft(usesLeft)));
```
WalshyDev, line 134: This wont always be the uses left lore line. Lots of plugins add their own lore line, Slimefun also adds one with Soulbound.

You should instead find the correct lore line and re-set it.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
```diff
@@ -33,54 +33,48 @@
...
     }
 
-    @Nonnull
-    public static String range(int blocks) {
+    public static @Nonnull String range(int blocks) {
         return "&7Range: &c" + blocks + " blocks";
     }
 
+    public static @Nonnull String usesLeft(int usesLeft) {
+        return "&e" + usesLeft + ' ' + (usesLeft > 1 ? "Uses" : "Use") + " &7left";
```
WalshyDev, line 77: We didn't use to support singular use right? If not then this will cause problems on servers supporting legacy items.

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
martinbrom: I don't think anything would break but Imma trust your judgment

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,156 @@
...
+ * @author Walshy
+ * @author TheBusyBiscuit
+ * @author martinbrom
+ *
+ * @see StormStaff
+ */
+public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
+
+    private static final NamespacedKey usageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
+    private static final Pattern REGEX = Pattern.compile(ChatColors.color("&e[0-9]+ Uses &7left"));
```
TheBusyBiscuit, line 45: These do **not** need to be final.

```suggestion
    private final NamespacedKey usageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
    private final Pattern lorePattern = Pattern.compile(ChatColors.color("&e[0-9]+ Uses &7left"));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,156 @@
...
+     * @return The {@link LimitedUseItem} for chaining of setters
+     */
+    public final @Nonnull LimitedUseItem setMaxUseCount(int count) {
+        Validate.isTrue(count > 0, "The maximum use count must be greater than zero!");
+
+        if (getState() == ItemState.UNREGISTERED) {
+            maxUseCount = count;
+            return this;
+        } else {
+            throw new IllegalStateException("You cannot modify the maximum use count after the Item was registered.");
```
TheBusyBiscuit, line 79: ummm, why not? 🤔

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,156 @@
...
+ * When the item runs out of "uses", it breaks.
+ *
+ * @author Linox
+ * @author Walshy
+ * @author TheBusyBiscuit
+ * @author martinbrom
+ *
+ * @see StormStaff
+ */
+public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
```
TheBusyBiscuit, line 42: Why is this abstract...?

```suggestion
public class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
```diff
@@ -33,54 +33,48 @@
...
-    public static String range(int blocks) {
+    public static @Nonnull String range(int blocks) {
         return "&7Range: &c" + blocks + " blocks";
     }
 
+    public static @Nonnull String usesLeft(int usesLeft) {
+        return "&e" + usesLeft + " Uses &7left";
+    }
+
 }
```
TheBusyBiscuit, line 80: *insert happy cookie crumbling noises*

TheBusyBiscuit (changes requested):
Looks fine overall, some things bother me with the LimitedUseItem class though

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
martinbrom: It is abstract because it does not implement the `ItemUseHandler`, which is something each limited item must define itself, I feel like there is no good default behavior.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
martinbrom: :)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
martinbrom: The comment says `final` but the suggestion removes the `static` keyword. Assuming the suggestion is correct, why tho? There will often be multiple instances of the `LimitedUseItem` and the values do not differ. It is safe for them to be static then, is it not?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
martinbrom: I feel like that's just waiting for problems to happen when we start modifying 'constant' values during runtime. I can, of course, remove this but the same check exists for energy capacity, energy consumption, and other values.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
TheBusyBiscuit: I see :+1:

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
TheBusyBiscuit: 1. `static` fields create a potential risk of memory leakage due to the way Bukkit's ClassLoader works, so it is best to keep those to a minimum.
2. Initialising the `NamespacedKey` static-ly is an incedribly bad idea. We do not have explicit control over when the class is loaded into memory and therefore no control over when this field is initialised. `SlimefunPlugin.instance()` may be `null` at that point, making this a class member ensures that it only gets loaded when we explicitly instantiate an object of that class.

The `Pattern` might make sense to instantiate static-ly but I would suggest this to be moved to `PatternUtils` then, so we have finer control over where our static fields lay.
Also `usageKey` may perhaps better be named `defaultUsageKey`, as it can be overridden by sub classes.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
TheBusyBiscuit: Energy values are a different matter though, these get passed to the energy network upon registration, so any change to them would need to be synchronized to the entire energy network handling and be dealt with.
Here I see rather few problems that would arise from this as long as it is kept as a positive integer. But I am fine with the warning here at first, as long as it is properly justified and documented.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,151 @@
...
+
+    @ParametersAreNonnullByDefault
+    private void updateItemLore(ItemStack item, ItemMeta meta, int usesLeft) {
+        List<String> lore = meta.getLore();
+
+        String newLine = ChatColors.color(LoreBuilder.usesLeft(usesLeft));
+        if (lore != null && !lore.isEmpty()) {
+            // find the correct line
+            for (int i = 0; i < lore.size(); i++) {
+                if (PatternUtils.USES_LEFT_LORE.matcher(lore.get(i)).matches()) {
```
WalshyDev, line 138: Do we want a regex for this? we could just as easily .endsWith("Uses left")

martinbrom: Good point, but since we're discussing, how would this get translated in the future? :smile:

WalshyDev: Easily if it's ends with, not as easily with regex :p

martinbrom: Just to clarify, I didn't imply that it would be easier with regex, I just didn't think of a way to do it with endsWith :smile: (in the one minute I thought about it)

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
```diff
@@ -33,54 +33,48 @@
...
     }
 
-    @Nonnull
-    public static String range(int blocks) {
+    public static @Nonnull String range(int blocks) {
         return "&7Range: &c" + blocks + " blocks";
     }
 
+    public static @Nonnull String usesLeft(int usesLeft) {
+        return "&e" + usesLeft + " Uses &7left";
```
WalshyDev, line 77: The suffix here should be a constant somewhere. So it can be used here and in LimitedUseItem.

WalshyDev (changes requested):
After this it LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/utils/LoreBuilder.java
martinbrom: I think it'd fit into the LoreBuilder, that fine?

WalshyDev (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,151 @@
...
+ * @author Linox
+ * @author Walshy
+ * @author TheBusyBiscuit
+ * @author martinbrom
+ *
+ * @see StormStaff
+ */
+public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
+
+    public static final String USES_LEFT_SUFFIX = ChatColors.color("Uses &7left");
```
TheBusyBiscuit, line 42: The previous implementation respected plural and singular.
This change will ruin that, e.g. "1 Uses left"

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private final NamespacedKey defaultUsageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
+
+    private int maxUseCount = -1;
```
TheBusyBiscuit, line 46:
```suggestion
    private final NamespacedKey defaultUsageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
    private int maxUseCount = -1;
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
martinbrom: I originally implemented it with plural and singular and Walshy said I should revert that??

martinbrom:
I just realized there was no need to match the singular version of the lore because the only time when that is visible is when the item has 1 use left and after using it one last time it gets destroyed and we don't need to update the lore anymore. Oh well...

src/main/java/io/github/thebusybiscuit/slimefun4/utils/PatternUtils.java
```diff
@@ -14,7 +16,8 @@
  */
 public final class PatternUtils {
 
-    private PatternUtils() {}
+    private PatternUtils() {
+    }
```
TheBusyBiscuit, line 20: Either one-line or with a comment inside. But not this please :/

```suggestion
    private PatternUtils() {}
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,149 @@
...
+ * @see StormStaff
+ */
+public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
+
+    private final NamespacedKey defaultUsageKey = new NamespacedKey(SlimefunPlugin.instance(), "uses_left");
+    private int maxUseCount = -1;
+
+    @ParametersAreNonnullByDefault
+    public LimitedUseItem(Category category, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        super(category, item, recipeType, recipe, null);
```
TheBusyBiscuit, line 48: Why is the recipeOutput `null` here?

```suggestion
        super(category, item, recipeType, recipe);
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -0,0 +1,149 @@
...
+    public void register(@Nonnull SlimefunAddon addon) {
+        if (getMaxUseCount() < 1) {
+            warn("The use count has not been configured correctly. It needs to be at least 1. The Item was disabled.");
+        } else {
+            super.register(addon);
+        }
+    }
+
+    @ParametersAreNonnullByDefault
+    protected void damageItem(Player p, ItemStack item) {
```
TheBusyBiscuit, line 95: Not sure if `damage` is the right word here. But cannot think of a better alternative either.

TheBusyBiscuit (changes requested):
2 changes till approval

src/main/java/io/github/thebusybiscuit/slimefun4/utils/PatternUtils.java
martinbrom: We really have to get some sort of ruleset for intellij, the autoformat on commit just leads to so many unnecessarily rejected reviews...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
martinbrom: Good catch, left over from the previous StormStaff code

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
martinbrom: Could be called `use` but thats no better

TheBusyBiscuit: How about `subtractOneFromOurUsesLeftAttribute()` :kappa:

TheBusyBiscuit (approved):
LGTM

svr333 (approved):
Looks good

PR #3118, opened 2021-06-14
Add diamond -> coal conversion

svr333:
## Description
Add one recipe to grindstone (and electric ore grinder)

## Proposed changes
- Add diamond to 24 coal conversion

[Discord suggestion 774](https://discord.com/channels/565557184348422174/627048923122499584/822003610140409876)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
@@ -91,10 +91,6 @@ public class GrindStone extends MultiBlockMachine {
 
         recipes.add(SlimefunItems.ENDER_LUMP_3);
         recipes.add(new SlimefunItemStack(SlimefunItems.ENDER_LUMP_2, 4));
-
-        recipes.add(SlimefunItems.CARBON);
-        recipes.add(new ItemStack(Material.COAL, 8));
-
         recipes.add(new ItemStack(Material.DIAMOND));
         recipes.add(new SlimefunItemStack(SlimefunItems.CARBON, 4));
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
@@ -91,6 +91,9 @@ public class GrindStone extends MultiBlockMachine {
 
         recipes.add(SlimefunItems.ENDER_LUMP_3);
         recipes.add(new SlimefunItemStack(SlimefunItems.ENDER_LUMP_2, 4));
+
+        recipes.add(new ItemStack(Material.DIAMOND));
+        recipes.add(new ItemStack(Material.COAL, 24));
     }
 
     @Override
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
@@ -92,8 +92,11 @@ public class GrindStone extends MultiBlockMachine {
         recipes.add(SlimefunItems.ENDER_LUMP_3);
         recipes.add(new SlimefunItemStack(SlimefunItems.ENDER_LUMP_2, 4));
 
+        recipes.add(SlimefunItems.CARBON);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
```diff
@@ -91,6 +91,9 @@ protected void registerDefaultRecipes(@Nonnull List<ItemStack> recipes) {
 
         recipes.add(SlimefunItems.ENDER_LUMP_3);
         recipes.add(new SlimefunItemStack(SlimefunItems.ENDER_LUMP_2, 4));
+
+        recipes.add(new ItemStack(Material.DIAMOND));
+        recipes.add(new ItemStack(Material.COAL, 24));
```
TheBusyBiscuit, line 96: Why 24 though?

svr333: I felt like 16 was a bit too few and 32 a bit too much
What would you suggest?

TheBusyBiscuit: I did the maths, and it seems like 1 diamond (synthesized) corresponds to 8 Compressed Carbon which each correspond to 4 Carbon which each correspond to 8 coal.
That means 1 diamond is roughly worth 4 stacks of coal.
Making this a 1:1 mapping seems ridicilous, I admit.
But I think 32 is a bit more on the fair side here while still yielding a big loss of coal compared to the other way around.

Another idea on how to spice this up is to introduce a middle item.
How about you grind diamonds into e.g. 8 diamond-coal-hybrid item and then these can each be ground into 4 coal.
Or similar.

svr333: All right

svr333: I was looking for a good name and then I realized we already have Carbons, we can just grind a diamond into 4 Carbon, and then just add a way to convert 1 carbon back into 8 coal.
That seem good?

Semisol (changes requested):
Looks good to me to be fair. I feel like the refund amount maybe might need an improvement but yeah.

But requesting changes because: The ore crusher can convert carbon to coal at a 1:1.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
```diff
@@ -91,6 +91,12 @@ protected void registerDefaultRecipes(@Nonnull List<ItemStack> recipes) {
 
         recipes.add(SlimefunItems.ENDER_LUMP_3);
         recipes.add(new SlimefunItemStack(SlimefunItems.ENDER_LUMP_2, 4));
+
+        recipes.add(SlimefunItems.CARBON);
+        recipes.add(new ItemStack(Material.COAL, 8));
+
```
Semisol, line 97: The ore crusher can already do this.
```suggestion
```

Semisol (approved):
Looks all good.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/GrindStone.java
```diff
@@ -91,6 +91,8 @@ protected void registerDefaultRecipes(@Nonnull List<ItemStack> recipes) {
 
         recipes.add(SlimefunItems.ENDER_LUMP_3);
         recipes.add(new SlimefunItemStack(SlimefunItems.ENDER_LUMP_2, 4));
+        recipes.add(new ItemStack(Material.DIAMOND));
+        recipes.add(new SlimefunItemStack(SlimefunItems.CARBON, 4));
```
TheBusyBiscuit, line 95:
```suggestion
        recipes.add(new SlimefunItemStack(SlimefunItems.ENDER_LUMP_2, 4));

        recipes.add(new ItemStack(Material.DIAMOND));
        recipes.add(new SlimefunItemStack(SlimefunItems.CARBON, 4));
```

TheBusyBiscuit (changes requested):

TheBusyBiscuit (approved):
You didn't see anything :eyes:

Semisol (approved):

WalshyDev (approved):

PR #3119, opened 2021-06-14
Add Trident recipe

svr333:
## Description
Add Trident Recipe

## Proposed changes
- Add trident recipe (4 tier 3 magic lumps, 4 water runes, 1 diamond sword)

[Discord suggestion 577](https://discord.com/channels/565557184348422174/627048923122499584/775973265763598336)

## Checklist


martinbrom:
I feel like this is way too cheap, tridents are not that easy to get your hands on. I'd throw in something like a nether star or sth for good measure.

Seggan:
mm I was planning to add this to SlimefunWarfare

Apeiros-46B:
I think 1 reinforced ingot, 2 nautilus shells, 2 water staves, 2 magical lump IIIs, and a nether star around a netherite sword would be reasonable

Seggan:
> mm I was planning to add this to SlimefunWarfare

so you think this should be done or add it to base slimefun?

svr333:
I dont see a reason why it shouldn't be added to base Slimefun, but in the end I don't really mind all that much.

Seggan:
🤷‍♂️ guess base slimefun it is

TheBusyBiscuit (approved):
The recipe is definitely debatable.
But looks good to me - I guess?

Sfiguz7 (approved):
We can always tinker with the recipe later on if people report it being too easy I guess, other than that looks good

PR #3134, opened 2021-06-21
Intregation: Orebfuscator

NgLoader:
## Description
Orebfuscator obscures blocks against x-rays, but if you destroy multiple blocks that are around the break block, only the break block will be updated.
Therefore, you have to send an update signal.

## Proposed changes
Added MultiBlockBreakEvent class when multiply blocks are breaking
Added OrebfuscatorIntregation class
Added Orebfuscator to plugin.yml (softdepend)
Added Orebfuscator to pom.xml (repo + dependency)
Edited IntregationManager (added orebfuscator check in onServerStart)

Resolves https://github.com/Imprex-Development/Orebfuscator/issues/82

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/integrations/IntegrationsManager.java
```diff
@@ -281,4 +288,7 @@ public boolean isChestTerminalInstalled() {
         return isChestTerminalInstalled;
     }
 
+    public boolean isOrebfuscatorInstalled() {
+		return isOrebfuscatorInstalled;
```
TheBusyBiscuit, line 292: I think there are tab characters in this line, we use spaces exclusively on this project.

src/main/java/io/github/thebusybiscuit/slimefun4/integrations/OrebfuscatorIntegration.java
```diff
@@ -0,0 +1,64 @@
...
+	public OrebfuscatorIntegration(@Nonnull SlimefunPlugin plugin) {
+        this.plugin = plugin;
+    }
+
+	/**
+	 * Init orebfuscation service and register listener
+	 */
+    public void register() {
+        RegisteredServiceProvider<OrebfuscatorService> provider = Bukkit.getServer().getServicesManager().getRegistration(OrebfuscatorService.class);
+        this.service = provider.getProvider();
```
TheBusyBiscuit, line 42: I don't see a need for declaring a variable here.

```suggestion
        this.service = plugin.getServer().getServicesManager().getRegistration(OrebfuscatorService.class).getProvider();
```

TheBusyBiscuit (changes requested):
Thanks for your pull request!

The integration looks very straightforward and fitting to implement.
Thank you for keeping it in-style with the other integrations too!

I highlighted two minor nitpicks but other than that it looks good to me.

src/main/java/io/github/thebusybiscuit/slimefun4/integrations/OrebfuscatorIntegration.java
NgLoader: It it better to read

src/main/java/io/github/thebusybiscuit/slimefun4/integrations/IntegrationsManager.java
NgLoader: Oh, sry, I will fix that

src/main/java/io/github/thebusybiscuit/slimefun4/integrations/OrebfuscatorIntegration.java
NgLoader: but when you like it more in a oneline we can change that

TheBusyBiscuit: Not a necessity in my opinion, really comes down to preference I guess.

But once the tab characters are removed I am gonna approve this.
It will likely be up to you or our second reviewer at that point XD
Thanks!

NgLoader: Okay I change it 👍🏻

src/main/java/io/github/thebusybiscuit/slimefun4/integrations/OrebfuscatorIntegration.java
```diff
@@ -0,0 +1,62 @@
...
+        Set<Block> blocks = new HashSet<>();
+        blocks.addAll(event.getAdditionalBlocks());
+        blocks.add(event.getPrimaryBlock());
+        this.service.deobfuscate(blocks);
+    }
+
+    @EventHandler
+    public void onReactorExplode(ReactorExplodeEvent event) {
+        this.service.deobfuscate(Arrays.asList(event.getLocation().getBlock()));
+    }
```
TheBusyBiscuit, line 61: One thing I forgot:

None of these Event Handlers affect the event's outcome, so they should all listen on the priority `MONITOR` and have `ignoreCancelled` set to `true`.

This helps to reduce unnecessary deobfuscations if the event did not actually process.

TheBusyBiscuit (changes requested):
Forgot one thing

src/main/java/io/github/thebusybiscuit/slimefun4/integrations/OrebfuscatorIntegration.java
NgLoader: Oh, okay

NgLoader: But then I need to set ``ignoreCancelled`` to ``false``

TheBusyBiscuit: `ignoreCancelled` means that the `EventHandler` will ignore cancelled events.
That means that your code won't be triggered if the event was cancelled. So `ignoreCancelled=true` is what you want.

Example:
If the BlockPlacer event gets cancelled, no block is placed. So we don't need to trigger an Orebfuscator update either since no change was made to the world.

Here is a javadocs ref:
https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/EventHandler.html

NgLoader: Oh xD
But the other line says ``Define if the handler ignores a cancelled event.`` so I was thinking when the value is ``true`` the event will be canceled

TheBusyBiscuit (dismissed):
LGTM

WalshyDev:
How many people actually use this given it's in Paper? Majority of servers use Paper these days and from a performance standpoint it being native is much better.

NgLoader:
OFC is much faster than paper and is able to use a cache with will increase the response time and not need to regenerate the chunk.
You can watch the currently installations [here](https://bstats.org/plugin/bukkit/Orebfuscator/8942)

NgLoader:
I have made some changes (readme.md updated)

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+import org.bukkit.block.Block;
+import org.bukkit.entity.Player;
+import org.bukkit.event.Cancellable;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+/**
+ * This {@link Event} is called when a {@link Player} breaks multiply blocks.
+ * 
+ * @author TheBusyBiscuit
```
TheBusyBiscuit, line 18: That's what I call identity theft :eyes:

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ */
+public class MultiBlockBreakEvent extends Event implements Cancellable {
```
TheBusyBiscuit, line 21: That name is completely misleading...
Slimefun has things called "MultiBlocks", so this is VERY misleading...

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * This returns the list of all breaking blocks.
+     * 
+     * @return A list of {@link Block}
+     */
+    @Nonnull
+    public List<Block> getBlocks() {
+		return this.blocks;
```
TheBusyBiscuit, line 54: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    public Block getClickedBlock() {
+        return clickedBlock;
+    }
+
+    /**
+     * This returns the list of all breaking blocks.
+     * 
+     * @return A list of {@link Block}
+     */
+    @Nonnull
```
TheBusyBiscuit, line 52: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    public List<Block> getBlocks() {
+		return this.blocks;
+	}
+
+    @Nullable
```
TheBusyBiscuit, line 57: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    public Player getPlayer() {
+		return this.player;
```
TheBusyBiscuit, line 59: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    public void setCancelled(boolean cancel) {
+        this.cancelled = cancel;
+    }
+
+    @Nonnull
+    public static HandlerList getHandlerList() {
+        return handlers;
+    }
+
+    @Nonnull
```
TheBusyBiscuit, line 77: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    public boolean isCancelled() {
+        return cancelled;
+    }
+
+    @Override
+    public void setCancelled(boolean cancel) {
+        this.cancelled = cancel;
+    }
+
+    @Nonnull
```
TheBusyBiscuit, line 72: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+        this.clickedBlock = clicked;
+        this.blocks = blocks;
+    }
+
+    /**
+     * This returns the specific {@link Block} that was breaked.
+     * 
+     * @return The {@link Block} that was breaked
+     */
+    @Nonnull
```
TheBusyBiscuit, line 42: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+    private static final HandlerList handlers = new HandlerList();
+
+    private final Player player;
+    private final Block clickedBlock;
+    private final List<Block> blocks;
+    private boolean cancelled;
+
+    @ParametersAreNonnullByDefault
+    public MultiBlockBreakEvent(@Nullable Player player, Block clicked, List<Block> blocks) {
+    	this.player = player;
```
TheBusyBiscuit, line 32: Formatting

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
```diff
@@ -0,0 +1,83 @@
...
+
+    private static final HandlerList handlers = new HandlerList();
+
+    private final Player player;
+    private final Block clickedBlock;
+    private final List<Block> blocks;
+    private boolean cancelled;
+
+    @ParametersAreNonnullByDefault
+    public MultiBlockBreakEvent(@Nullable Player player, Block clicked, List<Block> blocks) {
```
TheBusyBiscuit, line 31: Making Player nullable here is not really a good idea, there should be seperate events for this if anything.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/WoodcutterAndroid.java
```diff
@@ -48,6 +49,13 @@ protected boolean chopTree(Block b, BlockMenu menu, BlockFace face) {
             List<Block> list = Vein.find(target, MAX_REACH, block -> Tag.LOGS.isTagged(block.getType()));
 
             if (!list.isEmpty()) {
+                MultiBlockBreakEvent event = new MultiBlockBreakEvent(null, target, list);
+                Bukkit.getServer().getPluginManager().callEvent(event);
+
+                if (event.isCancelled()) {
+                    return true;
+                }
```
TheBusyBiscuit, line 57: Not a justified event call.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/LumberAxe.java
```diff
@@ -75,6 +77,13 @@ public ItemUseHandler onItemUse() {
                         logs.remove(block);
                     }
 
+                    MultiBlockBreakEvent event = new MultiBlockBreakEvent(e.getPlayer(), block, logs);
+                    Bukkit.getServer().getPluginManager().callEvent(event);
+
+                    if (event.isCancelled()) {
+                        return;
+                    }
```
TheBusyBiscuit, line 85: Not a justified event call.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/PickaxeOfVeinMining.java
```diff
@@ -47,7 +49,13 @@ public ToolUseHandler getItemHandler() {
...
             if (SlimefunTag.PICKAXE_OF_VEIN_MINING_BLOCKS.isTagged(e.getBlock().getType())) {
                 List<Block> blocks = Vein.find(e.getBlock(), maxBlocks.getValue(), b -> SlimefunTag.PICKAXE_OF_VEIN_MINING_BLOCKS.isTagged(b.getType()));
-                breakBlocks(e.getPlayer(), blocks, fortune, tool);
+
+                MultiBlockBreakEvent event = new MultiBlockBreakEvent(e.getPlayer(), e.getBlock(), blocks);
+                Bukkit.getServer().getPluginManager().callEvent(event);
+
+                if (!event.isCancelled()) {
+                    breakBlocks(e.getPlayer(), blocks, fortune, tool);
+                }
```
TheBusyBiscuit, line 58: Not a justified event call.

TheBusyBiscuit (changes requested):
I disagree with these new changes on so many levels.
Especially because it breaks the purpose of the event system.

Having one event for so many different use cases with nullable parameters is an incredibly bad idea.
Any player event should also extend `PlayerEvent` which this does not, despite it being a pseudo-playerevent.

You also claim to have changed our README??? But I luckily don't see any of that in the diff.
At this state I am not willing to merge this, please revert these changes and perhaps it can be brought up in the near future within another PR.

TheBusyBiscuit:
> How many people actually use this given it's in Paper? Majority of servers use Paper these days and from a performance standpoint it being native is much better.

We recommend Paper anyway and there will be a prompt to install Paper if Spigot is detected.


> OFC is much faster than paper and is able to use a cache with will increase the response time and not need to regenerate the chunk.
> we also 100% configuration able and we are coming with a ProximityHider.
> You can watch the currently installations [here](https://bstats.org/plugin/bukkit/Orebfuscator/8942)

Some evidence of that would be nice to see, such as benchmarks.
Though I don't really mind, there aren't too many servers on modern MC versions using this but any integration is a good addition in my opinion. If people don't use it, then it won't be loaded in anyway, so there is no negative impact from having this integration anyway.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
NgLoader: have you any good other name idea like ``BlocksBreakEvent``?

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
NgLoader: Oh, sorry xD forgot to change that

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/WoodcutterAndroid.java
NgLoader: Whats is for you a justified event call?
In ``ExplosiveTool`` it will be called like I call it

NgLoader:
> > How many people actually use this given it's in Paper? Majority of servers use Paper these days and from a performance standpoint it being native is much better.
Currently 48.7% of all ofc users are on paper, ofc is able to handle multiply packets async and through the cache (RAM/HARD DRIVE) we don't need to generate the chunk for every player new

> We recommend Paper anyway and there will be a prompt to install Paper if Spigot is detected.
OFC is able to run on paper thats not the problem I think.

> Some evidence of that would be nice to see, such as benchmarks.
> Though I don't really mind, there aren't too many servers on modern MC versions using this but any integration is a good addition in my opinion. If people don't use it, then it won't be loaded in anyway, so there is no negative impact from having this integration anyway.
Yea, we have made some locally benchmarks, we will provide some benchmarks later.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/MultiBlockBreakEvent.java
NgLoader: Okay

TheBusyBiscuit (changes requested):
I still rather have these changes reverted before merging.
And I suggest you to look into some best-practices for software development, mainly:
1. "Working with different branches" and avoiding making a pull request from your `master` branch
2. "Late requirements changes", when you create a pull request you define a scope which we will acknowledge. Changes are fine but constantly expanding the scope and the requirements is one of the key issues to avoid in software development and project management.
3. Respect the code-format with which you are working, you can usually find these things documented in the `CONTRIBUTING.md` file or similar. (https://github.com/Slimefun/Slimefun4/blob/master/CONTRIBUTING.md)
4. Bukkit's event system. Events work hierarchically, there are super classes and sub classes of event and each class is a valid Event handling point in of itself. Be wary of how you design your events and their attributes. As well as the corresponding mutability of these attributes and the implications that come with it.
5. Maintainability - I am not bringing these points up for no reason, introducing new events like these seems trivial but with such a large project that has been maintained for over 8 years now... and has a lot of dependents, any API changes (such as event additions) need to be thought through carefully. After all we will be the ones responsible for maintaining these additions for years to come.

So... I hope this clears up some of the underlying principles here that guard my decision.
I am still willing to merge this in its original state - without the event changes. I think any compatibility is worth having but this grew out of the original scope quite quickly and these newer changes rather belong into a dedicated follow-up pull request.
As mentioned before, API changes can trigger lengthy discussions and review periods, so there is now need to hold all changes back just because of this subset that I reckon should be postponed.

TheBusyBiscuit:
@WalshyDev Your opinion?

NgLoader:
> I still rather have these changes reverted before merging.
Ok, that's we can do

NgLoader:
I just reverted the 2 last commits.
Is it so better @TheBusyBiscuit ?

src/main/java/io/github/thebusybiscuit/slimefun4/integrations/OrebfuscatorIntegration.java
```diff
@@ -0,0 +1,63 @@
...
+        this.plugin = plugin;
+    }
+
+	/**
+	 * Init orebfuscation service and register listener
+	 */
+    public void register() {
+        this.service = Bukkit.getServer().getServicesManager().getRegistration(OrebfuscatorService.class).getProvider();
+
+        Bukkit.getServer().getPluginManager().registerEvents(this, this.plugin);
```
TheBusyBiscuit, line 43: Minor detail but if we can get the `Server` instance in an object-oriented fashion, it should be preffered over the static access.

```suggestion
        plugin.getServer().getPluginManager().registerEvents(this, this.plugin);
```

TheBusyBiscuit (changes requested):
Looks good to me, just one remark.

TheBusyBiscuit:
> Is it so better @TheBusyBiscuit ?

The german grammer is real here.
And yes, "it is better so".

TheBusyBiscuit (approved):

svr333 (approved):
Looks good

PR #3135, opened 2021-06-22
Changed ProtectionType from Enum to Class

Seggan:
## Description

Changed ProtectionType from enum to class so that addons can add their own. I haven't tested but I've checked out the bytecode and it appears that enum constants and fields are accessed the same way, so it shouldn't break anything.

## Proposed changes

Changed ProtectionType from an enum to a class

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
@@ -12,21 +12,22 @@ import org.bukkit.entity.Bee;
  * @see ProtectiveArmor
  *
  */
-public enum ProtectionType {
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ProtectionType.java
```diff
@@ -12,21 +12,22 @@
  * @see ProtectiveArmor
  *
  */
-public enum ProtectionType {
+@SuppressWarnings("InstantiationOfUtilityClass")
+public final class ProtectionType {
```
md5sha256, line 16: If the class is `final` it can't be extended

TheBusyBiscuit (commented):
Rejected.

PR #3138, opened 2021-06-25
Fix disappearing items with output chest by machines

### LINKED ISSUE #3136: Items processed by compressor will disappear when output chest is going to full
### ❗ Checklist

### 📍 Description

When output chest is going to full(not full yet) and compressor is processing items more than output chest can store, the rest of items that processed will disappear. But if output chest already full, and compressor is processing items, the items will go to compressor instead

### 📑 Reproduction Steps

1. Fill output chest with items(in this case, example: carbon) until almost full
2. Get coal block more than enough for crafting carbon
3. Right click the compressor

### 💡 Expected Behavior

When output chest is full and the compressor is still processing, the items will go to compressor.

### 📷 Screenshots / Videos

While output chest isnt full
https://user-images.githubusercontent.com/70990197/123103125-e8f0aa00-d45f-11eb-88b0-3ab44fee5bcc.mp4

While output chest is already full
https://user-images.githubusercontent.com/70990197/123103231-fe65d400-d45f-11eb-93b0-ba87bbab056b.mp4

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.16.x

### ⭐ Slimefun version

DEV- 939 (git 57f7553f)
image

### 🧭 Other plugins

_No response_

EpicPlayerA10:
## Description
This fixes disappearing items with machines connected to almost full output chest.
These machines are: Compressor, Magic Workbench, Armor Forge, Pressure Chamber.

## Proposed changes
Added to `Compressor`, `MagicWorkbench`, `ArmorForge` and `PressureChamber` check if the items can be added to the output chest after crafting.
If items can't be added then item will be added to another output chest (if it exists). And if items can't be added to any output chests then item will be added to dispenser. And if item can't be added to dispenser then the item will drop on the ground.

Resolves #3136 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -178,20 +178,18 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
      *          A crafted {@link ItemStack} from {@link MultiBlockMachine}
      * @param block
      *          Main {@link Block} of our {@link Container} from {@link MultiBlockMachine}
+     * @param blockInv
+     *          The {@link Inventory} of our {@link Container}
      *
      */
     @ParametersAreNonnullByDefault
-    protected void handleCraftedItem(ItemStack outputItem, Block block) {
-        BlockState state = PaperLib.getBlockState(block, false).getState();
-        Validate.isTrue(state instanceof Container);
-        Container container = (Container) state;
-        Inventory containerInv = container.getInventory();
-        Inventory outputInv = findOutputInventory(outputItem, block, containerInv);
+    protected void handleCraftedItem(ItemStack outputItem, Block block, Inventory blockInv) {
+        Inventory outputInv = findOutputInventory(outputItem, block, blockInv);
 
         if (outputInv != null) {
             outputInv.addItem(outputItem);
-        } else if (InvUtils.fits(containerInv, outputItem)) {
-            containerInv.addItem(outputItem);
+        } else if (InvUtils.fits(blockInv, outputItem)) {
+            blockInv.addItem(outputItem);
         } else {
             // fallback
             SlimefunUtils.spawnItem(block.getLocation(), outputItem, ItemSpawnReason.MULTIBLOCK_MACHINE_OVERFLOW, true);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -185,11 +185,13 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
 
         if (outputInv != null) {
             outputInv.addItem(outputItem);
-        } else if (InvUtils.fits(blockInv, outputItem)) {
-            blockInv.addItem(outputItem);
         } else {
-            // fallback
-            SlimefunUtils.spawnItem(block.getLocation(), outputItem, ItemSpawnReason.MULTIBLOCK_MACHINE_OVERFLOW, true);
+            ItemStack rest = blockInv.addItem(outputItem).get(0);
+
+            // fallback: drop item
+            if (rest != null) {
+                SlimefunUtils.spawnItem(block.getLocation(), rest, ItemSpawnReason.MULTIBLOCK_MACHINE_OVERFLOW, true);
+            }
         }
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -14,7 +14,7 @@ import org.bukkit.Material;
 import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
-import org.bukkit.block.data.type.Dispenser;
+import org.bukkit.block.Dispenser;
 import org.bukkit.entity.Player;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.ItemStack;
@@ -162,6 +162,29 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
             return dispInv;
         } else {
             return outputChest.orElse(null);
+        }
+    }
+
+    /**
+     * This method handles a output {@link ItemStack} from the {@link MultiBlockMachine} which has a crafting delay
+     *
+     * @param outputItem
+     *          A crafted {@link ItemStack} from {@link MultiBlockMachine}
+     * @param dispenser
+     *          Our {@link Dispenser} from {@link MultiBlockMachine}
+     *
+     */
+    protected void handleCraftedItem(ItemStack outputItem, Dispenser dispenser) {
+        Inventory dispInv = dispenser.getInventory();
+        Inventory outputInv = findOutputInventory(outputItem, dispenser.getBlock(), dispInv);
+
+        if (outputInv != null) {
+            outputInv.addItem(outputItem);
+        } else if (InvUtils.fits(dispInv, outputItem)) {
+            dispInv.addItem(outputItem);
+        } else {
+            // fallback
+            dispenser.getWorld().dropItemNaturally(dispenser.getLocation(), outputItem);
         }
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -14,7 +14,6 @@ import org.bukkit.Material;
 import org.bukkit.World;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
-import org.bukkit.block.BlockState;
 import org.bukkit.block.Container;
 import org.bukkit.block.Dispenser;
 import org.bukkit.entity.Player;
@@ -31,8 +30,6 @@ import io.github.thebusybiscuit.slimefun4.core.handlers.MultiBlockInteractionHan
 import io.github.thebusybiscuit.slimefun4.implementation.SlimefunPlugin;
 import io.github.thebusybiscuit.slimefun4.implementation.items.blocks.OutputChest;
 import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
-
-import io.papermc.lib.PaperLib;
 
 import me.mrCookieSlime.Slimefun.Lists.RecipeType;
 import me.mrCookieSlime.Slimefun.Objects.Category;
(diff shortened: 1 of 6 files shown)


J3fftw1:
> The `InvUtils` import stmt belongs lower (with the other io.xxx stuff). Other than that, the 'play sound' happens in every branch (other than the timing one) and therefore could be extracted but it might not be necessary, just thought I'd mention it.
> 
> You could also try to find another output chest if the output does not fit into the one that was found before the crafting process started (there can be more than 1 afaik).

Im in favor of extracting it since people that have addons with multi blocks can also easily adjust then. (I wont need to copy over this method then since all multiblock crafting will have this issue)

EpicPlayerA10:
> 
> 
> The `InvUtils` import stmt belongs lower (with the other io.xxx stuff). Other than that, the 'play sound' happens in every branch (other than the timing one) and therefore could be extracted but it might not be necessary, just thought I'd mention it.
> 
> You could also try to find another output chest if the output does not fit into the one that was found before the crafting process started (there can be more than 1 afaik).

I thought about items will be added to another output chest but i need to modify `findOutputChestFor` method in `OutputChest` class. And i think it could break some other machines.

martinbrom:
> I thought about items will be added to another output chest but i need to modify `findOutputChestFor` method in `OutputChest` class. And i think it could break some other machines.

why not just call it again? it won't return the original once since the items won't fit into that one

EpicPlayerA10:
> why not just call it again? it won't return the original output chest since the items won't fit into that one

Hmm, wait a second

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
```diff
@@ -90,8 +91,16 @@ private void craft(Player p, ItemStack output, Inventory inv, Block dispenser) {
                     if (current < 3) {
                         p.getWorld().playSound(p.getLocation(), Sound.BLOCK_ANVIL_USE, 1F, 2F);
                     } else {
+                        Inventory outputInv2 = findOutputInventory(output, dispenser.getBlock(), dispenser.getInventory());
                         p.getWorld().playSound(p.getLocation(), Sound.ENTITY_ARROW_HIT_PLAYER, 1F, 1F);
-                        outputInv.addItem(output);
+                        if (outputInv2 != null) {
+                            outputInv2.addItem(output);
+                        } else if (InvUtils.fits(dispenser.getInventory(), output)) {
```
WalshyDev, line 98: You should cache these getInventory calls, it will make a new one every time.

(ref: https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/block/CraftDispenser.java#30-37)

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
EpicPlayerA10: okay

TheBusyBiscuit (changes requested):
This all seems like a lot of duplicated code to me, might be worth to create a method in the super class and just call that

EpicPlayerA10:
> This all seems like a lot of duplicated code to me, might be worth to create a method in the super class and just call that

Better?

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -165,6 +165,29 @@ public void load() {
...
+    /**
+     * This method handles a output {@link ItemStack} from the {@link MultiBlockMachine} which has a crafting delay
+     *
+     * @param outputItem
+     *          A crafted {@link ItemStack} from {@link MultiBlockMachine}
+     * @param dispenser
+     *          Our {@link Dispenser} from {@link MultiBlockMachine}
+     *
+     */
+    protected void handleCraftedItem(ItemStack outputItem, Dispenser dispenser) {
```
WalshyDev, line 177: I would prefer we use `Container` instead of `Dispenser`. I don't see a reason to limit this to ONLY dispensers. This could easily be a chest, trapped chest, furnace even.

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
EpicPlayerA10: Okay, I will change it

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -165,6 +168,30 @@ public void load() {
...
+     * @param outputItem
+     *          A crafted {@link ItemStack} from {@link MultiBlockMachine}
+     * @param container
+     *          Our main {@link Container} from {@link MultiBlockMachine}
+     *
+     */
+    @ParametersAreNonnullByDefault
+    protected void handleCraftedItem(ItemStack outputItem, Container container) {
+        Inventory containerInv = container.getInventory();
+        Inventory outputInv = findOutputInventory(outputItem, container.getBlock(), containerInv);
```
TheBusyBiscuit, line 183: `Container#getBlock()` is a relatively heavy operation as it does not cache the `Block` internally, whereas we already gotten the `Block` object before-hand, making this completely avoidable.

In other words: This `Block` object should probably be passed via a parameter here to reduce the performance impact of this method.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/ArmorForge.java
```diff
@@ -70,9 +70,9 @@ private boolean isCraftable(Inventory inv, ItemStack[] recipe) {
     }
 
     @ParametersAreNonnullByDefault
-    private void craft(Player p, ItemStack output, Inventory inv, Block dispenser) {
+    private void craft(Player p, ItemStack output, Inventory inv, Dispenser dispenser) {
         Inventory fakeInv = createVirtualInventory(inv);
-        Inventory outputInv = findOutputInventory(output, dispenser, inv, fakeInv);
+        Inventory outputInv = findOutputInventory(output, dispenser.getBlock(), inv, fakeInv);
```
TheBusyBiscuit, line 75: Same here, `#getBlock()` is a bit expensive and completely avoidable by just passing the `Block` object via a parameter

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/MagicWorkbench.java
```diff
@@ -68,9 +68,10 @@ public void onInteract(Player p, Block b) {
         }
     }
 
-    private void craft(Inventory inv, Block dispenser, Player p, Block b, ItemStack output) {
+    @ParametersAreNonnullByDefault
+    private void craft(Inventory inv, Dispenser dispenser, Player p, Block b, ItemStack output) {
         Inventory fakeInv = createVirtualInventory(inv);
-        Inventory outputInv = findOutputInventory(output, dispenser, inv, fakeInv);
+        Inventory outputInv = findOutputInventory(output, dispenser.getBlock(), inv, fakeInv);
```
TheBusyBiscuit, line 74: Same here, `#getBlock()` is a bit expensive and completely avoidable by just passing the `Block` object via a parameter

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/MagicWorkbench.java
```diff
@@ -89,24 +90,23 @@ private void craft(Inventory inv, Block dispenser, Player p, Block b, ItemStack
...
     }
 
-    private void startAnimation(Player p, Block b, Inventory inv, ItemStack output) {
+    private void startAnimation(Player p, Block b, ItemStack output, Dispenser dispenser) {
         for (int j = 0; j < 4; j++) {
             int current = j;
             SlimefunPlugin.runSync(() -> {
                 p.getWorld().playEffect(b.getLocation(), Effect.MOBSPAWNER_FLAMES, 1);
                 p.getWorld().playEffect(b.getLocation(), Effect.ENDER_SIGNAL, 1);
-
```
TheBusyBiscuit, line 104: Why was this line removed, can you please put it back?

TheBusyBiscuit (changes requested):
This can be optimized.

Also, thanks for hooking into the new item-spawning system! :)

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -165,6 +171,33 @@ public void load() {
...
+     * @param outputItem
+     *          A crafted {@link ItemStack} from {@link MultiBlockMachine}
+     * @param block
+     *          Main {@link Block} of our {@link Container} from {@link MultiBlockMachine}
+     *
+     */
+    @ParametersAreNonnullByDefault
+    protected void handleCraftedItem(ItemStack outputItem, Block block) {
+        BlockState state = PaperLib.getBlockState(block, false).getState();
+        Validate.isTrue(state instanceof Container);
```
TheBusyBiscuit, line 186: Always provide a message alongside `Validate` statements.

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -165,6 +171,33 @@ public void load() {
...
+        Container container = (Container) state;
+        Inventory containerInv = container.getInventory();
```
TheBusyBiscuit, line 188: Well, now we have the same paradox as before, getting the BlockState is an expensive operation and we already gotten the BlockState at this point, so this will hurt performance. The BlockState needs to be passed as an argument as much as the Block does.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -165,6 +168,31 @@ public void load() {
...
+        Inventory outputInv = findOutputInventory(outputItem, block, blockInv);
+
+        if (outputInv != null) {
+            outputInv.addItem(outputItem);
+        } else if (InvUtils.fits(blockInv, outputItem)) {
+            blockInv.addItem(outputItem);
+        } else {
+            // fallback
+            SlimefunUtils.spawnItem(block.getLocation(), outputItem, ItemSpawnReason.MULTIBLOCK_MACHINE_OVERFLOW, true);
+        }
```
TheBusyBiscuit, line 193: Tiny improvement to improve performance even further, other than this I consider it approved.

```suggestion
        } else {
            ItemStack rest = blockInv.addItem(outputItem).get(0);

            // fallback: drop item
            if (rest != null) {
                SlimefunUtils.spawnItem(block.getLocation(), rest, ItemSpawnReason.MULTIBLOCK_MACHINE_OVERFLOW, true);
            }
        }
```

TheBusyBiscuit (changes requested):

TheBusyBiscuit (commented):
LGTM

TheBusyBiscuit (approved):
I said: LGTM...

svr333 (approved):
👍🏿

PR #3146, opened 2021-07-04
Minor changes to the ticker task

md5sha256:
## Description

Shutdown the executor in TickerTask properly. Add nextThreadName to TickerTaskFactory

## Proposed changes

- Shutdown the `asyncExecutor` when finishing up.
- Add a numerical id to the `Thread` names in `TikcerTaskFactory`

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -249,8 +249,17 @@ public final class TickerTask {
         /*
          * If the plugin has been disabled,
          * do not schedule the next execution.
+         *
+         * Additionally, shutdown the async executor
+         * if that has not already been done so.
+         * We do not this TickerTask instance to be
+         * reused following plugin shutdown. Therefore,
+         * we shutdown the ExecutorService.
          */
         if (!plugin.isEnabled()) {
+            if (!asyncExecutor.isShutdown()) {
+                asyncExecutor.shutdown();
+            }
             return;
         }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -423,6 +423,7 @@ public final class TickerTask {
      * it will not block for the async tasks
      */
     public synchronized void halt() {
+        SlimefunPlugin.logger().info("Stopping TickerTask, this may take a few seconds or up to a minute...");
         /*
          * Mark the ticker as halted so that #tick won't attempt to schedule another task
          */
@@ -434,11 +435,12 @@ public final class TickerTask {
         running = false;
         // Process async tasks and shutdown the asyncExecutor
         asyncExecutor.shutdown();
+        asyncExecutor.awaitTermination(1, TimeUnit.MINUTES);
         // Process remaining sync tasks
         while (!syncTasks.isEmpty()) {
             syncTasks.poll().run();
         }
-
+    SlimefunPlugin.logger().info("Stopped TickerTask!");
     }
 
     @ParametersAreNonnullByDefault
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -251,7 +251,6 @@ public final class TickerTask {
         /*
          * If the plugin has been disabled,
          * do not schedule the next execution.
-         *
          */
         if (!plugin.isEnabled()) {
             return;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -440,7 +440,7 @@ public final class TickerTask {
         while (!syncTasks.isEmpty()) {
             syncTasks.poll().run();
         }
-    SlimefunPlugin.logger().info("Stopped TickerTask!");
+        SlimefunPlugin.logger().info("Stopped TickerTask!");
     }
 
     @ParametersAreNonnullByDefault
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -240,7 +240,9 @@ public final class TickerTask {
          * We're not done with the synchronous tickers yet.
          * Give the server a breather and let it progress one tick.
          */
-        Bukkit.getScheduler().runTaskLater(plugin, this::processSyncTasks, 2L);
+        if (!halted) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerThreadFactory.java
```diff
@@ -43,7 +46,11 @@ int getThreadCount() {
      */
     @Override
     public Thread newThread(@Nonnull Runnable runnable) {
-        return new Thread(runnable, "Slimefun Ticker Thread");
+        return new Thread(runnable, nextThreadName());
+    }
+
+    private @Nonnull String nextThreadName() {
+        return String.format("Slimefun Ticker Thread-%d", threadNumber.getAndIncrement());
```
WalshyDev, line 53: I see no reason to use String#format here

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -249,8 +249,17 @@ private void finish() {
...
+         *
+         * Additionally, shutdown the async executor
+         * if that has not already been done so.
+         * We do not this TickerTask instance to be
+         * reused following plugin shutdown. Therefore,
+         * we shutdown the ExecutorService.
          */
         if (!plugin.isEnabled()) {
+            if (!asyncExecutor.isShutdown()) {
+                asyncExecutor.shutdown();
```
WalshyDev, line 261: We really should be shutting this down in onDisable

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
md5sha256: So an explicit call to a  `TickerTask#shutdownAsyncTasks` in `Plugin#onDisable`

WalshyDev: Yep.

WalshyDev: We should have a stop method which cleans up and stops everything. So, the executor gets shut down, pending tasks get cleared and the queues get processed and cleared.

md5sha256: Got it.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -416,9 +419,27 @@ public boolean isHalted() {
...
+         * Mark the ticker as halted so that #tick won't attempt to schedule another task
+         */
         halted = true;
+        tick();
+        /*
+         * Mark running after #tick so #processAsyncTasks gets scheduled
+         */
+        running = false;
+        // Process async tasks and shutdown the asyncExecutor
+        asyncExecutor.shutdown();
```
WalshyDev, line 437: We should set an awaitTermination. We need it for 2 reasons: 1. Things shouldn't be able to run forever and 2. It's needed to process tasks still in queue

```
     * Initiates an orderly shutdown in which previously submitted
     * tasks are executed, but no new tasks will be accepted.
     * Invocation has no additional effect if already shut down.
     *
     * <p>This method does not wait for previously submitted tasks to
     * complete execution.  Use {@link #awaitTermination awaitTermination}
     * to do that.
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -249,6 +251,7 @@ private void finish() {
         /*
          * If the plugin has been disabled,
          * do not schedule the next execution.
+         *
```
WalshyDev, line 254:
```suggestion
```

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
md5sha256: I was thinking that too, but wasn't sure what to have. Should we have it as 1 minute?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -416,9 +419,27 @@ public boolean isHalted() {
 
     /**
      * This method orders the {@link TickerTask} to be halted.
+     * Once halted the task <strong>cannot</strong> be un-halted.
+     * This method will block until all sync tasks are processed; however,
+     * it will not block for the async tasks
      */
     public synchronized void halt() {
+        /*
```
WalshyDev, line 426:
```suggestion
    public synchronized void halt() {
        SlimefunPlugin.logger().info("Stopping TickerTask, this may take a few seconds or up to a minute...");
        /*
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -416,9 +419,27 @@ public boolean isHalted() {
...
+         * Mark running after #tick so #processAsyncTasks gets scheduled
+         */
+        running = false;
+        // Process async tasks and shutdown the asyncExecutor
+        asyncExecutor.shutdown();
+        // Process remaining sync tasks
+        while (!syncTasks.isEmpty()) {
+            syncTasks.poll().run();
+        }
+
```
WalshyDev, line 441:
```suggestion
    SlimefunPlugin.logger().info("Stopped TickerTask!");
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
WalshyDev: Yeah a minute should be fine
```suggestion
        asyncExecutor.shutdown();
        asyncExecutor.awaitTermination(1, TimeUnit.MINUTES);
```

WalshyDev (changes requested):
Logging is good here too, if there's a delay people should know why

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -416,9 +418,29 @@ public boolean isHalted() {
...
+         */
+        running = false;
+        // Process async tasks and shutdown the asyncExecutor
+        asyncExecutor.shutdown();
+        asyncExecutor.awaitTermination(1, TimeUnit.MINUTES);
+        // Process remaining sync tasks
+        while (!syncTasks.isEmpty()) {
+            syncTasks.poll().run();
+        }
+    SlimefunPlugin.logger().info("Stopped TickerTask!");
```
WalshyDev, line 443:
```suggestion
        SlimefunPlugin.logger().info("Stopped TickerTask!");
```

WalshyDev (commented):
grrrr

WalshyDev (approved):
LGTM

TheBusyBiscuit (approved):
Sure, the blob approved.

PR #3148, opened 2021-07-05
Fix compile error in TickerTask

md5sha256:
## Checklist


TheBusyBiscuit (commented):
Wasn't there something like "you are supposed to re-interrupt the current Thread"?

md5sha256:
well, this is meant to be called on shutdown from main. In theory yeah if we wanted to propagate the exception that's what we would do. But given that this is a termination method I don't think we want to propagate.

TheBusyBiscuit (approved):

Issue #2901, opened 2021-03-22
Auto-crafters do not craft items even though the recipe product can fit in the chest


martinbrom:
## :round_pushpin: Description (REQUIRED)

Autocrafters only check for a completely empty spot in the chest and disregard any spots which contain the same item as the recipe product, see screenshot below.

## :bookmark_tabs: Steps to reproduce the Issue (REQUIRED)

1. Fill the auto crafter chest with correct ingredients
2. Fill one chest slot with the same item as the product of the selected recipe
3. Fill the rest of the chest with other items

![SF Screenshot](https://user-images.githubusercontent.com/16322387/112024428-786e8000-8b34-11eb-871c-a451263ac3c9.png)

## :bulb: Expected behavior (REQUIRED)

The recipe is used and the product is added to the chest

I feel like AbstractAutoCrafter.java#L342 should not only check for an empty slot but rather check if the product of the recipe fits.

## :scroll: Server Log

Unfortunately do not have access

## :open_file_folder: /error-reports/ Folder

Unfortunately do not have access

## :compass: Environment (REQUIRED)

 - Server Software: not sure
 - Minecraft Version: 1.16.5-R0.1-SNAPSHOT
 - Slimefun Version: DEV 847 (git 001e7654)

TheBusyBiscuit:
Auto Crafters require an empty slot to craft the item.
It is intended to work this way.

LilBC:
Awww... Sad, why tho? Is it just because the auto crafter will be too op if it works like that?

PR #3149, opened 2021-07-05
Fix Auto Crafter Empty Slot Check

LilBC:
## Description

There is an empty slot checking bug in the auto crafters. They only check if there is a slot in the chest that is completely empty. But if there is a stack of item in the chest that is not full but same item as the result of the recipe, the auto crafter will stop crafting because it thinks there are no empty slots.

## Proposed changes

I added a new method in the Abstract Auto Crafter file that checks if there is a stack of item that is the same type if there are no slots that are completely empty.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
@@ -398,11 +398,7 @@ public abstract class AbstractAutoCrafter extends SlimefunItem implements Energy
 
             for (ItemStack matchingItem : inv.getContents()) {
                 // Checks if any item stack that is not full that is the same type as the result.
-                if (
-                        ((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem))
-                        || ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true)))
-                        && matchingItem.getAmount() < matchingItem.getMaxStackSize()
-                ){
+                if (((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem)) || ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true))) && matchingItem.getAmount() < matchingItem.getMaxStackSize()){
                     return true;
                 }
             }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
@@ -10,6 +10,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
 import org.apache.commons.lang.Validate;
 import org.bukkit.Location;
 import org.bukkit.Material;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
```diff
@@ -391,6 +392,29 @@ protected boolean matchesAny(Inventory inv, Map<Integer, Integer> itemQuantities
...
+
+        if(inv.firstEmpty() < 0){
+
+            for (ItemStack matchingItem : inv.getContents()) {
+                // Checks if any item stack that is not full that is the same type as the result.
+                if (
+                        ((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem))
+                        || ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true)))
+                        && matchingItem.getAmount() < matchingItem.getMaxStackSize()
+                ){
```
J3fftw1, line 405: formatting :notLikeThis:

martinbrom:
Duplicate of https://github.com/Slimefun/Slimefun4/issues/2901

martinbrom:
@LilBC couldnt you use InvUtils.fits() instead?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
LilBC: Ops, sorry. I read the code style guidelines, I just didn't know what does line-wrapping means.

LilBC:
> Duplicate of #2901
> 
> I wanted to implement this as soon as the new autocrafters came out but 🍪 was against it. Felt like it would simplify using the autocrafters...
> 
> EDIT: thebusybot marked as duplicate even though I am not a member??

Ooops, my bad. I should've searched for this issue before I start coding. I guess I'll wait for 🍪 to see this and see what do they think... Probably gonna get closed. Sad... wasted too much time coding. :(

LilBC:
> @LilBC couldnt you use InvUtils.fits() instead?

Cool, I didn't know this method existed. I'll try to change to using it.

martinbrom:
> Edit: Hmm... I don't think it would work... I think it only works on Dirty Chest Menus ...

Not true, the contract of the method is `boolean fits(@NonNull Inventory inv, @NonNull ItemStack item, int... slots)`

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
```diff
@@ -414,8 +434,8 @@ public boolean craft(@Nonnull Inventory inv, @Nonnull AbstractRecipe recipe) {
             return false;
         }
 
-        // Check if we have an empty slot
-        if (inv.firstEmpty() != -1) {
+        // Check if the inventory have any available space
+        if (inventoryAvailable(inv, recipe.getResult())) {
```
martinbrom, line 438:
```suggestion
        if (InvUtils.fits(inv, recipe.getResult())) {
```

martinbrom (commented):
This would be the only change you'd need to do in case :cookie: changed his mind.
Also would need to add the import manually, the suggestions don't do that on their own.

Your code wouldn't work if the recipe produced more than 1 item. It's usually better to use existing code instead of writing everything yourself.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
```diff
@@ -391,6 +392,25 @@ protected boolean matchesAny(Inventory inv, Map<Integer, Integer> itemQuantities
...
+        if(inv.firstEmpty() < 0){
+
+            for (ItemStack matchingItem : inv.getContents()) {
+                // Checks if any item stack that is not full that is the same type as the result.
+                if (((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem)) || ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true))) && matchingItem.getAmount() < matchingItem.getMaxStackSize()){
+                    return true;
+                }
+            }
+            return false;
+        }
```
J3fftw1, line 406:
```suggestion
        if (inv.firstEmpty() < 0) {

            for (ItemStack matchingItem : inv.getContents()) {
                // Checks if any item stack that is not full that is the same type as the result.
                if (((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem)) 
                	|| ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true))) 
                	&& matchingItem.getAmount() < matchingItem.getMaxStackSize()
                ){
                    return true;
                }
            }
            return false;
        }
```

J3fftw1 (changes requested):

TheBusyBiscuit:
Rejecting this for the same reason as last time.
It's intended to work this way. Item comparisons are very expensive and this change would turn Auto-Crafters into absolute lag machines.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
LilBC: Ohhh ~ Okay! Thank you. Sorry I'm just kinda noob I didn't know how to use this method. Very noice.

LilBC:
> Rejecting this for the same reason as last time.
> It's intended to work this way. Item comparisons are very expensive and this change would turn Auto-Crafters into absolute lag machines.

Can I do it like after the player sets and saves the recipe, the autocrafter searches for and saves the first same item slot to only check that slot every tick. So it doesn't searchs for it and compare all of the items every tick. I think this feature is pretty important for automation since it makes automation so much easier.

PR #3155, opened 2021-07-07
Auto Crafter Empty Slot Check

LilBC:
## Description

Making auto crafters able to check and craft when there are no empty slots in the chest. I think this is a pretty important feature since it would make automation a lot easier. A while ago I made a pull request (#3149) making the auto crafters check every slot in the chest for any non-full slot with the same item as the result of the recipe. But Biscuit said item comparisons are expensive so that would be laggy.

## Proposed changes

I'm thinking should I add a new small option in the auto crafter menu that allows the player to choose an "output slot" so only the output slot is checked every tick.

I haven't started coding this yet, so ignore the commits added to this. I just wanna make a draft first to see what people do think about this. I can start coding it if most of us think it's not a bad idea.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
@@ -10,6 +10,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
 import org.apache.commons.lang.Validate;
 import org.bukkit.Location;
 import org.bukkit.Material;
@@ -391,6 +392,29 @@ public abstract class AbstractAutoCrafter extends SlimefunItem implements Energy
         return false;
     }
 
+    public boolean inventoryAvailable(@Nonnull Inventory inv, @Nonnull ItemStack resultItem) {
+
+        if(inv.firstEmpty() < 0){
+
+            for (ItemStack matchingItem : inv.getContents()) {
+                // Checks if any item stack that is not full that is the same type as the result.
+                if (
+                        ((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem))
+                        || ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true)))
+                        && matchingItem.getAmount() < matchingItem.getMaxStackSize()
+                ){
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        // Returns true if the first empty slot in the inventory is >= 0, which means there is an empty slot.
+        else {
+            return true;
+        }
+    }
+
     /**
      * This method performs a crafting operation.
      * It will attempt to fulfill the provided {@link AbstractRecipe} using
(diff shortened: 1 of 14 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
```diff
@@ -391,6 +392,25 @@ protected boolean matchesAny(Inventory inv, Map<Integer, Integer> itemQuantities
         return false;
     }
 
+    public boolean inventoryAvailable(@Nonnull Inventory inv, @Nonnull ItemStack resultItem) {
+
+        if(inv.firstEmpty() < 0){
```
J3fftw1, line 397:
```suggestion
        if (inv.firstEmpty() < 0) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
```diff
@@ -391,6 +392,25 @@ protected boolean matchesAny(Inventory inv, Map<Integer, Integer> itemQuantities
...
+
+            for (ItemStack matchingItem : inv.getContents()) {
+                // Checks if any item stack that is not full that is the same type as the result.
+                if (((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem)) || ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true))) && matchingItem.getAmount() < matchingItem.getMaxStackSize()){
+                    return true;
+                }
+            }
```
J3fftw1, line 404:
```suggestion
            for (ItemStack matchingItem : inv.getContents()) {
                // Checks if any item stack that is not full that is the same type as the result.
                if (((!(resultItem instanceof SlimefunItemStack) && resultItem.isSimilar(matchingItem)) 
                || ((resultItem instanceof SlimefunItemStack) && SlimefunUtils.isItemSimilar(matchingItem, resultItem, true))) 
                && matchingItem.getAmount() < matchingItem.getMaxStackSize()
                ){
                    return true;
                }
            }
```

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/autocrafters/AbstractAutoCrafter.java
```diff
@@ -10,6 +10,7 @@
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
```
martinbrom, line 13: should be lower, just use the reorder imports action in your IDE (hopefully there should be one)

martinbrom:
Isn't this the exact same code that :cookie: rejected?

My previous comment about recipes that output more than 1 item is still valid, the current implementation won't start the crafting process if all of the outputs do not fit, yours would.
You'd need to do sth like this instead. There are recipes where the result amount != 1.
```
matchingItem.getAmount() + resultItem.getAmount() <= matchingItem.getMaxStackSize()
```

LilBC:
Guys no, don't mind about the code for now! XD I'm sorry if I didn't made it clear, I still haven't started coding for it yet... This PR is just proposing the idea. Ahhh... Imma remove the changes in the code.

LilBC:
I'm sorry, this PR is getting really messy and unclear. I will close it and open a new one.

PR #3157, opened 2021-07-08
Auto Crafter Empty Slot Check

LilBC:
## Description

Making auto crafters able to check and craft when there are no empty slots in the chest. I think this is a pretty important feature since it would make automation a lot easier. A while ago I made a pull request (#3149) making the auto crafters check every slot in the chest for any non-full slot with the same item as the result of the recipe. But Biscuit said item comparisons are expensive so that would be laggy.

## Proposed changes

I'm thinking should I add a new small option in the auto crafter menu that allows the player to choose an "output slot" so only the output slot is checked every tick.

I haven't started coding this yet, so ignore the commits added to this. I just wanna make a draft first to see what people do think about this. I can start coding it if most of us think it's not a bad idea.


TheBusyBiscuit (commented):
This is literally an empty pull request...
Not sure if you created this by mistake or not but there is no point in having a pull request which adds nothing.

You can draft a pull request when you got actual code to show/propose.

PR #3158, opened 2021-07-08
Smart Filling Mode For Single Stack of Item

LilBC:
## Description

I'm making this PR draft so I can gather opinions for people to see if this would be an ok feature to be added into the game.

## Proposed changes

I want to add a new feature to the smart filling mode so that it not only transfers stack of items but also single items when there is only one stack of item. More specifically, when smart filling mode is enabled, it doesn't transfer the whole stack of item if there is only one stack of that type of item in the container. Instead, it will leave one item in the slot and transfer the rest. That reason this feature would be useful is because the one item left in the slot can serve as a filter, so that only that type of item can be put into that slot.


LilBC:
I'm not sure which file to edit though. CargoNetworkTask maybe?

TheBusyBiscuit (commented):
Another empty pull request.

PR #3169, opened 2021-07-11
Allow the Industrial Miner to produce raw ores rather than ore blocks

guizhanss:
## Description

 I suggested that the industrial miner should produce raw metals (iron, gold, and copper) instead of ores in 1.17, but it is not approved yet. 

## Proposed changes

Changed outcomes for industrial miner:
In 1.17, the basic industrial miner will produce raw metals (iron, gold, and copper) instead of metal ores.
In 1.17, the basic industrial miner will mine deepslate ores just like a pickaxe.
In 1.16, the basic industrial miner will produce gold nugget instead of nether gold ore.

[discord suggestion #1400](https://discord.com/channels/565557184348422174/693130800853418055/859575052173312030)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/IndustrialMiner.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/IndustrialMiner.java
@@ -126,22 +126,79 @@ public class IndustrialMiner extends MultiBlockMachine {
 
         Random random = ThreadLocalRandom.current();
 
-        switch (ore) {
-            case COAL_ORE:
-                return new ItemStack(Material.COAL);
-            case DIAMOND_ORE:
-                return new ItemStack(Material.DIAMOND);
-            case EMERALD_ORE:
-                return new ItemStack(Material.EMERALD);
-            case NETHER_QUARTZ_ORE:
-                return new ItemStack(Material.QUARTZ);
-            case REDSTONE_ORE:
-                return new ItemStack(Material.REDSTONE, 4 + random.nextInt(2));
-            case LAPIS_ORE:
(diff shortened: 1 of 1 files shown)


WalshyDev:
There's a ton of duplicate code here. We don't need like COAL_ORE in every single condition.

TheBusyBiscuit (approved):
LGTM

TheBusyBiscuit (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/IndustrialMiner.java
```diff
@@ -126,19 +126,50 @@ protected void registerDefaultFuelTypes() {
 
         Random random = ThreadLocalRandom.current();
 
+        if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_17)) {
+            // In 1.17, breaking metal ores should get raw metals. Also support deepslate ores.
+            switch (ore) {
+                case DEEPSLATE_COAL_ORE:
+                    return new ItemStack(Material.COAL);
```
Mooy1, line 133: can't coal drop more than 1?

ybw0014: This is the same as coal ore.

TheBusyBiscuit: @Mooy1 Only if you use fortune, coal ore always only dropped 1 coal item.

Minecraft wiki:
> "Coal ore mined using a pickaxe drops 1 unit of coal."

WalshyDev (approved):
LGTM

PR #3179, opened 2021-07-23
 Fix a flaky test

hwang-pku:
A flaky test is found in this project and this PR proposes to fix it.

1. How to reproduce the flaky test
- The project is built and tested under `javaJDK8`.
- The tool used for the flaky test detection is [NonDex](https://github.com/TestingResearchIllinois/NonDex).
- I ran `mvn edu.illinois:nondex-maven-plugin:1.1.2:nondex -Dtest=io.github.thebusybiscuit.slimefun4.api.items.settings.TestMaterialTagSetting#testAllowedValue`, which produces
> [ERROR] Failures:  
> [ERROR]   TestMaterialTagSetting.testAllowedValue:71 iterable contents differ at index [3], expected: <BIRCH_LOG> but was: <OAK_WOOD>

2. Why the test failed
- The method `org.bukkit.Tag.getValues()` called in the test  `io.github.thebusybiscuit.slimefun4.api.items.settings.TestMaterialTagSetting.testAllowedValue` returns a `set`, which does not guarantee the order of elements in it and in turn causes unpredictability in the order of elements in the `list` created from it. This causes the `assertIterableEquals` to fail, which depends on the order retained by the list.

3. The changes made
- `HashMap` is used to compare two lists independent of the order of the elements in them.
- In order to use `org.hamcrest.MatcherAssert.assertThat`, corresponding dependencies are added to `pom.xml`.


TheBusyBiscuit (changes requested):
The formatting is completely off here...
And I really don't see the point of adding the entire hamcrest library just for this single line.

If the formatting issues are fixed and a hamcrest-free implementation is used, I may consider it, thanks for making us aware though.

src/test/java/io/github/thebusybiscuit/slimefun4/api/items/settings/TestMaterialTagSetting.java
```diff
@@ -68,7 +69,7 @@ void testAllowedValue() {
         item.register(plugin);
 
         List<String> tagContents = tag.getValues().stream().map(Material::name).collect(Collectors.toList());
-        Assertions.assertIterableEquals(tagContents, setting.getValue());
+        Assertions.assertTrue(new HashSet<>(tagContents).equals(new HashSet<>(setting.getValue())));
```
WalshyDev, line 72:
```suggestion
        Assertions.assertEquals(tagContents, setting.getValue());
```

WalshyDev (changes requested):

hwang-pku:
Hi, it appears I used tabs instead of spaces so the formatting went off in github. I used junit instead to avoid unnecessary dependency. It works just the same.

src/test/java/io/github/thebusybiscuit/slimefun4/api/items/settings/TestMaterialTagSetting.java
hwang-pku: Hi, the point with converting the list to a hashSet here is to avoid comparing the two lists with respect to both elements and their orders. I tested the new line you're suggesting and it appears that `Assertions.assertEquals` still take into account the orders in the list, so the nondex still fails.

WalshyDev: Yes, my fault for not reading the description of this PR 😄

TheBusyBiscuit (approved):
Thank you for addressing these issues and for fixing this flaky test in the first place!

Sfiguz7 (approved):

PR #3184, opened 2021-07-29
Fix disappearing sensitive blocks on slimefun blocks

### LINKED ISSUE #3182: Block Placer makes blocks vanish
### ❗ Checklist

### 📍 Description

When placing a block, which needs to be supported (eg: Redstone Dust, Repeater, Redstone Torch) on top of a Block Placer, when the Block Placer is broken, the block on top does not drop into item form.

I only checked with Redstone Dust, Repeaters, and Redstone Torches.
I broke the block placer first using a Netherite Pickaxe, only with Vanilla Enchants

### 📑 Reproduction Steps

1. Place a Block Placer
2. Place Any of the following on top of Block Placer
    - Redstone Dust
    - Redstone Repeater
    - Redstone Torch
3. Break the Block Placer

### 💡 Expected Behavior

When the Block Placer is broken, the block on top is expected to drop as an Item, which it does not.

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

Slimefun DEV - 956

[Screenshot of Version](https://gyazo.com/815101758121ba1a5973b98ed00e90f8)

### 🧭 Other plugins

_No response_

EpicPlayerA10:
## Description
Fix disappearing sensitive blocks on slimefun blocks

## Proposed changes
`BlockListener#dropItems` calls `e.setDropItems(false)` which also cancels drops from sensitive block above.
I added to `BlockListener#checkForSensitiveBlockAbove` check for non-slimefun sensitive blocks.
I also added some other sensitive materials to `sensitive_materials.json`.

Resolves #3182 

## Checklist


Sfiguz7 (approved):
Well spotted

svr333 (commented):
`sensitive_materials.json` uses tabs instead of spaces, now might be a good time to switch?

svr333 (dismissed):
Looks good to me

Sfiguz7 (dismissed):

svr333 (approved):

src/main/resources/tags/sensitive_materials.json
```diff
@@ -1,8 +1,28 @@
...
-	]
+    "values" : [
+        "#minecraft:saplings",
+        "#minecraft:carpets",
+        "#slimefun:torches",
+        "#slimefun:pressure_plates",
+        "minecraft:cake",
+        "minecraft:redstone_wire",
+        "minecraft:repeater",
+        "minecraft:comparator",
```
Sfiguz7, line 10:
```suggestion
        "minecraft:comparator",
        "minecraft:lantern",
        {
            "id" : "minecraft:soul_lantern",
            "required" : false
        },
```

Sfiguz7 (changes requested):
Really sorry about the delay, I think this is missing lanterns. Other than that looks good

Sfiguz7 (dismissed):
LGTM

TheBusyBiscuit (changes requested):
Just blocking this until I get a chance to look over the code.

Honestly, some code documentation could help 😬

EpicPlayerA10:
Closing this PR for now. We need to find a better solution for this bug.

EpicPlayerA10:
But there is also a bug where blocks that are attached on the side of slimefun block e.g. banner won't drop. This fix will be harder to implement.

J3fftw1:
Yea fair their will always be edge cases and will always be pain to update. But i think the changes you made should still be applied

EpicPlayerA10:
Ah, you convinced me

J3fftw1:
Since you reopened it, it would be nice to support 1.19 blocks aswell.

J3fftw1:
and maybe we need to think of a way to support wall blocks like banners torches paintings etc,
not sure if they even bug if you place them on it. 
needs testing

EpicPlayerA10:
There is a cool method in bukkit `Block#canPlace` but it has been added since 1.18

J3fftw1:
> But there is also a bug where blocks that are attached on the side of slimefun block e.g. banner won't drop. This fix will be harder to implement.

Mmh my message apparently didnt send yesterday. We should probably look in a way to just send physics updates to blocks around it. Not sure if thats have at all since we need all sides of the block being updated

J3fftw1 (approved):

PR #3131, opened 2021-06-21
Radiation update

Semisol:
## Description

Implementing an idea I had.

## Proposed changes

1. Add a radiation exposure meter
2. Depending on how high it is, you get symptoms then instant death
3. If the player holds any radioactive items and is not protected, their exposure is increased by the sum of the following algorithm for all radioactive item stacks every execution.
`itemCount * radiationLevel.exposureModifier`
4. If they are protected/do not hold any radioactive item the meter is decremented by 1 every execution
5. The actionbar displays the current meter if not 0
6. If you have been exposed, you get a chat message
7. Vitamins and other items that cure radiation reset the meter to 0
8. Executions by default happen every second but can be edited in config

## Related Issues (if applicable)

None

## Checklist

- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [x] I followed the existing code standards and didn't mess up the formatting.
- [x] I did my best to add documentation to any public classes or methods I added.
- [x] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

Note: I have disabled the old radiation unit test due to not being compatible.

PR #3189, opened 2021-08-01
Cumulative Armor-Scheduling changes (Rainbow Armor + Radiation update)

TheBusyBiscuit:
This is a cumulative pull requests consisting of:
- [x] #3059 
- [x] #3131


Semisol:
After the new radiation system was added, `armor-update-interval` (which is in seconds) causes delayed application of the hazmat suit, which would result in you getting up to 10 seconds without it applying.
Should this delay be reduced, and if done how?
We have three options:
- Not do it at all
- Do it, and ask users to change it down to one/two
- Do it, and convert `armor-update-interval` to ticks which would give a 0.5s update interval, along with `radiation-update-interval` also being converted to ticks and being set to 20.
- Do a config update converting 10 to 1, somehow.

TheBusyBiscuit:
> After the new radiation system was added, `armor-update-interval` (which is in seconds) causes delayed application of the hazmat suit, which would result in you getting up to 10 seconds without it applying.
> Should this delay be reduced, and if done how?
> We have four options:
> 
> * Not do it at all
> * Do it, and ask users to change it down to one/two
> * Do it, and convert `armor-update-interval` to ticks which would give a 0.5s update interval, along with `radiation-update-interval` also being converted to ticks and being set to 20.
> * Do a config update converting 10 to 1, somehow.

I am not quite sure I understand what you mean

Semisol:
> I am not quite sure I understand what you mean

Essentially after the update this was how radiation applied (1 line = 1 second):
```
radiation !
radiation !
radiation !
radiation ! 
radiation !
radiation ! 
radiation !
radiation !
radiation !
radiation armortask
...
```
and, if you wear a hazmat suit, it won't take effect until the next armortask

TheBusyBiscuit:
I see, perhaps the "armor update" should be moved to the super class then, we could implement a "minimum delay" to reduce redundant runs then.
In other words, every one of the "armor" tasks triggers an update to the cached armor but updates will be skipped if the last update has been less than x ticks ago.

Semisol:
> I see, perhaps the "armor update" should be moved to the super class then, we could implement a "minimum delay" to reduce redundant runs then.
> In other words, every one of the "armor" tasks triggers an update to the cached armor but updates will be skipped if the last update has been less than x ticks ago.

Looks good to me. I'll PR in my other changes.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/SlimefunArmorTask.java
```diff
@@ -0,0 +1,88 @@
...
+     *            The {@link Player} wearing the piece of armor
+     * @param sfArmorPiece
+     *            {@link SlimefunArmorPiece} Slimefun instance of the piece of armor
+     * @param armorPiece
+     *            The actual {@link ItemStack} of the armor piece
+     */
+    @ParametersAreNonnullByDefault
+    protected void onArmorPieceTick(Player p, SlimefunArmorPiece sfArmorPiece, ItemStack armorPiece) {
+        for (PotionEffect effect : sfArmorPiece.getPotionEffects()) {
+            p.removePotionEffect(effect.getType());
```
Sfiguz7, line 84: Why does this explicitly remove effects of the same type from the player before applying its own? This forcefully overrides higher level same-type effects or longer effects if the effect on the player is of equal or lower level.
We don't want to replace, for example, Jump Boost 5 with Jump Boost 2; similarly, since Spigot now supports multiple PotionEffects of the same type, instead of removing Jump Boost 1 to apply Jump Boost 2 we can just apply Jump Boost 2 and if by the time JB2 is over JB1 still has time left before expiring it will take over on its own

Sfiguz7 (changes requested):
Just the one comment

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/SlimefunArmorTask.java
Semisol: A fix can be made by making potion effects last `updateInterval + 1s`, then they will expire almost immediately after they have been removed.
That's how the radiation system I implemented does it.

Sfiguz7:
Formatting will need to be checked, see #3293
I messed up a couple things, will see if I can fix them but just in case I am putting a comment in

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RainbowArmorTask.java
```diff
@@ -0,0 +1,56 @@
...
+        for (int i = 0; i < 4; i++) {
+            ItemStack item = p.getInventory().getArmorContents()[i];
+
+            if (item != null && item.hasItemMeta()) {
+                HashedArmorpiece armorPiece = profile.getArmor()[i];
+
+                armorPiece.getItem().ifPresent(sfArmorPiece -> {
+                    if (sfArmorPiece instanceof RainbowArmorPiece && sfArmorPiece.canUse(p, true)) {
+                        updateRainbowArmor(item, (RainbowArmorPiece) sfArmorPiece);
+                    }
```
JustAHuman-xD, line 40: Can we use instanceof casting here instead of casting the `sfArmorPiece` again when calling the method?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/SlimefunArmorTask.java
```diff
@@ -0,0 +1,87 @@
...
+ * @author martinbrom
+ * @author Semisol
+ *
+ */
+public class SlimefunArmorTask extends AbstractArmorTask {
+
+    @Override
+    protected void onTick() {
+        // Do nothing
+    }
```
JustAHuman-xD, line 31: Is there any reason to override this?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/SolarHelmetTask.java
```diff
@@ -0,0 +1,71 @@
...
+ * @author Semisol
+ *
+ */
+public class SolarHelmetTask extends AbstractArmorTask {
+
+    @Override
+    protected void onTick() {
+        // Do nothing
+    }
+
```
JustAHuman-xD, line 37: Again is it necessary to override?

src/test/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TestArmorTask.java
```diff
@@ -64,55 +56,9 @@ void testSlimefunArmor() throws InterruptedException {
...
-     * It performs a cartesian product on two boolean sets.
-     * 
-     * @return a {@link Stream} of {@link Arguments}
-     */
-    private static Stream<Arguments> cartesianBooleans() {
-        Stream<Boolean> stream = Stream.of(true, false);
-        return stream.flatMap(a -> Stream.of(true, false).map(b -> Arguments.of(a, b)));
-    }
-
 }
```
JustAHuman-xD, line 64: Is there a plan to re-do the tests with all of the armor and radiation related changes?

JustAHuman-xD (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/radiation/RadiationSymptom.java
```diff
@@ -0,0 +1,67 @@
+package io.github.thebusybiscuit.slimefun4.core.radiation;
+
+import javax.annotation.Nonnull;
+
+import org.apache.commons.lang.Validate;
```
JustAHuman-xD, line 5:
```suggestion
import org.apache.commons.lang.Validate;

```

src/main/java/io/github/thebusybiscuit/slimefun4/core/radiation/RadiationSymptom.java
```diff
@@ -0,0 +1,67 @@
...
+    WITHER_HIGH(75, PotionEffectType.WITHER, 3),
+    IMMINENT_DEATH(100, PotionEffectType.HARM, 49);
+
+    private final int minExposure;
+    private final PotionEffect potionEffect;
+
+    RadiationSymptom(int minExposure, @Nonnull PotionEffectType type, int level) {
+        Validate.notNull(type, "The effect type cannot be null");
+        Validate.isTrue(minExposure > 0, "The minimum exposure must be greater than 0.");
+        Validate.isTrue(level >= 0, "The status effect level must be non-negative.");
```
JustAHuman-xD, line 37: Preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/core/radiation/RadiationSymptom.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * This method applies the symptom to a player.
+     *
+     * @param p
+     *            The player
+     */
+
+    public void apply(@Nonnull Player p) {
+        Validate.notNull(p, "The player cannot be null");
```
JustAHuman-xD, line 51: Preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/armor/RainbowArmorPiece.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param item The {@link SlimefunItemStack} that describes the visual features of our {@link RainbowArmorPiece}
+     * @param recipeType the {@link RecipeType} that determines how this {@link RainbowArmorPiece} is crafted
+     * @param recipe An Array representing the recipe of this {@link RainbowArmorPiece}
+     * @param dyeColors An Array representing the {@link DyeColor}s this {@link RainbowArmorPiece} will cycle between
+     */
+    @ParametersAreNonnullByDefault
+    public RainbowArmorPiece(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, DyeColor[] dyeColors) {
+        super(itemGroup, item, recipeType, recipe, new PotionEffect[0]);
+
+        Validate.notEmpty(dyeColors, "RainbowArmorPiece colors cannot be empty!");
```
JustAHuman-xD, line 41: preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/AbstractArmorTask.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Schedules this {@link AbstractArmorTask} to run every {@code tickInterval} ticks
+     *
+     * @param plugin
+     *            The {@link Slimefun}
+     * @param tickInterval
+     *            Delay between two "runs" of this task in ticks
+     */
+    public final void schedule(@Nonnull Slimefun plugin, long tickInterval) {
+        Validate.notNull(plugin, "The plugin instance cannot be null!");
```
JustAHuman-xD, line 52: Preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/utils/RadiationUtils.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public final class RadiationUtils {
+
+    private static final StatusEffect RADIATION_EFFECT = new StatusEffect(new NamespacedKey(Slimefun.instance(), "radiation"));
+    private static final int MAX_EXPOSURE_LEVEL = 100;
+
+    private RadiationUtils() {}
+
+    public static void clearExposure(@Nonnull Player p) {
+        Validate.notNull(p, "The player cannot be null");
```
JustAHuman-xD, line 29: Preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/utils/RadiationUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+        RADIATION_EFFECT.clear(p);
+    }
+
+    public static int getExposure(@Nonnull Player p) {
+        Validate.notNull(p, "The player must not be null");
```
JustAHuman-xD, line 35: Preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/utils/RadiationUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+        return RADIATION_EFFECT.getLevel(p).orElse(0);
+    }
+
+    public static void addExposure(@Nonnull Player p, int exposure) {
+        Validate.notNull(p, "The player cannot be null");
```
JustAHuman-xD, line 41: Preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/utils/RadiationUtils.java
```diff
@@ -0,0 +1,53 @@
...
+
+        int level = Math.min(RADIATION_EFFECT.getLevel(p).orElse(0) + exposure, MAX_EXPOSURE_LEVEL);
+        RADIATION_EFFECT.addPermanent(p, level);
+    }
+
+    public static void removeExposure(@Nonnull Player p, int exposure) {
+        Validate.notNull(p, "The player should not be null");
```
JustAHuman-xD, line 48: Preconditions

JustAHuman-xD (changes requested):

J3fftw1:
Handled in #3892

Issue #3190, opened 2021-08-01
Raw ores not working with the luck multiplier of Slimefun Furnaces


marvjosh1:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

When raw ores (iron, gold, copper) are smelted in enhanced / reinforced / carbonado furnaces, each raw ore is smelted into 1 ingot only.

### 📑 Reproduction Steps

1. Place down an enhanced / reinforced / carbonado furnace with a luck multiplier greater than 1
2. Place some fuel in it
3. Place each kind raw ores in it
4. Watch it being smelted into 1 ingot only

### 💡 Expected Behavior

The expected behaviour is that raw ores are being multipled to give an amount of ingots around the luck multiplier when being smelted.

### 📷 Screenshots / Videos

https://youtu.be/n5gefHoJ9UI

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

https://i.imgur.com/iKtj1hl.png
https://i.imgur.com/i7ywuHv.png

ybw0014:
Only ores benefit from the luck multiplier.

TheBusyBiscuit:
I would argue that this should definitely apply to Raw Ores here. In 1.17 that has become the default drop.

variananora:
> I would argue that this should definitely apply to Raw Ores here. In 1.17 that has become the default drop.

Can it be togglable? Because Iron Ore + Fortune + Furnace Luck is kinda too much?

Skizzles:
> > I would argue that this should definitely apply to Raw Ores here. In 1.17 that has become the default drop.
> 
> Can it be togglable? Because Iron Ore + Fortune + Furnace Luck is kinda too much?

They bring up a very good point, possibly a lower fortune rate for raw ore compared to standard?

TheBusyBiscuit:
I think a toggle would be better.
I reckon it would best to just add a toggle for all the ores (raw or not) in a follow-up build.
This way it can be tweaked to an even finer degree.

PR #3192, opened 2021-08-01
Add Enhanced furnaces' luck multiplier for raw ores in 1.17

ybw0014:
## Description

Mr Cookie said:
> I would argue that this should definitely apply to Raw Ores here. In 1.17 that has become the default drop.

## Proposed changes

add enhanced furnaces' luck multiplier for raw metals in 1.17.

#3190

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/EnhancedFurnaceListener.java
```diff
@@ -67,7 +68,14 @@ public void onItemSmelt(FurnaceSmeltEvent e) {
 
             if (state instanceof Furnace) {
                 FurnaceInventory inventory = ((Furnace) state).getInventory();
-                int amount = inventory.getSmelting().getType().toString().endsWith("_ORE") ? ((EnhancedFurnace) sfItem).getRandomOutputAmount() : 1;
+
+                boolean multiplier = inventory.getSmelting().getType().toString().endsWith("_ORE");
```
martinbrom, line 72: might be cleaner to use tags instead for both checks

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/EnhancedFurnaceListener.java
```diff
@@ -67,7 +69,15 @@ public void onItemSmelt(FurnaceSmeltEvent e) {
...
             if (state instanceof Furnace) {
                 FurnaceInventory inventory = ((Furnace) state).getInventory();
-                int amount = inventory.getSmelting().getType().toString().endsWith("_ORE") ? ((EnhancedFurnace) sfItem).getRandomOutputAmount() : 1;
+                Material smelting = inventory.getSmelting().getType();
+                boolean multiplier = SlimefunTag.ORES.isTagged(smelting);
+
+                // In 1.17, the raw ores should benefit from luck multiplier
+                if (SlimefunPlugin.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_17) && SlimefunTag.RAW_METALS.isTagged(smelting)) {
+                    multiplier = true;
+                }
```
TheBusyBiscuit, line 78: This can be simplified a lot.
In earlier minecraft versions, this tag is going to be empty anyway, so the version check is kinda redundant.

```suggestion
                boolean multiplier = SlimefunTag.ORES.isTagged(smelting) || SlimefunTag.RAW_METALS.isTagged(smelting);
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/EnhancedFurnaceListener.java
```diff
@@ -67,7 +69,10 @@ public void onItemSmelt(FurnaceSmeltEvent e) {
 
             if (state instanceof Furnace) {
                 FurnaceInventory inventory = ((Furnace) state).getInventory();
-                int amount = inventory.getSmelting().getType().toString().endsWith("_ORE") ? ((EnhancedFurnace) sfItem).getRandomOutputAmount() : 1;
+                Material smelting = inventory.getSmelting().getType();
+                boolean multiplier = SlimefunTag.ORES.isTagged(smelting) || SlimefunTag.RAW_METALS.isTagged(smelting);
```
TheBusyBiscuit, line 73: Looking at this, it may perhaps even be better to just create a new tag here which combines both of these.

ybw0014: what should the new tag be called? `luck_multiplier_materials` or something else

TheBusyBiscuit (changes requested):

Sfiguz7 (approved):

TheBusyBiscuit (approved):

PR #3204, opened 2021-08-17
Fix pickup delay not being set

### LINKED ISSUE #3203: Plugin Incompatibility - Altar Probe when using stacker and other item plugins. Possible fixes
### ❗ Checklist

### 📍 Description

This is more of a suggestion based off multiple issues found with Altar Probe items happening with multiple item based plugins like WildStacker, RoseStacker, (insert stacker name here), drop2inventory. While there are suggestions on discord, this is directly related to bug reports.

After speaking with developers of some of the plugins involved a lot of them suggested lower priority levels for these events, after speaking in the discord you guys had stated against that which is fine.

One of the devs of said plugins said:

> I think the best solution would be that they give their items the maximum possible pickup delay, which makes items to not be pickable.
> Also, they should use a lower priority, this way no other plugin will override their pickup denial. Furthermore, they should look at PlayerAttemptPickupItemEvent (or something like that), which is actually being called before PlayerPickupItemEvent, and cancel that event as well.
> I dont want to patch it on my end, because as you said it already happened with other plugins, and the best approach is that they make their plugin compatible with all the other plugins, and not force everyone to add support for their plugin
> 
> I will find the pickup delay I was talking about and will post it here later, so the Slimefun developers can chnage their code (the pickup-delay approach is much more optimizes in terms of performance as well)

If none of these options work for you guys, what would be a solution so this can be resolved for multiple plugins involved?

Thanks

### 📑 Reproduction Steps

Install WildStacker
Place an item on an ancient altar pedestal
Walk towards the item
See the item picked up

### 💡 Expected Behavior

I expected the item to not be picked up

### 📷 Screenshots / Videos

image

### 📜 Server Log

N/A

None found

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

Slimefun 960

### 🧭 Other plugins

WildStacker
RoseStacker
Drop2Inventory

WalshyDev:
## Description
Set a pickup delay of Short.MAX_VALUE (which in the client means "never pickup"). A lot of third-party plugins (including ones like WildStacker) check for this value to know if it's a custom item or if it should be modified/stacked. Setting this prevents a lot of third-party plugins from messing with the items.

## Proposed changes
Set the pickup delay to "never pickup" in the `markAsNoPickup` method.

Resolves #3203

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -81,9 +81,11 @@ public final class SlimefunUtils {
      */
     public static void markAsNoPickup(@Nonnull Item item, @Nonnull String context) {
         item.setMetadata(NO_PICKUP_METADATA, new FixedMetadataValue(SlimefunPlugin.instance(), context));
-        // Max the pickup delay - This makes it so no Player can pick up items ever without need for an event
-        // It is also one indication used by third party plugins to know if it's a custom item.
-        // Fixes #3203
+        /*
+         * Max the pickup delay - This makes it so no Player can pick up items ever without need for an event.
+         * It is also an indication used by third-party plugins to know if it's a custom item.
+         * Fixes #3203
+         */
         item.setPickupDelay(Short.MAX_VALUE);
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -81,6 +81,11 @@ public final class SlimefunUtils {
      */
     public static void markAsNoPickup(@Nonnull Item item, @Nonnull String context) {
         item.setMetadata(NO_PICKUP_METADATA, new FixedMetadataValue(SlimefunPlugin.instance(), context));
+        // Max the pickup delay - This makes it so no Player can pick up items ever without need for an event
+        // It is also one indication used by third party plugins to know if it's a custom item.
+        // Fixes #3203
+        item.setPickupDelay(Short.MAX_VALUE);
+        SlimefunPlugin.instance().getLogger().info("Set pickup delay");
     }
 
     /**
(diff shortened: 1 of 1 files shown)


svr333 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -81,6 +81,10 @@ public static boolean hasNoPickupFlag(@Nonnull Item item) {
      */
     public static void markAsNoPickup(@Nonnull Item item, @Nonnull String context) {
         item.setMetadata(NO_PICKUP_METADATA, new FixedMetadataValue(SlimefunPlugin.instance(), context));
+        // Max the pickup delay - This makes it so no Player can pick up items ever without need for an event
+        // It is also one indication used by third party plugins to know if it's a custom item.
+        // Fixes #3203
```
TheBusyBiscuit, line 86: multi-line comments exist :eyes:

TheBusyBiscuit (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
svr333: well if we're going to request changes then might as well do
```diff
-		 // Max the pickup delay - This makes it so no Player can pick up items ever without need for an event
-		 // It is also one indication used by third party plugins to know if it's a custom item.
+		 // Max the pickup delay - This makes it so no Player can pick up items ever without need for an event.
+        // It is also an indication used by third party plugins to know if it's a custom item.
```

WalshyDev, line 86: *sigh* you guys are a pain
```suggestion
        /*
         * Max the pickup delay - This makes it so no Player can pick up items ever without need for an event.
         * It is also an indication used by third-party plugins to know if it's a custom item.
         * Fixes #3203
         */
```

svr333 (approved):

Sfiguz7 (approved):

PR #3224, opened 2021-09-01
Guide shows wrong machine

Bencey:
Enderlump 1 is made with grindstone not the enhanced crafting table


WalshyDev:
What is this PR... you broke an import and are changing a recipe that has existed for years. You CAN use the grind stone but it is not the only machine.

TheBusyBiscuit:
Yep, that was likely a copy/paste error, it should be `RecipeType.GRIND_STONE` there.
However you broke the file by messing with the imports.

Bencey:
Oh crap. Just looked at my PR. I didnt mean to touch the import.

PR #3229, opened 2021-09-02
Fixed ender lump guide entry

Bencey:
Fixed the recipe for enderlump as the guide says to use enhanced crafting table (Which is incorrect)

Previously made a PR via https://github.com/Slimefun/Slimefun4/pull/3224 but accidentally touched an import so recreating the PR


Mooy1:
Isn't it supposed to work in both?

Bencey:
> Isn't it supposed to work in both?

Nope tested it yesterday. T2+ you make in the crafting station but for tier 1 you need to grind up eyes of ender in the grindstone

Phoenix-Starlight:
Indeed, for enderlumps (tier 1) it is made via grindstone.

TheBusyBiscuit (approved):

PR #3139, opened 2021-06-26
⚠️ BREAKING CHANGES ⚠️ - Relocation of important API classes.

TheBusyBiscuit:
## 🌐 Description
This Pull Request introduces major changes which **will break almost every addon in existence**.
But before you panic, do not worry. This Pull Request will **not** be merged anytime soon.
Please read carefully.

### 1. Why... why do you do this to us? 😢 
As you know, Slimefun is an ever-growing, ever-expanding project.
It has grown steadily since over 8 years and is constantly being rewritten over and over again.
With so many people working on the project, mistakes happen.
Slimefun has a reputation and bad code practices is a part of that.
We have made many bad design decisions over these 8 years, though frankly which developer hasn't over such a long time period.

We are also constantly working on improving the codebase and getting rid of these bad decisions by refactoring the codebase.
And we almost did our best to avoid breaking changes within the many dependents of the project, most noteworthy the many addons that exist for Slimefun.
However sometimes we find ourselves in a corner. A corner from which we cannot escape without breaking changes.
And we have reached many corners like that.

### 2. What has changed? 😨
I will try my best to sum up all the things that have changed within the API, so you can prepare accordingly.
We changed our package signature from `me.mrCookieSlime.Slimefun` to `io.github.thebusybiscuit.slimefun4` a while ago.
But there are still a lot of classes who have missed the move, because they would break the API.

1. `Category` has been renamed to `ItemGroup`.
2. All Category / ItemGroup variants have been relocated to `io.github.thebusybiscuit.slimefun4.api.items.groups`
3. The `SlimefunItem` class has been relocated to `io.github.thebusybiscuit.slimefun4.api.items`
4. The `SlimefunItemStack` class has been relocated to `io.github.thebusybiscuit.slimefun4.api.items`
5. The `ItemHandler` class has been relocated to `io.github.thebusybiscuit.slimefun4.api.items`
6. The `RecipeType` class has been relocated to `io.github.thebusybiscuit.slimefun4.api.recipes`
7. Research classes have been moved from `io.github.thebusybiscuit.slimefun4.core.researching` to `io.github.thebusybiscuit.slimefun4.api.researches`
8. The main class `SlimefunPlugin` was renamed to `Slimefun`
9. [CS-CoreLib2](https://github.com/TheBusyBiscuit/CS-CoreLib2) was removed and replaced by [dough](https://github.com/baked-libs/dough)

And lastly, any method signatures that involved the aforementioned classes was also updated accordingly.

[...]

PR #3232, opened 2021-09-04
Rename category -> itemGroup

svr333:
## Description
This is a follow up to #3139, a lot of comments and variables and error messages needed to be updated still.
This is what that does

Two more follow up questions:
- in `wiki.json`, `Christmas-Seasonal-Category` is mentioned, change to `Christmas-Seasonal-ItemGroup`?
- do I update `messages.yml` (and other files) keys? (eg: `guide.tooltips.open-category` -> `guide.tooltips.open-itemgroup`)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
@@ -424,10 +424,10 @@ public class ProgrammableAndroid extends SlimefunItem implements InventoryBlock,
         });
 
         int index = 0;
-        int itemGroupIndex = 45 * (page - 1);
+        int categoryIndex = 45 * (page - 1);
 
         for (int i = 0; i < 45; i++) {
-            int target = itemGroupIndex + i;
+            int target = categoryIndex + i;
 
             if (target >= scripts.size()) {
                 break;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
@@ -78,8 +78,8 @@ public class ProgrammableAndroid extends SlimefunItem implements InventoryBlock,
     private final int tier;
 
     @ParametersAreNonnullByDefault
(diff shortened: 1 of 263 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
```diff
@@ -424,10 +424,10 @@ protected void openScriptDownloader(Player p, Block b, int page) {
         });
 
         int index = 0;
-        int categoryIndex = 45 * (page - 1);
+        int itemGroupIndex = 45 * (page - 1);
 
         for (int i = 0; i < 45; i++) {
-            int target = categoryIndex + i;
+            int target = itemGroupIndex + i;
```
TheBusyBiscuit, line 430: This does not refer to item groups, it's the different categories for android scripts

TheBusyBiscuit (changes requested):
Looks good, just one small mistake.

Thanks!

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/androids/ProgrammableAndroid.java
svr333: ah all right

svr333:
Also what about these
> - in `wiki.json`, `Christmas-Seasonal-Category` is mentioned, change to `Christmas-Seasonal-ItemGroup`?
> - do I update `messages.yml` (and other files) keys? (eg: `guide.tooltips.open-category` -> `guide.tooltips.open-itemgroup`)

TheBusyBiscuit:
> Also what about these @TheBusyBiscuit
> 
> > * in `wiki.json`, `Christmas-Seasonal-Category` is mentioned, change to `Christmas-Seasonal-ItemGroup`?
> > * do I update `messages.yml` (and other files) keys? (eg: `guide.tooltips.open-category` -> `guide.tooltips.open-itemgroup`)

Regarding the wiki reference, the wiki page itself would need to be renamed first (if thats wanted).
Message strings... well, they *could* be updated but they would need to be updated in every class and every locale file. If you wanna go through all of that, sure. Otherwise we can just leave it as is or do it later.

svr333:
I think I'm going to keep the Christmas seasonal category as-is, because I think the in-game text also still says category over itemgroup, and in this case it does make more sense  
its a christmas category

Ill do the other changes in a bit

TheBusyBiscuit (approved):

WalshyDev (approved):
God GitHub hates large PRs but I couldn't find any typos or anything so looks good

PR #3235, opened 2021-09-05
Modified smart filling

Shadowboy00:
Changed smart filling nodes to not fill in empty slots before the configured items.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoUtils.java
@@ -338,26 +338,27 @@ final class CargoUtils {
         int minSlot = range[0];
         int maxSlot = range[1];
 
+        // store the first empty slot for smart filling mode
+        int slotCandidate = -1;
+
         for (int slot = minSlot; slot < maxSlot; slot++) {
             // Changes to this ItemStack are synchronized with the Item in the Inventory
             ItemStack itemInSlot = contents[slot];
-
-            if (itemInSlot == null) {
-                inv.setItem(slot, stack);
-                return null;
-            } else {
-                int currentAmount = itemInSlot.getAmount();
-                int maxStackSize = itemInSlot.getType().getMaxStackSize();
-
-                if (!smartFill && currentAmount == maxStackSize) {
-                    // Skip full stacks - Performance optimization for non-smartfill nodes
-                    continue;
-                }
-
-                if (SlimefunUtils.isItemSimilar(itemInSlot, wrapper, true, false)) {
-                    if (currentAmount < maxStackSize) {
+            
+            if (smartFill) {
+                if (itemInSlot == null) {
+                    if (slotCandidate == -1) {
+                        slotCandidate = slot;
+                    }
+                } else {
+                    int currentAmount = itemInSlot.getAmount();
+                    int maxStackSize = itemInSlot.getType().getMaxStackSize();
+                    // stop when found the first available slot
+                    if (SlimefunUtils.isItemSimilar(itemInSlot, wrapper, true, false)) {
+                        if (currentAmount == maxStackSize) {
+                            return stack;
+                        }
                         int amount = currentAmount + stack.getAmount();
-
                         if (amount > maxStackSize) {
                             stack.setAmount(amount - maxStackSize);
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit:
This was intentional.
Merging this would ruin the performance of cargo nodes even further.

Shadowboy00:
Now that cargo system has got a performance boost, would this be reconsidered please, it'd be a big help for automations.

Sefiraat:
Personally, I think the last thing we should do with the performance gain is add new and largely not needed features [fix?] to absorb it. I think server owners can breathe easy knowing their users can build a bit more freely without worrying and perhaps this should wait until the full cargo rewrite?

Shadowboy00:
Hi Sefiraat, thanks for your comment. This addition will only affect the smart filling nodes, which is a very small portion compared to the the non-smart filling nodes, so I don't project this will have much influence on the performance. And it will solve a big problem for those who uses it (also can remove the "This is not perfect" from the description xD).

J3fftw1:
@Shadowboy00 mind updating your branch i will run timings after work to see how much of a performance hit this actually makes even after the update so we can compare the results to see if its good to add it

Shadowboy00:
@J3fftw1 Sure thing, I'll let you know when it's done.

Shadowboy00:
@J3fftw1 It's done, please have a look.

J3fftw1:
[after-your-changes.txt](https://github.com/Slimefun/Slimefun4/files/7194926/after-your-changes.txt)
[before-your-changes.txt](https://github.com/Slimefun/Slimefun4/files/7194927/before-your-changes.txt)
These are the timing results i think the says enough
https://owo.whats-th.is/5kZ62ZH.png

Shadowboy00:
Thanks for the results, may I ask the setup for the tests, i.e. the settings for the input node and the output node, and whether the chests are empty or not, as in practical use cases the smart filling nodes will not need to go through all the slots of a container as it will be filled with the required items and stop, thus yielding the same performance as before.

Cases analysis:
Worst case: goes through all slots (could happen before as well, just fill the container's slots with other stuff but not full stacks and leave the last slot open)
Normal case: same as before (maybe check 1 more slot on average)
Best case: same as before
Also the ratio of smart-filling nodes is much less, so I think it is still viable.

J3fftw1:
i just use world that Walshy send to me for testing. This world comes from a server with a lot of cargo nodes not exactly sure what server its from. But they send the location of 1 of the biggest bases on it i use that for testing. I just stood in a location and ran sf timings --verbose in console and it gave those results. What is happening in the base i got no clue maybe @Sefiraat knows whats going on but i cant tell for sure

Sefiraat:
It's a copy of my personal base with about as many different combinations of cargo configurations as you could imagine. Smart fill accounts for a good 20-30 (ish) nodes. Perhaps more.
These are used for auto crafting and - worth noting here - I do not encounter the issue you descripe as the issue being fixed here. This is likely as my supply is static and i always ship outputs into barrels with void excess on should I overrun so I can never (fingers crossed!) back up. Not all my nodes do this but most do.

Dont just think practical use cases. Players are not going to think about the how/why. They will just build and use smart fill when auto crafting as a near-default.

Shadowboy00:
Thanks for the comments, yes the code wasn't fully optimized for performance yet, would you be able to share the world to me so I can test out myself. In the meanwhile I've tried to improve the performance by removing the unnecessary checks I used in my code, feel free to test it out. But sharing the test world with me would be helpful, and I can get back to you once I tried everything I can.

Sefiraat:
Apologies, it's not my world it's only given to a few people courtesy of the owners and I dont want to spread it around further.

If you autocraft every step to a max solar panel with smart fill on all nodes where appropritate should give you a good base to work from.

Again, however, why not just wait until the re-write? This is not a wholey prevalent problem and can wait until a full solution is devised?

Shadowboy00:
OK I understand, it's just I'm faced with the problem and I tried to solve it, that's it. I be grateful If you could test the code one more time and if the performance is still poor I'll give up on the idea, thanks.

J3fftw1:
I will post some test results in a bit

Shadowboy00:
Thanks

J3fftw1:
Btw are you on the discord that makes talkimg about this 10 times easier

Shadowboy00:
Yes, I've sent a friend request to you

PR #3238, opened 2021-09-06
Fix metrics not being able to download

WalshyDev:
We used REPO_NAME in quite a few places which I didn't expect but here we are. Anyway, this fixes that.


TheBusyBiscuit (approved):
:LUL:

PR #3253, opened 2021-09-13
Hercules' Pickaxe supports 1.17 deepslate ores and copper

### LINKED ISSUE #3251: Hercules Pickaxe not turning Copper Ore, or Iron/Gold Deepslate Ore into dust.
### ❗ Checklist

### 📍 Description

When attempting to mine Iron Deepslate Ore or Gold Deepslate Ore with the Hercules' Pickaxe, instead of turning into dust, it drops normal items; Raw Iron and Raw Gold. Also, when mining Copper Ore or Copper Deepslate Ore with the Hercules' Pickaxe, and the same occurs.

### 📑 Reproduction Steps

1) Find/Place one of the three ores mentioned above (Iron/Gold/Copper Deepslate Ores, or just regular Copper Ore).
2) Obtain, in some way, the Hercules' Pickaxe
3) Mine the Ore.

### 💡 Expected Behavior

I was expecting for the Iron Deepslate Ore to drop Iron Dust, the Gold Deepslate Ore to drop Gold Dust, and either variant of the Copper Ore to drop Copper Dust.

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

Slimefun DEV - 968 (git 79c6c139)

### 🧭 Other plugins

_No response_

ybw0014:
## Description

The Hercules' Pickaxe should function like Ore Crusher, so the drops of mining deepslate ores or copper ores should be their corresponding dusts.

## Proposed changes

Add support for Hercules' Pickaxe with 1.17 deepslate ores and copper ores.

Resolves #3251

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
@@ -3,6 +3,8 @@ package io.github.thebusybiscuit.slimefun4.implementation.items.tools;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.api.MinecraftVersion;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 import org.bukkit.Material;
 import org.bukkit.inventory.ItemStack;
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
```diff
@@ -25,16 +27,36 @@ public HerculesPickaxe(ItemGroup itemGroup, SlimefunItemStack item, RecipeType r
...
+                        case DEEPSLATE_IRON_ORE:
+                            drops.add(new CustomItemStack(SlimefunItems.IRON_DUST, 2));
+                            break;
+                        case DEEPSLATE_GOLD_ORE:
+                            drops.add(new CustomItemStack(SlimefunItems.GOLD_DUST, 2));
+                            break;
+                        case COPPER_ORE:
+                        case DEEPSLATE_COPPER_ORE:
+                            drops.add(new CustomItemStack(SlimefunItems.COPPER_DUST, 2));
+                            break;
```
TheBusyBiscuit, line 44: Missing `default:` statement

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
ybw0014: this section is about 1.17 ores, if the block type does not match here, it should then go to next switch

TheBusyBiscuit: This switch is still missing a `default: break;` statement though.
Having no 'default' statement usually implies that there are no unhandled possible values.
But this `switch` does not cover all values of the `Material` enum, hence why it definitely needs a `default` statement from a semantical view.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
```diff
@@ -25,16 +27,36 @@ public HerculesPickaxe(ItemGroup itemGroup, SlimefunItemStack item, RecipeType r
...
+                    case IRON_ORE:
+                        drops.add(new CustomItemStack(SlimefunItems.IRON_DUST, 2));
+                        break;
+                    case GOLD_ORE:
+                        drops.add(new CustomItemStack(SlimefunItems.GOLD_DUST, 2));
+                        break;
+                    default:
+                        for (ItemStack drop : e.getBlock().getDrops(tool)) {
+                            drops.add(new CustomItemStack(drop, drop.getAmount() * 2));
+                        }
```
TheBusyBiscuit, line 58: Also missing `break;`

TheBusyBiscuit (approved):
LGTM I suppose.

Though the more I look at it, the more I start to question why this item even exists...

svr333 (approved):
Looks good, maybe regarding @TheBusyBiscuit's comment, we can deprecate the item

PR #3256, opened 2021-09-14
Fix saddle duplication

### LINKED ISSUE #3254: Saddle Duplication w/ Talisman of the Hunter
### ❗ Checklist

### 📍 Description

When using a Talisman of the Hunter, if you kill a pig, strider (probably horse, didn't test), and the talisman procs, it will duplicate the saddle, giving you 2.

### 📑 Reproduction Steps

1. Give yourself a Talisman of the Hunter
2. Spawn in pigs or striders.
3. Put saddles on them all.
4. Kill them until the talisman procs.

### 💡 Expected Behavior

I'd expect the talisman to duplicate the mob drops, not the saddle.

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

No error

No error

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

image

### 🧭 Other plugins

N/A

EpicPlayerA10:
## Description
Fixed saddle duplication.

## Proposed changes
Added to `TalismanListener#getExtraDrops` check for `Steerable` entity which can have saddle

Resolves #3254

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
@@ -20,6 +20,7 @@ import org.bukkit.entity.Item;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
 import org.bukkit.entity.Projectile;
+import org.bukkit.entity.Steerable;
 import org.bukkit.entity.Trident;
 import org.bukkit.event.Event;
 import org.bukkit.event.EventHandler;
@@ -196,6 +197,14 @@ public class TalismanListener implements Listener {
                 for (ItemStack item : horse.getInventory().getStorageContents()) {
                     items.remove(item);
                 }
+            }
+        }
+
+        if (entity instanceof Steerable) {
+            Steerable steerableEntity = (Steerable) entity;
+
+            if (steerableEntity.hasSaddle()) {
+                items.remove(new ItemStack(Material.SADDLE));
             }
         }
 


svr333 (approved):
Looks good

Sfiguz7 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
```diff
@@ -199,6 +200,14 @@ public void onKill(EntityDeathEvent e) {
             }
         }
 
+        if (entity instanceof Steerable) {
+            Steerable steerableEntity = (Steerable) entity;
+
+            if (steerableEntity.hasSaddle()) {
+                items.remove(new ItemStack(Material.SADDLE));
```
WalshyDev, line 207: I'm not sure this will always work. You're making a new object rather than getting the existing which means it NEEDS to match this.

What happens if the saddle is renamed? What if it has a custom model data?

We should probably loop the items and remove if type == SADDLE instead

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/TalismanListener.java
EpicPlayerA10: But when you kill a pig with for example renamed saddle then pig will drop you a normal saddle without name, lore, tags etc.

svr333:
Looking at this it seems that this does work, seeing as he did test this (though we can and maybe should re-test this)

The only question I have is, when you put a chest on a donkey, are those items not duped too?

TheBusyBiscuit (changes requested):
`Steerable` is a very new interface anyway and this will break backwards compatibility.
So, I will be closing this for now until we find a better fix for this issue.

PR #3259, opened 2021-09-15
Fixes #3248

### LINKED ISSUE #3248: AutoEnchanter overrides enchantments
### ❗ Checklist

### 📍 Description

Auto Enchanter overrides lower level on the same enchantment
Example: Pickaxe already have Unbreaking 10 and you add Unbreaking 3 book, you will get Unbreaking 3 on the final product

https://discord.com/channels/565557184348422174/565569972697301020/886602517071425637

### 📑 Reproduction Steps

1. Use auto enchanter
2. Put higher level enchantment on the main item
3. Put lower level enchantment on enchanted book
4. Lower level overrides the higher level

### 💡 Expected Behavior

Auto Enchanter should keep the highest level enchantments

### 📷 Screenshots / Videos

https://youtu.be/MXJ_LuQxJFs

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

```
> sf versions
[21:07:57 INFO]: This Server uses the following setup of Slimefun:
Paper git-Paper-259 (MC: 1.17.1)
Slimefun DEV - 968 (git 79c6c139)
Metrics-Module #28
Java 16

No Addons installed
> pl
[21:07:57 INFO]: Plugins (3): Slimefun, Vault, VoidGen
> ver
[21:07:58 INFO]: Checking version, please wait...
[21:08:00 INFO]: This server is running Paper version git-Paper-259 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 810d200)
You are running the latest version
Previous version: git-Paper-250 (MC: 1.17.1)
```

### 🧭 Other plugins

_No response_

Sfiguz7:
## Description

As of now, the AutoEnchanters blindly apply whatever is on the enchanted book onto the item, disregarding any check other than compatibility ones.
A small discussion on the Discord lead to two main ideas: some people like the fact everything gets overridden, meaning an extra layer of difficulty is added as more thought and planning has to be put in instead of blindly slapping books on the item. On the other hand, some people like the idea of the enchanter not completely disregarding what is on the item to begin with, and it to be able to only update the enchantments it needs to.

This PR is motivated by this dichotomy: an additional item setting has been added that allows servers to choose whether they do want the override or not (defaulting to yes so that the old behavior is conserved). If the enchantment to be applied is of lower or equal level it is ignored meaning the enchanting time is also reduced.

## Proposed changes

Added itemSetting to AutoEnchanters to choose whether existing enchants should be overridden no matter what or checked for an improvement first. Defaults to old behavior for consistency.

TESTING SAMPLE:
Sword and book used
https://imgur.com/a/Q1k7tK0

Time and result with setting to true (default):
https://imgur.com/a/dVjzpCX

Time and result with setting to false (new behavior):
https://imgur.com/a/RetSUwJ

Resolves #3248 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
@@ -6,7 +6,6 @@ import java.util.Map;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import io.github.thebusybiscuit.slimefun4.api.items.ItemSetting;
 import org.bukkit.Bukkit;
 import org.bukkit.Material;
 import org.bukkit.enchantments.Enchantment;
@@ -17,6 +16,7 @@ import io.github.bakedlibs.dough.inventory.InvUtils;
 import io.github.thebusybiscuit.slimefun4.api.events.AsyncAutoEnchanterProcessEvent;
 import io.github.thebusybiscuit.slimefun4.api.events.AutoEnchantEvent;
 import io.github.thebusybiscuit.slimefun4.api.items.ItemGroup;
+import io.github.thebusybiscuit.slimefun4.api.items.ItemSetting;
 import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
 import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItemStack;
 import io.github.thebusybiscuit.slimefun4.api.recipes.RecipeType;
@@ -40,7 +40,7 @@ import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
  */
 public class AutoEnchanter extends AbstractEnchantmentMachine {
 
-    private final ItemSetting<Boolean> overrideExistingEnchantsLvl= new ItemSetting<>(this, "override-existing-enchants-lvl", true);
+    private final ItemSetting<Boolean> overrideExistingEnchantsLvl= new ItemSetting<>(this, "override-existing-enchants-lvl", false);
 
     @ParametersAreNonnullByDefault
     public AutoEnchanter(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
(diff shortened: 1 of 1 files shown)


svr333:
Uh #3255

Sfiguz7:
Ah you had done it already. Hmm I don't know if it should be a setting or just default, I don't mind either way. I will keep this open so we can choose which one to actually merge

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -6,6 +6,7 @@
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.api.items.ItemSetting;
```
ybw0014, line 9: this import should not be here

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -103,6 +108,12 @@ protected MachineRecipe enchant(BlockMenu menu, ItemStack target, ItemStack ench
             }
         }
 
+        // If override is false, remove those with lower level so we don't override existing enchants
+        // This also removes those with the same level so they aren't accounted for enchanting time
```
TheBusyBiscuit, line 112: Multiline comments exist.

Sfiguz7:
We should decide whether we want it to be optional (this PR) or permanent (#3255) before I fix the two things. Shall we have a chat in #developers?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting/AutoEnchanter.java
```diff
@@ -39,9 +40,13 @@
  */
 public class AutoEnchanter extends AbstractEnchantmentMachine {
 
+    private final ItemSetting<Boolean> overrideExistingEnchantsLvl= new ItemSetting<>(this, "override-existing-enchants-lvl", true);
```
svr333, line 43:
```suggestion
    private final ItemSetting<Boolean> overrideExistingEnchantsLvl= new ItemSetting<>(this, "override-existing-enchants-lvl", false);
```

svr333 (changes requested):

Sfiguz7:
For those with access, the discussion took place [here](https://discord.com/channels/565557184348422174/565570276038017044/888122263872872498) (just in case, for future reference)

svr333 (approved):

TheBusyBiscuit (approved):
lgtm

PR #3275, opened 2021-09-22
Fixes #3133

### LINKED ISSUE #3133: ItemsAdder items transferred via cargo
### ❗ Checklist

### 📍 Description

Slimefun cargo adds texture to the renamed item

### 📑 Reproduction Steps

I have in the ItemsAdder item named Coin with a texture on the stick. image
I'll put the coin in the first chest
I'll put a stick renamed to Coin in the second chest
I'll have the stick renamed Coin transferred via cargo
The stick acquires the texture of the coin
(Coin (ENG) = Mince (CZ))

We tried it with other items and it works like this with everything.

### 💡 Expected Behavior

The item transferred through the cargo does not acquire texture

### 📷 Screenshots / Videos

The videos were more than 10MB and could not be uploaded, I give a link to youtube
https://youtu.be/I1baDR_b1YQ
https://youtu.be/1-mUxS-7ADA

### 📜 Server Log

N/A

N/A

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.16.x

### ⭐ Slimefun version

image

### 🧭 Other plugins

image

Sfiguz7:
## Description

Check #3133 for details. Cargo would not check for CustomModelData so different ModelData items with the same name and lore would stack when sent via cargo, when they should not.
## Proposed changes

- Added an additional check for CustomModelData: this will help no matter the resource pack, and in particular will fix the issue with ItemsAdder.

Resolves #3133
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -51,6 +51,7 @@ import io.github.thebusybiscuit.slimefun4.utils.itemstack.ItemStackWrapper;
  *
  * @author TheBusyBiscuit
  * @author Walshy
+ * @author Sfiguz7
  */
 public final class SlimefunUtils {
 
@@ -63,7 +64,8 @@ public final class SlimefunUtils {
      * This method quickly returns whether an {@link Item} was marked as "no_pickup" by
      * a Slimefun device.
      *
-     * @param item The {@link Item} to query
+     * @param item
+     *            The {@link Item} to query
      * @return Whether the {@link Item} is excluded from being picked up
      */
     public static boolean hasNoPickupFlag(@Nonnull Item item) {
@@ -74,8 +76,10 @@ public final class SlimefunUtils {
      * This will prevent the given {@link Item} from being picked up.
      * This is useful for display items which the {@link AncientPedestal} uses.
      *
-     * @param item    The {@link Item} to prevent from being picked up
-     * @param context The context in which this {@link Item} was flagged
+     * @param item
+     *            The {@link Item} to prevent from being picked up
+     * @param context
+     *            The context in which this {@link Item} was flagged
      */
     public static void markAsNoPickup(@Nonnull Item item, @Nonnull String context) {
         item.setMetadata(NO_PICKUP_METADATA, new FixedMetadataValue(Slimefun.instance(), context));
@@ -90,7 +94,8 @@ public final class SlimefunUtils {
     /**
      * This method checks whether the given {@link ItemStack} is considered {@link Soulbound}.
      *
-     * @param item The {@link ItemStack} to check for
+     * @param item
+     *            The {@link ItemStack} to check for
      * @return Whether the given item is soulbound
      */
     public static boolean isSoulbound(@Nullable ItemStack item) {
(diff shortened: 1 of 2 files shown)


WalshyDev (approved):
REEE didn't mean to do this

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -319,17 +321,23 @@ private static boolean equalsItemMeta(@Nonnull ItemMeta itemMeta, @Nonnull ItemM
...
-            return true;
-        } else {
+        } else if (checkLore) {
             Optional<List<String>> itemLore = meta.getLore();
 
-            if (itemMeta.hasLore() && itemLore.isPresent()) {
-                return equalsLore(itemMeta.getLore(), itemLore.get());
-            } else {
-                return !itemMeta.hasLore() && !itemLore.isPresent();
+            if (itemMeta.hasLore() && itemLore.isPresent() && !itemMeta.getLore().equals(itemLore.get())) {
```
WalshyDev, line 327: Why is this `.equals` and the old `equalsLore`?

WalshyDev (changes requested):

Sfiguz7:
> REEE didn't mean to do this

TOO LATE HA (I am not sure why it's now equals, let me check what fancy stuff equalsLore does...)

WalshyDev (dismissed):
Alright looks good now

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -319,17 +321,23 @@ private static boolean equalsItemMeta(@Nonnull ItemMeta itemMeta, @Nonnull ItemM
...
-                return !itemMeta.hasLore() && !itemLore.isPresent();
+            if (itemMeta.hasLore() && itemLore.isPresent() && !equalsLore(itemMeta.getLore(), itemLore.get())) {
+                return false;
+            } else if (itemMeta.hasLore() != itemLore.isPresent()) {
+                return false;
             }
         }
+
+        // Fixes #3133: name and lore are not enough
+        OptionalInt itemCustomModelData = meta.getCustomModelData();
```
TheBusyBiscuit, line 335: While we are at it, `meta` should be renamed to `snapshot`, now that the class was renamed.

TheBusyBiscuit (changes requested):
Looks good (as far as I can tell). Just one variable name is a bit confusing (not your fault).

However, we need to make sure that custom model data is NEVER checked if its a Slimefun item. So, are we positive that this is the case?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
Sfiguz7: > Just one variable name is a bit confusing (not your fault).
Which variable and what would you like the name to be?
I can rename meta easily.

As for CustomModelData, the equalsItemMeta method I have modified is only ever called inside isItemSimilar, and only after a check for the IDs has been performed (lines 281-285 or, due to the cargo filter issue with wrappers, 290-300) - I double checked :)

svr333 (approved):
lgtm

WalshyDev (approved):
God these methods are a mess, this LGTM though

TheBusyBiscuit (approved):

Issue #1907, opened 2020-05-14
Electric Gold Pan can be manipulated to only produce one type of item


NCBPFluffyBear:
## Description (Required)
By clogging the second output slot, the electric gold pan will be forced to only produce the item in the first output slot (as long as it is Sifted Ore, Flint, Coal, or Iron Nuggets). This is similar to the old dust washer bug where you could force it to produce a specific dust.

## Steps to reproduce the Issue (Required)
1. Force an item into the second output slot by using Shift + Click, or by letting the second output slot overflow.
2. Place the item (Sifted Ore, Clay, Flint, or Iron Nuggets) that you wish to produce in the first output slot.
3. Run the Electric Gold Pan with electricity (Don't use cargo).

## Expected behavior (Required)
The gold pan should stop producing items when its output slots are occupied.

## Server Log / Error Report
No error, this is (I assume) just a bug that was not considered when coding the Electric Gold Pan.

## Environment (Required)
Minecraft Version: 1.15.2 Spigot
Slimefun Version: vDEV - 502 (git 62f54c0)
CS-CoreLib Version: vDEV - 85 (git 4a7fce52)

Issue #2324, opened 2020-09-14
Electric Gold Pan stops panning if both output slots are full, even if next item exists in output slot


MTDeath:
## :round_pushpin: Description (REQUIRED)

Added gravel to Electric Gold Pan (EGP), machine made flint in first output slot and sifted ore in 2nd slot.  EGP stopped processing items, I removed sifted ore from 2nd output slot and EGP made sifted ore in 2nd slot again.  Happens when connected to Cargo or not.

## :bookmark_tabs: Steps to reproduce the Issue (REQUIRED)

Setup Power and place Electric Gold Pan (tested on I,II,III), add Gravel, allow machine to process the gravel and once both output slots are full the machine will stop processing gravel.  If you remove one item, the same item may show up as next output but it would not increment the count.
https://youtu.be/NfXJXS8Fexw

## :bulb: Expected behavior (REQUIRED)

If the 2 output slots are full but the next item exits in the output slots, it should increment the count on the item not stop processing.

## :scroll: Server Log

no errors in server log

## :open_file_folder: Error Reports

no errors in folder

## :compass: Environment (REQUIRED)

 - Server Software (Spigot/Paper):
 - Minecraft Version: Paper v1_16_R2
 - Slimefun Version: vDev - 675
 - CS-CoreLib Version: vDev - 79
![SlimefunVersion](https://user-images.githubusercontent.com/55070828/93071640-61fbdf00-f64e-11ea-839a-a5309775994b.png)

TheBusyBiscuit:
That is completely intended, if the inventory has no space for new items it will stop producing them.
The output is chosen randomly, so the only way to know that there is definitely space for it is to check for an empty slot.
In the past Gold Pans were heavily abused by filling both slots with the desired item and then getting only that item you are looking for, this is the compromise here.

TheBusyBot:
Your issue has been closed by an admin, it may fall under one or more of the following categories.
**Please wait for an admin to tick off the points that apply.**
<hr>

* [ ] You did not follow our template. Please follow the Issue template to help us identify your issue more effectively.
* [ ] You did not provide any information about your versions (We absolutely need the exact version numbers that are installed on your Server, \"latest\" is not helpful)
* [ ] You did not provide a proper description to the problem. Try to write at least 4-6 sentences and/or provide screenshots or videos on how to reproduce this.
* [ ] We were unable to reproduce issue, if you think your issue still persists then please comment down below and give a better description on how to reproduce it.
[...]

PR #3277, opened 2021-09-23
Allow admins to control the behavior of the Electric Gold Pan via the…

KyleEstein:
… legacy-gold-pan configuration option.

## Description

As reported in Issue #2324 if a server has cargo networks disabled then the change made to fix Issue #1907 makes Electric Gold Pans practically useless. This change adds a "legacy-gold-pan" option to allow a server admin to revert the behavior of the Electric Gold Pan to what it was before commit 4efc10db1576957547ca18a837a089d269fff770.

## Proposed changes

Add a new config option boolean named "legacy-gold-pan". If the option is set then the Electric Gold Pan will behave as it did prior to the update for Issue #1907

## Checklist


Sfiguz7 (approved):

src/main/resources/config.yml
```diff
@@ -13,6 +13,7 @@ options:
   legacy-ore-washer: false
   legacy-dust-washer: false
   legacy-ore-grinder: true
+  legacy-gold-pan: false
```
WalshyDev, line 16: I hate this "legacy" naming options and I do not think this should be in the config. It's specific to the gold pan so it should be in items.yml

Please give this a better name and make it an ItemSetting

WalshyDev (changes requested):

src/main/resources/config.yml
KyleEstein: Can you please provide a better name that would be acceptable? I do not wish to waste time creating a different but equally unacceptable alternative. Unless the legacy options for the ore washer, electric dust washer, and ore grinder are updated as well, Slimefun will end up with different methods to accomplish basically the same thing for different machines.

WalshyDev: Just think of an actually descriptive name. Like what the hell is "legacy" supposed to do? Without being told, you don't know.

The existing options have been there for a long time (literally years). All other options are in items.yml, this is the standard. If you want to revert behaviour for other items you check items.yml.

KyleEstein: > The existing options have been there for a long time (literally years).

Yeah, I know. I created them. So it's my fault they exist (well, not legacy-ore-grinder, not sure who added that one) and if it's not right I'll do what I can to correct it. I noticed there is an option that starts with override-, so maybe something along those lines since we want to override the current behavior of the machine and make it behave like it did in an older version of Slimefun (the "older version" part is what made me choose the "legacy-*" option names in the first place.) But the reason we need the old behavior is because without cargo networks the current behavior makes the machine practically useless, so perhaps it should have a name related to "cargo-net-disabled"? Do either of these cause a good descriptive option name to pop into your head?

KyleEstein:
I rebased and now there's 8 commits so I don't know if I've completely hosed this PR and should just create a new one or what. Anyway, I added the ItemSetting "override-output-limit" with a nice description in ElectricGoldPan.java. Let me know if this PR is just broken now and if I need to try again.

src/main/resources/config.yml
KyleEstein: Converted to an ItemSetting named "override-output-limit". But did I break this PR with my rebase? Let me know if I should close this PR and create a fresh one.

svr333 (commented):
I can see why someone would use this setting, so good for me
Though I dont think the changelog is supposed to be changed

KyleEstein:
> I can see why someone would use this setting, so good for me Though I dont think the changelog is supposed to be changed

I made that change to resolve the Merge Conflict that was tagged by TheBusyBot. I don't use git and github often so every time I interact with it it feels like I have to learn it all over again. Any suggestions on what I should do to clean up this PR are appreciated. I'm half expecting a comment like "Yeah, this PR is hopelessly broken. Just close it and and create a new one."

Solverbolt:
Not entirely sure where to post such a request, i used the search just to find the general idea. but I noticed that the Electric Pan Machine will not interact or work with anything that can empty its inventory, including everything from the cargo section.  Any way to change that, so that maybe using a hopper or output chest will clear its inventory. Currently having to babysit the thing after putting a stack of gravel in defeats the purpose.

svr333 (approved):

TheBusyBiscuit (approved):
We can argue about "intuitive naming" here but otherwise it looks fine to me

PR #3288, opened 2021-09-26
Added ability to set capacity for Solar Gens, like all other gens

Sfiguz7:
## Description

All of the generators in Slimefun have a default capacity BUT the ability to change it, with some having it >0 by default and thus effectively also working as an energy buffer. Solar Generators, on the otherhand, had no way to do this as the capacity was hardcoded to be 0 and unchangeable. This PR increases solar gens extensibility, by making it comparable to the other generators.
## Proposed changes

- Added a capacity member to Solar Generators
- Added capacity as a constructor parameter
- Added a constructor defaulting to the old behavior so nothing breaks
- Changed getCapacity() to return this new member instead of the hardcoded 0

N/A
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/generators/SolarGenerator.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/generators/SolarGenerator.java
@@ -35,13 +35,20 @@ public class SolarGenerator extends SlimefunItem implements EnergyNetProvider {
 
     private final int dayEnergy;
     private final int nightEnergy;
+    private final int capacity;
 
     @ParametersAreNonnullByDefault
-    public SolarGenerator(ItemGroup itemGroup, int dayEnergy, int nightEnergy, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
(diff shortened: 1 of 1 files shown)


svr333 (approved):
LGTM

WalshyDev (approved):
LGTM, I always like extending the API

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/generators/SolarGenerator.java
```diff
@@ -35,13 +35,20 @@
 
     private final int dayEnergy;
     private final int nightEnergy;
+    private final int capacity;
 
     @ParametersAreNonnullByDefault
-    public SolarGenerator(ItemGroup itemGroup, int dayEnergy, int nightEnergy, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+    public SolarGenerator(ItemGroup itemGroup, int dayEnergy, int nightEnergy, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, int capacity) {
```
TheBusyBiscuit, line 41: The `capacity` argument is never validated :NotLikeThis:

PR #3289, opened 2021-09-28
Switching explosive tool modes is now similar to Multi Tool and persistent

qwertyuioplkjhgfd:
## Description

This was an approved suggestion in the Slimefun discord server and I thought I could implement it.

## Proposed changes

I have changed ``ExplosiveTool`` from a ``SimpleSlimefunItem`` to a ``SlimefunItem``.
I have added a system similar to the ``MultiTool`` to allow shift-clicking to change the modes of the explosive pickaxe. 
I have added a message in ``messages.yml`` (``messages.explosive-pick.mode-change``) to let the player know what mode their tool is on.

[This suggestion on the discord](https://discord.com/channels/565557184348422174/627048923122499584/891159523388182528)

## Checklist


TerslenK:
looks cool

WalshyDev:
I'm not sure how I feel about this. I don't like adding modes to this, it isn't an electric thing. Alongside that, this changes for the shovel too.

Also not a fan of storing this data in a map, we don't clean this up so it will just pile memory up forever. Instead PDC would be a better option

qwertyuioplkjhgfd:
PDC?

Sefiraat:
> PDC?
> 
> (also, the shifting mechanic was part of the shovel too since the shift check is done in `ExplosiveTool`, and the multitool uses a similar map that also piles up memory)

PDC is PersistantDataContainer where you can store the state (imo Better as it’ll carry over across restarts better) 

> PDC?
> 
> (also, the shifting mechanic was part of the shovel too since the shift check is done in `ExplosiveTool`, and the multitool uses a similar map that also piles up memory)

PDC = PersistantDataContainer which imo would be better anyway as this method would reset to default each restart where as a PDC method could allow it to persist through.

qwertyuioplkjhgfd:
Would it be better for the MultiTool to use a PDC as well?

WalshyDev:
> Would it be better for the MultiTool to use a PDC as well?

Yes it would but that isn't in scope for this PR, it should instead be separate.

qwertyuioplkjhgfd:
I made ``ExplosiveTool`` abstract since an explosive tool is an abstract classification of an object. Slimefun never creates an ``ExplosiveTool`` instance or registers it as an item (and I don't think any addons would either) so making it abstract shouldn't break anything. Ideally the class would be renamed ``AbstractExplosiveTool`` but that might break some addons that extend it so I haven't done that. 

Maybe the explosive option should be handled in a separate class (like ``FireworksOption``) instead of stuffing the persistence functionality inside ``ExplosiveTool`` but it should be fine as is

TheBusyBiscuit:
I feel like this only ever makes sense if a player can actually see which mode the pick is currently in...
Also, you are storing this info on the player rather than on the tool, players might assume this is a per-tool setting.

And I agree, while the class *can* be made abstract, it would have had to be renamed.

But as it stands now, I don't think we can/should accept this feature.
The current sneak-to-avoid mechanic is somewhat more user-friendly than this.
Sorry for the late reply from  my side, but I will go ahead and reject this feature, thanks for your PR though!

PR #3291, opened 2021-09-30
Fix other items being consumed when switching items after consuming juice

### LINKED ISSUE #3264: When drinking use and quickly switching ithe item slots to another, both juices get consumed and the player recieves a singular glass blottle
### ❗ Checklist

### 📍 Description

When drinking a juice and in the last instance of drinking it switching to another hotbar slot with a juice, bot get consumed and the player recieves a glass bottle.

### 📑 Reproduction Steps

1. Place 2 juices in your inventory
2. Start drinking one of them
3. The last minute before the player is finishing drinking it, switch to the hotbar slot with the other juice

### 💡 Expected Behavior

The expected behaviour is that only one of the juices would be consumed, and the player doesnt get a glass bottle from it.

### 📷 Screenshots / Videos

https://www.youtube.com/watch?v=GE1DBlG2I84

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

https://i.imgur.com/Hgp4T9g.png
https://i.imgur.com/AuIoUYW.png

### 🧭 Other plugins

_No response_

qwertyuioplkjhgfd:
## Description

This fixes a bug described in #3264 and makes the cooler consistent with drinking the juice normally. 

## Proposed changes

Instead of doing all the complex logic of refinding the itemstack of the juice just to remove the held item after a delay (allowing the player to switch stacks), just set the item of the consumption to nothing and apply the effects manually so that no glass bottle is created. Also makes drinking juice consistent with the cooler automatically feeding the player. 

Resolves #3264

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
@@ -61,10 +61,8 @@ public class Juice extends SimpleSlimefunItem<ItemConsumptionHandler> {
              * Minecraft has been broken when it comes to Saturation potions for a long time
              */
             for (PotionEffect effect : effects) {
-                if (effect.getType().equals(PotionEffectType.SATURATION)) {
-                    p.addPotionEffect(effect);
-                    break;
-                }
+                p.addPotionEffect(effect);
+                break;
             }
 
             e.setItem(null);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
@@ -6,8 +6,6 @@ import java.util.List;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import org.bukkit.Material;
-import org.bukkit.entity.Player;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.inventory.meta.ItemMeta;
 import org.bukkit.inventory.meta.PotionMeta;
@@ -19,11 +17,9 @@ import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
 import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItemStack;
 import io.github.thebusybiscuit.slimefun4.api.recipes.RecipeType;
 import io.github.thebusybiscuit.slimefun4.core.handlers.ItemConsumptionHandler;
-import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 import io.github.thebusybiscuit.slimefun4.implementation.items.SimpleSlimefunItem;
 import io.github.thebusybiscuit.slimefun4.implementation.items.backpacks.Cooler;
 import io.github.thebusybiscuit.slimefun4.implementation.listeners.CoolerListener;
-import io.github.thebusybiscuit.slimefun4.utils.SlimefunUtils;
 
 /**
  * This class represents a {@link SlimefunItem} that can be stored inside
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
```diff
@@ -65,39 +61,12 @@ public ItemConsumptionHandler getItemHandler() {
              * Minecraft has been broken when it comes to Saturation potions for a long time
              */
             for (PotionEffect effect : effects) {
-                if (effect.getType().equals(PotionEffectType.SATURATION)) {
-                    p.addPotionEffect(effect);
-                    break;
-                }
+                p.addPotionEffect(effect);
+                break;
```
WalshyDev, line 65: Why change this? You're now always going to just re-add the first effect which could be anything

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
```diff
@@ -65,39 +61,12 @@ public ItemConsumptionHandler getItemHandler() {
...
             }
 
-            removeGlassBottle(p, item);
+            e.setItem(null);
```
WalshyDev, line 68: This will remove all 10 if they're stacked, why change this?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CoolerListener.java
```diff
@@ -124,7 +124,6 @@ private boolean consumeJuice(@Nonnull Player p, @Nonnull ItemStack coolerItem, @
                     p.addPotionEffect(effect);
                 }
 
-                p.setSaturation(6F);
```
WalshyDev, line 127: Why remove this?

WalshyDev (changes requested):
This PR makes no sense right now, you have just completely broken the item

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
qwertyuioplkjhgfd: Because setting the item of the event to ``null`` basically makes it so that the player is eating air, and so no vanilla effects will be applied from the original potion. This ensures that the effects do get applied in the first place.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CoolerListener.java
qwertyuioplkjhgfd: This makes being fed by the cooler consistent with just drinking the juice normally

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
Sfiguz7: What does the setting the item to null have to do with this change?
You have just changed the behavior to "Add saturation effect if it was saturation" to "Add effect whatever that was".
This does not check the item of the event at all. Since all juices only have saturation you don't even see the difference, if you reverted this to the original everything else you changed would still work fine so I am confused why you even changed it in the first place.

qwertyuioplkjhgfd: Actually, golden apple juice gives absorption. It is because of golden apple juice that I added this.

Sfiguz7: Yes but this was only to fix absorption. Minecraft applies all effects from potions just fine EXCEPT it forgets about saturation, so we add it once more. That's all commented right above the spot you edited:
"/*
 * Fix for Saturation on potions is no longer working,
 * Minecraft has been broken when it comes to Saturation potions for a long time
 */"

By doing this, you are reapplying all of the effects - which changes nothing since you are applying saturation too which was what this fix addressed, but you are applying for example absorption twice when it should only be applied once. It's redundant

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
qwertyuioplkjhgfd: I guess if there was something in the server that allowed stacking juices, this would break it, but the original removeGlassBottle method wouldn't work properly in that case either if the player had glass bottles in their inventory already or if the player had two juices in one stack.

Sfiguz7: The removeGlassBottle is dumb I agree; we should probably just cancel the event, apply the effects, remove one from e.getItem() since we already know it's the correct hand

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CoolerListener.java
Sefiraat: If this PR is to fix an issue, i'd say stay in scope of that issue and then quesiton if the behaviour needs changing separately. The cooler is a different thing and it may well be intended for it to behave differently? IDK and I dont really have an opinion but I do think keep PRs in scope.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/food/Juice.java
qwertyuioplkjhgfd: Setting the item to null will not apply absorption

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CoolerListener.java
qwertyuioplkjhgfd: Agreed

PR #3292, opened 2021-09-30
Removed useless overriding

Sfiguz7:
## Description

#3288 was nice but SolarGenerators override the isChargeable method for no reason. This actually breaks the creation of solar gens with a capacity because the child method is called and that returns false no matter what.
Since the method is actually useless (the super method checks the actual capacity) I have removed it

## Proposed changes

- Removed useless override of isChargeable
- Updated changelog

N/A
## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@
 * Fixed #3273
 * Fixed an exploit regarding the Smithing Table
 * Fixed #3265
-* (API) Made Solar Gens able to have a capacity/energy buffer
+* `SolarGenerator` has a new constructor to accept capacity
 
 ## Release Candidate 28 (06 Sep 2021)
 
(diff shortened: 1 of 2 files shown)


WalshyDev (approved):
Weird, even with this method as it was before the recent change, it would always be false. I guess it was made to not be overriden but no idea why. API expandability is good

svr333 (approved):
LGTM

CHANGELOG.md
```diff
@@ -45,6 +45,8 @@
 * Fixed #3248
 * Fixed #3273
 * Fixed an exploit regarding the Smithing Table
+* Fixed #3265
+* (API) `SolarGenerator` has a new constructor to accept capacity
```
TheBusyBiscuit, line 49: why is this listed under "Fixes"...

PR #3295, opened 2021-10-01
Fix explosive bow knockback being way too high if close to the entity

Sfiguz7:
## Description

As the code is now, the Explosive Bow seems to work fine as long as there is enough distance from the impact point and nearby entities, especially if they are on the same level. When shooting an entity from close by though, especially if directly above, you get shot in the air even 40-60 blocks up which is definitely not supposed to happen.

## Proposed changes

- Cleaned up the extra redundant calculations
- Modified knockback calculation to achieve reasonable pushes

N/A
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
@@ -60,9 +60,13 @@ public class ExplosiveBow extends SlimefunBow {
                 distanceVector.setY(distanceVector.getY() + 0.6);
 
                 Vector velocity = entity.getVelocity();
+                System.out.println("velocity: " + velocity);
+                System.out.println(e.getDamage());
 
                 double distanceSquared = distanceVector.lengthSquared();
+                System.out.println("distanceSq: " + distanceSquared);
                 double damage = e.getDamage() * (1 - (distanceSquared / (2 * range.getValue() * range.getValue())));
+                System.out.println("damage: " + damage);
 
                 if (!entity.getUniqueId().equals(target.getUniqueId())) {
                     EntityDamageByEntityEvent event = new EntityDamageByEntityEvent(e.getDamager(), entity, EntityDamageEvent.DamageCause.ENTITY_EXPLOSION, damage);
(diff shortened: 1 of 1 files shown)


Sfiguz7:
This has been way more of a pain than it should've. Special thanks to @svr333 for testing, @Sefiraat for commenting random stuff in my test builds and @WalshyDev for being extra picky on the values

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -56,21 +56,20 @@ public BowShootHandler onShoot() {
             for (Entity nearby : entites) {
                 LivingEntity entity = (LivingEntity) nearby;
 
-                Vector distanceVector = entity.getLocation().toVector().subtract(target.getLocation().toVector());
-                distanceVector.setY(distanceVector.getY() + 0.6);
-
-                Vector velocity = entity.getVelocity();
+                Vector distanceVector = entity.getLocation().toVector().subtract(target.getLocation().toVector())
+                    .add(new Vector(0, 0.75, 0));
```
WalshyDev, line 60: Format consistentlyy
```suggestion
                Vector distanceVector = entity.getLocation().toVector()
                    .subtract(target.getLocation().toVector())
                    .add(new Vector(0, 0.75, 0));
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
```diff
@@ -56,21 +56,20 @@ public BowShootHandler onShoot() {
...
 
                 double distanceSquared = distanceVector.lengthSquared();
-                double damage = calculateDamage(distanceSquared, e.getDamage());
+                double damage = e.getDamage() * (1 - (distanceSquared / (2 * range.getValue() * range.getValue())));
```
WalshyDev, line 63: We should definitely max this to some value.

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
Sfiguz7: It is, the second term is between 0.5 and 1 so the entire thing is between e.getDamage() and e.getDamage()/2

svr333 (approved):

WalshyDev (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/weapons/ExplosiveBow.java
WalshyDev: neeeeeeeeeeeeeeeeeeerrrrrrrrrrrrrrrrrrrrrrrrrrrrddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd

PR #2912, opened 2021-03-27
Unlock researches in recipe view

RobotHanzo:
## Description
Implemented suggestion 972, offers the ability to unlock items within the recipe book

## Demonstration
[YouTube](https://youtu.be/tJliZAfvBFU)

## Changes
Made non-unlocked items inside a recipe clickable and unlocks the relevant research, sufficient `Validate.notNull`s are used to prevent the most edge cases.

## Checklist
- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [x] I followed the existing code standards and didn't mess up the formatting.
- [ ] I did my best to add documentation to any public classes or methods I added.
- [ ] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

PR #3296, opened 2021-10-02
Unlock researches in recipe view

RobotHanzo:
## Description
A newer and improved version of #2912 (Implemented suggestion 972, offers the ability to unlock items within the recipe book)
With the ability to translate strings related to it

## Demonstration
https://youtu.be/wtYmTkPmMlc

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
@@ -298,9 +298,17 @@ public class SurvivalSlimefunGuide implements SlimefunGuideImplementation {
             menu.addItem(index, new CustomItemStack(ChestMenuUtils.getNoPermissionItem(), sfitem.getItemName(), message.toArray(new String[0])));
             menu.addMenuClickHandler(index, ChestMenuUtils.getEmptyClickHandler());
         } else if (isSurvivalMode() && research != null && !profile.hasUnlocked(research)) {
-            menu.addItem(index, new CustomItemStack(ChestMenuUtils.getNotResearchedItem(), ChatColor.WHITE + ItemUtils.getItemName(sfitem.getItem()), "&4&l" + Slimefun.getLocalization().getMessage(p, "guide.locked"), "",
-                    Slimefun.getLocalization().getMessage(p, "guide.unlock.click"), "",
-                    Slimefun.getLocalization().getMessage(p, "guide.unlock.cost").replace("%cost%", String.valueOf(research.getCost()))));
+            String unlockCost = Slimefun.getLocalization()
+                    .getMessage(p, "guide.unlock.cost")
+                    .replace("%cost%", String.valueOf(research.getCost()));
+            menu.addItem(index, new CustomItemStack(ChestMenuUtils.getNotResearchedItem(),
+                    ChatColor.WHITE + ItemUtils.getItemName(sfitem.getItem()),
+                    "&4&l" + Slimefun.getLocalization().getMessage(p, "guide.locked"),
+                    "",
+                    Slimefun.getLocalization().getMessage(p, "guide.unlock.click"),
+                    "",
+                    unlockCost
+            ));
             menu.addMenuClickHandler(index, (pl, slot, item, action) -> {
                 research.unlockFromGuide(this, p, profile, sfitem, itemGroup, page);
                 return false;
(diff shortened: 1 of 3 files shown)


svr333:
PR looks good on a first look but Ill check again later

svr333 (approved):
Looks good to me

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
```diff
@@ -554,7 +568,7 @@ private void displayItem(ChestMenu menu, PlayerProfile profile, Player p, Object
 
         for (int i = 0; i < 9; i++) {
             ItemStack recipeItem = getDisplayItem(p, isSlimefunRecipe, recipe[i]);
-            menu.addItem(recipeSlots[i], recipeItem, clickHandler);
+            menu.addItem(recipeSlots[i], recipeItem, (recipeItem instanceof UnlockableItemStack) ? lockedClickHandler : clickHandler);
```
WalshyDev, line 571: Extract this ternary into a variable just above, would be cleaner

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
```diff
@@ -632,12 +646,19 @@ private void addBackButton(ChestMenu menu, int slot, Player p, PlayerProfile pro
         if (isSlimefunRecipe) {
             SlimefunItem slimefunItem = SlimefunItem.getByItem(item);
 
-            if (slimefunItem == null) {
+            if (slimefunItem == null || slimefunItem.canUse(p, false)) {
```
WalshyDev, line 649: Think you meant to negate that second expression

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/UnlockableItemStack.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An extension of {@link ItemStack} that indicates the item is unlockable in the guide
+ *
+ * @author RobotHanzo
+ */
+public class UnlockableItemStack extends SlimefunItemStack {
```
WalshyDev, line 21: No need for this class to be public

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
RobotHanzo: Wdym? That is meant to return the original item if the player has unlocked the item

svr333: he means if slimefun item is null OR slimefunItem can NOT be used

right now it says if slimefun item is null or slimefunItem can be used

So in other words I think hes correct here

RobotHanzo: My current if statement here means
if slimefunItem is null (it is not a slimefun item) OR slimefun item can be used (it is already unlocked)
return the original itemstack
else
return a barrier

svr333: ah right then it does make sense; maybe add a small comment to indicate this because it can be confusing?

RobotHanzo: The line right below it indicated everything?

svr333: right my bad, its fine as is

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
```diff
@@ -554,7 +576,13 @@ private void displayItem(ChestMenu menu, PlayerProfile profile, Player p, Object
 
         for (int i = 0; i < 9; i++) {
             ItemStack recipeItem = getDisplayItem(p, isSlimefunRecipe, recipe[i]);
-            menu.addItem(recipeSlots[i], recipeItem, clickHandler);
+            MenuClickHandler handler;
+            if (recipeItem instanceof UnlockableItemStack) {
+                handler = lockedClickHandler;
+            } else {
+                handler = clickHandler;
+            }
```
svr333, line 584:
```suggestion
            MenuClickHandler handler = (recipeItem instanceof UnlockableItemStack) ? lockedClickHandler : clickHandler;

```
if statement works too but Id just stick to the ternary (though it is your choice)

svr333 (dismissed):
still looks good to me

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
RobotHanzo: Walshy thought the if statement is better so

svr333: no walshy just said to extract the ternary to a variable so the method call isnt a big mess
you can still use a ternary if pleased

svr333 (approved):

TheBusyBiscuit (approved):

PR #3297, opened 2021-10-02
Update wiki.json

svr333:
## Description
Updates the wiki.json to be up to date with the current wiki

## Remarks
For the talismans, I had to add the page manually when creating the items (both normal and ender).  
The `REACTOR_COLLANT_CELL` typo is intended, it is also typod like this in Slimefun.

I have added all the missing pages, what I have not done is checked if all the current links are still correct (maybe another time)


TheBusyBiscuit (approved):
I'mma trust ya on this and assume all the links are correct.

svr333:
I went through them one by one opening the sf guide, seeing what pages were missing, and the going to the wiki repo to check what links the pages had

so it should be fine

TheBusyBiscuit:
@Slimefun/code-reviewers bump

WalshyDev (approved):

WalshyDev:
@TheBusyBiscuit do you want to push an update to the changelog for this and merge? I think it'd be nice to include in the changelog but also not sure it's needed either.

TheBusyBiscuit:
Nah, don't wanna mess up the changelog of #3367.
Can add it there though and do a "ci skip" here

PR #3298, opened 2021-10-03
Check for sneaking when breaking logs on lumber axe

qwertyuioplkjhgfd:
## Description

This was an approved suggestion on the discord and I thought it would be helpful for many players.

## Proposed changes

When breaking a log with the lumber axe, check if the player is sneaking so that it doesn't cut down all logs.

[This suggestion on the discord](https://discord.com/channels/565557184348422174/627048923122499584/892393743133577287)

## Checklist


svr333 (approved):
I like this addition, LGTM

On second thought, maybe this change should be reflected in the lore?

Sfiguz7 (approved):
LGTM and although it may be nice for users to have it in the lore I don't feel like it's necessary, since the Explo pick does not have it either. It would also break items on older versions or if they have backwards compat on so this is fine in my opinion

PR #3307, opened 2021-10-06
Solar Generators - Return night energy in non overworld dimensions

svr333:
## Description
When using solar panels outside of the overworld, it will return 0 as power used currently.
This change makes it so it will return the night energy instead.

## Proposed changes
- Return night energy in other dimensions but overworld

[Discord Suggestion 12](https://discord.com/channels/565557184348422174/627048923122499584/698913136673292299)

## Checklist


Sefiraat:
I agree but a nether equivalent of the solar gens OR having this as a config option would be nice as I have experienced players regretting building in the nether due to this and it seems a shame to constrain or limit just because of realism - esp. as this was an approved suggestion.

PR #3309, opened 2021-10-07
Implement Middle Click support for Slimefun Items

svr333:
## Description
This PR will add support to retrieve SlimefunItems by middle clicking on them.

## Proposed changes
- Add a listener to check for middle clicks on blocks

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -0,0 +1,95 @@
...
+            boolean isOutsideInventoryClick = e.getCursor().getType() == b.getType() && e.getCurrentItem().getType() == Material.AIR;
+
+            /*
+             * This is an edge case where the player is looking at a WALL_HEAD (eg. cargo)
+             * and then the boolean above wont match because WALL_HEAD != PLAYER_HEAD.
+             * This check makes up for that lack.
+             */
+            boolean isPlayerWallhead = b.getType() == Material.PLAYER_WALL_HEAD && e.getCursor().getType() == Material.PLAYER_HEAD;
+
+            if (isOutsideInventoryClick || isPlayerWallhead) {
```
TheBusyBiscuit, line 65: I feel like there will be more edge cases like the player head coming up in the future.
So it would be a good idea to move this into a `private boolean` method so it can easily be modified later.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -0,0 +1,95 @@
...
+                Optional<String> blockId = Slimefun.getBlockDataService().getBlockData(b);
```
TheBusyBiscuit, line 66: This would only work for Tile entities though, for general id grabbing you would need to use `BlockStorage`

TheBusyBiscuit (changes requested):
Thanks for adding these comments, I think I understand the proposed changes a lot better now

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -0,0 +1,102 @@
...
+            // vanilla block -> ignore
+            if (sfItem == null) {
+                return;
+            }
+            
+            /* 
+                * Before giving the item to the user, check if you can swap
+                * to the item instead (user already has item in hotbar).
+                * This is sometimes bypassed by the client itself (not fixable though).
+                */
```
TheBusyBiscuit, line 65:
```suggestion
            /* 
             * Before giving the item to the user, check if you can swap
             * to the item instead (user already has item in hotbar).
             * This is sometimes bypassed by the client itself (not fixable though).
             */
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -0,0 +1,102 @@
...
+            for (int i = 0; i < 9; i++) {
+                SlimefunItem hotbarItem = SlimefunItem.getByItem(player.getInventory().getItem(i));
+                if (hotbarItem != null && hotbarItem.getId() == sfItem.getId()) {
```
TheBusyBiscuit, line 68: Can also do `sfItem.isItem(...)` here which should be better and more optimized

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -0,0 +1,102 @@
...
+            // get the block the player is looking at for later
+            Block b = player.getTargetBlockExact(5);
+
+            if (!isActualMiddleClick(e, b)) {
+                return;
+            }
+
+            // find the actual slimefun item the user is looking at
+            String id = BlockStorage.checkID(b);
+            SlimefunItem sfItem = SlimefunItem.getById(id);
```
TheBusyBiscuit, line 54:
```suggestion
            SlimefunItem sfItem = BlockStorage.check(b);
```

TheBusyBiscuit (changes requested):

TheBusyBiscuit (dismissed):
LGTM

svr333:
@Sfiguz7 whenever you arent too busy with exames, would you mind reviewing
you were there when I was coding this, might be useful insight

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -0,0 +1,100 @@
...
+        * after this we can make sure that it is a middle click outside of the inventory
+        * currentItem should also be air, otherwise it is not outside of the inventory
+        */
+        boolean isOutsideInventoryClick = e.getCursor().getType() == b.getType() && e.getCurrentItem().getType() == Material.AIR;
+
+        /*
+        * This is an edge case where the player is looking at a WALL_HEAD (eg. cargo)
+        * and then the boolean above wont match because WALL_HEAD != PLAYER_HEAD.
+        * This check makes up for that lack.
+        */
```
WalshyDev, line 95:
```suggestion
         * On a middle click outside the user inventory, cursor will be set
         * to the actual block that is middle clicked, while currentItem will be AIR.
         *
         * This check is really weird due to the weird nature of this event's behaviour.
         * It checks if the block the player is looking at is of the same type as the cursor;
         * after this we can make sure that it is a middle click outside of the inventory
         * currentItem should also be air, otherwise it is not outside of the inventory
         */
        boolean isOutsideInventoryClick = e.getCursor().getType() == b.getType() && e.getCurrentItem().getType() == Material.AIR;

        /*
         * This is an edge case where the player is looking at a WALL_HEAD (eg. cargo)
         * and then the boolean above wont match because WALL_HEAD != PLAYER_HEAD.
         * This check makes up for that lack.
         */
```

WalshyDev (approved):
LGTM

svr333:
according to sefi/var there have been some changes to middle click in 1.18 so Ill have to do some more testing before merge

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -0,0 +1,100 @@
...
+                    return;
+                }
+            }
+
+            // Give the item, doing it like this will not alter any other cases.
+            e.setCursor(sfItem.getItem().clone());
+        }
+    }
+
+    private boolean isActualMiddleClick(InventoryCreativeEvent e, Block b) {
```
TheBusyBiscuit, line 79: params are nonnull by default :eyes:

TheBusyBiscuit (changes requested):
Pretty much approving this but I'd rather want this to be build 1001.
Luckily, I found an extremely dangerous bug that needs to be fixed before I would ever be able to merge it ;)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
svr333: NOOOO MY BUILD 1000 :(((((((((((((((

TheBusyBiscuit (approved):
LGTM

WalshyDev (approved):
LGTM - Won't merge yet

Sfiguz7 (approved):

PR #3311, opened 2021-10-07
Implement the ability for debug logging

WalshyDev:
## Description
This is a simple addition. It allows for us to add in debug logging when we want to find specific info. For example, if someone's cargo is taking a while we can ask them to do:
`/sf debug slimefun_cargo_input_testing` 

This will enable some debug logs to print in the console. Then we simply ask them for the logs and see what is happening. This will be very helpful.

This also can be used for addons, we do not maintain an enum or anything. It's a simple string to specify the debug "test".

## Proposed changes
* Created `Debug` class - used to log and set the test
* Created debug command to enable and disable testing
* Added some debug logs for cargo

N/A

## Checklist

### DIFF
--- a/.github/workflows/yaml-linter.yml
+++ b/.github/workflows/yaml-linter.yml
@@ -18,6 +18,6 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v2
     - name: YAML Linter
-      uses: ibiqlik/action-yamllint@v3.0.4
+      uses: ibiqlik/action-yamllint@v3.1.0
       with:
         config_file: '.github/configs/yaml-linter.yml'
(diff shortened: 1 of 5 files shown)


WalshyDev:
> Looks good to me, some nitpickings left and two things I wanna address:
> 
> 1. What about tab completion?
> 2. Should there maybe be something like `/sf debug info` which tells you which test case is currently running?

1. Well... tab completion isn't easy considering this should support addons or such too. Like, we could tab complete ours only I guess, just feels a bit weird.
2. Yep, would be good. I'd probably just do `/sf debug` shows it though

TheBusyBiscuit:
What if we have a static registration method for test cases?

TheBusyBiscuit:
Alright, what about having just `disable` and `<test case>` as options?

WalshyDev:
What is `<test case>` here? A literal String "<test case>", seems like a weird thing to do in tab-complete. Current wouldn't make sense either.

I can add `disable` though

TheBusyBiscuit:
Oh, also there is a missing `package-info.java` file!

PR #3312, opened 2021-10-08
Fixes for #3260 and removal of unnecesary parameter of method getCharge()

### LINKED ISSUE #3260: Generators power multiple energy network with unnerfed rate
### ❗ Checklist

### 📍 Description

### 📑 Reproduction Steps

1. Put the solar generator
3. Put the capacitor

### 💡 Expected Behavior

### 📷 Screenshots / Videos

image

### 📜 Server Log

N/A

N/A

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.16.x

### ⭐ Slimefun version

DEV - 968

### 🧭 Other plugins

N/A

TweepCoding:
## Description

This pull request's objective is to fix the problem with energy duplication and removing an unnecessary parameter in a function.

## Proposed changes

What this pull request does is adding a value to the BlockData
of both generators and consumers of energy that tells them which network they belong on, and allows said network
to only recieve energy and consume energy from machines within its own network

Also, removed the location parameter in getCharge() of EnergyNetComponent, since it's not needed to get the charge
of a component and it's only use is to get the underlying Config data of said component.

Resolves #3260
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
@@ -3,6 +3,7 @@ package io.github.thebusybiscuit.slimefun4.core.attributes;
 import java.util.logging.Level;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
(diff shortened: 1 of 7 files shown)


TheBusyBiscuit:
Closing this as stale.

Issue #1151, opened 2019-10-04
Saw Table making wood without any wood logs


SuperSlimey5:
If i right click the Saw Table without anything it gives me wood logs.

![2019-10-04_18 22 31](https://user-images.githubusercontent.com/47258749/66192222-57f9fe00-e6d3-11e9-83ff-74faa9a4dcb1.png)
![2019-10-04_18 22 27](https://user-images.githubusercontent.com/47258749/66192223-57f9fe00-e6d3-11e9-92cf-528c698e377b.png)

![Capture](https://user-images.githubusercontent.com/47258749/66192159-37ca3f00-e6d3-11e9-8a90-cea8e2ed2341.JPG)

WalshyDev:
Update SF and CS-CoreLib.

PR #3347, opened 2021-11-09
Added Advanced and Empowered Android Woodcutter (approved suggestion)

FN-FAL113:
## Description
Based from the approved suggestion ```#1151``` - chaelzinho_15#1592, this should add the needed android that utilizes nuclear fuels.

## Proposed changes
- Added a Empowered Programmable Android (Woodcutter)
- Added a research for the Empowered Programmable Android (Woodcutter)
- Added a Advanced Programmable Android (Woodcutter)
- Added a research for the Advanced Programmable Android (Woodcutter)

Approved suggestion ```#1151``` on discord (05/03/2021)

## Checklist


svr333 (dismissed):
LGTM

TerslenK:
will you gonna add advanced one aswell?

FN-FAL113:
> will you gonna add advanced one aswell?

Will do after I'm done with some of my studies

svr333:
I didn't realize Advanced woodcutter wasn't a thing yet

FN-FAL113:
Indeed, the suggestion was based off using nuclear fuels which is the empowered one but why not add advanced for it too :)

svr333 (approved):
Looks fine to me

TheBusyBiscuit (changes requested):
This PR is pretty useless in my opinion.
These new androids add literally no additional function other than "being Tier 2/3" and the fuel change that comes with it...

Same reason why I have rejected that idea several times in the past.
Unless they add any benefit, there's no reason to add it. Take the other Androids for example:

Fisherman: Higher chance of catching fish/loot
Butcher: Extra damage per tier
Farmer: Ability to harvest ExoticGarden plants

Same reason as to why there is no Tier 2/3 Miner android...

FN-FAL113:
I'll take a look on what I can do

FN-FAL113:
Will higher chance of pushing amount (amount * 2 at a chance of 15 * getTier()) of logs to the android be enough for an additional function?

svr333:
I feel like having a small chance of double output should be a fine upgrade, though I think even without it, it should be fine to have the upgrades. Some people just prefer to use higher tier fuel for androids.

But if it must, then Im fine with double drops

TheBusyBiscuit:
double-drops seems like a rather "meh" route here, it essentially creates blocks out of thin air, which may not be desirable for everyone.
For woodcutters I'd personally perhaps have added a low chance of getting apples (or nether warts for nether trees?).
But I feel like it would be more annoying junk for people who have sophisticated auto-setups.
Range and things like that would also require to nerf the existing Woodcutter android which would be even more annoying for existing players.

It's not exactly an easy feature to think of and that's the reason why there has not been such an android to this day.
But I would really appreciate if there was a solution to this issue before merging this.

FN-FAL113:
I'm closing this anytime if I can't think a good function for the androids, Since I thought at first people only wanted fuel types for the android based on the suggestion.

FN-FAL113:
Can't focus on this pr anymore, been sad my pet dog is I'll.

PR #3355, opened 2021-11-25
Added option to allow slimefun Items to be useable without research

CuddleTheif:
## Description

Added a new option to the config file called dont-need-research-to-use. If set to true slimefun items (with the exception of multiblock structures) will be usable by players even without reaserching the item first. They will still need the research to use multiblock structures and to make any slimefun items.

## Proposed changes

Slight change to canUse method in the SlimefunItem to to check the config setting if needed to be researched. 
Also added a parameter to check ignoring the new config setting for multilocks and for crafting.

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
@@ -1066,8 +1066,8 @@ public class SlimefunItem implements Placeable {
      * @param sendMessage
      *            Whether to send that {@link Player} a message response.
      * @param forceResearch
-     *            If you should the research is required even if require-research-on-use is false
-     * 
+     *            Whether the research is required even if require-research-on-use is false
+     *
      * @return Whether this {@link Player} is able to use this {@link SlimefunItem}.
      */
     public boolean canUse(@Nonnull Player p, boolean sendMessage, boolean forceResearch) {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
@@ -1021,6 +1021,7 @@ public class SlimefunItem implements Placeable {
         }
     }
 
+	
     /**
      * This method checks if the given {@link Player} is able to use this {@link SlimefunItem}.
      * A {@link Player} can use it if the following conditions apply:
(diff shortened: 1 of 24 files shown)


WalshyDev:
What's the reason for this change? :thinking: 
Can you provide a scenario in which this would be useful?

CuddleTheif:
For servers that what some players to create items and trade, sell, etc those items to other players. 
Also for when you have an elevator in your base and you want your friends to visit but they can't travel cuz they haven't gotten that far. That kind of stuff

Boomer-1:
for whatever it's worth, server owners can already disable research for specific items. 

As for just removing researches without server configuration, I would highly caution against this. that dramatically alters the game play of slimefun

WalshyDev:
Yeah as Boomer states, you can already disable researches overall and specific ones. I can't see a case where this is is required over the already existing features.

CuddleTheif:
But I;m not suggesting you disable the research. Just for the sake of using the items. You still need the research to make them. and it wouldn't be without server configuration. I added it as a config setting that would default be off.

Boomer-1:
this would still allow players to not have to do any research to use the items and still dramatically alters the game. all it would take is one player to make things, everyone else buys them. I still would highly discourage this request.

CuddleTheif:
Couple things. 
1. Yes it would "alter gameplay" but that is the point. For people who would prefer this type of gameplay where the player has the choice to either research and make it themselves or just buy it from another player. That's why it would be a setting. So if a server wants this type of gameplay they can and if they don't they leave it off. Just like the settings for disabling research.
2. There are already exists an item that allows players to share all their research with another player. So just as easily a player could make a ton of those and sell those and it have a 'worse' effect. So it's not even that big of a change 
3. This is an addition I wanted and a server actual requested from me and is currently using and enjoying.

src/main/resources/config.yml
```diff
@@ -28,6 +28,7 @@ researches:
   free-in-creative-mode: true
   enable-fireworks: true
   disable-learning-animation: false
+  dont-need-research-to-use: false
```
Sefiraat, line 31: This wording isn’t very straight forward, perhaps require-research-on-use and default this to true instead?

Sefiraat (commented):
Generally I don’t think the PR is useful and only serves to create shortcuts in a linear gameplay where the most requested thing is more content. However as long as it’s a config option that defaults to being off, then it’s a direct choice and while I think server owners will regret it, choice is choice.

src/main/resources/config.yml
svr333: definitely default the behavior to what it currently is

CuddleTheif: That makes sense, easy change. (Just as a note to svr333 it was set to the default behavior to what it currently is he is just suggesting a name change and fitting value with the new name but still keep it to the default behavior to what it currently).

Boomer-1:
the optional setting is better. Just for the record, that tomb of knowledge is the #1 disabled item in public servers iirc.

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
```diff
@@ -1043,6 +1044,33 @@ public void error(@Nonnull String message, @Nonnull Throwable throwable) {
...
+     * 
+     * If any of these conditions evaluate to <code>false</code>, then an optional message will be
+     * sent to the {@link Player}.
+     * 
+     * @param p
+     *            The {@link Player} to check
+     * @param sendMessage
+     *            Whether to send that {@link Player} a message response.
+     * @param forceResearch
+     *            If you should have the reasearch even if require-research-on-use is true
```
svr333, line 1069:
```suggestion
     * @param forceResearch
     *            If you should the research is required even if require-research-on-use is false
```

svr333 (changes requested):
Im gonna wait before I review further to see whether or not this will be accepted into Slimefun

Seeing as @WalshyDev doesn't really seem to like the idea?

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
svr333: my english left me there but I edited the suggestion so it is nicer, sorry

TheBusyBiscuit:
I'm sorry if this comes a bit late but I've been fairly busy.
I don't feel comfortable merging a PR like this as it adds a substantial amount of complexity to an already fairly loaded feature stack. Parts of this should be rewritten to make it more expandable.

But in its current state, this PR is not to be merged.
The implementation is messy but would come with huge implications for the overall system but also gameplay.
And adding onto this, it is not entirely certain if this is a desirable change in the first place.

I appreciate the effort but I am rejecting this for now.

PR #3362, opened 2021-11-30
Add Hologram Controller

Apeiros-46B:
## Description
I made this pull request because of a few discord suggestions and because I also want this feature.

## Proposed changes
I have added a new item called the Hologram Controller, which is unlocked together with the Hologram Projector. Right clicking on an Energy Regulator, Cargo Manager, GEO Miner, Nuclear Reactor, or Nether Star Reactor with a Hologram Controller in the hand will toggle the hologram.

Discord suggestion 899
Discord suggestion 1107
Discord suggestion 1121
Discord suggestion 1175

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
@@ -50,6 +50,8 @@ public class CargoNet extends AbstractItemNetwork implements HologramOwner {
     protected final Map<Location, Integer> roundRobin = new HashMap<>();
     private int tickDelayThreshold = 0;
 
+    public boolean showHologram = true;
+
     @Nullable
     public static CargoNet getNetworkFromLocation(@Nonnull Location l) {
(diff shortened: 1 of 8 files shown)


TheBusyBiscuit (changes requested):
:NotLikeThis:

This PR is so inherently redundant...
You implemented a feature into almost every implementation of `HologramOwner`.
I know it is an interface, so you cannot have a field there but this is the wrong approach nonetheless.

The item class itself is also super redundant.

I also think the term "Controller" is not quite right here, how about "Hologram Switch".

On another note, I see absolutely no sign of persistence here, the networks will forget this setting upon server restart.

Apeiros-46B:
I didn't even notice most of these issues before you mentioned them :NotLikeThis:
I don't have much time right now but I'll try to fix most of the issues you mentioned asap

TheBusyBiscuit (changes requested):
In its current state, this is not really mergeable.

Apeiros-46B:
Should be mergeable now, let me know if you notice any issues

TheBusyBiscuit:
Closing this as stale.

PR #3363, opened 2021-12-01
Register filled flask as a Slimefun Item

qwertyuioplkjhgfd:
## Description

Currently, ``FILLED_FLASK_OF_KNOWLEDGE`` is the only Slimefun Item that has a ``SlimefunItemStack`` but is never registered as a ``SlimefunItem`` in ``SlimefunItemSetup``. This meant that calls to ``SlimefunItem.getById("FILLED_FLASK_OF_KNOWLEDGE")`` would return ``null``. This pull request registers a SlimefunItem for the ItemStack with minimal behavioral changes.

## Proposed changes

- Create ``FilledKnowledgeFlask``, which is a ``SlimefunItem`` that implements ``NotConfigurable`` and is hidden in the Slimefun guide.
- Create a new ``RecipeType``: ``RecipeType.INTERACT``, which signifies that the item is obtained by interacting with the ingredient.
- Create and register a ``FilledKnowledgeFlask`` instance in ``SlimefunItemSetup``.

The overall effects of these changes are:
- When clicking a filled flask of knowledge in the Slimefun guide, it will now display the method of obtaining a filled flask, which is interacting with an empty flask of knowledge. 
- ``FILLED_FLASK_OF_KNOWLEDGE`` will be added to the ``SlimefunRegistry``.

``SlimefunItem.getById("FILLED_FLASK_OF_KNOWLEDGE")`` no longer returns ``null``.

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/HiddenItem.java
```diff
@@ -0,0 +1,26 @@
...
+import io.github.thebusybiscuit.slimefun4.core.attributes.NotConfigurable;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+
+import org.bukkit.inventory.ItemStack;
+
+/**
+ * The {@link HiddenItem} is a {@link NotConfigurable} {@link SlimefunItem}
+ * that is hidden from the Slimefun guide.
+ *
```
svr333, line 16:
```suggestion
 * The {@link HiddenItem} is a {@link NotConfigurable} {@link SlimefunItem}
 * that is hidden from the Slimefun guide.
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/HiddenItem.java
```diff
@@ -0,0 +1,26 @@
...
+ */
+public class HiddenItem extends SlimefunItem implements NotConfigurable {
+
+    @ParametersAreNonnullByDefault
+    public HiddenItem(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        super(itemGroup, item, recipeType, recipe);
+        this.setHidden(true);
+    }
+
+}
```
svr333, line 26:
```suggestion
}
```

svr333 (changes requested):

poma123 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/HiddenItem.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * The {@link HiddenItem} is a {@link NotConfigurable} {@link SlimefunItem}
+ * that is hidden from the Slimefun guide.
+ */
+public class HiddenItem extends SlimefunItem implements NotConfigurable {
+
+    @ParametersAreNonnullByDefault
+    public HiddenItem(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        super(itemGroup, item, recipeType, recipe);
+        this.setHidden(true);
```
svr333, line 22: couldnt we default the recipe to an empty `ItemStack[]` (or 9 nulls) meaning the ctor would get this declaration:
`public HiddenItem(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType)`
since this will always be hidden?

That way we dont have to initialize them in `SlimefunItemSetup` with a lot of nulls

qwertyuioplkjhgfd: well right now it has the empty flask as an ingredient, i guess it would be better to have it just be nulls, would that be better?

qwertyuioplkjhgfd:
how do you squash

svr333:
squashing is done right before we merge into another branch, so one of the maintainers has to do it

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/HiddenItem.java
svr333: well the thing is the recipe doesnt matter right?
so we could make the constructor of HiddenItem something like:

```java
    public HiddenItem(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType) {
        super(itemGroup, item, recipeType, new ItemStack[] {});
        this.setHidden(true);
```

this means that if we then create an item somewhere else, we dont even have to initalize a recipe, because its not used anyway

poma123: > the thing is the recipe doesnt matter right?

I don't really agree. It does matter as we have a new RecipeType that shows players what item should be right-clicked with to obtain the result item (in this case the filled knowledge flask).
This only makes sense if we do show the recipe item (in this case the empty flask)

svr333: fair enough, but in that case should we just have a singular ItemStack then, instead of an array?

svr333:
```java
    public HiddenItem(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack recipe) {
        super(itemGroup, item, recipeType, new ItemStack[] { recipe, null, null, null, null, null, null, null, null });
        this.setHidden(true);
```

poma123: What about `recipeItem` instead?

qwertyuioplkjhgfd: That wouldn't work actually. Hidden items are not necessarily of RecipeType INTERACT. This would limit hidden items to only be able to have a single ingredient, which I'm not sure is the intention.

qwertyuioplkjhgfd:
bump; not sure if the requested changes are intended

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/HiddenItem.java
poma123: Ah, good point. We should then leave it as it was before this, I don't think that the 8 nulls would be a huge problem anyway. We have a bunch of them in that file already, I think it's okay.

svr333 (approved):
good for me then

poma123 (changes requested):
> When clicking a filled flask of knowledge in the Slimefun guide, it will now display the method of obtaining a filled flask, which is interacting with an empty flask of knowledge.

I mean if it is hidden from the guide, how the heck would it display IN the guide how to obtain it if it is hidden already..? It's clearly not in the guide since it is a HiddenItem... This doesn't make any sense to me.

svr333:
> I mean if it is hidden from the guide, how the heck would it display IN the guide how to obtain it if it is hidden already..? It's clearly not in the guide since it is a HiddenItem... This doesn't make any sense to me.

well yeah thats the first remark I made as well, you know if the item is hidden the recipe is never shown

qwertyuioplkjhgfd:
> > When clicking a filled flask of knowledge in the Slimefun guide, it will now display the method of obtaining a filled flask, which is interacting with an empty flask of knowledge.
> 
> I mean if it is hidden from the guide, how the heck would it display IN the guide how to obtain it if it is hidden already..? It's clearly not in the guide since it is a HiddenItem... This doesn't make any sense to me.

it will show the recipe if you click on a filled flask in a recipe of another item, such as the talisman of the wise.

poma123 (approved):
Okay then.

svr333 (approved):

TheBusyBiscuit (approved):
LGTM

PR #3377, opened 2021-12-18
Improved the API for BiomeMap

WalshyDev:
## Description
BiomeMaps will become a big part of addons supporting 1.18+ and the current implementation doesn't make it easy or fun to implement. So, this PR will expand the API a bit in order to easily support third-party plugins/addons :)

Here is what you'd need to do, what you can do now and an even easier way:
```java
try {
    // What you'd need to do :/
    final BiomeMap<Integer> map = BiomeMap.fromJson(
        new NamespacedKey(this, "test"),
        String.join("", Files.readAllLines(new File(getClass().getResource("/biome-maps/test.json").getFile()).toPath())),
        JsonElement::getAsInt
    );

    // now supported!
    final BiomeMap<Integer> map1 = BiomeMap.fromResource(
        new NamespacedKey(this, "test1"),
        this,
        "/biome-maps/test.json",
        JsonElement::getAsInt
    );

    System.out.println(map1.getOrDefault(Biome.BADLANDS, 0));
    System.out.println(map1.getOrDefault(Biome.BEACH, 0));

    // Simple :)
    final BiomeMap<Integer> map2 = BiomeMap.getIntMapFromResource(
        new NamespacedKey(this, "test2"),
        this,
        "/biome-maps/test.json"
    );
} catch (Exception e) {
    e.printStackTrace();
}
```

## Proposed changes
* Implement a new `BiomeMap.fromResource` method which accepts a JavaPlugin
* Implement some common implementation APIs
  * `getIntMapFromResource(NamespacedKey, JavaPlugin, String path)`
  * `getLongMapFromResource(NamespacedKey, JavaPlugin, String path)`
  * `getStringMapFromResource(NamespacedKey, JavaPlugin, String path)`

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/biomes/BiomeMap.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/biomes/BiomeMap.java
@@ -22,6 +22,7 @@ import com.google.gson.JsonElement;
 import io.github.thebusybiscuit.slimefun4.api.exceptions.BiomeMapException;
 import io.github.thebusybiscuit.slimefun4.api.geo.GEOResource;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+import org.bukkit.plugin.java.JavaPlugin;
 
 /**
  * {@link BiomeMap}s are used to map data values to {@link Biome} constants.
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/biomes/BiomeMap.java
```diff
@@ -150,13 +151,34 @@ public String toString() {
 
     @ParametersAreNonnullByDefault
     public static <T> @Nonnull BiomeMap<T> fromResource(NamespacedKey key, String path, BiomeDataConverter<T> valueConverter) throws BiomeMapException {
+        return fromResource(key, Slimefun.instance(), path, valueConverter);
+    }
```
TheBusyBiscuit, line 155: I completely forgot that different resources will need a different `JavaPlugin` instance. :NotLikeThis:

In my opinion, we can just drop this method and pass `Slimefun.instance()` for our own resources, it'd be too tempting to expose this method here.

WalshyDev: Agreed

src/main/java/io/github/thebusybiscuit/slimefun4/utils/biomes/BiomeMap.java
```diff
@@ -22,6 +22,7 @@
 import io.github.thebusybiscuit.slimefun4.api.exceptions.BiomeMapException;
 import io.github.thebusybiscuit.slimefun4.api.geo.GEOResource;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
```
TheBusyBiscuit, line 24: Import no longer needed
```suggestion
```

TheBusyBiscuit (changes requested):
One unused import otherwise looking good

TheBusyBiscuit (approved):
LGTM

Sfiguz7 (approved):

PR #3381, opened 2021-12-21
make parachute class respect constructor recipetype parameter

qwertyuioplkjhgfd:
## Description

right now, regardless of what recipetype you construct parachute with, it will always be enhanced crafting table
this fixes that

## Proposed changes

make parachute respect the constructor parameter

N/A

## Checklist


TheBusyBiscuit (approved):
lmao

Sfiguz7 (approved):
Oh dam ahahaha Nice spot!

PR #3383, opened 2021-12-22
check for meta even if ids match

### LINKED ISSUE #3379: Cargo ignores lore of slimefun items, causing backpacks to stack.
### ❗ Checklist

### 📍 Description

When sending newly crafted backpacks of the same size thru cargo from one chest to another the backpacks became stacked and all had the same id #.

### 📑 Reproduction Steps

Put a dozen or so newly crafted backpacks into a chest. 
setup a cargo node to blacklist to send to another chest
put a same sized backpack into the cargo node on whitelist
as items transfer they will stack.

### 💡 Expected Behavior

I would expect backpacks to keep their lore and not stack.

### 📷 Screenshots / Videos

https://youtu.be/O00yWZRFdno

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

dev 987

### 🧭 Other plugins

_No response_

qwertyuioplkjhgfd:
## Description

this prevents backpacks from being marked as similar even if they have different ids

## Proposed changes

check for item meta once the slimefun ids are matched

related: #3379 
do note that this only fixes backpacks since the id is stored in the lore, but dankpacks still stacks as their id is stored in nbt

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -276,7 +276,8 @@ public final class SlimefunUtils {
         } else if (checkAmount && item.getAmount() < sfitem.getAmount()) {
             return false;
         } else if (sfitem instanceof SlimefunItemStack && item instanceof SlimefunItemStack) {
-            return ((SlimefunItemStack) item).getItemId().equals(((SlimefunItemStack) sfitem).getItemId());
+            return ((SlimefunItemStack) item).getItemId().equals(((SlimefunItemStack) sfitem).getItemId()) &&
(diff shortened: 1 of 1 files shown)


WalshyDev:
Is there ever a scenario where display name matters? getDisplayName is a very intensive call and only getLore seems to matter here. I feel like this is the wrong approach and just ruins the whole performance improvement for no reason

Alongside that, we should _only_ run this when needed. Not just for all items (again, just ruins the entire performance improvement)

Sefiraat:
Also I assigned the issue to myself as I have started work on this already. Please refrain from picking up work where people are assigned. You just end up wasting either your time of mine which isn’t very nice in the long run.

PR #3397, opened 2021-12-31
Allow addons to specify wiki page to items

ybw0014:
## Description

Some addons have wiki pages too, however current codes only allow core slimefun items to show the wiki item in their recipes.
I add a feature to allow addons to speficy the wiki prefix, and call `SlimefunItem#addWikipage()` to show the wiki icon.

## Proposed changes

- Add `getWikiPrefix()` in `SlimefunAddon` to specify the wiki prefix of addon.
- Rename `addOfficialWikipage()` to `addWikipage()` in `SlimefunItem`, and call `SlimefunAddon#getWikiPrefix()` to get the wiki prefix.
- Add `WikiSetup` class to setup wiki from "wiki.json".
- Change the location of `addWikipage` in `Talisman` because it now needs to be called after registering.

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/SlimefunAddon.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/SlimefunAddon.java
@@ -1,6 +1,5 @@
 package io.github.thebusybiscuit.slimefun4.api;
 
-import java.util.Optional;
 import java.util.logging.Logger;
 
 import javax.annotation.Nonnull;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/SlimefunAddon.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/SlimefunAddon.java
@@ -1,5 +1,6 @@
 package io.github.thebusybiscuit.slimefun4.api;
 
+import java.util.Optional;
 import java.util.logging.Logger;
 
 import javax.annotation.Nonnull;
@@ -23,6 +24,7 @@ import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
  * an Addon.
  * 
  * @author TheBusyBiscuit
+ * @author ybw0014
  *
  */
 public interface SlimefunAddon {
@@ -96,5 +98,12 @@ public interface SlimefunAddon {
         PluginDescriptionFile description = getJavaPlugin().getDescription();
         return description.getDepend().contains(dependency) || description.getSoftDepend().contains(dependency);
     }
+
+    /**
+     * This method returns the prefix of the wiki link of this addon.
+     *
+     * @return The prefix of wiki link, null if wiki does not exist
+     */
+    default @Nullable String getWikiPrefix() { return null; }
 
 }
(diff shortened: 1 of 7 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/setup/WikiSetup.java
```diff
@@ -0,0 +1,62 @@
+package io.github.thebusybiscuit.slimefun4.api.setup;
```
TheBusyBiscuit, line 1: First off this is not allowed to be inside the `api` package

src/main/java/io/github/thebusybiscuit/slimefun4/api/setup/WikiSetup.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public static void fromJson(JavaPlugin addon) {
+        if (!(addon instanceof SlimefunAddon)) {
+            throw new IllegalArgumentException("This is not a Slimefun addon");
+        }
+
+        addon.getLogger().log(Level.INFO, "Loading Wiki pages...");
+
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(addon.getClass().getResourceAsStream("/wiki.json"), StandardCharsets.UTF_8))) {
+            JsonElement element = JsonParser.parseString(reader.lines().collect(Collectors.joining("")));
```
TheBusyBiscuit, line 47: Do not use static `JsonParser` methods

src/main/java/io/github/thebusybiscuit/slimefun4/api/setup/WikiSetup.java
```diff
@@ -0,0 +1,62 @@
...
+public final class WikiSetup {
+
+    private WikiSetup() {}
+
+    /**
+     * This method will load wiki pages from "/wiki.json" file.
+     *
+     * @param addon The {@link SlimefunAddon} instance that is setting up wiki pages
+     */
+    public static void fromJson(JavaPlugin addon) {
```
TheBusyBiscuit, line 39: Poorly named method

src/main/java/io/github/thebusybiscuit/slimefun4/api/SlimefunAddon.java
```diff
@@ -97,4 +98,11 @@ default boolean hasDependency(@Nonnull String dependency) {
         return description.getDepend().contains(dependency) || description.getSoftDepend().contains(dependency);
     }
 
+    /**
+     * This method returns the prefix of the wiki link of this addon.
+     *
+     * @return The prefix of wiki link, null if wiki does not exist
+     */
+    default @Nullable String getWikiPrefix() { return null; }
```
TheBusyBiscuit, line 106: No one-liner methods

TheBusyBiscuit (changes requested):
This is a hard pass.

First off, you cannot use static `JsonParser` methods as this will break compatibility with all Minecraft versions but 1.18 (or 1.17 maybe).

But mainly, the ingame wiki is always referenced as "the official Slimefun wiki" in all locale strings.
Setting the official wiki string to an arbitrary string can be easily abused.

I agree that addons should have wiki support again at some point but it was intentionally removed and if it were to be re-added then properly differently to this.

Issue #3408, opened 2022-01-07
Feature request - Call entity damage event when storm staff causes lightning damage


ChanceSD:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

As requested by @Sfiguz7 
Basically to have some way for other plugins to identify that lightning from storm staff came from a player so it can be blocked as PvP. Suggestions were metadata or calling the damage event.

### 📑 Reproduction Steps

N/a

### 💡 Expected Behavior

N/a

### 💻 Server Software

Spigot

### 🎮 Minecraft Version

1.18.x

### ⭐ Slimefun version

N/a

TheBusyBiscuit:
@Sfiguz7 You very well know feature requests do not belong on the bug tracker...

Sfiguz7:
I do but we don't really have a place except here. I don't think this should be made a suggestion, it is not exactly an addition just a compatibility fix that would make Slimefun's accessibility better overall, and I still wanted to have it written somewhere to reference when PRing

variananora:
A little off topic but I think this could 'fix' #3396

TheBusyBiscuit:
Closing this issue as it's not really a bug report and its already served its purpose of reminding @Sfiguz7 to make a PR.
The discussion can continue in #3411 if there is a need for further discussion in the first place.

PR #3411, opened 2022-01-07
Storm Staff PVP overhaul to improve accessibility

Sfiguz7:
## Description
This PR is twofold: on one hand, it transforms all the damage dealt by lightnings to damage dealt by the caster; on the other, it adds the ability to enable StormStaff usage in non PVP areas (as any damage is now automatically nullified, having the EntityDamageByEntityEvent been "transformed" to attribute the damage to the caster).

## Proposed changes
- Substituted any EntityDamageByEntityEvent caused by StormStaff lightnings with ones with the caster as damager
- Added an ItemSetting to allow enabling usage of Storm Staves in non PVP area (damage to players is prevented)

#3396
#3408 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/staves/StormStaff.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/staves/StormStaff.java
@@ -43,7 +43,8 @@ public class StormStaff extends LimitedUseItem {
     private static final NamespacedKey usageKey = new NamespacedKey(Slimefun.instance(), "stormstaff_usage");
     private final ItemSetting<Boolean> allowNonPVPUse = new ItemSetting<>(this, "allow-non-pvp-use", false);
 
-    public static final Map<UUID, UUID> stormStaffLightnings = new HashMap<>();
+    // This is a Map<LightningStrikeUuid, PlayerUuid> to keep track of the caster/lightning strike relationship for listeners
+    private static final Map<UUID, UUID> stormStaffLightnings = new HashMap<>();
 
     @ParametersAreNonnullByDefault
     public StormStaff(ItemGroup group, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
@@ -105,6 +106,10 @@ public class StormStaff extends LimitedUseItem {
         }
 
         damageItem(p, item);
+    }
+
+    public static @Nonnull Map<UUID, UUID> getStormStaffLightnings() {
+        return stormStaffLightnings;
     }
 
 }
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/staves/StormStaff.java
```diff
@@ -36,12 +41,16 @@
     public static final int MAX_USES = 8;
 
     private static final NamespacedKey usageKey = new NamespacedKey(Slimefun.instance(), "stormstaff_usage");
+    private final ItemSetting<Boolean> allowNonPVPUse = new ItemSetting<>(this, "allow-non-pvp-use", false);
+
+    public static final Map<UUID, UUID> stormStaffLightnings = new HashMap<>();
```
WalshyDev, line 46: No. Do not use public global variables NotLikeThis

Make a getter and also a make a comment above explaining what the key and value is since it isn't clear

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/entity/StormStaffLightningListener.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @EventHandler
+    public void onLightningDamage(EntityDamageByEntityEvent e) {
+        if (e.getDamager() instanceof LightningStrike) {
+            LightningStrike lightningStrike = (LightningStrike) e.getDamager();
+            UUID lightningStrikeUuid = lightningStrike.getUniqueId();
+            if (StormStaff.stormStaffLightnings.containsKey(lightningStrikeUuid)) {
+                e.setCancelled(true);
+                Player caster = Bukkit.getPlayer(StormStaff.stormStaffLightnings.get(lightningStrikeUuid));
```
WalshyDev, line 39: Don't do containsKey and get. `containsKey` calls get. Do the call once and check properly

WalshyDev (changes requested):

WalshyDev (changes requested):
I'm not sure why the ItemSetting exists, the way I see it there are 2 problems:
* The staff cannot at all if PVP is disabled
* Plugins cannot identify who did this damage

These could both be solved if the EntityDamageByEntityEvent just handled the current if.
* Check if the lightning is hitting a player with protection modules or if PVP is disabled
  * If the lightning isn't hitting a player, we don't need to cancel and re-fire and stuff. The staff can still work when not hitting players this way
* Re-fire with a modified event

Sfiguz7:
Welp before I change it, I'd like to hear what @TheBusyBiscuit thinks since he assigned himself too. If he agrees sure, otherwise I;d rather have the conversation and decision happen without moving stuff back and forth

TheBusyBiscuit (changes requested):
I don't think using a `Map<UUID, UUID>` here is really needed.
This could be made a lot leaner by just using lazy metadata values.

Aside from that I am not sure of the side effects that calling this additional event will cause.
Since plugins may rightfully not expect PvP events with `Damagecause.LIGHTNING`

Also I got no clue what @WalshyDev is proposing there

TheBusyBiscuit:
It's been over a year, closing this as stale.

Issue #3341, opened 2021-11-06
Cargo is not filtering Vanilla Enchanted Books (All treated as same with/without Lore filter)


NathZen:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

Cargo doesn't fillter Enchanted Books per enchant. All Enchant were merged into 1 chest that has any enchanted book filter in it

### 📑 Reproduction Steps

1. Place the chest and fill it with several different enchanted books
2. Connect the Cargo Node (Input) to the chest and enable the blacklist in it.
3. Place some chest and connect the Advanced Cargo Node (Output) to it
4. Insert 1 type of enchanted book per cargo on each chest into the whitelist filter.

### 💡 Expected Behavior

The cargo distinguish the Enchanted Book

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.16.x

### ⭐ Slimefun version

![image](https://user-images.githubusercontent.com/61259322/140599331-6fc49c1c-ac41-4bf6-abab-59febb635228.png)

kylemarham:
Can this bug be reviewed please? 13 months later no update.

Boomer-1:
there is no issue here. They work fine. I've had a setup for over a year going from level 3 to 15 and not once had an issue

Boomer-1:
if you can provide video evidence of it not working i will leave this open. All node configurations must be captured clearly, showing every setup and book in each one,, and verification  the nodes are connected to the cargo network,

kylemarham:
Please see the below link

https://streamable.com/zs5ssd

The expected behaviour is that Smite 4 is inserted into the Advanced Cargo Node (Output), but the cargo node feeds Smite 3 books into the book binder instead.

Boomer-1:
Their server is 6 months behind. Ill leave this open waiting on confirmation of the issue on dev 1035.

kylemarham:
Can confirm this is still an issue on version 1035

https://streamable.com/r05bnm

J3fftw1:
I think the best way to solve this issue is to add a same enchant filter instead of a lore check.
Enchants arent really lore so we have to check it differently aswell
[...]

PR #3417, opened 2022-01-11
DistinctiveItem/isSimilar + Backpack canStack

### LINKED ISSUE #3379: Cargo ignores lore of slimefun items, causing backpacks to stack.

Sefiraat:
## Description

Supercedes #3384 

SlimefunItemStacks that share an ID are currently automatically stacked when running through cargo. Items that vary by lore or another meta aspect are being 'merged' with items that happen to match on ID but are, otherwise, distinct. See issue #3379. This likely only effects backpacks within core Slimefun but also effects any similar SlimefunItems using similar distinctions - DankPacks and Tinkers parts (I'm so self centred!) being an example.

## Proposed changes
New `ItemAttribute` called `DistinctiveItem` to act as an identifier for items that match this criterion.
Distinctive item has a method that allows an implementation to say how two same-ID items can be distinguished.
SlimefunBackpack has been amended to implement this interface.
`SlimefunUtils#isItemSimilar` now checks if the stack's SlimefunItem is a DistinctiveItem and, if so, return the result of the `canStack` method.

Discussion with Walshy shows this is making an extra call to get the ID from the Meta that was already made in the previous `getItemData` that may mean this PR needs more work.

Resolves #3379

## Checklist


Sefiraat:
Comment from the previous PR I'm closing:

So, near as I can tell, the 5th branch (line 280) is never called by items moving through cargo though it will be reached when directly calling isItemSimilar. Given that meta is required to check distinction this will could add a lot of overhead to any other machine using this. Given the above, do we still want to check distinction outside of cargo - given that we do not currently and it, in itself, hasn't shown issues. Input welcome for how to add this in.

Also - changes in commit 92b45ad tested under the same conditions with expected and wanted results.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -321,6 +332,19 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
         }
     }
 
+    private static @Nullable String getMetaId(ItemMeta itemMeta) {
```
WalshyDev, line 335: Not sure why this doesn't return an Optional. Makes this code very weird. Not sure we even need this method at all tbh

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -321,6 +332,19 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+        Optional<String> optionalId = Slimefun.getItemDataService().getItemData(itemMeta);
+        return optionalId.orElse(null);
+    }
+
+    private static @Nonnull Optional<DistinctiveItem> getDistinctiveItem(@Nonnull String id) {
```
WalshyDev, line 340: Not sure this method is needed. Feels like it should just be moved inside isItemSimilar

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -282,10 +283,10 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
             ItemMeta itemMeta = item.getItemMeta();
 
             if (sfitem instanceof SlimefunItemStack) {
-                Optional<String> id = Slimefun.getItemDataService().getItemData(itemMeta);
+                String id = getMetaId(itemMeta);
 
-                if (id.isPresent()) {
-                    return id.get().equals(((SlimefunItemStack) sfitem).getItemId());
+                if (id != null) {
+                    return id.equals(((SlimefunItemStack) sfitem).getItemId());
```
WalshyDev, line 289: We probably want a distinctive check here too. Since this is also just an ID check currently

not sure when this code is called but hey, will probably solve issues in addons or other parts of the code

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
Sefiraat: isItemSimilar is already so big I really didn't want to add this directly inside of it - since this comment I am now using this an additional 2 times also to I think this may be best kept outside, just to keep this monster method readable :D

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
Sefiraat: Added

Sefiraat:
I have added the check to both the SlimefunItemStack/SlimefunItemStack branch and to the ItemStack/SlimefunItemStack branches which should now cover all possible routes

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -302,10 +324,20 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
                  */
                 ItemMeta possibleSfItemMeta = sfitem.getItemMeta();
                 Debug.log(TestCase.CARGO_INPUT_TESTING, "  sfitem is ItemStackWrapper - possible SF Item: {}", sfitem);
+                String id = Slimefun.getItemDataService().getItemData(itemMeta).orElse(null);
 
                 // Prioritize SlimefunItem id comparison over ItemMeta comparison
-                if (Slimefun.getItemDataService().hasEqualItemData(possibleSfItemMeta, itemMeta)) {
+                if (Slimefun.getItemDataService().getItemData(possibleSfItemMeta).orElse("").equals(id)) {
```
WalshyDev, line 330: Move the getItemData().orElse out of the if. Then also make the `orElse` be null, I don't like making a new string here.

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -278,16 +279,37 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+                /*
+                 * Some items can't rely on just IDs matching and will implement Distinctive Item
+                 * in which case we want to use the method provided to compare
+                 */
+                if (sfItemStack instanceof DistinctiveItem && itemStack instanceof DistinctiveItem) {
+                    return ((DistinctiveItem) sfItemStack).canStack(sfItemStack.getItemMeta(), itemStack.getItemMeta());
+                }
+                return true;
+            }
+            return false;
```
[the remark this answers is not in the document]
Sefiraat: No :P

StarWishsama:
This change will cause backpack unable to upgrade

Boomer-1:
any chance this is being pushed out? It may fix a couple other issues. #3341 and #3379

J3fftw1:
> any chance this is being pushed out? It may fix a couple other issues. #3341 and #3379

Patience :pansmug:
@Sefiraat is working on a different fix which is way better then this iirc

Boomer-1:
Awesome. it could fix a multitude of issues at once.

Sefiraat:
It needs a touch up to fix issues with backpacks upgrading but it's not going to be worked on until I know if the general direction is viable/approvable. This isn't the PR in which I was working on a better idea though I'm afraid, that was one focused on networks :)

Sefiraat:
It was mentioned a long while ago in discord that this is likely not the correct approach, but I'm not really sure I have a better one. I know this PR is live on the Chinese fork, not sure how they've worked around the backpack upgrading however

ybw0014:
> It was mentioned a long while ago in discord that this is likely not the correct approach, but I'm not really sure I have a better one. I know this PR is live on the Chinese fork, not sure how they've worked around the backpack upgrading however

https://github.com/StarWishsama/Slimefun4/commit/1802199eb677f24279ed359dd026a46f7176c80c

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -282,16 +341,43 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
         } else if (checkAmount && item.getAmount() < sfitem.getAmount()) {
             return false;
         } else if (sfitem instanceof SlimefunItemStack && item instanceof SlimefunItemStack) {
-            return ((SlimefunItemStack) item).getItemId().equals(((SlimefunItemStack) sfitem).getItemId());
+            SlimefunItemStack stackOne = (SlimefunItemStack) sfitem;
+            SlimefunItemStack stackTwo = (SlimefunItemStack) item;
```
J3fftw1, line 345: The variables aren't needed since we can just use the instanceof and add the variable there.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -282,16 +341,43 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+            if (stackOne.getItemId().equals(stackTwo.getItemId())) {
+                /*
+                 * PR #3417
+                 *
+                 * Some items can't rely on just IDs matching and will implement Distinctive Item
+                 * in which case we want to use the method provided to compare
+                 */
+                if (checkDistinction && stackOne instanceof DistinctiveItem && stackTwo instanceof DistinctiveItem) {
+                    return ((DistinctiveItem) stackOne).canStack(stackOne.getItemMeta(), stackTwo.getItemMeta());
```
J3fftw1, line 354: no need to cast here we can make the variable again with the Instanceof check

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -523,4 +630,27 @@ public static boolean canPlayerUseItem(@Nonnull Player p, @Nullable ItemStack it
...
+     * If the MC version is 1.16 or above
+     * this will call {@link Inventory#isEmpty()} (Which calls MC code resulting in a faster method).
+     *
+     * @param inventory
+     *            The {@link Inventory} to check.
+     *
+     * @return True if the inventory is empty and false otherwise
+     */
+    public static boolean isInventoryEmpty(@Nonnull Inventory inventory) {
+        if (Slimefun.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_16)) {
```
J3fftw1, line 645: remove this check since we only support 1.16 and up these days

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -523,4 +630,27 @@ public static boolean canPlayerUseItem(@Nonnull Player p, @Nullable ItemStack it
     public static @Nullable Item spawnItem(Location loc, ItemStack item, ItemSpawnReason reason) {
         return spawnItem(loc, item, reason, false);
     }
+
+    /**
+     * Helper method to check if an Inventory is empty (has no items in "storage").
+     * If the MC version is 1.16 or above
```
J3fftw1, line 636: remove this message as well

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -523,4 +630,27 @@ public static boolean canPlayerUseItem(@Nonnull Player p, @Nullable ItemStack it
...
+     *            The {@link Inventory} to check.
+     *
+     * @return True if the inventory is empty and false otherwise
+     */
+    public static boolean isInventoryEmpty(@Nonnull Inventory inventory) {
+        if (Slimefun.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_16)) {
+            return inventory.isEmpty();
+        } else {
+            for (ItemStack is : inventory.getStorageContents()) {
+                if (is != null && !is.getType().isAir()) {
```
J3fftw1, line 649: Can we give this variable a better name?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -327,6 +426,14 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
         }
     }
 
+    private static @Nonnull Optional<DistinctiveItem> getDistinctiveItem(@Nonnull String id) {
+        SlimefunItem slimefunItem = SlimefunItem.getById(id);
+        if (slimefunItem instanceof DistinctiveItem) {
+            return Optional.of((DistinctiveItem) slimefunItem);
```
J3fftw1, line 432: no need for casting here please add a variable

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
Sefiraat: Ha yeah this PR was from before we added patterns into SF lol - showing how old it is. Ill push something later today

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/DistinctiveItem.java
```diff
@@ -0,0 +1,31 @@
...
+     * with one another.
+     *
+     * @param itemMetaOne
+     *                    The {@link ItemMeta} of the first stack being compared.
+     * @param itemMetaTwo
+     *                    The {@link ItemMeta} of the second stack being compared.
+     *
+     * @return Whether the two {@link ItemMeta}s are distinct
+     */
+    boolean canStack(@Nonnull ItemMeta itemMetaOne, @Nonnull ItemMeta itemMetaTwo);
```
WalshyDev, line 30: I'm wondering if we want SlimefunItem or at least ItemStack here rather than Meta. Feels like that expands the realm of possibility a lot more. We can also change later but that of course means a deprecation cycle.

WalshyDev (commented):
LGTM, just need to fix Jeff's comments and want an opinion too.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/DistinctiveItem.java
Sefiraat: IIRC my thought process was to provide the itemmeta to stop people pulling out the itemmeta’s again should they need to adding more overhead to an already poorly performant process. I could provide the stacks alongside however and make it clear in the docs? I could also form up a compound object to wrap that all into or something?

WalshyDev: Yeah good thought not wanting to add overhead. We know how... fun getItemMeta is...

Yeah I think 2 params is best.
Not sure a wrapper is needed, I think that's probably just more annoying tbh.

Sefiraat: For the purposes here the items only really ‘merge’ and dupe when the slimefunitems are the same and just something within the meta makes them distinct be it lore *yuck* or PDC being the common factor. So hopefully this covers most use cases, it certainly does for all the known ones atm.

Sefiraat: In which case I’ll implement the feedback tomorrow with the aim to merge asap!

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
Sefiraat: Out of scope - I know this appears in my diff but has not been worked on as part of this PR, just moved around. I don't want more in here than is necessary.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
Sefiraat: Out of scope - I know this appears in my diff but has not been worked on as part of this PR, just moved around. I don't want more in here than is necessary.

Sefiraat:
Given how long it has been since this PR was first raised, if someone else other than myself wants to give the preview build a whirl before it's merged. Hopefully we can get this in finally over the weekend!

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/DistinctiveItem.java
```diff
@@ -0,0 +1,31 @@
...
+     * This method is called by {@link SlimefunUtils#isItemSimilar} when two {@link SlimefunItemStack}
+     * IDs match on a DistinctiveItem and should return if the two items can stack
+     * with one another.
+     *
+     * @param itemMetaOne
+     *                    The {@link ItemMeta} of the first stack being compared.
+     * @param itemMetaTwo
+     *                    The {@link ItemMeta} of the second stack being compared.
+     *
+     * @return Whether the two {@link ItemMeta}s are distinct
```
WalshyDev, line 28:
```suggestion
     * @return Whether the two {@link ItemMeta}s are stackable
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -281,17 +340,42 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
         } else if (checkAmount && item.getAmount() < sfitem.getAmount()) {
             return false;
-        } else if (sfitem instanceof SlimefunItemStack && item instanceof SlimefunItemStack) {
-            return ((SlimefunItemStack) item).getItemId().equals(((SlimefunItemStack) sfitem).getItemId());
+        } else if (sfitem instanceof SlimefunItemStack stackOne && item instanceof SlimefunItemStack stackTwo) {
+            if (stackOne.getItemId().equals(stackTwo.getItemId())) {
+                /*
+                 * PR #3417
+                 *
+                 * Some items can't rely on just IDs matching and will implement Distinctive Item
```
WalshyDev, line 348:
```suggestion
                 * Some items can't rely on just IDs matching and will implement {@link DistinctiveItem}
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -281,17 +340,42 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
-                if (id.isPresent()) {
-                    return id.get().equals(((SlimefunItemStack) sfitem).getItemId());
+                String id = Slimefun.getItemDataService().getItemData(itemMeta).orElse(null);
+
+                if (id != null) {
+                    if (checkDistinction) {
+                        /*
+                         * PR #3417
+                         *
+                         * Some items can't rely on just IDs matching and will implement Distinctive Item
```
WalshyDev, line 369:
```suggestion
                         * Some items can't rely on just IDs matching and will implement {@link DistinctiveItem}
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -304,12 +388,25 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+                String possibleItemId = Slimefun.getItemDataService().getItemData(possibleSfItemMeta).orElse(null);
                 // Prioritize SlimefunItem id comparison over ItemMeta comparison
-                if (Slimefun.getItemDataService().hasEqualItemData(possibleSfItemMeta, itemMeta)) {
+                if (id != null && id.equals(possibleItemId)) {
                     Debug.log(TestCase.CARGO_INPUT_TESTING, "  Item IDs matched!");
+
+                    /*
+                     * PR #3417
+                     *
+                     * Some items can't rely on just IDs matching and will implement Distinctive Item
```
WalshyDev, line 403:
```suggestion
                     * Some items can't rely on just IDs matching and will implement {@link DistinctiveItem}
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -523,4 +628,27 @@ public static boolean canPlayerUseItem(@Nonnull Player p, @Nullable ItemStack it
     public static @Nullable Item spawnItem(Location loc, ItemStack item, ItemSpawnReason reason) {
         return spawnItem(loc, item, reason, false);
     }
+
+    /**
+     * Helper method to check if an Inventory is empty (has no items in "storage").
```
WalshyDev, line 633:
```suggestion
     * Helper method to check if an {@likk Inventory} is empty (has no items in "storage").
```

WalshyDev (commented):
Just some Javadoc improvements left from me

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
Sefiraat: While this is only docs, I really don't want to fall out of scope, that method is in the diff but has not been changed as part of this PR, likely just got jumbled at some point.

Also, {@likk} lol

WalshyDev (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/backpacks/SlimefunBackpack.java
```diff
@@ -82,4 +85,14 @@ public ItemUseHandler getItemHandler() {
         };
     }
 
+    @Override
```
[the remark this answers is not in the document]
J3fftw1: nvm its still early

J3fftw1 (approved):
Not worth much but here it is LGTM

WalshyDev (approved):

JustAHuman-xD (approved):
LGTM

PR #3421, opened 2022-01-12
ExposedStorage

Sefiraat:
## Description

This purpose of this PR is to provide a way for specific SlimefunItems to provide a summary of their contents. This would be used when Transport Slots and/or #getContents simply doesn't cut the mustard! 
Examples of blocks that could implement this interface would be Barrels to expose their full content as opposed to just the one stack in the output slot, Multi-page inventories to return all held items not just the currently selected one and the likes of DankPacks and likely many more possible applications I've not thought of just yet.

## Proposed changes

- New Interface extending ItemAttribute.
- A method `getStoredItems` that will return `Map<ItemStack, Integer>` used to tell calling methods what the SlimefunItems currently contains.
- A method `withdrawItem` that takes a template ItemStack and a requested amount and returns how many items were successfully removed for the calling method to then handle. (I have swapped between ItemStack, ItemStack Array and int a few times and this one felt the most useful in the end - very open to input as to the return type for this method)
- A method `depositItemStacks` that takes an ItemStack array for the SlimefunItem to try to consume into the storage. Currently returns nothing as the calling method can just look at the array after calling to check what's left - open to having the return the modified array instead?)
- A method `depositItemStack` that just calls `depositItemStacks`. Return type may change based on feedback to the above.
- A method `contains` to return a boolean if the SlimefunItem contains a matching ItemStack within itself.

See Sefiraat/Networks@3e36b9f in which I changed my Network Shells (v e r y r o u g h l y) to use this interface and had the network read the items in a much neater way than I was doing previously. Working perfectly from testing undertaken.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExposedStorage.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExposedStorage.java
@@ -1,0 +1,92 @@
+package io.github.thebusybiscuit.slimefun4.core.attributes;
+
+import javax.annotation.Nonnull;
+
+import org.bukkit.Location;
+import org.bukkit.inventory.Inventory;
+import org.bukkit.inventory.ItemStack;
+
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
+
+import java.util.Map;
+
+/**
+ * This interface, when attached to a {@link SlimefunItem}, provides methods for the item to
+ * expose its contents, for when {@link Inventory#getContents} may not be a suitable reflection
+ * of the Storages content.
+ * Examples include deep storage barrels or inventories with multiple pages.
+ * 
+ * It also adds methods for requesting items and adding items directly into storage
+ * 
+ * @author Sefiraat
+ * 
+ */
+public interface ExposedStorage extends ItemAttribute {
+
+    /**
+     * This method should return the full contents of the block in question
+     * summarized into a map of {@link ItemStack} and total amount stored.
+     * The {@link ItemStack} is representative only and shouldn't be used
+     * directly.
+     *
+     * An {@link ItemStack} should only appear once in the Map regardless of
+     * how many may be in storage. E.g. if you have 2 stacks of stone, you would
+     * do Map.put(stoneStack, 128);
+     *
+     * @param location
+     *          The {@link Location} of the storage block that the content request
+     *
+     * @return A {@link Map} that lists all exposed items and their quantity
+     */
+    @Nonnull Map<ItemStack, Integer> getStoredItems(@Nonnull Location location);
+
+    /**
+     * This method should is used to withdraw an item from the storage.
+     * The volume of the matching {@link ItemStack} within storage should be reduced
+     * during/after withdrawal.
+     *
+     * @param location
+     *          The {@link Location} of the storage block that the withdrawal request is for
+     * @param similarStack
+     *          An {@link ItemStack} that will be used to match against stored items.
+     * @param amount
+     *          The amount of items requested for withdrawal, capped
+     *
+     * @return The number of items successfully removed from the storage to up {@param amount}
+     */
+    int withdrawItem(@Nonnull Location location, @Nonnull ItemStack similarStack, int amount);
+
+    /**
+     * This method is used to submit a single {@link ItemStack} into the
+     * {@link ExposedStorage}. By default, this just calls {@link #depositItemStacks}
+     *
+     * The {@link ItemStack} should have its amount reduced as appropriate during
+     * consumption, so it can be checked by the requester to return items to their
+     * starting point.
+     *
+     * @param location
+     *          The {@link Location} of the storage block that the deposit request is for
+     * @param itemToDeposit
+     *          The {@link ItemStack} to be added into storage
+     */
+    default void depositItemStack(@Nonnull Location location, @Nonnull ItemStack itemToDeposit) {
+        depositItemStacks(location, new ItemStack[]{itemToDeposit});
+    }
+
+    /**
+     * This method is used to submit an array of {@link ItemStack}s into the
+     * {@link ExposedStorage}.
+     *
+     * Each {@link ItemStack} should have its amount reduced as appropriate during
+     * consumption, so it can be checked by the requester to return items to their
+     * starting point.
+     *
+     * @param location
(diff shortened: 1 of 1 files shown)


WalshyDev:
First thoughts: This is an interesting PR and I understand the reasoning why. I have a few questions however:

1. Do you believe this is better than just improving the current inv/content system?
  1. Sub question: What do we do about the current methods for cargo and such

2. Getting the contents returns a Map<Item, Quantity> however deposit takes an array. I think this should probably be consistent but not sure in what way. Not a fan of either honestly (performance + tedious). Any suggestions there?

3. Is there any implementation inside Slimefun which should be done before this is landed?

Sefiraat:
> First thoughts: This is an interesting PR and I understand the reasoning why. I have a few questions however:
> 
> 1. Do you believe this is better than just improving the current inv/content system?

Likely not, but any improvements to the existing system would likely unify how any Slimefun inventory is looked at by anything which would, in turn, mean a big change to cargo. Given cargo is due an overhaul in the future I'd assume this is better smaller in scope for now but potentially used for direction/scoping later on?

>    
>    * Sub question: What do we do about the current methods for cargo and such

Nothing, yet :)

> 2. Getting the contents returns a Map<Item, Quantity> however deposit takes an array. I think this should probably be consistent but not sure in what way. Not a fan of either honestly (performance + tedious). Any suggestions there?

Yeah true, I fell into step here because of the way Networks works but deposit can easily be changed to `Map<ItemStack, Inventory>`, contents is kinda locked in as a Map however, unless there is another data structure that can be used that may be faster?

> 3. Is there any implementation inside Slimefun which should be done before this is landed?

Not really, most esoterically designed storages live in addons, those I know of:
- Personal Storages (may be tricky)
- FluffyMachines barrels
- Fluffy's foundry (deposit only?)
- Infinity Expansion barrels
- SimpleStorage multi page inventories
- DankTech packs
- Tinkers smeltery (perhaps not)
- Network memory cells

The only one I have found within core is the Trash can which could implement this to return empty content and deposit items could trigger voiding directly or put items into the inventory to be ticked away, but this is barely an improvement over the exposed cargo slots currently so i'd assume no

> 4. What do you think the performance implications are and should implementations be considered thread-safe or not?

This leaves any comparison/retrieval methods up to the addon implementing the interface which could lead to some issues but, in most cases, these addons must have something similar enough to be able to interact with the block internally. It should be easy enough to tell within a profiler if - for example - a Networks tick is taking longer due to a method being called from another addon to fix it at source. Annnnnd.... yeah this would need to be treadsafe - i'll add something to the documentation but is there an annotation or something else that may make it more prevalent?

WalshyDev:
> Likely not, but any improvements to the existing system would likely unify how any Slimefun inventory is looked at by anything which would, in turn, mean a big change to cargo. Given cargo is due an overhaul in the future I'd assume this is better smaller in scope for now but potentially used for direction/scoping later on?

Yeah makes sense

> Yeah true, I fell into step here because of the way Networks works but deposit can easily be changed to Map<ItemStack, Inventory>, contents is kinda locked in as a Map however, unless there is another data structure that can be used that may be faster?

Map is awkward because you need to do like:
```java
final Map<ItemStack, Integer> map = new HashMap<>();
final ItemStack item = // grab item
map.put(item, // grab qantity)

return map
```

then grabbing and using too likely is gonna be a bit awkward.

Now what is better? No idea frankly. I think we should definitely think more about this though

> This leaves any comparison/retrieval methods up to the addon implementing the interface which could lead to some issues but, in most cases, these addons must have something similar enough to be able to interact with the block internally. It should be easy enough to tell within a profiler if - for example - a Networks tick is taking longer due to a method being called from another addon to fix it at source. Annnnnd.... yeah this would need to be treadsafe - i'll add something to the documentation but is there an annotation or something else that may make it more prevalent?

`@ThreadSafe`. If this needs to be thread-safe then that should definitely be documented and made clear. I was suspecting as such but I was sure given the current code someone would do it in a non-thread-safe way.

I think when the storage rewrite happens everything will be made a lot nicer in this API too.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExposedStorage.java
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * This method should return a total count of how many items are within the ExposedStorage
+     *
+     * @param location
+     *          The {@link Location} of the storage block to count items from
+     *
+     * @return The number of items held in the block
+     */
+    long getAmount(@Nonnull Location location);
```
WalshyDev, line 63: These should be consistent. Mixing int & longs is a pain

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExposedStorage.java
```diff
@@ -0,0 +1,114 @@
...
+     * @param location
+     *          The {@link Location} of the storage block that the withdrawal request is for
+     * @param similarStack
+     *          An {@link ItemStack} that will be used to match against stored items.
+     * @param amount
+     *          The amount of items requested for withdrawal, capped
+     *
+     * @return The number of items successfully removed from the storage to up {@param amount}
+     */
+    int withdrawItem(@Nonnull Location location, @Nonnull ItemStack similarStack, int amount);
```
WalshyDev, line 79: same here

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExposedStorage.java
```diff
@@ -0,0 +1,114 @@
...
+     * Each {@link ItemStack} should have its amount reduced as appropriate during
+     * consumption, so it can be checked by the requester to return items to their
+     * starting point.
+     *
+     * @param location
+     *          The {@link Location} of the storage block that the deposit request is for
+     * @param itemsToDeposit
+     *          The {@link ItemStack} array to be added into storage
+     */
+    void depositItemStacks(@Nonnull Location location, @Nonnull ItemStack[] itemsToDeposit);
```
WalshyDev, line 111: As I noted originally it's a bit weird to use a Map in one place and an array in another. I can see someone wanting to put 1000 gold blocks in, would this mean I'd need an array full of 64x itemstacks?
I think this should be a Map, considering we have stuff like Map.of these days (yay for finally supporting java 9) I think the usage is ok

WalshyDev (changes requested):

PR #3425, opened 2022-01-15
Ability to use isLenient flag of BiomeMapParser

poma123:
## Description

The reason behind my changes is that I need to use the isLenient flag in order to load all valid biomes from my _editable_ biomemap file without the process being stopped by an exception. After that, missing (unconfigured) biomes would be used with a default value.

## Proposed changes

- Made BiomeMapParser class public
- Added a new fromJson method to BiomeMap class with the isLenient parameter

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/biomes/BiomeMap.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/biomes/BiomeMap.java
@@ -149,6 +149,15 @@ public class BiomeMap<T> implements Keyed {
     }
 
     @ParametersAreNonnullByDefault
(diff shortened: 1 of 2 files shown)


TheBusyBiscuit (dismissed):
LGTM

It's proven to be reliable thus far, so exposing that API should be fine in my book.

poma123:
Yay! So wouldn't be a dedicated #fromJson function in BiomeMap better with an isLenient parameter (or something similar)? Or is this good enough and simply I am the one who overcomplicates things as always? 😅

TheBusyBiscuit:
I mean a `fromJson` method is surely nice and easier to write.
But I think there isn't any reason not to expose the parser itself, especially if we are going to add more flags besides "lenient" to it.

poma123:
Okay then! I will look into it asap!

TheBusyBiscuit (approved):
Still looks good to me.

WalshyDev (approved):
LGTM

PR #3431, opened 2022-01-16
Update dough

WalshyDev:
## Description
Updated dough to 1.1.2

## Proposed changes
Updated pom.xml and code in IntegrationsManager

Also updated the changelog, this did fix a few issues with protection things so would be good to be added. However, not sure if it should be in changes or fixes since it's both (though indirectly for fixes)

N/A

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,11 +43,11 @@
 * Added Spore Blossom as a fuel type for the Bio-Generator
 * Added a new item setting for Freezers to allow them to use a 9:1 "vanilla" ratio instead of 1:1 (1:1 by default, like before)
 * (API) Added `PlayerProfile#hasUnlockedEverything()` to check if a player has unlocked all researches
+* Added support for the plugin "HuskTowns"
 
 #### Changes
 * (API) `BiomeMapParser` is now `public`
 * (API) `BiomeMap.fromJson` now allows you to specify if you want the BiomeMap to be parsed leniently
-* Updated dough, fixing some issues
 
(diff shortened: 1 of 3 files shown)


svr333 (dismissed):

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 #### Changes
 * (API) `BiomeMapParser` is now `public`
 * (API) `BiomeMap.fromJson` now allows you to specify if you want the BiomeMap to be parsed leniently
+* Updated dough, fixing some issues
```
TheBusyBiscuit, line 50: We do not list dependency updates here.
```suggestion
```

CHANGELOG.md
```diff
@@ -47,6 +47,7 @@
 #### Changes
```
TheBusyBiscuit, line 47: The following needs to be added to the changelog

#### Additions
* Added support for the plugin "HuskTowns"

#### Fixes
* Fixed "LogBlock" integration
* Fixed "Lands" integration

TheBusyBiscuit (changes requested):
You beat me to it. I was just about to update this. But the changelog needs to be fixed before merging this.

CHANGELOG.md
WalshyDev: Alright, didn't think you'd want them added here since it was in dough but will do

TheBusyBiscuit (approved):

svr333 (approved):

PR #3451, opened 2022-01-30
Allow Fluid Pump to fill water bottles

FN-FAL113:
## Description
Approved suggestion on discord that proposes to be able to fill empty bottles with water

## Proposed changes
- Adds support for empty glass bottles

Approved suggestion ```#1967```

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
@@ -212,8 +212,8 @@ public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements Invent
         return null;
     }
 
-    private @Nonnull ItemStack getFilledBottle(@Nonnull Block fluid){
-        switch(fluid.getType()){
+    private @Nonnull ItemStack getFilledBottle(@Nonnull Block fluid) {
+        switch (fluid.getType()) {
             case WATER:
             case BUBBLE_COLUMN:
                 ItemStack waterBottle = new ItemStack(Material.POTION);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
@@ -1,6 +1,7 @@
 package io.github.thebusybiscuit.slimefun4.implementation.items.electric.machines;
 
 import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -55,6 +56,8 @@ public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements Invent
     private static final int ENERGY_CONSUMPTION = 32;
     private static final int RANGE = 42;
 
+    private final ItemStack WATER_BOTTLE = new ItemStack(Material.POTION);
+
     private final int[] border = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 13, 31, 36, 37, 38, 39, 40, 41, 42, 43, 44, 22 };
     private final int[] inputBorder = { 9, 10, 11, 12, 18, 21, 27, 28, 29, 30 };
     private final int[] outputBorder = { 14, 15, 16, 17, 23, 26, 32, 33, 34, 35 };
@@ -62,22 +65,15 @@ public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements Invent
     private final ItemStack emptyBucket = ItemStackWrapper.wrap(new ItemStack(Material.BUCKET));
     private final ItemStack emptyBottle = ItemStackWrapper.wrap(new ItemStack(Material.GLASS_BOTTLE));
 
-    private static final ItemStack waterBottle = new ItemStack(Material.POTION);
-
-    static {
-        PotionMeta meta = (PotionMeta) waterBottle.getItemMeta();
-        if(meta != null) {
-            meta.setBasePotionData(new PotionData(PotionType.WATER));
-            waterBottle.setItemMeta(meta);
-        }
-    }
-
     @ParametersAreNonnullByDefault
     public FluidPump(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
         super(itemGroup, item, recipeType, recipe);
 
         addItemHandler(onBreak());
         createPreset(this, this::constructMenu);
+        PotionMeta meta = (PotionMeta) this.WATER_BOTTLE.getItemMeta();
+        meta.setBasePotionData(new PotionData(PotionType.WATER));
+        this.WATER_BOTTLE.setItemMeta(meta);
     }
 
     @Nonnull
@@ -170,9 +166,9 @@ public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements Invent
                     return;
                 }
                 if (SlimefunUtils.isItemSimilar(menu.getItemInSlot(slot), emptyBottle, true, false)) {
-                    ItemStack glassBottle = getFilledBottle(fluid);
+                    ItemStack bottle = getFilledBottle(fluid);
 
-                    if (!menu.fits(glassBottle, getOutputSlots())) {
+                    if (!menu.fits(bottle, getOutputSlots())) {
                         return;
                     }
 
@@ -181,8 +177,10 @@ public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements Invent
                     if (nextFluid != null) {
                         removeCharge(b.getLocation(), ENERGY_CONSUMPTION);
                         menu.consumeItem(slot);
-                        menu.pushItem(glassBottle, getOutputSlots());
-                        nextFluid.setType(Material.AIR);
+                        menu.pushItem(bottle, getOutputSlots());
+                        if(ThreadLocalRandom.current().nextInt(100) < 30) {
+                            nextFluid.setType(Material.AIR);
+                        }
                     }
 
                     return;
@@ -222,7 +220,7 @@ public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements Invent
         switch(fluid.getType()){
             case WATER:
             case BUBBLE_COLUMN:
-                return waterBottle;
+                return WATER_BOTTLE.clone();
             default:
                 return new ItemStack(Material.GLASS_BOTTLE);
         }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
@@ -56,14 +56,13 @@ public class FluidPump extends SimpleSlimefunItem<BlockTicker> implements Invent
     private static final int ENERGY_CONSUMPTION = 32;
     private static final int RANGE = 42;
 
-    private final ItemStack WATER_BOTTLE = new ItemStack(Material.POTION);
-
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -153,6 +167,24 @@ protected void tick(@Nonnull Block b) {
                         nextFluid.setType(Material.AIR);
                     }
 
+                    return;
+                }
+                if (SlimefunUtils.isItemSimilar(menu.getItemInSlot(slot), emptyBottle, true, false)) {
+                    ItemStack glassBottle = getFilledBottle(fluid);
```
Sefiraat, line 173: bottle, not glass bottle - glass would indicate empty, waterBottle found indicate filled so imo bottle indicates either possible state?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -153,6 +167,24 @@ protected void tick(@Nonnull Block b) {
...
+                        return;
+                    }
+
+                    Block nextFluid = findNextFluid(fluid);
+
+                    if (nextFluid != null) {
+                        removeCharge(b.getLocation(), ENERGY_CONSUMPTION);
+                        menu.consumeItem(slot);
+                        menu.pushItem(glassBottle, getOutputSlots());
+                        nextFluid.setType(Material.AIR);
```
Sefiraat, line 185: This is an interesting one, and would need input from others, but water bottles don't consume the water block they are used on when used in a vanilla context, should it here? Alternatively a Cauldron will lose 1/3 a bucket of water when filling up a bottle so perhaps it would be nice to add a 1 in 3 chance to consume the water, just to add a little interesting quirk into the machine? Debatable :D

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -185,6 +217,17 @@ private Block findNextFluid(@Nonnull Block fluid) {
         return null;
     }
 
+    @Nonnull
+    private ItemStack getFilledBottle(@Nonnull Block fluid){
+        switch(fluid.getType()){
+            case WATER:
+            case BUBBLE_COLUMN:
+                return waterBottle;
```
Sefiraat, line 225: Fairly sure we are  going to want a `.clone()` here?

Sefiraat (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -57,6 +60,17 @@
     private final int[] outputBorder = { 14, 15, 16, 17, 23, 26, 32, 33, 34, 35 };
 
     private final ItemStack emptyBucket = ItemStackWrapper.wrap(new ItemStack(Material.BUCKET));
+    private final ItemStack emptyBottle = ItemStackWrapper.wrap(new ItemStack(Material.GLASS_BOTTLE));
+
+    private static final ItemStack waterBottle = new ItemStack(Material.POTION);
```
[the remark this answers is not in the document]
Sefiraat: Also statics before non-statics, this should be between RANGE and border. Although, this doesn't really need to be static either and could easily just be private and then setup in the constructor to save having the static block...?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
FN-FAL113: agree on that, no need to debate for it I guess since its a vanilla mechanics xD later on players might just complain from what I did, will just add a low chance

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
FN-FAL113: will do, forgot to clone the object

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -52,18 +56,24 @@
     private static final int ENERGY_CONSUMPTION = 32;
     private static final int RANGE = 42;
 
+    private final ItemStack WATER_BOTTLE = new ItemStack(Material.POTION);
```
Sefiraat, line 59: I probably confused things here by adding an alternatiuve - If a field is static then it needs to be in SCREAMING_SNAKE_CASE but if not, then camelCase (waterBottle) is fine. Likewise if a field is a static constant then it goes above all other declared fields, if it is final then it goes with other private finals and, in this case, below emptyBottle - My question here was whether it needed to be static or not. 

I would now move this below emptyBottle, change to waterBottle and keep it non-static as you have it currently

FN-FAL113: im the one who got confused sefi, my bad xD. will do the changes in a moment. I really need to review conventions and standards

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -57,13 +61,18 @@
...
 
     @ParametersAreNonnullByDefault
     public FluidPump(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
         super(itemGroup, item, recipeType, recipe);
 
         addItemHandler(onBreak());
         createPreset(this, this::constructMenu);
+        PotionMeta meta = (PotionMeta) this.waterBottle.getItemMeta();
+        meta.setBasePotionData(new PotionData(PotionType.WATER));
+        this.waterBottle.setItemMeta(meta);
```
TheBusyBiscuit, line 75: There isn't really any point in storing this water bottle as a field, when you clone it inside `getFilledBottle` anyway.
You are effectively creating a new instance anyway, so there is absolutely no need to store one in memory all the time.
Just assemble the `ItemStack` within `getFilledBottle` directly.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -153,6 +162,26 @@ protected void tick(@Nonnull Block b) {
                         nextFluid.setType(Material.AIR);
                     }
 
+                    return;
+                }
+                if (SlimefunUtils.isItemSimilar(menu.getItemInSlot(slot), emptyBottle, true, false)) {
```
TheBusyBiscuit, line 167: Semantically, an item cannot be an empty bucket AND and empty bottle.
You should probably also store `menu.getItemInSlot(slot)` as a variable perhaps
```suggestion
                } else if (SlimefunUtils.isItemSimilar(menu.getItemInSlot(slot), emptyBottle, true, false)) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -153,6 +162,26 @@ protected void tick(@Nonnull Block b) {
...
+
+                    Block nextFluid = findNextFluid(fluid);
+
+                    if (nextFluid != null) {
+                        removeCharge(b.getLocation(), ENERGY_CONSUMPTION);
+                        menu.consumeItem(slot);
+                        menu.pushItem(bottle, getOutputSlots());
+                        if(ThreadLocalRandom.current().nextInt(100) < 30) {
+                            nextFluid.setType(Material.AIR);
+                        }
```
TheBusyBiscuit, line 182: Formatting, also... why make it randomized
```suggestion
                        menu.pushItem(bottle, getOutputSlots());

                        if (ThreadLocalRandom.current().nextInt(100) < 30) {
                            nextFluid.setType(Material.AIR);
                        }
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -185,6 +214,17 @@ private Block findNextFluid(@Nonnull Block fluid) {
         return null;
     }
 
+    @Nonnull
+    private ItemStack getFilledBottle(@Nonnull Block fluid){
+        switch(fluid.getType()){
```
TheBusyBiscuit, line 219: Formatting
```suggestion
    private @Nonnull ItemStack getFilledBottle(@Nonnull Block fluid) {
        switch (fluid.getType()) {
```

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
Sefiraat: https://github.com/Slimefun/Slimefun4/pull/3451#discussion_r795150129

This was my reasoning though as stated something to discuss rather than just add.

FN-FAL113: If I were to respect vanilla mechanics then I would just add an random chance since water bottles doesn't consume water off from an water source but then sefi stated it does on cauldron, kinda middle in there but I like the fact sefi stated this as in favor of staying in vanilla

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -185,8 +212,21 @@ private Block findNextFluid(@Nonnull Block fluid) {
         return null;
     }
 
-    @Nonnull
-    private ItemStack getFilledBucket(@Nonnull Block fluid) {
+    private @Nonnull ItemStack getFilledBottle(@Nonnull Block fluid){
+        switch(fluid.getType()){
```
TheBusyBiscuit, line 216:
```suggestion
    private @Nonnull ItemStack getFilledBottle(@Nonnull Block fluid) {
        switch (fluid.getType()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/FluidPump.java
```diff
@@ -153,6 +160,26 @@ protected void tick(@Nonnull Block b) {
...
+                    }
+
+                    Block nextFluid = findNextFluid(fluid);
+
+                    if (nextFluid != null) {
+                        removeCharge(b.getLocation(), ENERGY_CONSUMPTION);
+                        menu.consumeItem(slot);
+                        menu.pushItem(bottle, getOutputSlots());
+
+                        if(ThreadLocalRandom.current().nextInt(100) < 30) {
```
TheBusyBiscuit, line 178:
```suggestion
                        if (ThreadLocalRandom.current().nextInt(100) < 30) {
```

TheBusyBiscuit (changes requested):
Still formatting

TheBusyBiscuit (approved):
Looks good to me.
Just unsure about the randomized block consumption, perhaps an `ItemSetting` could be added here in a future PR

FN-FAL113:
Somehow I messed up the branching by deleting the forked repo for this pull request due to the industrial miner pr I made last time, if sefi request changes I'll just make another pull request :NotLikeThis:

Sefiraat (approved):
In which case, LGTM.

PR #3452, opened 2022-01-31
Industrial miner unintentional fuel consuming fix

### LINKED ISSUE #3336: industrial miner small bug
### ❗ Checklist

### 📍 Description

when you build an industrial miner, put there lava and turn it on, it can tell you that pistons should be facing up, so you fix the pistons which you accidentaly placed wrong and want to turn it on again, but you realize that the lava was already consumed!

### 📑 Reproduction Steps

see description

### 💡 Expected Behavior

lava should not be consumed if the machine is not built right

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

DEV 976

### 🧭 Other plugins

_No response_

FN-FAL113:
## Description
Fixes the small bug from advanced industrial miner during warm-up phase, it should respect the ```#stop()``` boolean value if there are any errors during ```#setPistonState``` validation

## Proposed changes
- Prevent industrial miner from consuming fluid fuels if there are any error messages during ```#setPistonState```

- Resolves #3336 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/MiningTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/MiningTask.java
@@ -312,10 +312,17 @@ class MiningTask implements Runnable {
                 ItemStack item = inv.getContents()[i];
 
                 if (fuelType.test(item)) {
-                    ItemUtils.consumeItem(item, false);
+                    /*
+                     * Fixes #3336
+                     * Respects the state of the miner if there are
+                     * no any errors during #setPistonState
+                     */
+                    if (running) {
+                        ItemUtils.consumeItem(item, false);
 
-                    if (miner instanceof AdvancedIndustrialMiner) {
-                        inv.addItem(new ItemStack(Material.BUCKET));
+                        if (miner instanceof AdvancedIndustrialMiner) {
+                            inv.addItem(new ItemStack(Material.BUCKET));
+                        }
                     }
 
                     return fuelType.getTicks();


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/MiningTask.java
```diff
@@ -312,10 +312,17 @@ private int grabFuelFrom(@Nonnull Inventory inv) {
...
-                    ItemUtils.consumeItem(item, false);
-
-                    if (miner instanceof AdvancedIndustrialMiner) {
-                        inv.addItem(new ItemStack(Material.BUCKET));
+                    /*
+                     * Fixes #3336
+                     * Respects the state of the miner if there are
+                     * no any errors during #setPistonState
+                     */
+                    if (running) {
```
TheBusyBiscuit, line 320: You can connect these two statements using `&&`, no need to introduce another nesting level here.

TheBusyBiscuit (changes requested):
Also please learn to use branches.
You are committing this from the `master` branch, which is already problematic itself.
But now you are sharing this branch with #3451. This will cause problems if we request other changes there.

Try to create a new branch for every PR, otherwise you WILL run into issues.
Preferabbly using a semantic system like we do, e.g. `fix/xyz` or `feature/xyz` as branch names

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/MiningTask.java
FN-FAL113: oopsie my bad, that's a thing I haven't learned from gh where using branches properly, Good to know though. Should I make a new PR with a new branch or keep this one

PR #3459, opened 2022-02-05
Hercules Pickaxe drops fix

### LINKED ISSUE #3455: Hercules Pickaxe duplication on 1.17 ores
### ❗ Checklist

### 📍 Description

When breaking new 1.17 ores that can be converted to Slimefun dust using Hercules Pickaxe it will drop their dust and raw ore version at the same time.

List of working ores:
  - Copper Ore
  - Deepslate Copper Ore
  - Deepslate Iron Ore
  - Deepslate Gold Ore

I believe this is cause by
https://github.com/Slimefun/Slimefun4/blob/008090cc6f50b15a503ef1f151cea20931016cd0/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java#L33-L47

When adding those item it doesn't break until it's over and will add their natural drop in 
https://github.com/Slimefun/Slimefun4/blob/008090cc6f50b15a503ef1f151cea20931016cd0/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java#L58-L60

Side note: Not sure why those are `* 2` tho. Destroying other ores with Hercules Pickaxe can give you twice their drop, also works with fortune enchantment. (Should I create other bug report on this? or is this intended behaviour?)

### 📑 Reproduction Steps

1. Get Hercules Pickaxe
2. Mine blocks listed above
3. They will drop dust and raw ore at the same time.

### 💡 Expected Behavior

The raw ore should not be dropped.

### 📷 Screenshots / Videos

https://youtu.be/tweHnXMhICY

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.18.x

### ⭐ Slimefun version

```
Paper git-Paper-181 (MC: 1.18.1)
Slimefun DEV - 999 (git e02eedd9)
Metrics-Module #28
Java 17

Installed Addons: (2)
  CMILib v1.1.1.0
  CMI v9.1.0.6
```

### 🧭 Other plugins

_No response_

FN-FAL113:
## Description
Fixes the drops for the hercules pickaxe, a simple boolean that prevents it from doing the default case for the second switch/case. 

## Proposed changes
- Prevents the switch/case from doing the default fallback during block mining which can do natural drops

- Resolves #3455 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
@@ -27,38 +27,49 @@ public class HerculesPickaxe extends SimpleSlimefunItem<ToolUseHandler> {
     @Override
     public @Nonnull ToolUseHandler getItemHandler() {
         return (e, tool, fortune, drops) -> {
+            boolean hadDoubledDrops = false;
             Material mat = e.getBlock().getType();
 
             if (SlimefunTag.ORES.isTagged(mat)) {
                 if (Slimefun.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_17)) {
-                    switch (mat) {
+                    switch(mat) {
                         case DEEPSLATE_IRON_ORE:
                             drops.add(new CustomItemStack(SlimefunItems.IRON_DUST, 2));
+                            hadDoubledDrops = true;
                             break;
                         case DEEPSLATE_GOLD_ORE:
                             drops.add(new CustomItemStack(SlimefunItems.GOLD_DUST, 2));
+                            hadDoubledDrops = true;
                             break;
                         case COPPER_ORE:
                         case DEEPSLATE_COPPER_ORE:
                             drops.add(new CustomItemStack(SlimefunItems.COPPER_DUST, 2));
+                            hadDoubledDrops = true;
                             break;
                         default:
                             break;
                     }
                 }
 
-                switch (mat) {
-                    case IRON_ORE:
-                        drops.add(new CustomItemStack(SlimefunItems.IRON_DUST, 2));
-                        break;
-                    case GOLD_ORE:
-                        drops.add(new CustomItemStack(SlimefunItems.GOLD_DUST, 2));
-                        break;
-                    default:
-                        for (ItemStack drop : e.getBlock().getDrops(tool)) {
-                            drops.add(new CustomItemStack(drop, drop.getAmount() * 2));
-                        }
-                        break;
+                /*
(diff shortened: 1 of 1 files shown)


variananora:
I think we should also remove the `* 2` since it will duplicate any ores with silk touch... https://github.com/StarWishsama/Slimefun4/issues/474 video for reference.

variananora (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
```diff
@@ -50,13 +50,13 @@ public HerculesPickaxe(ItemGroup itemGroup, SlimefunItemStack item, RecipeType r
...
                     case IRON_ORE:
                         drops.add(new CustomItemStack(SlimefunItems.IRON_DUST, 2));
-                        break;
+                        return;
                     case GOLD_ORE:
                         drops.add(new CustomItemStack(SlimefunItems.GOLD_DUST, 2));
-                        break;
+                        return;
                     default:
                         for (ItemStack drop : e.getBlock().getDrops(tool)) {
```
EpicPlayerA10, line 58: I think this loop is useless. You are adding to `drops` item which can drop by default without adding it.

EpicPlayerA10 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/HerculesPickaxe.java
FN-FAL113: > I think this `for` loop is useless. You are adding to `drops` items which can drop by default without adding them to list.

Tested it ingame, drops are cancelled somewhere, if the default fallback is removed and a diff ore is mined which has no dust associated with them it will not drop anything using the hercules pickaxe. So its better to have the for loop to do its default drop

EpicPlayerA10: It should drop vanilla items by default if  `drops` list is empty
https://github.com/Slimefun/Slimefun4/blob/1773ae0a4a33b28c56ee031f24e7297145db7e65/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java#L169-L176

FN-FAL113: I Might as well test it out without the default loop ingame, Didn't touched the code since it was there before this pr

PR #3461, opened 2022-02-06
Add SlimefunBlockDropEvent 

CarmJos:
## Description

In some cases, developers want to prevent slimefun's block drops or make drops directly into some containers.

But in previous codes, this is not possible because it uses the `#dropItemNaturally(...)` method.

## Proposed changes

I added a event called `SlimefunBlockDropEvent` to provide a way for developers edit the drops.

I have tested the event by my own plugin.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunBlockDropEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunBlockDropEvent.java
@@ -17,14 +17,14 @@ import java.util.List;
  * @author CarmJos
  * @see io.github.thebusybiscuit.slimefun4.implementation.listeners.BlockListener
  */
-public class SlimeBlockDropEvent extends BlockEvent {
+public class SlimefunBlockDropEvent extends BlockEvent {
 
     private static final HandlerList handlers = new HandlerList();
 
     private final Player player;
     private List<ItemStack> drops;
 
-    public SlimeBlockDropEvent(Player player, Block theBlock, List<ItemStack> drops) {
+    public SlimefunBlockDropEvent(Player player, Block theBlock, List<ItemStack> drops) {
         super(theBlock);
         this.player = player;
         this.drops = drops;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -46,211 +46,211 @@ import java.util.concurrent.ThreadLocalRandom;
  */
 public class BlockListener implements Listener {
 
-	public BlockListener(@Nonnull Slimefun plugin) {
-		plugin.getServer().getPluginManager().registerEvents(this, plugin);
-	}
+    public BlockListener(@Nonnull Slimefun plugin) {
+        plugin.getServer().getPluginManager().registerEvents(this, plugin);
+    }
 
-	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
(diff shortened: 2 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -1,15 +1,16 @@
...
+import io.github.thebusybiscuit.slimefun4.api.events.SlimefunBlockDropEvent;
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
+import io.github.thebusybiscuit.slimefun4.core.attributes.NotPlaceable;
+import io.github.thebusybiscuit.slimefun4.core.handlers.BlockBreakHandler;
+import io.github.thebusybiscuit.slimefun4.core.handlers.BlockPlaceHandler;
+import io.github.thebusybiscuit.slimefun4.core.handlers.ToolUseHandler;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+import io.github.thebusybiscuit.slimefun4.utils.tags.SlimefunTag;
+import me.mrCookieSlime.Slimefun.api.BlockStorage;
+import org.bukkit.Bukkit;
```
Sefiraat, line 13: Looks like a reformat or auto import has changed the order/formatting here - best revert back to how it was.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -192,10 +194,8 @@ private void dropItems(BlockBreakEvent e, List<ItemStack> drops) {
      * Sensitive {@link Block Blocks} are pressure plates or saplings, which should be broken
      * when the block beneath is broken as well.
      *
-     * @param p
-     *            The {@link Player} who broke this {@link Block}
-     * @param b
-     *            The {@link Block} that was broken
+     * @param p The {@link Player} who broke this {@link Block}
+     * @param b The {@link Block} that was broken
```
Sefiraat, line 198: Looks like a reformat or auto import has changed the order/formatting here - best revert back to how it was.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
CarmJos: fixed

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
CarmJos: fixed

CarmJos:
Please review this request. @Sefiraat

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunBlockDropEvent.java
```diff
@@ -0,0 +1,71 @@
...
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.bukkit.event.block.BlockEvent;
+import org.bukkit.inventory.ItemStack;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+
+/**
+ * This {@link Event} is called whenever a {@link Player} break a Slimefun block which should drop some {@link ItemStack}.
```
svr333, line 15:
```suggestion
 * This {@link Event} is called whenever a {@link Player} breaks a Slimefun block which should drop some {@link ItemStack}.
```

svr333 (commented):
Looks good to me, though PR's are paused currently anyway so

svr333:
😕

PR #3473, opened 2022-02-26
Improve Round-Robin Accuracy

qwertyuioplkjhgfd:
## Description

Currently, round-robin can lead to an uneven distribution with filled chests or output nodes with filters. This is because while the index gets incremented, the actual output node is just the "next available" node. 
<details>
<summary>Example</summary>
<br>

```
X - invalid output node (filled chest or filter)
O - available output node

output node array:
XXXXXXXXOO
```

During the first 8 ticks, the round robin index would be on an X, and the next available output node would be chosen (the first O). Then the 9th tick outputs onto the first O directly, and the 10th tick outputs onto the second O. Then the round robin index rolls back over to 0 and puts 9 ticks worth of items into the first O again. In this case, the first O is heavily biased and gets much more items than the second O. This PR aims to fix that, and similar uneven distributions. 
</details>

## Proposed changes

When an item is transferred in round-robin mode, set the round-robin index to the index of the node being transferred to, so that invalid output nodes are skipped and available output nodes are distributed across the same number of sf ticks per cycle. 

[video demonstrating](https://www.youtube.com/watch?v=jTKyd820Ygg)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
@@ -171,12 +171,14 @@ class CargoNetworkTask implements Runnable {
         int index = 0;
         Collection<Location> destinations;
         if (roundrobin) {
+            // The current round-robin index of the (unsorted) outputNodes list,
+            // or the index at which to start searching for valid output nodes
+            index = network.roundRobin.getOrDefault(inputNode, 0);
             // Use an ArrayDeque to perform round-robin sorting
             // Since the impl for roundRobinSort just does Deque.addLast(Deque#removeFirst)
             // An ArrayDequeue is preferable as opposed to a LinkedList:
             // - The number of elements does not change.
             // - ArrayDequeue has better iterative performance
-            index = network.roundRobin.getOrDefault(inputNode, 0);
             Deque<Location> tempDestinations = new ArrayDeque<>(outputNodes);
             roundRobinSort(index, tempDestinations);
             destinations = tempDestinations;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
```diff
@@ -190,8 +192,12 @@ private ItemStack distributeItem(ItemStack stack, Location inputNode, List<Locat
             if (target.isPresent()) {
                 ItemStackWrapper wrapper = ItemStackWrapper.wrap(item);
                 item = CargoUtils.insert(network, inventories, output.getBlock(), target.get(), smartFill, item, wrapper);
+                index++;
 
                 if (item == null) {
+                    if (roundrobin) {
+                        network.roundRobin.put(inputNode, index % outputNodes.size());
```
TheBusyBiscuit, line 199: Why the modulo here?
Can you please elaborate and possibly add a comment to this part.

TheBusyBiscuit (dismissed):
Really took me quite some time to understand this code.
But I think I understand it now and it looks good to me, just the modulo part I don't fully get yet.

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNetworkTask.java
qwertyuioplkjhgfd: The modulo is because the round robin index wraps around the output nodes, so once the index gets to the end of the output nodes it goes back to the start

JasperChaseTOQ:
any update on this?

TheBusyBiscuit (approved):
My brain hurts. But I think I understand this now and I approve it.

JasperChaseTOQ:
🥳

svr333:
I like the elegant way of handling the randomness, lgtm

svr333 (approved):

PR #3479, opened 2022-03-05
Fix pedestal item disappearing

### LINKED ISSUE #3476: Item disappears when taking item back on pedestal while inventory is full.
### ❗ Checklist

### 📍 Description

When inventory is full and you take something on the pedestal, the item will disappear.

### 📑 Reproduction Steps

1. place pedestal
2. put an item on it
3. make inventory full
4. take out the item on the pedestal

### 💡 Expected Behavior

The item will not able to be taken/ dropped on the ground

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.18.x

### ⭐ Slimefun version

image
image
image
image

### 🧭 Other plugins

_No response_

FN-FAL113:
## Description
A small bug fix for the ancient pedestal when retrieving placed items.

## Proposed changes
- Drop item instead if the player's inventory is full and no remaining stack space when retrieving the item back from the pedestal

- Resolves #3476 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/AncientAltarListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/AncientAltarListener.java
@@ -156,8 +156,19 @@ public class AncientAltarListener implements Listener {
             Slimefun.runSync(() -> removedItems.remove(uuid), 30L);
 
             entity.remove();
-            p.getInventory().addItem(pedestalItem.getOriginalItemStack(entity));
             p.playSound(pedestal.getLocation(), Sound.ENTITY_ITEM_PICKUP, 1F, 1F);
+
+            /*
+             * Fixes #3476
+             * Drop the item instead if the player's
+             * inventory is full
+             */
+            if (p.getInventory().firstEmpty() != -1) {
+                p.getInventory().addItem(pedestalItem.getOriginalItemStack(entity));
+
+            } else {
+                p.getWorld().dropItem(pedestal.getLocation().add(0, 1, 0), pedestalItem.getOriginalItemStack(entity));
+            }
         }
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/AncientAltarListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/AncientAltarListener.java
@@ -161,12 +161,12 @@ public class AncientAltarListener implements Listener {
             /*
              * Fixes #3476
              * Drop the item instead if the player's inventory is full and
-             * no stack space left else add the item in the inventory
+             * no stack space left else add remaining items from the returned map value
              */
-            for (Map.Entry<Integer, ItemStack> stackEntry : p.getInventory().addItem(pedestalItem.getOriginalItemStack(entity)).entrySet()) {
-                if (stackEntry != null) {
-                    p.getWorld().dropItem(pedestal.getLocation().add(0, 1, 0), stackEntry.getValue().clone());
-                }
+            Map<Integer,ItemStack> remainingItemMap = p.getInventory().addItem(pedestalItem.getOriginalItemStack(entity));
+            
+            for (ItemStack item : remainingItemMap.values()) {
+                p.getWorld().dropItem(pedestal.getLocation().add(0, 1, 0), item.clone());
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/AncientAltarListener.java
```diff
@@ -156,8 +156,18 @@ private void usePedestal(@Nonnull Block pedestal, @Nonnull Player p) {
...
             entity.remove();
-            p.getInventory().addItem(pedestalItem.getOriginalItemStack(entity));
             p.playSound(pedestal.getLocation(), Sound.ENTITY_ITEM_PICKUP, 1F, 1F);
+
+            /*
+             * Fixes #3476
+             * Drop the item instead if the player's
+             * inventory is full
+             */
+            if (p.getInventory().firstEmpty() != -1) {
```
Sefiraat, line 166: This will not check if the item actually fits in the target inventory. Best to add the items but then check the returned Map<Integer, ItemStack> and drop what if left behind. This way item retrieval will work with a full inventory but with space in a stack for the item being removed

FN-FAL113: It does add stacked items to another stacked ones but does this imply It must drop the remaining items that cannot be stacked onto other items anymore?

FN-FAL113: The pedestal only accepts one item at a time though and do not support placing stacked items, I could just get the returned map assign it to a variable and for each value is gonna be dropped

FN-FAL113: Very well, it makes it much simple and since the pedestal only supports one amount of  placed item at a time and it can only return a single amount item

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/AncientAltarListener.java
```diff
@@ -156,8 +156,18 @@ private void usePedestal(@Nonnull Block pedestal, @Nonnull Player p) {
...
             entity.remove();
-            p.getInventory().addItem(pedestalItem.getOriginalItemStack(entity));
             p.playSound(pedestal.getLocation(), Sound.ENTITY_ITEM_PICKUP, 1F, 1F);
+
+            /*
+             * Fixes #3476
+             * Drop the item instead if the player's inventory is full and
+             * no stack space left else add the item in the inventory
+             */
+            for (Map.Entry<Integer, ItemStack> stackEntry : p.getInventory().addItem(pedestalItem.getOriginalItemStack(entity)).entrySet()) {
```
Sefiraat, line 166: I think declaring the map separately from the for loop here would be far easier to read and far neater.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/AncientAltarListener.java
```diff
@@ -156,8 +156,18 @@ private void usePedestal(@Nonnull Block pedestal, @Nonnull Player p) {
...
+                if (stackEntry != null) {
```
Sefiraat, line 167: I’m fairly sure when iterating the entry set you can’t get a nullable entry? Could be wrong as I’m not at my desk but I feel like this isn’t needed?

Sefiraat: Also, may just be best to iterate the values directly, don’t think the int it required in this context at all.

FN-FAL113: > I’m fairly sure when iterating the entry set you can’t get a nullable entry? Could be wrong as I’m not at my desk but I feel like this isn’t needed?
Yes thats correct, I added it for debugging and safety ig when I was testing it ingame, I can fully remove it in the next commit

Sefiraat (approved):
Looks fine to me :)

FN-FAL113:
Oh this should be missing some imports will fix later

Sefiraat (approved):
Looks good to me :)

PR #3488, opened 2022-03-11
Fixed light blocks causing cargo nodes to not be placed.

Silent-Program:
## Description

I noticed that when placing cargo nodes inside light blocks it causes an error. 
## Proposed changes

Allow cargo nodes to be placed inside of light blocks.
## Checklist

- [x]  I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ ]  I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CargoNodeListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CargoNodeListener.java
@@ -2,7 +2,11 @@ package io.github.thebusybiscuit.slimefun4.implementation.listeners;
 
 import javax.annotation.Nonnull;
 
+import io.github.thebusybiscuit.slimefun4.api.MinecraftVersion;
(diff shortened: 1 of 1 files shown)


Silent-Program:
I am somewhat new to all of this, so please let me know any changes you need me to make.

qwertyuioplkjhgfd:
I think you're putting the check in the wrong place. This will let you place cargo nodes on the top or bottom of a block if the replaced block is a light block. Did you thoroughly check your changes?

Sefiraat:
As above, what you have done here is not correct. Before I close this, you mentioned an error which you have not provided? Can you please provide details of this error so we can see what it is you're actually trying to achieve?

Silent-Program:
> 

My bad, I realize what I did wrong there and will fix it.  By error, I meant a message saying that you can not place it there.  I will change the code to fix the issue and then submit a new PR

Sefiraat:
Figured that may be the case ;) Good luck :D

PR #3489, opened 2022-03-11
Added warning when players try to place cargo nodes in spaces occupied by "light" blocks

Silent-Program:
## Description

When placing cargo node on chest while light block is present it tells the player you may not do so.
## Proposed changes

Allow players to place cargo nodes inside of light blocks (you can place vanilla blocks inside of light blocks)
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CargoNodeListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CargoNodeListener.java
@@ -2,7 +2,11 @@ package io.github.thebusybiscuit.slimefun4.implementation.listeners;
 
 import javax.annotation.Nonnull;
 
+import io.github.thebusybiscuit.slimefun4.api.MinecraftVersion;
+import org.bukkit.Bukkit;
+import org.bukkit.Material;
 import org.bukkit.block.Block;
+import org.bukkit.block.data.type.Light;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.block.BlockPlaceEvent;
@@ -15,7 +19,7 @@ import io.github.thebusybiscuit.slimefun4.implementation.items.cargo.CargoNode;
 /**
  * This {@link Listener} is solely responsible for preventing Cargo Nodes from being placed
  * on the top or bottom of a block.
- * 
+ *
  * @author TheBusyBiscuit
  *
  */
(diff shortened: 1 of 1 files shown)


Silent-Program:
This is the second PR as I put the check in the wrong place and bonked up the on top of the chest check.  But I double-checked that that is no longer broken and indeed it is not.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CargoNodeListener.java
```diff
@@ -1,41 +1,45 @@
...
 import org.bukkit.block.Block;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.block.BlockPlaceEvent;
 import org.bukkit.inventory.ItemStack;
 
-import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
-import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
-import io.github.thebusybiscuit.slimefun4.implementation.items.cargo.CargoNode;
+import javax.annotation.Nonnull;
```
variananora, line 14: import order

variananora (commented):
I advise you to see our code standards here: https://github.com/Slimefun/Slimefun4/blob/master/CONTRIBUTING.md

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CargoNodeListener.java
```diff
@@ -1,41 +1,45 @@
...
+		if ((b.getY() != e.getBlockAgainst().getY() || !e.getBlockReplacedState().getType().isAir()) && isCargoNode(e.getItemInHand())) {
+			
+			Slimefun.getLocalization().sendMessage(e.getPlayer(), "machines.CARGO_NODES.must-be-placed", true);
+			e.setCancelled(true);
+		}
+	}
+	
+	private boolean isCargoNode(@Nonnull ItemStack item) {
+		return SlimefunItem.getByItem(item) instanceof CargoNode;
+	}
```
variananora, line 44: use spaces, not tabs

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/CargoNodeListener.java
```diff
@@ -1,41 +1,45 @@
...
+	public CargoNodeListener(@Nonnull Slimefun plugin) {
+		plugin.getServer().getPluginManager().registerEvents(this, plugin);
+	}
+	
+	@EventHandler(ignoreCancelled = true)
+	public void onCargoNodePlace(BlockPlaceEvent e) {
+		Block b = e.getBlock();
+		
+		if (Slimefun.getMinecraftVersion().isAtLeast(MinecraftVersion.MINECRAFT_1_17)) {
+			if (e.getBlockReplacedState().getType() == Material.LIGHT) return;
```
variananora, line 33: use brackets

variananora (changes requested):

Sefiraat:
Wont this still allow placing above or below a block? I may be screwing up the logic and I cant test personally but it looks like a light block above/below a block will then allow a cargo node to be placed in a location where it shouldn't be.

TheBusyBiscuit:
Closing this as stale

PR #3522, opened 2022-04-06
fix #3504

### LINKED ISSUE #3504: Cargo doesnt exist
### ❗ Checklist

### 📍 Description

There is no cargo in the latest version of slimefun, the only error I have in the console is in the paste, I do not have more errors.: https://paste.walshy.dev/TT5xl96pZ6FV

### 📑 Reproduction Steps

----------

### 💡 Expected Behavior

Not working after update

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

https://paste.walshy.dev/TT5xl96pZ6FV

https://paste.walshy.dev/TT5xl96pZ6FV

### 💻 Server Software

Airplane

### 🎮 Minecraft Version

1.16.x

### ⭐ Slimefun version

Slimefun DEV - 1008 (git 020094e0) 

### 🧭 Other plugins

_No response_

J3fftw1:
## Description

This pr will change the way Slimefun handles colours materials
This will remove the ColoredMaterial class and will change completely in to tags
Right now i just have changed Cargo to support this.
When this proposal actually gets accepted i will change all other classes to also support tags

With the current implementation cargo actually already works this is fully tested

## Proposed changes

Remove ColoredMaterials class
Add different tags for colored materials 
Add a way to get a ordered list instead of a set
Change rainbowblocks, cargo and all others that currently use ColoredMaterial class to tags

fix #3504

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/cargo/AbstractCargoNode.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/cargo/AbstractCargoNode.java
@@ -4,8 +4,12 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.utils.tags.SlimefunTag;
 import org.apache.commons.lang.Validate;
 import org.bukkit.Location;
+import org.bukkit.Material;
+import org.bukkit.NamespacedKey;
+import org.bukkit.Tag;
 import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
 import org.bukkit.event.block.BlockPlaceEvent;
@@ -29,6 +33,9 @@ import me.mrCookieSlime.Slimefun.api.BlockStorage;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenuPreset;
 import me.mrCookieSlime.Slimefun.api.item_transport.ItemTransportFlow;
+
+import java.util.ArrayList;
+import java.util.Set;
 
 /**
  * This abstract class is the super class of all cargo nodes.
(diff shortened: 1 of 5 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/cargo/AbstractCargoNode.java
```diff
@@ -92,6 +99,8 @@ protected void addChannelSelector(Block b, BlockMenu menu, int slotPrev, int slo
         boolean isChestTerminalInstalled = Slimefun.getIntegrations().isChestTerminalInstalled();
         int channel = getSelectedChannel(b);
 
+
+
```
WalshyDev, line 103:
```suggestion
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ColoredMaterial.java
```diff
@@ -199,29 +199,29 @@
...
+//            Material.GRAY_CANDLE,
+//            Material.LIGHT_GRAY_CANDLE,
+//            Material.CYAN_CANDLE,
+//            Material.PURPLE_CANDLE,
+//            Material.BLUE_CANDLE,
+//            Material.BROWN_CANDLE,
+//            Material.GREEN_CANDLE,
+//            Material.RED_CANDLE,
+//            Material.BLACK_CANDLE
+//    });
```
WalshyDev, line 224: why is this commented out? what's the alternative?

WalshyDev (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ColoredMaterial.java
J3fftw1: there isnt yet

This was for testing

Candles will be turned into tags aswell
Will do after i get feedback that something like this actually gets accepted

WalshyDev: Don't see why we wouldn't. It is a breaking API change though.

Please cleanup your git history.

PR #3526, opened 2022-04-11
Add event to allow changing outputs of the AutoDisenchanter

Geolykt:
## Description

Right now there is no way to transfer enchantments via the AutoDisenchanter if the enchantments are added by a plugin that does not extends bukkit's Enchantments class.

## Proposed changes

Create AutoDisenchanterComputeOutputEvent, which fires whenever an AutoDisenchanter attempts to disenchant an item. Call the event whenever a AutoDisenchanter crafts, and cancel/allow the disenchantment process if needed.
Furthermore AutoDisenchantEvent was deprecated as AutoDisenchanterComputeOutputEvent is basically doing the same thing

N/A

## Checklist

- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them. [It shouldn't anyways, ABI compat is preserved]

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AutoDisenchanterComputeOutputEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AutoDisenchanterComputeOutputEvent.java
@@ -1,0 +1,178 @@
+package io.github.thebusybiscuit.slimefun4.api.events;
+
+import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNonnullByDefault;
+
+import org.apache.commons.lang.Validate;
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.bukkit.inventory.ItemStack;
+
+import io.github.thebusybiscuit.slimefun4.implementation.items.electric.machines.enchanting.AutoDisenchanter;
+
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
+
+/**
+ * An {@link Event} that is fired just after {@link AutoDisenchantEvent} iff it was not cancelled.
+ * This event will start out as being cancelled unless it contains a bukkit enchantment.
+ * The enchantment transfer logic is already performed by slimefun before this event is fired, though this
+ * logic does not affect input items.
+ * The main usecase of this event is to control the output items or to control the time the disenchanting process takes.
+ *
+ * @author Geolykt
+ */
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AutoDisenchanterComputeOutputEvent.java
```diff
@@ -0,0 +1,178 @@
...
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+import org.bukkit.inventory.ItemStack;
+
+import io.github.thebusybiscuit.slimefun4.implementation.items.electric.machines.enchanting.AutoDisenchanter;
+
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
+
+/**
+ * An {@link Event} that is fired just after {@link AutoDisenchantEvent} iff it was not cancelled.
```
J3fftw1, line 16:
```suggestion
 * An {@link Event} that is fired just after {@link AutoDisenchantEvent} if it was not cancelled.
```

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/AutoDisenchanterComputeOutputEvent.java
Geolykt: This was intended as iff means "if and on if"

J3fftw1: ah okay didnt know that

svr333: I have never seen somebody use iff, Id just go for if here

Geolykt: If alone is too ambigious, so I'll go with if and only if

Geolykt:
If you ever intend on changing anything within the https://github.com/Slimefun/Slimefun4/tree/master/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/enchanting package, please notify me so I can update the ASM Transformers I use to implement my needed changes regardless of this PR being pulled or not

PR #3536, opened 2022-04-16
research return same instance

TheSilentPro:
## Description

Research should return the same instance for builder pattern for slimefun items.

## Proposed changes

returned same instance in `Research#addItems(SlimefunItem)`

/

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/researches/Research.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/researches/Research.java
@@ -158,12 +158,14 @@ public class Research implements Keyed {
      * @param items
      *            Instances of {@link SlimefunItem} to bind to this {@link Research}
      */
-    public void addItems(SlimefunItem... items) {
+    public Research addItems(SlimefunItem... items) {
         for (SlimefunItem item : items) {
             if (item != null) {
                 item.setResearch(this);
             }
         }
+
+        return this;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


svr333 (approved):
I dont see an issue with this, good for chaining ig

svr333 (changes requested):
need to update the java docs above though

Geolykt:
Just as a quick note that it does mess with ABI so it will break any plugin that uses this method

TheSilentPro:
> Just as a quick note that it does mess with ABI so it will break any plugin that uses this method

No? It just returns an instance of the Research. The method is the same and nothing else is changed.
Plugins that already use this shouldn't need to change anything.

Geolykt:
> > Just as a quick note that it does mess with ABI so it will break any plugin that uses this method
> 
> No? It just returns an instance of the Research. The method is the same and nothing else is changed. Plugins that already use this shouldn't need to change anything.

The issue is that methods are not only called by name, but also by descriptor. The descriptor contains the arguments (duh) but also the return type. It was `([Lio/github/thebusybiscuit/slimefun4/api/items/SlimefunItem;)V` but now it is `([Lio/github/thebusybiscuit/slimefun4/api/items/SlimefunItem;)Lio/github/thebusybiscuit/slimefun4/api/researches/Research;`. As such one will need to recompile the plugin - a subtle change that can cause issues

(that being said it could be possible that return types are ignored by the JVM so I could be talking bullshit here but I highly doubt that the implementation does that)

TheBusyBiscuit (changes requested):
Yeah I cannot approve this as it would break most addons with immediate effect.
Loosely looking over the comments, others have pointed this out too.

But to put it simply: Lots of compilers identify methods not just by name but by signature. The method signature includes the name but also things such as return types or parameters which makes overloading of methods possible. And when you create code that references a method, the compiler will remember the method signature that is supposed to be used.

So making changes to the signature will essentially lead to all addons that use this method throwing an error as the JVM will look for a method signature that does not exist.

I can understand why you would want to use a builder pattern here but I don't think this change justifies a breaking update here.

PR #3564, opened 2022-05-20
Add ensure compatibility in pull request template

qwertyuioplkjhgfd:
## Description

Sometimes PRs can be created and merged which (unintentionally) break compatibility with different Minecraft versions, such as the addition of candles to an enum. This aims to serve as a reminder to consider if the changes can affect other Minecraft versions.

## Proposed changes

Add an entry into the pull request checklist about supporting multiple Minecraft versions.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,6 +17,7 @@
 <!-- Don't worry, these are not requirements. They only serve as guidance. -->
 - [ ] I have fully tested the proposed changes and promise that they will not break everything into chaos.
 - [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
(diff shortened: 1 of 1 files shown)


J3fftw1:
Might be worth actually saying latest Minecraft version, so we dont have to also change this every 6 months for a new version release since this will be overlooked while making a pr

Sefiraat (changes requested):
As jeff said - `1.14* -> Latest Supported` would suit?

TheBusyBiscuit:
> As jeff said - `1.14* -> Latest Supported` would suit?

@Sefiraat We won't be supporting 1.14 / 1.15 for all eternity either. So, this is bound to be edited every so often anyway. To me, it would be fine to state the versions explicitly.

WalshyDev (approved):

Sefiraat (approved):
Fair enough, in which case LTGM :)

TheBusyBiscuit (approved):

Sefiraat (approved):

PR #3571, opened 2022-05-28
rework on range configuration

TerslenK:
## Description
since range was hardcoded i tried to make it more configurable

## Proposed changes
changed constructor of connectors and change range thing

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
@@ -12,6 +12,7 @@ import java.util.logging.Level;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import io.github.thebusybiscuit.slimefun4.implementation.items.cargo.CargoConnectorNode;
 import org.bukkit.Location;
 import org.bukkit.block.Block;
 import org.bukkit.inventory.ItemStack;
@@ -41,7 +42,6 @@ import me.mrCookieSlime.Slimefun.api.BlockStorage;
  */
 public class CargoNet extends AbstractItemNetwork implements HologramOwner {
 
-    private static final int RANGE = 5;
     private static final int TICK_DELAY = Slimefun.getCfg().getInt("networks.cargo-ticker-delay");
 
     private final Set<Location> inputNodes = new HashSet<>();
@@ -83,7 +83,7 @@ public class CargoNet extends AbstractItemNetwork implements HologramOwner {
 
     @Override
     public int getRange() {
-        return RANGE;
+        return CargoConnectorNode.RANGE;
     }
 
     @Override
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/EnergyConnector.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/EnergyConnector.java
@@ -28,10 +28,13 @@ import io.github.thebusybiscuit.slimefun4.implementation.items.SimpleSlimefunIte
  *
  */
 public class EnergyConnector extends SimpleSlimefunItem<BlockUseHandler> implements EnergyNetComponent {
+    public static int RANGE;
 
     @ParametersAreNonnullByDefault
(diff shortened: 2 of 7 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
```diff
@@ -12,6 +12,7 @@
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import io.github.thebusybiscuit.slimefun4.implementation.items.cargo.CargoConnectorNode;
```
Sefiraat, line 15: Your import order needs changing to match conventions (not just this instance but all)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
```diff
@@ -220,6 +220,8 @@
  *
  */
 public final class SlimefunItemSetup {
+    private static final int energyrange = Slimefun.getCfg().getInt("networks.energy-connector-range");
```
Sefiraat, line 223: Naming conventions here, energyRange, cargoRange

TerslenK: okay

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/EnergyConnector.java
```diff
@@ -28,10 +28,13 @@
  *
  */
 public class EnergyConnector extends SimpleSlimefunItem<BlockUseHandler> implements EnergyNetComponent {
+    public static int RANGE;
 
     @ParametersAreNonnullByDefault
-    public EnergyConnector(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, ItemStack recipeOutput) {
+    public EnergyConnector(ItemGroup itemGroup, int range, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, ItemStack recipeOutput) {
```
J3fftw1, line 34: right now this is a breaking change
what i would do is, making another constructor that would make the int range optional

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
TerslenK: k mb

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/EnergyConnector.java
TerslenK: can you give me a example? i never heard something like 2nd constructor

svr333:
```java
public EnergyConnector(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, ItemStack recipeOutput, int range = 100) {
	EnergyConnector(itemGroup, item, recipeType, recipe, recipeOutput);
        // And also set range to 'range' variable
}```

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/CargoNet.java
TerslenK: done

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
TerslenK: done

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/EnergyConnector.java
martinbrom: Using a constructor to assign a public static field is really sketchy. How can you be sure someone doesn't use it before you assign it? Also, what if there are two `EnergyConnector` tiers with different ranges? Then this code totally falls apart...

Can you make it work by the `CargoNet` and `EnergyNet` using the cfg too?

martinbrom (changes requested):

PR #3583, opened 2022-06-10
Update HeatedPressureChamber.java

JasperChaseTOQ:
## Description

i would like this to be implemented as i dont see any reason for it not to be implemented and it will help greatly as a visual
## Proposed changes

added extends RecipeDisplayItem so it will now display what recipes it can add.
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/HeatedPressureChamber.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/HeatedPressureChamber.java
@@ -7,8 +7,6 @@ import java.util.List;
 
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import io.github.thebusybiscuit.slimefun4.core.attributes.RecipeDisplayItem;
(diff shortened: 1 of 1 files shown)


Sefiraat:
Uhm, RecipeDisplayItem is an interface and cannot be extended except by another interface. Also FYI a class can only ever extend a single class. I think you want `implements` here

JasperChaseTOQ:
> Uhm, RecipeDisplayItem is an interface and cannot be extended except by another interface. Also FYI a class can only ever extend a single class. I think you want `implements` here

your right. smh..... that was dumb of me my bad

Sefiraat:
Also worth noting that when you implement the interface, you will need to implement the various methods the interface has. You should also ensure the plug-in compiles before you push :)

JasperChaseTOQ:
> Also worth noting that when you implement the interface, you will need to implement the various methods the interface has. You should also ensure the plug-in compiles before you push :)
implemented the methods and testing right now

JasperChaseTOQ:
> Also worth noting that when you implement the interface, you will need to implement the various methods the interface has. You should also ensure the plug-in compiles before you push :)

um it wont compile for me but it also wont compile the commits before mine so its not from my commits at least :/

PR #3589, opened 2022-06-11
Replace Validate checks

### LINKED ISSUE #3586: Replace `Validate` checks from `commons-lang` with `Preconditions` checks from Guava
See #3585 for context.

In MC 1.19 Spigot has removed `commons-lang` from its dependencies.
We really shouldn't be shading this library just for this one feature and instead switch to what Spigot uses.

So any `Validate.notNull(x, msg)` will become `Preconditions.checkNotNull(x, msg)` and any `Validate.isTrue(x, msg)` shall become `Preconditions.checkArgument(x, msg)`.
It would be really helpful if somebody could make a pull request for this and remove `commons-lang` from our `pom.xml`.

ybw0014:
## Description

As #3586 said, replace all Validate checks with Precondition.

## Proposed changes

- Replace all `Validate.notNull(Object, String)` with `Preconditions.checkArgument(Object, String)`. (Changing to `Preconditions.checkNotNull` will throw `NullPointerException`)

Resolves #3586

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
@@ -796,8 +796,10 @@ public class SlimefunItem implements Placeable {
      *            Any {@link ItemHandler} that should be added to this {@link SlimefunItem}
      */
     public final void addItemHandler(ItemHandler... handlers) {
-        Validate.notEmpty(handlers, "You cannot add zero handlers...");
-        Validate.noNullElements(handlers, "You cannot add any 'null' ItemHandler!");
+        Preconditions.checkArgument(handlers != null && handlers.length > 0, "You cannot add zero handlers...");
+        for (ItemHandler handler : handlers) {
+            Preconditions.checkNotNull(handler, "You cannot add any 'null' ItemHandler!");
+        }
 
         // Make sure they are added before the item was registered.
         if (state != ItemState.UNREGISTERED) {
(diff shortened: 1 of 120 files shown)


ybw0014:
Not sure about how you would like to replace `Validate.noNullElements`, so discuss it here or https://discord.com/channels/565557184348422174/565570276038017044/985310097955758081.

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
```diff
@@ -795,8 +796,10 @@ public void load() {
      *            Any {@link ItemHandler} that should be added to this {@link SlimefunItem}
      */
     public final void addItemHandler(ItemHandler... handlers) {
-        Validate.notEmpty(handlers, "You cannot add zero handlers...");
-        Validate.noNullElements(handlers, "You cannot add any 'null' ItemHandler!");
+        Preconditions.checkArgument(handlers != null && handlers.length > 0, "You cannot add zero handlers...");
+        for (ItemHandler handler : handlers) {
+            Preconditions.checkArgument(handler != null, "You cannot add any 'null' ItemHandler!");
```
WalshyDev, line 801: This should definitely be a util method.

I don't think it warrants needing to be in dough so I'd just add it under our utils package.

WalshyDev (changes requested):

TheBusyBiscuit (changes requested):
I really find `Preconditions.checkNotNull` better semantic wise but it's a shame that it throws an NPE...

I was really hoping we could avoid writing our own util for it, but it really seems like `Preconditions` doesn't cover our use cases too well...

Perhaps a `Validate` class should be added to `dough-common` then... or Slimefun's util package.
What do the others think?

J3fftw1:
> I really find `Preconditions.checkNotNull` better semantic wise but it's a shame that it throws an NPE...
> 
> I was really hoping we could avoid writing our own util for it, but it really seems like `Preconditions` doesn't cover our use cases too well...
> 
> Perhaps a `Validate` class should be added to `dough-common` then... or Slimefun's util package. What do the others think?

I would say push this to Dough so plugins that use Dough only and not Slimefun can make use of it aswell.

ybw0014:
I'll open a new pr when https://github.com/baked-libs/dough/pull/184 is merged.

PR #3591, opened 2022-06-13
Add Enchanted Golden Apple recipe (suggestion #2147 from punished_Garett)

Fhoz:
## Description

Simple addition of an approved suggestion's item, it uses 24k gold blocks instead of the vanilla gold blocks.

## Proposed changes

Adding Enchanted Golden Apple, crafted in the Enhanced Crafting Table (and normal workbench since it's a VanillaItem)

Not applicable

## Checklist


JustAHuman-xD (approved):
LGTM

svr333 (approved):
Code was good the only thing this brought up is "is this too OP issue"

but I guess people can always disable it

J3fftw1:
I dont think this should just be merged in either with a notice or have it disabled by default.
But its not worth delaying this PR even more.
We could always PR instantly after this

J3fftw1:
We need to decide what to do with this PR. its a simple PR which has been open for ages.
Do we want this in or not.
The suggestion github has 31 upvotes and hasnt been approved.

WalshyDev:
I'm ok with this, lots of people want it and it's not that impacting. Server owners can disable if they wish.

PR #3592, opened 2022-06-16
Fix for Book Binder

JasperChaseTOQ:
changed book binder energy consumption from 16 to 8 as energy consumption is per tick and not second. the lore shows 16/s therefore the j/t should be 8/t instead of 16/t as each tick is half a second.

## Description

changed book binder energy consumption from 16 to 8 as energy consumption is per tick and not second. the lore shows 16/s therefore the j/t should be 8/t instead of 16/t as each tick is half a second.

## Proposed changes

changed book binder energy consumption from 16 to 8
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/SlimefunItemSetup.java
@@ -1819,7 +1819,7 @@ public final class SlimefunItemSetup {
         new BookBinder(itemGroups.electricity, SlimefunItems.BOOK_BINDER, RecipeType.ENHANCED_CRAFTING_TABLE,
         new ItemStack[] {null, new ItemStack(Material.ENCHANTING_TABLE), null, new ItemStack(Material.BOOKSHELF), SlimefunItems.HARDENED_METAL_INGOT, new ItemStack(Material.BOOKSHELF), SlimefunItems.SYNTHETIC_SAPPHIRE, SlimefunItems.SMALL_CAPACITOR, SlimefunItems.SYNTHETIC_SAPPHIRE})
         .setCapacity(256)
-        .setEnergyConsumption(16)
+        .setEnergyConsumption(8)
         .setProcessingSpeed(1)
         .register(plugin);
 


JasperChaseTOQ:
shall i change the pull request to put just change the lore instead then?

JasperChaseTOQ:
yeah lets wait then

J3fftw1:
> Given that it's been functioning this way for how many years and NO ONE noticed until now, I'd just change the lore.

Book binder isnt even a year old

JasperChaseTOQ:
oh its only that old? also i feel sefiraat has a point that 8/t is too little for what does honestly even the while were at it (since we either have to change the lore or the consumption since one or the other is incorrect) why dont we change both and increase it to make it harder and change it from a medium machine to either advanced or end game cause thats more suitable for what it does in my opinion.    ....however thats besides the main point of the PR. its to fix the bug, just a thought.

Boomer-1:
i thought it had been out for years, but it's Jan of 2021. either way, the point is correct, is it needs fixing one way or another

WalshyDev:
I would agree it should use quite a bit of energy. Changing the lore is something which will still break items on those with compat enabled.

bStats isn't loading for me so I can't view the impact but I'd hope it's very minimal at this point anyway

So the options are:
1. Change energy to 8/t
2. Change lore to 16/t (or another value) and break servers with compat enabled

I'll let others weigh in their opinions on that.

J3fftw1:
> I would agree it should use quite a bit of energy. Changing the lore is something which will still break items on those with compat enabled.
> 
> bStats isn't loading for me so I can't view the impact but I'd hope it's very minimal at this point anyway

109 servers have compat enabled

Sefiraat:
> I would agree it should use quite a bit of energy. Changing the lore is something which will still break items on those with compat enabled.
> 
> bStats isn't loading for me so I can't view the impact but I'd hope it's very minimal at this point anyway
> 
> So the options are:
> 
> 1. Change energy to 8/t
> 2. Change lore to 16/t (or another value) and break servers with compat enabled
> 
> I'll let others weigh in their opinions on that.

Personally, I vote changing the power - am I right in thinking the compat wont effect machines placed and in use, only those items still held as stacks?

image

Either way, its 100 / 1800~ servers and for those 100 the resolution would be to swap out the items to those players effected.

Side note, perhaps it's worth while getting a snapshot of the number of players on those servers still with compat on? While it's 100 servers, it may be a collection of dead/dying servers?

Sfiguz7:
I believe we should change the lore. We only support 1.14+ and have been for a while, 1.14 being the version that came out mid summer 2019: having compat enabled would only make sense if a server is 1.13 or very very early 1.14 which I doubt anyone really is, I believe most servers with compat on simply did not get the memo.

Considering we are planning on lifting support for those versions not too far in the future anyway (#staff-chat: https://discord.com/channels/565557184348422174/565570355851165711/985240413055119500) I don't really think this is as bad of a breaking change, if anything it will raise awareness once again about backwards compatibility being unnecessary and unnecessarily heavy on their servers.

Boomer-1:
change the lore, I vote in the affirmative

JasperChaseTOQ:
im also in favor im changing the lore. and im in agreement with sefiraat and sfiguz that its not that much of a breaking change and i personally feel its kinda like those who havent updated to java 16 yet from java 8  (they have had more the enough time to update and i dont really feel theres much reason if any not to) but maybe do as sefiraat suggested first and getting a snapshot of the number of players who play on the servers with compat on first to see just how much of an effect it will really have plus its also good information to have going foward imo.

PR #3602, opened 2022-07-02
Update to Java 16

TheBusyBiscuit:
## Description
This PR updates Slimefun to Java 16.
This is a breaking update for any servers that run older versions of Java.
**Please update your Java version immediately!**

This PR is still in progress. There are a lot of advantages that come with the usage of Java 16 and it would be nice to take advantage of most if not all of them.

I compiled a list of relatively easy to implement patterns here:
https://sonarcloud.io/project/issues?resolved=false&rules=java%3AS5194%2Cjava%3AS6201%2Cjava%3AS6204%2Cjava%3AS6206&types=CODE_SMELL&id=Slimefun_Slimefun4

Anyone can contribute!
Feel free to PR to this branch with a few changes from the list above ❤️

The following newly available patterns are desirable:

### "instanceof + cast"
```java
  if (x instanceof String string) {
    string.something()
  }
  
  // instead of:

  if (o instanceof String) {
    String string = (String) o;
    string.something()
  }
```

### switch expressions
```java
  return switch (x) {
    case 1 -> 1;
    case 2 -> 2;
  }

  // instead of:

  switch (x) {
    case 1:
      return 1;
    case 2:
      return 2;
  }
```

and various other patterns, check the sonarcloud link for more details.
Of course any other sonarcloud issue is always up for grabs too!

## Checklist


Sefiraat:
FYI, I am happy to do those stream changes noted in the sonar list but I can't be entirely sure when a collection is required as immutable or not. While I could go through and check every usage I feel that may be better served by those who know the code base far better than myself.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/MiningTask.java
```diff
@@ -268,8 +267,8 @@ private boolean push(@Nonnull ItemStack item) {
             if (chest.getType() == Material.CHEST) {
                 BlockState state = PaperLib.getBlockState(chest, false).getState();
 
-                if (state instanceof Chest) {
-                    Inventory inv = ((Chest) state).getBlockInventory();
+                if (state instanceof Chest chestState) {
```
Sefiraat, line 270: I think this may be a conflict that was resolved as my PR was merged in but I had called all of these patterns `instanceof Chest chest` so perhaps either change this to chest or change the same patterns in `VanillaInventoryDropHandler` and `OutputChest` ?

Sefiraat (approved):
~~Just one comment on the Chest patterns but even then it's still fine.~~ So LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/miner/MiningTask.java
TheBusyBiscuit: I renamed this because this class already has a field named ´chest´

Sefiraat: Quite right, didn't pay enough attention ha

TheBusyBiscuit:
@WalshyDev Do you perhaps have time to give this a look? :eyes:

Sfiguz7 (approved):
LGTM

Sefiraat (approved):
👍

PR #3608, opened 2022-07-06
Allow addons to disable warning

Sefiraat:
## Description
Currently when an addon adds an item to another addon's ItemGroup, a warning is thrown to the console. This allows addons to allow this behaviour if they want to (defaulted to current behaviour) so the warnings will not be triggered. This doesn't strictly change the behaviour of the items being added as this check didn't actually stop the items being added.

## Proposed changes
A boolean that can be set to disable the warning. Defaults to false.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
@@ -185,7 +185,7 @@ public class ItemGroup implements Keyed {
             return;
         }
 
-        if (isRegistered() && !item.getAddon().getName().equals(this.addon.getName()) && !isCrossAddonItemGroup()) {
+        if (isRegistered() && !isCrossAddonItemGroup() && !item.getAddon().getName().equals(this.addon.getName())) {
             item.warn("This item does not belong into ItemGroup " + this + " as that group belongs to " + this.addon.getName());
         }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
@@ -44,7 +44,7 @@ public class ItemGroup implements Keyed {
     protected final NamespacedKey key;
     protected final ItemStack item;
     protected int tier;
-    protected boolean allowOtherAddonItems = false;
+    protected boolean crossAddonItemGroup = false;
 
     /**
      * Constructs a new {@link ItemGroup} with the given {@link NamespacedKey} as an identifier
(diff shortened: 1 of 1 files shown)


WalshyDev (dismissed):
Sounds and looks good to me

ProfElements:
Yeah this has been affecting DynaTech for a minute because of Mob Data Cards in infinityexpansion

Sefiraat:
> Yeah this has been affecting DynaTech for a minute because of Mob Data Cards in infinityexpansion

That’s a good example, I’ll add this to IX asap assuming it’s approved :)

svr333 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
```diff
@@ -184,7 +185,7 @@ public void add(@Nonnull SlimefunItem item) {
             return;
         }
 
-        if (isRegistered() && !item.getAddon().getName().equals(this.addon.getName())) {
+        if (isRegistered() && !item.getAddon().getName().equals(this.addon.getName()) && !this.allowOtherAddonItems) {
```
TheBusyBiscuit, line 188: This should NOT use the field but rather the getter, otherwise overriding that method would have no effect.
And since you did not declare that method final, it can in fact be overridden.

TheBusyBiscuit (changes requested):
LGTM in general, thanks for this addition!
Just one bad practice that I spotted, otherwise very good.

I would also like to propose a better name for this attribute/method though. "CrossAddonItemGroup"? E.g. `isCrossAddonItemGroup()` and `setCrossAddonItemGroup`.
If it doesn't sound any better than it would be fine as is.

Sefiraat:
> LGTM in general, thanks for this addition! Just one bad practice that I spotted, otherwise very good.
> 
> I would also like to propose a better name for this attribute/method though. "CrossAddonItemGroup"? E.g. `isCrossAddonItemGroup()` and `setCrossAddonItemGroup`. If it doesn't sound any better than it would be fine as is.

I have no strong feelings about the name, and yours is less of a mouthful xD

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
```diff
@@ -184,7 +185,7 @@ public void add(@Nonnull SlimefunItem item) {
             return;
         }
 
-        if (isRegistered() && !item.getAddon().getName().equals(this.addon.getName())) {
+        if (isRegistered() && !item.getAddon().getName().equals(this.addon.getName()) && !isCrossAddonItemGroup()) {
```
TheBusyBiscuit, line 188:
```suggestion
        if (isRegistered() && !isCrossAddonItemGroup() && !item.getAddon().getName().equals(this.addon.getName())) {
```

TheBusyBiscuit (changes requested):
Just one small micro-optimization

TheBusyBiscuit (approved):
LGTM

svr333 (approved):
looks good

PR #3609, opened 2022-07-07
Added reinforced deepslate to `unbreakable_materials` tag

JasperChaseTOQ:
## Description

added reinforced deepslate to unbreakable materials tag because its a block with high controversy about whether it should be minable or not as it takes forever and a half to mine and its absolute unobtainable and cant be destroyed any other way so therefore i believe that it should be added to this so that it cant be accidently broken with an explosive pickaxe as if a player wasn't intending to it would mess with them as they couldn't replace it
## Proposed changes

added reinforced deepslate to unbreakable materials tag
## Checklist

### DIFF
--- a/src/main/resources/tags/unbreakable_materials.json
+++ b/src/main/resources/tags/unbreakable_materials.json
@@ -12,6 +12,10 @@
 		{
 			"id" : "minecraft:jigsaw",
 			"required" : false
+		},
+		{
+			"id" : "minecraft:reinforced_deepslate",
+			"required" : false
 		}
 	]
 }


Sefiraat:
It is, however, breakable. It can be broken by a player and, while it doesn’t drop anything, the block is removed from the world. It doesn’t belong in this tag I’m afraid.

JasperChaseTOQ:
> It is, however, breakable. It can be broken by a player and, while it doesn’t drop anything, the block is removed from the world. It doesn’t belong in this tag I’m afraid.

do you have an idea of a tag or something where it would belong instead then?

Sefiraat:
If the purpose is to stop reinforced deepslate from being effected by an exploding tool, then a new tag like EXPLOSIVE_TOOL_EXCLUDED which can include the unbreakable blocks and this but, imo it's only 5 hardness levels above obsidian so I feel like this just isn't needed? Server owners can add to the list themselves if they deem that they don't want it to break?

JasperChaseTOQ:
closing this as it does not belong in this tag as mentioned by sefiraat

PR #3616, opened 2022-07-12
Multimeter additional info

Sefiraat:
## Description
This isn't directly for Slimefun. Many addons have blocks that display information about their current operational status within a GUI inventory. While this is a good way to do this from the player's perspective, this creates a BlockMenu that is kept in RAM indefinitely which has recently been the cause of some issues. While the block storage rewrite will likely be the best 'fix' for this, for now this is to allow addons to remove information-only GUIs for their machines.

The biggest 'offender' for this is InfinityExpansion which in my most recent example server had 70k~ of these inventories which are, effectively, wasting RAM - and a lot of it.
If accepted I will be using this in Infinity, Networks and Tinker and will recommend to any other addons with similar blocks.

This solution will allow addons to display the information they are trying to put forward through the MultiMeter

## Proposed changes
Add a new ItemAttribute that addons can implement to add additional messages to the MultiMeter.

This could easily be a new tool instead. While I think it fits well in the MultiMeter, I'm happy to make a new item for this purpose.

## Checklist


TheBusyBiscuit (changes requested):
This feels like unnecessary weight in my opinion.
Why not add a default method to `EnergyComponent` like `sendMultimeterInfo(Player player, Location location)`, move the stuff for the default message there and override if necessary?

Sure, then every item would need to be required to be an energy component for the multimeter to be able to read it. But then again, a multimeter working on something that is not an energy component seems un-intuitive anyway.

Sefiraat:
> This feels like unnecessary weight in my opinion. Why not add a default method to `EnergyComponent` like `sendMultimeterInfo(Player player, Location location)`, move the stuff for the default message there and override if necessary?
> 
> Sure, then every item would need to be required to be an energy component for the multimeter to be able to read it. But then again, a multimeter working on something that is not an energy component seems un-intuitive anyway.

I did have two non-energycomponent items in my scope. Would a new item called a Probe be accepted instead? If not ill do as you suggest as near all the weight in the blockstorage reports I've worked on are EnergyComponents and should be enough in the main

svr333 (dismissed):
I like the simplicity of this implementation, I think it'll suffice for addon usage atm

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/Multimeter.java
```diff
@@ -54,6 +54,7 @@ public ItemUseHandler getItemHandler() {
                     p.sendMessage("");
                     Slimefun.getLocalization().sendMessage(p, "messages.multimeter", false, str -> str.replace("%stored%", stored).replace("%capacity%", capacity));
                     p.sendMessage("");
+                    component.sendMultimeterInfo(p, l);
```
TheBusyBiscuit, line 57: Maybe extract this into the default method:
`Slimefun.getLocalization().sendMessage(p, "messages.multimeter", false, str -> str.replace("%stored%", stored).replace("%capacity%", capacity));`

```suggestion
                    component.sendMultimeterInfo(p, l);
                    p.sendMessage("");
```

This way the padding is kept consistently and it allows you to specify whether your extra info should be added above or below the default text.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
```diff
@@ -199,4 +200,17 @@ default void removeCharge(@Nonnull Location l, int charge) {
         }
     }
 
+    /**
+     * Override this method to display additional information to the {@link Player} when a
+     * {@link io.github.thebusybiscuit.slimefun4.implementation.items.electric.gadgets.Multimeter} is used.
```
TheBusyBiscuit, line 205: Why is this written out...?
Add an import instead, you use the short version below anyway.
```suggestion
     * {@link Multimeter} is used.
```

TheBusyBiscuit (changes requested):

Sfiguz7 (dismissed):
LGTM, it is actually a good idea to move off of GUIs. It does make it slightly more inconvenient for players but the advantage in terms of server resources outweighs that imo

svr333 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/EnergyNetComponent.java
```diff
@@ -199,4 +201,30 @@ default void removeCharge(@Nonnull Location l, int charge) {
         }
     }
 
+    /**
+     * Override this method to change what information shown to the {@link Player} when a
```
WalshyDev, line 205:
```suggestion
     * Override this method to change what information is shown to the {@link Player} when a
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/Multimeter.java
```diff
@@ -43,17 +43,12 @@ public ItemUseHandler getItemHandler() {
...
 
-                    Location l = e.getClickedBlock().get().getLocation();
-                    String stored = NumberUtils.getCompactDouble(component.getCharge(l)) + " J";
-                    String capacity = NumberUtils.getCompactDouble(component.getCapacity()) + " J";
-
-                    Player p = e.getPlayer();
-                    p.sendMessage("");
-                    Slimefun.getLocalization().sendMessage(p, "messages.multimeter", false, str -> str.replace("%stored%", stored).replace("%capacity%", capacity));
-                    p.sendMessage("");
+                    e.cancel();
```
WalshyDev, line 50: nit: I'd put this at the start/start of the if, bit weird to have this in the middle (and easy to get lost -- "is this cancelled?" - not quick to find when surrounded by code)

WalshyDev (commented):
LGTM just quick english fix + nit :)

J3fftw1:
marked this as stale no activity in almost 2 months.
Will close this after a week, if its still stale.

PR #3632, opened 2022-07-28
Update messages.yml in fa 

a2py:
i am  from iran and this translation is wrong

### DIFF
--- a/src/main/resources/languages/fa/messages.yml
+++ b/src/main/resources/languages/fa/messages.yml
@@ -9,7 +9,7 @@ guide:
   locked: 'قفل شده است'
   work-in-progress: 'اين قابليت هنوز کامل نشده!'
   search:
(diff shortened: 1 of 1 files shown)


J3fftw1:
This should be done through crowdin not a pull request

src/main/resources/languages/fa/messages.yml
```diff
@@ -33,7 +33,7 @@ guide:
   credits:
     roles:
       wiki: '&3ویرایش کننده ویکی'
-      translator: '&9مترجم'
+      translator: '&9(amir22games;) )مترجم'
```
TheBusyBiscuit, line 36: Just no...

TheBusyBiscuit (commented):
We use Crowdin for translations.
And manipulating messages to insert your name is not acceptable!

PR #3633, opened 2022-07-28
Update researches.yml

a2py:
I can speak farsi (Persian) better than your translator


J3fftw1:
This should be done through crowdin not a pull request
Please check the following wiki
https://github.com/Slimefun/Slimefun4/wiki/Translating-Slimefun

PR #3654, opened 2022-08-13
- Adjust ticker times to show correct time on config changes for custom-ticker-delay

simon44556:
- Adjust ticker execution steps based on custom-ticker-delay changes

## Description

This PR adjust the tick rate based on the config value for custom-ticker-delay.
It will try to keep the tickers execution at a similar speed as to when the default value is used.

I also adjust the displayed time based on the difference of execution time.

## Proposed changes

Changed CraftingOperation to take into account the new ticks config. Normalize the value to the default one of one operation per 10 ticks. 

Changed NumberUtils. Added logic that calculates the time new changes will take to finish the operation.

Not sure if there is na issue open.

Here is how to reproduce the issue:
1. Adjust custom-ticker-delay to 20.
2. Run a ticker ingame.
3. The display timer will be wrong the execution slows down by 1/2 so the secounds and minutes are not synced to the actual values.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
@@ -45,7 +45,7 @@ public class CraftingOperation implements MachineOperation {
          * Normalize the tickrate to the config value custom-ticker-delay.
          * This makes the machines run nearly the same speed on changing the delay.
          */
-        int tickRate = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
+        int tickRate = Slimefun.getTickerTask().getTickRate();
         int normalizedTickRate = (int) Math.round((tickRate / 10.0D) * num);
 
         currentTicks += Math.max(normalizedTickRate, num);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
@@ -7,7 +7,6 @@ import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.core.machines.MachineOperation;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
-
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineRecipe;
 
 /**
@@ -49,7 +48,7 @@ public class CraftingOperation implements MachineOperation {
         int tickRate = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
         int normalizedTickRate = (int) Math.round((tickRate / 10.0D) * num);
 
-        currentTicks += normalizedTickRate;
+        currentTicks += Math.max(normalizedTickRate, num);
     }
 
     public @Nonnull ItemStack[] getIngredients() {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
@@ -7,7 +7,6 @@ import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.core.machines.MachineOperation;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
-
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineRecipe;
 
 /**
@@ -22,7 +21,6 @@ public class CraftingOperation implements MachineOperation {
     private final ItemStack[] results;
 
     private final int totalTicks;
-    private int tickRate;
     private int currentTicks = 0;
 
     public CraftingOperation(@Nonnull MachineRecipe recipe) {
@@ -37,8 +35,6 @@ public class CraftingOperation implements MachineOperation {
         this.ingredients = ingredients;
         this.results = results;
         this.totalTicks = totalTicks;
-
-        this.tickRate = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
     }
 
     @Override
@@ -49,7 +45,8 @@ public class CraftingOperation implements MachineOperation {
          * Normalize the tickrate to the config value custom-ticker-delay.
          * This makes the machines run nearly the same speed on changing the delay.
          */
-        int normalizedTickRate = (int) Math.round((tickRate / 10.0d) * num);
+        int tickRate = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
+        int normalizedTickRate = (int) Math.round((tickRate / 10.0D) * num);
 
         currentTicks += normalizedTickRate;
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
@@ -6,6 +6,7 @@ import org.apache.commons.lang.Validate;
 import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.core.machines.MachineOperation;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 
 import me.mrCookieSlime.Slimefun.Objects.SlimefunItem.abstractItems.MachineRecipe;
 
@@ -21,10 +22,13 @@ public class CraftingOperation implements MachineOperation {
     private final ItemStack[] results;
 
     private final int totalTicks;
+    private int tickRate;
     private int currentTicks = 0;
 
     public CraftingOperation(@Nonnull MachineRecipe recipe) {
         this(recipe.getInput(), recipe.getOutput(), recipe.getTicks());
+
+        this.tickRate = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
     }
 
     public CraftingOperation(@Nonnull ItemStack[] ingredients, @Nonnull ItemStack[] results, int totalTicks) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -171,9 +176,31 @@ private NumberUtils() {}
...
 
-    public static @Nonnull String getTimeLeft(int seconds) {
+    public static @Nonnull String getTimeLeft(int ticksLeft) {
         String timeleft = "";
 
+        /**
+         * This is the amount calculated in addProgress(). 
+         * This will be the scale used for timer.
+         */
+        double timeScale = (TICK_RATE / 10.0d) * 1;
```
J3fftw1, line 186: Is the * 1 necessary here. it will always stay the same.

simon44556: You are correct this is not required. I removed it now.

J3fftw1: i was doubting my self for a tad. opened my calc and did 0.1*1

svr333: lol

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -25,6 +25,10 @@
  *
  */
 public final class NumberUtils {
+    /**
+     * Save config value for tick rate
+     */
+    private static final int TICK_RATE = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
```
TheBusyBiscuit, line 31: This is not safe to call in a static context.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -35,7 +39,8 @@ public final class NumberUtils {
     /**
      * We do not want any instance of this to be created.
      */
-    private NumberUtils() {}
+    private NumberUtils() {
+    }
```
TheBusyBiscuit, line 43: Please restore the original formatting here

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -171,9 +176,31 @@ private NumberUtils() {}
...
+            seconds = (int) (ticksLeft * (normalScale / 2 + 1.0d));
+        }
+        // Adjust for negative change 
+        else if (normalScale > 0) {
+            seconds = (int) (ticksLeft * (normalScale + 1.0d));
+        }
+        // No change 
+        else {
+            seconds = ticksLeft;
+        }
```
TheBusyBiscuit, line 202: Please adhere to our code style.

```suggestion
        if (normalScale < 0) {
            //Adjust for positive change
            seconds = (int) (ticksLeft * (normalScale / 2 + 1.0D));
        } else if (normalScale > 0) {
            // Adjust for negative change 
            seconds = (int) (ticksLeft * (normalScale + 1.0D));
        } else {
            // No change 
            seconds = ticksLeft;
        }
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -171,9 +176,31 @@ private NumberUtils() {}
...
 
-    public static @Nonnull String getTimeLeft(int seconds) {
+    public static @Nonnull String getTimeLeft(int ticksLeft) {
         String timeleft = "";
 
+        /**
+         * This is the amount calculated in addProgress(). 
+         * This will be the scale used for timer.
+         */
+        double timeScale = (TICK_RATE / 10.0d);
```
TheBusyBiscuit, line 186:
```suggestion
        double timeScale = TICK_RATE / 10.0D;
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -171,9 +176,31 @@ private NumberUtils() {}
...
+        double normalScale = ((timeScale) - Math.round(timeScale));
```
TheBusyBiscuit, line 187: Redundant braces.
```suggestion
        double normalScale = timeScale - Math.round(timeScale);
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
```diff
@@ -35,12 +37,21 @@ public CraftingOperation(@Nonnull ItemStack[] ingredients, @Nonnull ItemStack[]
         this.ingredients = ingredients;
         this.results = results;
         this.totalTicks = totalTicks;
+
+        this.tickRate = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
```
TheBusyBiscuit, line 41: Reading a config setting in a constructor here is not efficient...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
```diff
@@ -21,6 +22,7 @@ public class CraftingOperation implements MachineOperation {
     private final ItemStack[] results;
 
     private final int totalTicks;
+    private int tickRate;
```
TheBusyBiscuit, line 25: Why is this not final?

TheBusyBiscuit, line 25: Also why is this defines as a class member???
This value is not dependent on the state of instances from this class...

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
```diff
@@ -35,12 +37,21 @@ public CraftingOperation(@Nonnull ItemStack[] ingredients, @Nonnull ItemStack[]
...
     @Override
     public void addProgress(int num) {
         Validate.isTrue(num > 0, "Progress must be positive.");
-        currentTicks += num;
+
+        /**
+         * Normalize the tickrate to the config value custom-ticker-delay.
+         * This makes the machines run nearly the same speed on changing the delay.
+         */
+        int normalizedTickRate = (int) Math.round((tickRate / 10.0d) * num);
```
TheBusyBiscuit, line 52:
```suggestion
        int normalizedTickRate = (int) Math.round((tickRate / 10.0D) * num);
```

TheBusyBiscuit (changes requested):
Lots of changes required.
Many of them are related to formatting.
For reference: Double-values are either displayed as `x.x` or with a capital letters.
Lowercase letters should not be used for explicit number typing.

Also there is no reason for you to store the tick rate as a class member as that shouldn't change on a per-instance-level.

simon44556:
Ok I think that now it should be ok.

I went through all the comments and redid the formating and the config access.

svr333:
Tip for future PRs, you can mark any Requested Change as Resolved, that way it's clear what has been addressed and what wasn't

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -171,9 +170,29 @@ private NumberUtils() {}
...
+        if (normalScale < 0) {
+            //Adjust for positive change
+            seconds = (int) (ticksLeft * (normalScale / 2 + 1.0D));
+        } else if (normalScale > 0) {
+            // Adjust for negative change 
+            seconds = (int) (ticksLeft * (normalScale + 1.0D));
+        } else {
+            // No change 
+            seconds = ticksLeft;
+        }
```
svr333, line 194: Not a necessary change but we can get rid of the else bracket by defaulting seconds to ticksLeft

```suggestion
        int seconds = ticksLeft;

        if (normalScale < 0) {
            //Adjust for positive change
            seconds = (int) (ticksLeft * (normalScale / 2 + 1.0D));
        } else if (normalScale > 0) {
            // Adjust for negative change 
            seconds = (int) (ticksLeft * (normalScale + 1.0D));
        }
```

svr333 (commented):
PR looks good to me, thanks for your contribution

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
simon44556: Good idea yes.

svr333 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/CraftingOperation.java
```diff
@@ -40,7 +40,15 @@ public CraftingOperation(@Nonnull ItemStack[] ingredients, @Nonnull ItemStack[]
     @Override
     public void addProgress(int num) {
         Validate.isTrue(num > 0, "Progress must be positive.");
-        currentTicks += num;
+
+        /**
+         * Normalize the tickrate to the config value custom-ticker-delay.
+         * This makes the machines run nearly the same speed on changing the delay.
+         */
+        int tickRate = Slimefun.getCfg().getInt("URID.custom-ticker-delay");
```
TheBusyBiscuit, line 48:
```suggestion
        int tickRate = Slimefun.getTickerTask().getTickRate();
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/NumberUtils.java
```diff
@@ -171,9 +170,26 @@ private NumberUtils() {}
...
 
-    public static @Nonnull String getTimeLeft(int seconds) {
+    public static @Nonnull String getTimeLeft(int ticksLeft) {
         String timeleft = "";
 
+        /**
+         * This is the amount calculated in addProgress(). 
+         * This will be the scale used for timer.
+         */
+        double timeScale = Slimefun.getCfg().getInt("URID.custom-ticker-delay") / 10.0D;
```
TheBusyBiscuit, line 180:
```suggestion
        double timeScale = Slimefun.getTickerTask().getTickRate() / 10.0D;
```

TheBusyBiscuit (changes requested):
As I mentioned in my last review: Config readouts can be expensive operations!

It is much safer and more efficient to just use the provided getter for reading the cached value instead.

J3fftw1 (approved):
Will make needed changes so its not a breaking API endpoint.

PR #3672, opened 2022-09-12
Change the registry to use world UUIDs

gecko10000:
## Description
I have recently run into an issue where a world, when created quickly enough after deletion of a world with the same name, will not store data. I have tested this extensively and narrowed the issue down to [this](https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java#L102) ticker code and [this](https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java#L116) check for the world's existence. Since the ticker code only runs every half second, it does not remove the world from the registry until up to a Slimefun tick later and therefore causes issues in the constructor.

An easy way to recreate this issue is to delete a world and immediately create a new one with the same name.

## Proposed changes
This PR changes the registry to use UUIDs instead of the world names themselves. This _does not_ affect the storage location in files and therefore these are not breaking changes.

This may not be the ideal way to do it; it may be better to remove the world from the registry [here](https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/WorldListener.java#L28) instead.

No issue opened

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
@@ -87,6 +87,7 @@ import io.github.thebusybiscuit.slimefun4.implementation.listeners.MiningAndroid
 import io.github.thebusybiscuit.slimefun4.implementation.listeners.MultiBlockListener;
 import io.github.thebusybiscuit.slimefun4.implementation.listeners.NetworkListener;
 import io.github.thebusybiscuit.slimefun4.implementation.listeners.PlayerProfileListener;
+import io.github.thebusybiscuit.slimefun4.implementation.listeners.RadioactivityListener;
 import io.github.thebusybiscuit.slimefun4.implementation.listeners.SeismicAxeListener;
 import io.github.thebusybiscuit.slimefun4.implementation.listeners.SlimefunBootsListener;
 import io.github.thebusybiscuit.slimefun4.implementation.listeners.SlimefunBowListener;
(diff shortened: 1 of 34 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
```diff
@@ -1,12 +1,7 @@
 package io.github.thebusybiscuit.slimefun4.implementation;
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
svr333, line 4: No star imports please (very possible your IDE did this automatically)

svr333 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
gecko10000: Yes, it did.

svr333 (approved):
This does look indeed like it isnt a breaking change and would still work with all worlds.
So in that case it looks good to me!

J3fftw1:
im not sure how i closed this throught git.
I was updating his branch.

PR #3725, opened 2022-12-14
Disable Slimefun when CS-CoreLib is installed

ybw0014:
## Description

(Idea from @J3fftw1 )
Since CS-CoreLib is no longer needed, and will cause issues to some addons, we should disable Slimefun and let users remove CS-CoreLib in order to run Slimefun and addons properly.

## Proposed changes

Change the CS-CoreLib warning level to severe.
Disable Slimefun when CS-CoreLib is installed.

## Checklist

- [x] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.14.* - 1.19.*).


J3fftw1:
> Think this is a solid idea

Thanks

ybw0014:
> I assume you've tested the change, however small and discrete? If so can you just mark the ticks in the PR otherwise test then tick.

I didn't have time to test it yesterday as it was made before I sleep. will test it today.

svr333 (approved):
lgtm

Sefiraat (approved):
I see the ticks so I’m Happy, this is a good idea that will help greatly I think :)

PR #3733, opened 2023-01-03
[Fix] BlockPlacer placing disabled items

justtos:
## Description

BlockPlacer being able to place `disabled` items

## Proposed changes

In the `onBlockDispense` method, adds a simple check if the item is disabled or not.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
@@ -104,7 +104,7 @@ public class BlockPlacer extends SlimefunItem {
 
                 if (item != null) {
                     // Check if this Item can even be placed down
-                    if (!(item instanceof NotPlaceable)) {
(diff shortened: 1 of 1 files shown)


svr333 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/BlockPlacer.java
```diff
@@ -104,7 +104,7 @@ private BlockDispenseHandler onBlockDispense() {
 
                 if (item != null) {
                     // Check if this Item can even be placed down
-                    if (!(item instanceof NotPlaceable)) {
+                    if (!(item instanceof NotPlaceable) && !item.isDisabled()) {
```
WalshyDev, line 107: This will check world-specific disablement in addition to global.

```suggestion
                    if (!(item instanceof NotPlaceable) && !item.isDisabledIn(dispenser.getWorld())) {
```

WalshyDev (changes requested):

svr333 (approved):

WalshyDev (approved):

PR #3750, opened 2023-01-24
fix `/sf versions` permission

8MiYile:
## Description

everyone can use `/sf versions` command
client plugin [SlimefunEssentials](https://github.com/JustAHuman-xD/SlimefunEssentials) mayve use this command
get addons list

## Proposed changes

`plugins.yml` 
`slimefun.command.versions`
`default: true`

## Checklist

- [x] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.14.* - 1.19.*).

### DIFF
--- a/src/main/resources/plugin.yml
+++ b/src/main/resources/plugin.yml
@@ -58,7 +58,7 @@ permissions:
     default: op
   slimefun.command.versions:
     description: Allows you to do /sf versions
-    default: op
+    default: true
   slimefun.command.backpack:
     description: Allows you to do /sf backpack
     default: op


J3fftw1:
Like i said on Discord you can manually apply this permission to people.

8MiYile:
@J3fftw1 
I know, but not all servers add this permission, after all, Some administrators are lazy

J3fftw1:
> @J3fftw1 I know, but not all servers add this permission, after all, Some administrators are lazy

thats sadly on that server. A lot of servers prefer to not show the version thats why its not defaulted to true.

8MiYile:
Open this permission or add a new command `/sf getaddonlist`

J3fftw1:
`/sf verions` already does what you want but this permission needs to be given.

TheBusyBiscuit (commented):
This is definitely not a "fix" nor is it a desirable change in my opinion.
The default has been consistent for **6 years** (looked it up, this was literally last changed exactly 6 years ago next week lol).
And never do I recall any server owner actively wanting the default to be changed. After all, configurations like these are up to the server admins to make. And changing the default would affect hundreds of servers without any heads up.

I don't see this change justified enough for that to take place.

PR #3752, opened 2023-01-25
Added bamboo as a fuel type for androids

J3fftw1:
## Description

Added bamboo as a fuel type for androids

We might wanna change this again when 1.20 releases since all the new types of bamboo blocks.
Also do we wanna include all other burnable material since there is a huge list?

## Proposed changes

Added bamboo as a fuel type for androids

## Checklist


TheBusyBiscuit (approved):
lgtm

svr333 (approved):

PR #3760, opened 2023-02-05
fix taking damage on head collision while wearing elytra cap

### LINKED ISSUE #3759: Elytra Cap doesn't negate damage when crashing to ground or walls using elytra
### ❗ Checklist

### 📍 Description

Elytra Cap doesn't negate damage when crashing to ground or walls using elytra, while the description of the item said otherwise

### 📑 Reproduction Steps

1. Equip an elytra cap
2. Equip an elytra
3. Fly using elytra
4. Crash into walls or ground
5. See that you got damaged

### 💡 Expected Behavior

It should negate the damage

### 📷 Screenshots / Videos

https://youtu.be/TjE_2hCBhRw

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

```
Paper git-Paper-386 (MC: 1.19.3)
Slimefun DEV - 1043 (git 4bfc4f06)
Metrics-Module #28
Java 17

No Addons installed
```

### 🧭 Other plugins

_No response_

J3fftw1:
## Description

This pr somewhat resolves #3759.
It doesnt fully for some reason when you collide with the floor. it doesnt detect you are gliding.
#isGliding() just doesnt trigger. I dont see any way to work around this sadly.
If anyone does feel free to let me know.

Apparently the ElytraCap never worked

## Proposed changes

Change the checked armor slot from 0 to 3 (0 = boots, 1 = leggings, 2  = chestplate and 3 = helmet) (credits to @variananora 
Removed player variable and added it to the instance check.
Changed #setDamage(0) to #setCancelled(true) This seems to work more reliable.
Changed negated #isPresent to #isEmpty()

Resolves #3759  (not fully tho)

## Checklist

- [x] I pressed space


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ElytraImpactListener.java
```diff
@@ -34,35 +35,31 @@ public ElytraImpactListener(@Nonnull Slimefun plugin) {
...
+            Optional<PlayerProfile> optional = PlayerProfile.find(p);
 
-            if (p.isGliding()) {
-                Optional<PlayerProfile> optional = PlayerProfile.find(p);
-
-                if (!optional.isPresent()) {
-                    PlayerProfile.request(p);
-                    return;
-                }
+            if (optional.isEmpty()) {
```
Sefiraat, line 46: By removing the `isGliding()` check and having `DamageCause#FALL` and `DamageCause#FLY_INTO_WALL` I would imagine this would mean the elytra would now protect you against normal fall damage without activating the elytra. I cant test right now so I will either test tomorrow or if someone else can get here first to test it out, lmk. If it doesn't do this then I am happy with the PR.

Sefiraat: Derp

Sefiraat (approved):
Looks good to me 👍

J3fftw1:
Not sure if we wanna close the issue linked or reopen a new 1 since this is a partial fix.
Also not sure if we can fix the other part and check if the player has flown in the past.

Sfiguz7 (commented):
Everything looks good to me, sad about the hit the ground part. Maybe this event gets called first? If so, when hitting the ground from the server's POV we get that the elytra is disabled first, meaning the player is not gliding anymore when the damage gets calculated.
https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/entity/EntityToggleGlideEvent.html

Mind checking if this is the case? If so, we can add another listener to keep track of players who have just hit the ground but have not yet received damage - say by adding the players to a temporary collection that does not get saved onDisable (so that it gets cleared over restarts) and running a task to remove the player from said collection on the next tick (so it cannot be abused). If it isn't hey at least it is a partial fix, I can't think of any other way to check what is happening, we can just merge it as is

J3fftw1:
> Everything looks good to me, sad about the hit the ground part. Maybe this event gets called first? If so, when hitting the ground from the server's POV we get that the elytra is disabled first, meaning the player is not gliding anymore when the damage gets calculated. [hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/entity/EntityToggleGlideEvent.html](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/event/entity/EntityToggleGlideEvent.html)
> 
> Mind checking if this is the case? If so, we can add another listener to keep track of players who have just hit the ground but have not yet received damage - say by adding the players to a temporary collection that does not get saved onDisable (so that it gets cleared over restarts) and running a task to remove the player from said collection on the next tick (so it cannot be abused). If it isn't hey at least it is a partial fix, I can't think of any other way to check what is happening, we can just merge it as is

Im pretty sure i checked this. 
not 100% sure more like 60 since its been a while
but i think i remember that there is no event called when this happens but im not sure i will verifiy this later this week.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ElytraImpactListener.java
```diff
@@ -18,51 +22,63 @@
...
     }
 
+    @EventHandler
+    public void onGlideToggle(EntityToggleGlideEvent event) {
+        Entity entity = event.getEntity();
+        if (entity instanceof Player player && player.isGliding()) {
+            UUID uuid = player.getUniqueId();
+            gliding.add(uuid);
+        }
+        Slimefun.instance().getServer().getScheduler().runTaskLater(Slimefun.instance(), gliding::clear, 1);
```
Sefiraat, line 50: Can we just have some docs/comments somewhere detailing why we add them/clear them 1 tick later from the group? We know its required and we know why but just reading the code its reads a bit werid not knowing about the 'forced toggle' minecraft does

J3fftw1: Not sure if i made the right comment i cant exactly remember why we do it

Sfiguz7:
So is this functional in the end? I forget if we ever had a conversation about it after my last review.
If it works (build testers wink wink), let's cull the PRs amount with yet another merge!

Sefiraat:
I was happy with the function, I’ll try to to a final fuller review later today if I can

Sefiraat (approved):

JustAHuman-xD (approved):
LGTM

Issue #3774, opened 2023-02-20
Soulbound items disappear when restarting the server mid-deathscreen


KiinoOFC:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

When a player dies with Soulbound items. After restarting the server, the items disappear

### 📑 Reproduction Steps

1. Have items with Soulbound
2. Then you die
3. Restart the server
4. Join the server
5. Checks if Soulbound items are there

### 💡 Expected Behavior

Soulbound items do not disappear

### 📷 Screenshots / Videos

Before restarting
![image](https://user-images.githubusercontent.com/58773688/220183635-57c019f7-0cf6-40e1-ac38-346f636c1f1b.png)

After restart
![image](https://user-images.githubusercontent.com/58773688/220183779-11f7e723-31f3-4714-aaae-bd20a30229d8.png)

### 📜 Server Log

no errors to log

### 📂 `/error-reports/` folder

no errors to folder

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

Paper git-Paper-411 (MC: 1.17.1)
Slimefun DEV - 1043 (git 4bfc4f06)
Metrics-Module #28
Java 17

Boomer-1:
please provide your full plugin list

N1kyZ:
OBS: Have to restart server on death screen

no data pack installed
plugins:
![image](https://user-images.githubusercontent.com/58773688/220218301-dbf5ce44-9f76-4b40-bca7-7fe5c22e990c.png)

Boomer-1:
over many repeated attempts i cannot duplicate this with 1.17.1 and dev 1043

Boomer-1:
what mods are you using?

N1kyZ:
no mod

Boomer-1:
well, you have at least one based on the picture you provided of your inventory. something you've installed, in the server or the client is interfering. without all the info we can't help

N1kyZ:
I managed to fix my OBS, maybe this will help

https://user-images.githubusercontent.com/58773688/220232637-c8984e5a-a6c0-45c2-a760-aab205f5870a.mp4

[...]

PR #3781, opened 2023-02-24
fix soulbound listener bugs when death

Crsuh2er0:
## Description

To fix soulbound listener bugs when death.

## Proposed changes

Set items in player inventory 1 tick after death.

Rewsolves #3774 

## Checklist

- [√] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [√] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [√] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.16.* - 1.19.*).
- [√] I followed the existing code standards and didn't mess up the formatting.
- [√] I did my best to add documentation to any public classes or methods I added.
- [√] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [√] I added sufficient Unit Tests to cover my code.

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/SoulboundListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/SoulboundListener.java
@@ -7,6 +7,7 @@ import java.util.UUID;
 import javax.annotation.Nonnull;
 
 import org.bukkit.entity.Player;
+import org.bukkit.entity.Slime;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.Listener;
 import org.bukkit.event.entity.PlayerDeathEvent;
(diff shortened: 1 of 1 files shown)


J3fftw1:
Wouldn’t this cause a null pointer since some players might not respawn directly then the inventory is null.

the best way to fix the issue is to save the soul bound items to data-storage i guess when the player leaves or when the server shuts down

Crsuh2er0:
> Wouldn’t this cause a null pointer since some players might not respawn directly then the inventory is null.
> 
> the best way to fix the issue is to save the soul bound items to data-storage i guess when the player leaves or when the server shuts down

@NotNull HumanEntity#getInventory
I have tested it and it works well on 1.16.5 and 1.19.2.

J3fftw1:
This would however still cause an exception if the player leaves instantly. 
There is a 1 in 20 chance of it happening and we are already null checking the items. We could just add another null check for player.

Crsuh2er0:
Now we use runTask to set the items back to the player's inventory immediately after the event and the exception won't be caused.

Sfiguz7:
Since this is a draft and we have #3830 opened I will close the PR. Should the other not be approved we can think it over again and maybe reopen this instead

PR #3786, opened 2023-03-02
Fix #3361

### LINKED ISSUE #3361: Soulbound not disabled when disabled in config
### ❗ Checklist

- [x] I am using the official english version of Slimefun and did not modify the jar.

### 📍 Description

When I disabled soulbound items in the config already soulbound enchanted items from the ancient rune still functioned. No other config option exists for disabling already soulbound enchanted items.

### 📑 Reproduction Steps

1. Have ancient rune of soulbind enabled
2. Drop the rune with a piece of armor to enchant it
3. Disable the rune
4. Use the soulbound item

### 💡 Expected Behavior

Expected behaviour would be the soulbounded item would not function just like any SF item that is disabled.

### 📷 Screenshots / Videos

https://discord.com/channels/565557184348422174/565569972697301020/914758306130706442 <-- Related convo in the SF discord

### 📜 Server Log

Nothing in Logs

No errors

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

Versions

### 🧭 Other plugins

_No response_

JustAHuman-xD:
## Description
When a Item is soulbound via the Soulbound Rune it will still act as soulbound when the rune is disabled. However if someone crafts for example, a soulbound axe it will only act as soulbound if the axe is not disabled. This makes the item only act as soulbound if the rune is not disabled.

## Proposed changes
Added a reference to the soulbound rune slimefun item
Check if it is not null, not disabled, and not disabled in the specific world

Resolves #3361 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -121,7 +121,7 @@ public final class SlimefunUtils {
         if (item != null && item.getType() != Material.AIR) {
             ItemMeta meta = item.hasItemMeta() ? item.getItemMeta() : null;
 
-            if (!SlimefunItems.SOULBOUND_RUNE.getItem().isDisabled() && hasSoulboundFlag(meta)) {
+            if (SlimefunItems.SOULBOUND_RUNE.getItem() != null && !SlimefunItems.SOULBOUND_RUNE.getItem().isDisabled() && hasSoulboundFlag(meta)) {
                 return true;
             }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -11,6 +11,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.implementation.SlimefunItems;
 import org.apache.commons.lang.Validate;
 import org.bukkit.ChatColor;
 import org.bukkit.Location;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -120,7 +123,8 @@ public static boolean isSoulbound(@Nullable ItemStack item, @Nullable World worl
         if (item != null && item.getType() != Material.AIR) {
             ItemMeta meta = item.hasItemMeta() ? item.getItemMeta() : null;
 
-            if (hasSoulboundFlag(meta)) {
+            SlimefunItem rune = SlimefunItems.SOULBOUND_RUNE.getItem();
+            if (hasSoulboundFlag(meta) && rune != null && !rune.isDisabled() && (world == null || !rune.isDisabledIn(world))) {
```
TheBusyBiscuit, line 127: This should probably be nested inside the if-statement. If the flag is not present then there is no need to perform any additional checks.

Or a much better approach would be to put the `hasSoulboundFlag` at the end. Since it might be more expensive computation-wise. Depends on what the cheaper operation is. The null check and `.isDisabled()` are probably the cheapest ones. So let's try it that way I guess. This should allow the code to disregard unneeded operations

```suggestion
            SlimefunItem rune = SlimefunItems.SOULBOUND_RUNE.getItem();
            if (rune != null && !rune.isDisabled() && (world == null || !rune.isDisabledIn(world)) && hasSoulboundFlag(meta)) {
```

TheBusyBiscuit (changes requested):

JustAHuman-xD:
Yeah I was wondering which would be cheaper, wasn't sure.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -11,6 +11,7 @@
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.implementation.SlimefunItems;
```
Sefiraat, line 14: This import needs to be moved to order correctly :)

Sefiraat (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
JustAHuman-xD: Fixed!

Sefiraat (approved):
👍

svr333 (approved):

TheBusyBiscuit (approved):
LGTM

PR #3788, opened 2023-03-02
Adding ExternallyInteractable

Sefiraat:
## Description

Something I am working on in other addons is the ability to trigger/interact with other SlimefunItems when placed in the world in a generic way, i.e. not relying on depending on each individual possible addon/specifying ID's to do so. 

## Proposed changes

The idea here is that addons can implement the Interface and, within the overridden method, complete whatever actions it would like to do when being interacted with and return a result object denoting a generic summary what happened as part of the interaction. 

Currently I have only included the success of the interaction and any possible items generated as part of said interaction (as this has been initially thought up for networks). If there are any ideas for other things to be added that would also be useful, I can whack that in as well.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
@@ -6,6 +6,7 @@ import org.bukkit.inventory.ItemStack;
 
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -60,13 +61,20 @@ public interface ExternallyInteractable {
         }
 
         /**
+         * Adds an or several {@link ItemStack}'s into the result
+         */
+        public void addResultItems(ItemStack... itemStacks) {
+            Collections.addAll(resultItems, itemStacks);
+        }
+
+        /**
          * Returns the {@link ItemStack}(s) produced as a result of this interaction, if any.
          *
-         * @return A list of {@link ItemStack}(s) to process.
+         * @return An unmodifiable {@link Set} of {@link ItemStack}(s) created due to the interaction.
          */
         @Nonnull
         public Set<ItemStack> getResultItems() {
-            return resultItems;
+            return Collections.unmodifiableSet(resultItems);
         }
     }
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
@@ -5,6 +5,7 @@ import org.bukkit.Location;
 import org.bukkit.inventory.ItemStack;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 import java.util.Collections;
 import java.util.HashSet;
@@ -28,8 +29,7 @@ public interface ExternallyInteractable {
      *
      * @return The {@link InteractionResult} denoting the result of the interaction.
      */
-    @Nonnull
-    InteractionResult onInteract(@Nonnull Location location);
+    @Nonnull InteractionResult onInteract(@Nonnull Location location);
 
     /**
      * This class represents the result of an interaction on an {@link ExternallyInteractable} item.
@@ -37,8 +37,8 @@ public interface ExternallyInteractable {
     class InteractionResult {
 
         private final boolean interactionSuccessful;
-        @Nonnull
-        private final Set<ItemStack> resultItems = new HashSet<>();
+        private @Nullable String resultMessage;
+        private final @Nonnull Set<ItemStack> resultItems = new HashSet<>();
 
         /**
          * Creates a new InteractionResult.
@@ -81,9 +81,36 @@ public interface ExternallyInteractable {
          *
          * @return An unmodifiable {@link Set} of {@link ItemStack}(s) created due to the interaction.
          */
-        @Nonnull
-        public Set<ItemStack> getResultItems() {
+        public @Nonnull Set<ItemStack> getResultItems() {
             return Collections.unmodifiableSet(resultItems);
+        }
+
+        /**
+         * Sets a custom result message for this interaction.
+         *
+         * @param
+         *       resultMessage The message to be sent with the Result
+         */
+        public void setResultMessage(@Nullable String resultMessage) {
+            this.resultMessage = resultMessage;
+        }
+
+        /**
+         * Returns whether this result has a result message or not.
+         *
+         * @return True if a result message is present
+         */
+        public boolean hasResultMessage() {
+            return this.resultMessage != null;
+        }
+
+        /**
+         * Returns the custom result message for this result.
+         *
+         * @return A String of the provided custom result message.
+         */
+        public @Nullable String getResultMessage() {
+            return resultMessage;
         }
     }
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
@@ -1,0 +1,72 @@
+package io.github.thebusybiscuit.slimefun4.core.attributes;
+
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
+import org.bukkit.Location;
+import org.bukkit.inventory.ItemStack;
+
+import javax.annotation.Nonnull;
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Implement this interface for any {@link SlimefunItem} to provide methods for
+ * external code to 'interact' with the item when placed as a block in the world.
+ *
+ * @author TheBusyBiscuit
+ *
+ */
+public interface ExternallyInteractable {
+
+    /**
+     * This method should be used by the implementing class to fulfill the actions needed
+     * when being interacted with returning the result of the interaction.
+     *
+     * @param location
+     *                 The {@link Location} of the Block being targeted for the interaction.
+     *
+     * @return The {@link InteractionResult} denoting the result of the interaction.
+     */
+    @Nonnull
+    InteractionResult onInteract(@Nonnull Location location);
+
+    /**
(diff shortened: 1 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.annotation.ParametersAreNonnullByDefault;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Implement this interface for any {@link SlimefunItem} to provide methods for
+ * external code to 'interact' with the item when placed as a block in the world.
+ *
+ * @author TheBusyBiscuit
```
TheBusyBiscuit, line 17: I do not remember writing this code...
#identitytheft

Sefiraat: Dr. CopyPasta strikes again xD

Sfiguz7 (dismissed):
LGTM

SchnTgaiSpock:
> If there are any ideas for other things to be added that would also be useful, I can whack that in as well.

Maybe if an interaction was unsuccessful, it could also return the reason? Like not researched, not enough of some resource, etc...; it could then be used for smth like a message to the player

Sfiguz7 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Implement this interface for any {@link SlimefunItem} to provide methods for
+ * external code to 'interact' with the item when placed as a block in the world.
+ *
+ * @author Sefiraat
+ *
```
WalshyDev, line 19:
```suggestion
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
```diff
@@ -0,0 +1,116 @@
...
+ */
+public interface ExternallyInteractable {
```
WalshyDev, line 21:
```suggestion
 */
@FunctionalInterface
public interface ExternallyInteractable {
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
```diff
@@ -0,0 +1,116 @@
...
+     *                 The {@link Location} of the Block being targeted for the interaction.
+     *
+     * @return The {@link InteractionResult} denoting the result of the interaction.
+     */
+    @Nonnull InteractionResult onInteract(@Nonnull Location location);
+
+    /**
+     * This class represents the result of an interaction on an {@link ExternallyInteractable} item.
+     */
+    class InteractionResult {
```
WalshyDev, line 37: we should avoid inner classes as much as possible. Any reason we don't want to put this in its own class?

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
```diff
@@ -0,0 +1,116 @@
...
+
+        private final boolean interactionSuccessful;
+        private @Nullable String resultMessage;
```
WalshyDev, line 40: Use case?

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
```diff
@@ -0,0 +1,116 @@
...
+        private final @Nonnull Set<ItemStack> resultItems = new HashSet<>();
```
WalshyDev, line 41: I don't want this to turn into a huge class with 10 different possible outputs, may be worth making this more abstract so we don't need to refactor this later.
Not sure if I want to block on this yet.

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/ExternallyInteractable.java
Sefiraat: It just allows the object being interacted with to define a string message - could be anything. Then it's up to the interacting class how/if it deal withs it. 
So for example in my Light Controller block, it would just say something like 'Failed due to not having enough power' or 'Failed due to missing connection' and it could then be fed back to the user from the class that called for the interaction

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/InteractionResult.java
```diff
@@ -0,0 +1,61 @@
...
+     * Sets a custom result message for this interaction.
+     *
+     * @param resultMessage The message to be sent with the Result
+     */
+    public void setResultMessage(@Nullable String resultMessage) {
+        this.resultMessage = resultMessage;
+    }
+
+    /**
+     * Returns whether this result has a result message or not.
```
J3fftw1, line 45: Lets change this up (for some reason i cant do code suggestion but hey o

Lets change this to 
Returns wheter this result has a result message or is null.

This is code so null makes more sense then not

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/InteractionResult.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @return True if a result message is present
```
J3fftw1, line 47: Do we want True to be capital T here since 'true' is never 'True'

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/InteractionResult.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public boolean hasResultMessage() {
+        return this.resultMessage != null;
+    }
+
+    /**
+     * Returns the custom result message for this result.
```
J3fftw1, line 54: add or null

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/InteractionResult.java
```diff
@@ -0,0 +1,61 @@
...
+     *
+     * @return A String of the provided custom result message.
```
J3fftw1, line 56: again or null

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/ItemInteractionResult.java
```diff
@@ -0,0 +1,64 @@
...
+ * This class represents the result of an interaction on an {@link ExternallyInteractable} item.
+ */
+public class ItemInteractionResult extends InteractionResult {
+
+    private final @Nonnull Set<ItemStack> resultItems = new HashSet<>();
+
+    /**
+     * Creates a new InteractionResult.
+     *
+     * @param successful Whether the interaction was successful or not.
```
J3fftw1, line 21: or null again

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/ItemInteractionResult.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    public ItemInteractionResult(boolean successful) {
+        super(successful);
+    }
+
+    /**
+     * Creates a new InteractionResult.
+     *
+     * @param successful Whether the interaction was successful or not.
```
J3fftw1, line 30: or null again

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/InteractionResult.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    @ParametersAreNonnullByDefault
+    public InteractionResult(boolean successful) {
+        this.interactionSuccessful = successful;
+    }
+
+    /**
+     * Returns whether the interaction was successful or not.
+     *
+     * @return boolean denoting whether the interaction was successful or not.
```
J3fftw1, line 29: or null again

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/InteractionResult.java
Sefiraat: I won't put it here as that's the purpose of the annotation.

J3fftw1 (approved):

WalshyDev (approved):
(this was meant to be sent this morning, oops)

src/main/java/io/github/thebusybiscuit/slimefun4/core/attributes/interactions/ItemInteractionResult.java
```diff
@@ -0,0 +1,64 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * This class represents the result of an interaction on an {@link ExternallyInteractable} item.
+ */
+public class ItemInteractionResult extends InteractionResult {
```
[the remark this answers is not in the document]
Sefiraat: It was in just the one class previously with a note from yourself to abstract it out, is there a middle ground, if so I may need something more specific in direction other than not liking it? Happy to make changes if needed, and likely before it's merged in to save potential breaking changes down the line, but I'm not sure what the mid-step here would be?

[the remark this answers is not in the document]
WalshyDev: I'm not completely sure what the actual solution is. Maybe a Consumer which is called and then returns whatever, generic to decide what object/whatever? I don't know.

I just don't like the fact we'd have to constantly add more and more and more.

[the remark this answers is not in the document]
WalshyDev: I don't have the solution and that's why I'm fine with this for now. I can just feel us having to refactor later.

Sfiguz7 (approved):
Left some comments in Sefi's DMs, overall I share Walshy's opinion on all fronts.
As for the technical side, LGTM

PR #3794, opened 2023-03-12
Make Fire Staff not disenchantable by default

### LINKED ISSUE #3768: Elemental Staff - Fire should not be usable in disenchanters
### ❗ Checklist

### 📍 Description

this is a bug regarding Elemental Staff - Fire  

### 📑 Reproduction Steps

if the bug start when player put stacks

### 💡 Expected Behavior

i have suggestion add only 1 stick per time

### 📷 Screenshots / Videos

https://www.youtube.com/watch?v=A6NEFDFA2Qw

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

Dev - 1043

### 🧭 Other plugins

_No response_

Phoenix-Starlight:
## Description

To make the Elemental Staff - Fire not disenchantable by default.
## Proposed changes

Creation of a Slimefun item class and altering the item setup to use this class.
This won't work with existing `Items.yml` so existing servers need to edit the file to reflect this change.

Resolves #3768 
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/staves/FireStaff.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/staves/FireStaff.java
@@ -1,0 +1,29 @@
+package io.github.thebusybiscuit.slimefun4.implementation.items.magical.staves;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+
+import org.bukkit.entity.Player;
+import org.bukkit.inventory.ItemStack;
+
+import io.github.thebusybiscuit.slimefun4.api.items.ItemGroup;
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItemStack;
+import io.github.thebusybiscuit.slimefun4.api.recipes.RecipeType;
+import io.github.thebusybiscuit.slimefun4.core.handlers.ItemUseHandler;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+
+/**
+ * The {@link FireStaff} is a magical {@link SlimefunItem}, enchanted with Fire Aspect 5.
+ * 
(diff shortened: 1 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/magical/staves/FireStaff.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public class FireStaff extends SlimefunItem {
+
+    @ParametersAreNonnullByDefault
+    public FireStaff(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        super(itemGroup, item, recipeType, recipe);
+        disenchantable = false;
+    }
+
+}
```
TheBusyBiscuit, line 29: Could probably be generalized into `EnchantedStaff` or `EnchantedStick`

TheBusyBiscuit (changes requested):

TheBusyBiscuit (commented):
Sorry for making you change the name again.
What do you think about `PermanentlyEnchantedItem`?

I think that might be more indicative of the fact that it isn't just enchanted but cannot be disenchanted.

TheBusyBiscuit (approved):
LGTM

svr333 (approved):

Sfiguz7:
Should we maybe add other items to this? Like the infused elytra

Phoenix-Starlight:
Perhaps, in my view as long as enchantments do not exceed vanilla limits
they can be disenchanted, however I understand the view where it should be
by default not dis-enchantable. (Too easy method to get enchantments etc.)

Sfiguz7:
Eh I guess for the other stuff servers can still decide to change the `is_enchantable` flag in items.yml, the situation is clearly different for the fire staff since it keeps its "effects" even after being disenchanted (contrarily to armor tools and weapons).
I think this works just fine then!

PR #3795, opened 2023-03-13
Fix Disabled Items Shifting Multiblock Recipes

### LINKED ISSUE #3286: Disabling some Slimefun items shifted some multiblocks recipe
### ❗ Checklist

### 📍 Description

In this case I'll disable Stone Chunks from `items.yml` and try using Compressor

This also happened here: https://github.com/TheBusyBiscuit/ExoticGarden/issues/210

I think this is because when registering recipe it didn't check for the disabled items.

https://github.com/Slimefun/Slimefun4/blob/fc207b6a83866cca138e4a0b43d02e1e55a5c3b0/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/Compressor.java#L37-L47

I haven't tried anything else but looking at things I think I can make this work one Ore Grinder and Ore Crusher too..

### 📑 Reproduction Steps

1. Install Slimefun
2. Disable Stone Chunks from `items.yml`
3. Look at the guide, the recipe just got shifted
4. Try using the compressor just using 1 coal
5. It will output 8 coal

### 💡 Expected Behavior

The recipe should not be shifted.

### 📷 Screenshots / Videos

https://youtu.be/sgOPYilkLis

<details>
<summary>Side by side</summary>

Untitled

</details>

<details>

<summary>Normal Recipe</summary>

image

</details>

<details>
<summary>Shifted Recipe</summary>

image

</details>

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

```
> sf versions
[10:34:39 INFO]: This Server uses the following setup of Slimefun:
Paper git-Paper-279 (MC: 1.17.1)
Slimefun DEV - 974 (git 36c6d154)
Metrics-Module #28
Java 16

No Addons installed
```

### 🧭 Other plugins

_No response_

JustAHuman-xD:
## Description
If server owners disable specific items it can end up shifting multiblock recipes leading to either minimal or disasterous consequences. This PR aims to solve that.

## Proposed changes
Inside of MultiBlockMachine#load, instead of looping through each itemstack individually (input and output indiviudally), it now handles input and output at the same time. And checks if both are not disabled before adding both to the recipes. Previously it just went through stack at a time, so if an input was disabled the output would become the input and the next recipes input would be come the output. Now if either the input or output are disabled, neither item gets added which stops the shifting.

Resolves #3286

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -9,7 +9,10 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import com.google.common.base.Preconditions;
+
 import org.apache.commons.lang.Validate;
+
 import org.bukkit.Material;
 import org.bukkit.World;
 import org.bukkit.block.Block;
@@ -104,7 +107,8 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
     @Override
     public void load() {
         super.load();
-        
+
+        Preconditions.checkArgument(displayRecipes.size() % 2 == 0, "This MultiBlockMachine's display recipes were illegally modified!");
         for (int i = 0; i < displayRecipes.size(); i += 2) {
             ItemStack inputStack = displayRecipes.get(i);
             ItemStack outputStack = displayRecipes.size() >= i + 2 ? displayRecipes.get(i + 1) : null;
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -118,7 +118,6 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
 
             SlimefunItem inputItem = SlimefunItem.getByItem(inputStack);
             SlimefunItem outputItem = SlimefunItem.getByItem(outputStack);
-            
             if ((inputItem == null || !inputItem.isDisabled()) && (outputItem == null || !outputItem.isDisabled())) {
                 recipes.add(new ItemStack[] { inputStack });
                 recipes.add(new ItemStack[] { outputStack });
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -111,7 +111,13 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
         Preconditions.checkArgument(displayRecipes.size() % 2 == 0, "This MultiBlockMachine's display recipes were illegally modified!");
         for (int i = 0; i < displayRecipes.size(); i += 2) {
             ItemStack inputStack = displayRecipes.get(i);
-            ItemStack outputStack = displayRecipes.size() >= i + 2 ? displayRecipes.get(i + 1) : null;
+            ItemStack outputStack:
+            if (displayRecipes.size() >= i + 2) {
+                outputStack = displayRecipes.get(i + 1);
+            } else {
+                outputStack = null;
+            }
+
             SlimefunItem inputItem = SlimefunItem.getByItem(inputStack);
             SlimefunItem outputItem = SlimefunItem.getByItem(outputStack);
             
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -111,11 +111,9 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
         Preconditions.checkArgument(displayRecipes.size() % 2 == 0, "This MultiBlockMachine's display recipes were illegally modified!");
         for (int i = 0; i < displayRecipes.size(); i += 2) {
             ItemStack inputStack = displayRecipes.get(i);
-            ItemStack outputStack;
+            ItemStack outputStack = null;
             if (displayRecipes.size() >= i + 2) {
                 outputStack = displayRecipes.get(i + 1);
-            } else {
-                outputStack = null;
             }
 
             SlimefunItem inputItem = SlimefunItem.getByItem(inputStack);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
@@ -111,7 +111,7 @@ public abstract class MultiBlockMachine extends SlimefunItem implements NotPlace
         Preconditions.checkArgument(displayRecipes.size() % 2 == 0, "This MultiBlockMachine's display recipes were illegally modified!");
         for (int i = 0; i < displayRecipes.size(); i += 2) {
             ItemStack inputStack = displayRecipes.get(i);
-            ItemStack outputStack:
+            ItemStack outputStack;
             if (displayRecipes.size() >= i + 2) {
                 outputStack = displayRecipes.get(i + 1);
             } else {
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -104,12 +104,20 @@ public void postRegister() {
     @Override
     public void load() {
         super.load();
-
-        for (ItemStack recipeItem : displayRecipes) {
-            SlimefunItem item = SlimefunItem.getByItem(recipeItem);
-
-            if (item == null || !item.isDisabled()) {
-                recipes.add(new ItemStack[] { recipeItem });
```
TheBusyBiscuit, line 112: I hate this code I wrote so much.

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -104,12 +104,20 @@ public void postRegister() {
...
+        
+        for (ItemStack inputStack : displayRecipes) {
+            int index = displayRecipes.indexOf(inputStack);
```
TheBusyBiscuit, line 109: No need for this expensive `.indexOf` operation here, we are looping through this list anyway. So just create your own counter and increment it on every run.

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -104,12 +104,20 @@ public void postRegister() {
...
+            if (index % 2 != 0) {
+                continue;
+            }
```
TheBusyBiscuit, line 112: Might also be worth to just do `for (int i = 0; i < displayRecipes.size(); i +=2)` here to save all this stuff.

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -104,12 +104,16 @@ public void postRegister() {
...
     public void load() {
         super.load();
-
-        for (ItemStack recipeItem : displayRecipes) {
-            SlimefunItem item = SlimefunItem.getByItem(recipeItem);
-
-            if (item == null || !item.isDisabled()) {
-                recipes.add(new ItemStack[] { recipeItem });
+        
+        for (int i = 0; i < displayRecipes.size(); i += 2) {
```
TheBusyBiscuit, line 108: You should also check if the size of this list is a multiple of 2 before.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
JustAHuman-xD: why? (Sorry if that came out blatant just genuinely confused)

JustAHuman-xD:
Can you give some context as to that change that was suggested, I still don't understand the point of it

TheBusyBiscuit:
> Can you give some context as to that change that was suggested, I still don't understand the point of it

Sure, we should just make sure that the size of this list is divisible by two.
If someone - in whatever way - manipulates the list and adds a single entry, all recipes will not only be shifted but now it would also cause an index out of bounds exception because you are getting the `i + 1`th element.

So it would be in our best interest to catch that scenario, make sure the size is a multiple of two and if not, throw out a warning that the list was illegally modified.

Ideally, we would have an actual recipe system here instead of this damn array but it is what it is...

JustAHuman-xD:
So that won't ever error, or shift. The size check would prevent that, I've tested it. As for the I + 1, that's what the size ternary check is for. If you mean if a recipe was added with only one input, and nothing added for the output and that's what causes the shifting the yeah I see what you mean. So I could make it send a messag

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
JustAHuman-xD: Adding a case for if the list is not a multiple of 2.

Sfiguz7 (dismissed):
LGTM, as discussed in vc

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -105,11 +108,16 @@ public void postRegister() {
...
 
-        for (ItemStack recipeItem : displayRecipes) {
-            SlimefunItem item = SlimefunItem.getByItem(recipeItem);
-
-            if (item == null || !item.isDisabled()) {
-                recipes.add(new ItemStack[] { recipeItem });
+        Preconditions.checkArgument(displayRecipes.size() % 2 == 0, "This MultiBlockMachine's display recipes were illegally modified!");
+        for (int i = 0; i < displayRecipes.size(); i += 2) {
+            ItemStack inputStack = displayRecipes.get(i);
+            ItemStack outputStack = displayRecipes.size() >= i + 2 ? displayRecipes.get(i + 1) : null;
```
Sefiraat, line 114: Please don't use a ternary here :)

J3fftw1, line 114: Please dont use ternary here its against guidelines and its hard to understand here.
https://github.com/Slimefun/Slimefun4/blob/master/CONTRIBUTING.md#7-style-preferences

JustAHuman-xD:
Also was the double comment about the ternary necessary. 🥲 You guys are in a vc actively xd

J3fftw1:
> Also was the double comment about the ternary necessary. 🥲 You guys are in a vc actively xd

didnt know he was commenting.

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -105,11 +108,22 @@ public void postRegister() {
...
-            SlimefunItem item = SlimefunItem.getByItem(recipeItem);
+        Preconditions.checkArgument(displayRecipes.size() % 2 == 0, "This MultiBlockMachine's display recipes were illegally modified!");
+        for (int i = 0; i < displayRecipes.size(); i += 2) {
+            ItemStack inputStack = displayRecipes.get(i);
+            ItemStack outputStack;
+            if (displayRecipes.size() >= i + 2) {
+                outputStack = displayRecipes.get(i + 1);
+            } else {
+                outputStack = null;
+            }
```
J3fftw1, line 119:
```suggestion
            ItemStack outputStack = null;
            if (displayRecipes.size() >= i + 2) {
                outputStack = displayRecipes.get(i + 1);
            }
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -105,11 +108,19 @@ public void postRegister() {
     public void load() {
         super.load();
 
-        for (ItemStack recipeItem : displayRecipes) {
-            SlimefunItem item = SlimefunItem.getByItem(recipeItem);
+        Preconditions.checkArgument(displayRecipes.size() % 2 == 0, "This MultiBlockMachine's display recipes were illegally modified!");
+        for (int i = 0; i < displayRecipes.size(); i += 2) {
```
WalshyDev, line 112:
```suggestion

        for (int i = 0; i < displayRecipes.size(); i += 2) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/multiblocks/MultiBlockMachine.java
```diff
@@ -105,11 +108,19 @@ public void postRegister() {
...
+            ItemStack outputStack = null;
+            if (displayRecipes.size() >= i + 2) {
+                outputStack = displayRecipes.get(i + 1);
+            }
 
-            if (item == null || !item.isDisabled()) {
-                recipes.add(new ItemStack[] { recipeItem });
+            SlimefunItem inputItem = SlimefunItem.getByItem(inputStack);
+            SlimefunItem outputItem = SlimefunItem.getByItem(outputStack);
+            if ((inputItem == null || !inputItem.isDisabled()) && (outputItem == null || !outputItem.isDisabled())) {
```
WalshyDev, line 121:
```suggestion
            // If the input/output is not a Slimefun item or it's not disabled then it's valid.
            if ((inputItem == null || !inputItem.isDisabled()) && (outputItem == null || !outputItem.isDisabled())) {
```

WalshyDev (approved):

Sfiguz7 (approved):
Justin's fault if it breaks

PR #3798, opened 2023-03-15
Add new methods in BlockStorage and TickerTask for clearing all block info in a chunk

### LINKED ISSUE #3432: Performance impact when clearing blockdata from a whole chunk
### ❗ Checklist

### 📍 Description

When using SuperiorSkyblock, it Hooks into Slimefun to remove items on Island reset, this causes a large tps drop when resetting or disbanding an island.

### 📑 Reproduction Steps

Install SSB
Place down a large amount of Slimefun content
Attempt to disband / reset an Island

### 💡 Expected Behavior

Ideally we (the server) and @OmerBenGera would love if this could be more optimized. So we are reaching out to receive assistance with doing so.

Issue:
https://github.com/BG-Software-LLC/SuperiorSkyblock2/issues/850

Slimefun hook:
https://github.com/BG-Software-LLC/SuperiorSkyblock2/blob/003ed9c0e4264b925c98898ebb619f13d4beb46a/Hook_Slimefun/src/main/java/com/bgsoftware/superiorskyblock/hooks/support/SlimefunHook.java#L87

### 📷 Screenshots / Videos

https://i.imgur.com/Q0ZMh7P.png

### 📜 Server Log

https://pastebin.com/kYyu871z

N/A

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.18.x

### ⭐ Slimefun version

Dev 996

(Happens regardless of addons installed or not)

### 🧭 Other plugins

SuperiorSkyblock2

md5sha256:
## Description

Saw #3432 and wanted to implement changes based on TheBusyBiscuit's line of thinking

## Proposed changes

Adds two new methods in `BlockStorage` 
- `void clearAllBlockInfoAtChunk(Chunk chunk, boolean destroy)`
- `void clearAllBlockInfoAtChunk(World world, int chunkX, int chunkZ, boolean destroy)`

Adds two new methods in `TickerTask`
- `void queueDelete(Collection<Location> locations, boolean destroy)`
- `void queueDelete(Map<Location, Boolean> locations)`

Given that both `BlockStorage` and `TickerTask` are going to be rewritten anyway, I haven't added any more convenience methods that I otherwise could for bulk operations. 

In summary, the changes allow the caller to perform a bulk version of `clearBlockInfo` for locations in a specific chunk. I'm not sure how much this will reduce hitting locks within the `ConcurrentHashMap` implementation, however, I am sure this will prevent multiple table-resizing operations from occurring as the use of `Map#putAll` means the hashtable is only resized once if needed. 

I haven't performed any benchmarks just yet but the theory should be that it's faster to use `BlockStorage#clearAllBlockInfoAtChunk` than it would be to perform the same manually i.e: 
```java
public void clearAllBlockInfoAtChunk(Chunk chunk) {
     BlockStorage blockStorage = BlockStorage.getStorage(chunk.getWorld());
    if (blockStorage == null) {
        return;
    }
    int chunkX = chunk.getX();
    int chunkZ = chunk.getZ();
    Map<Location, Config> storage = blockStorage.getRawStorage();
    for (Location location : storage.keySet()) {
         if (location.getBlockX() >> 4 == chunkX && location.getBlockZ() >> 4 == chunkZ) {
              BlockStorage.clearBlockInfo(location);
         }
    }
}
```
I intend on attaching benchmarks as I'm not entirely sure if this will be faster. I've decided to make the PR just so I can have some feedback on implementing something which can perhaps reduce the impact of #3432

Resolves #3432

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -238,11 +238,14 @@ public class TickerTask implements Runnable {
 
     @ParametersAreNonnullByDefault
     public void queueDelete(Collection<Location> locations, boolean destroy) {
-        Map<Location, Boolean> toDelete = new HashMap<>(locations.size());
+        Validate.notNull(locations, "Locations must not be null");
+
+        Map<Location, Boolean> toDelete = new HashMap<>(locations.size(), 1.0F);
         for (Location location : locations) {
+            Validate.notNull(location, "Locations must not contain null locations");
             toDelete.put(location, destroy);
         }
-        queueDelete(toDelete);
+        deletionQueue.putAll(toDelete);
     }
 
     @ParametersAreNonnullByDefault
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -1,6 +1,8 @@
 package io.github.thebusybiscuit.slimefun4.implementation.tasks;
 
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Map;
@@ -231,6 +233,26 @@ public class TickerTask implements Runnable {
         Validate.notNull(l, "Location must not be null!");
 
         deletionQueue.put(l, destroy);
+    }
+
+
+    @ParametersAreNonnullByDefault
+    public void queueDelete(Collection<Location> locations, boolean destroy) {
+        Map<Location, Boolean> toDelete = new HashMap<>(locations.size());
+        for (Location location : locations) {
+            toDelete.put(location, destroy);
+        }
+        queueDelete(toDelete);
+    }
+
+    @ParametersAreNonnullByDefault
+    public void queueDelete(Map<Location, Boolean> locations) {
+        Validate.notNull(locations, "Locations must not be null");
+        for (Map.Entry<Location, Boolean> entry : locations.entrySet()) {
+            Validate.notNull(entry.getKey(), "Location in locations cannot be null");
+            Validate.notNull(entry.getValue(), "Boolean toDestroy in locations cannot be null");
+        }
+        deletionQueue.putAll(locations);
     }
 
     /**
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
@@ -6,6 +6,7 @@ import java.io.StringWriter;
 import java.nio.file.Files;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
@@ -20,6 +21,7 @@ import javax.annotation.ParametersAreNonnullByDefault;
 
 import org.apache.commons.lang.Validate;
 import org.bukkit.Bukkit;
+import org.bukkit.Chunk;
 import org.bukkit.Location;
 import org.bukkit.World;
 import org.bukkit.block.Block;
@@ -605,6 +607,24 @@ public class BlockStorage {
 
     public static void clearBlockInfo(Location l, boolean destroy) {
         Slimefun.getTickerTask().queueDelete(l, destroy);
+    }
+
+    public static void clearAllBlockInfoAtChunk(Chunk chunk, boolean destroy) {
+        clearAllBlockInfoAtChunk(chunk.getWorld(), chunk.getX(), chunk.getZ(), destroy);
+    }
+
+    public static void clearAllBlockInfoAtChunk(World world, int chunkX, int chunkZ, boolean destroy) {
+        BlockStorage blockStorage = getStorage(world);
+        if (blockStorage == null) {
+            return;
+        }
+        Map<Location, Boolean> toClear = new HashMap<>(blockStorage.storage.size());
+        for (Location location : blockStorage.storage.keySet()) {
+            if (location.getBlockX() >> 4 == chunkX && location.getBlockZ() >> 4 == chunkZ) {
+                toClear.put(location, destroy);
+            }
+        }
+        Slimefun.getTickerTask().queueDelete(toClear);
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
```diff
@@ -607,6 +608,25 @@ public static void clearBlockInfo(Location l, boolean destroy) {
...
+    public static void clearAllBlockInfoAtChunk(World world, int chunkX, int chunkZ, boolean destroy) {
+        BlockStorage blockStorage = getStorage(world);
+        if (blockStorage == null) {
+            return;
+        }
+        Map<Location, Boolean> toClear = new HashMap<>();
+        Map<Location, Config> storage = blockStorage.getRawStorage();
+        for (Location location : storage.keySet()) {
+            if (location.getBlockX() >> 4 == chunkX && location.getBlockZ() >> 4 == chunkZ) {
+                toClear.put(location, destroy);
```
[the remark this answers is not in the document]
md5sha256: The issue is that eventually it has to become a map, whether I do that in the clear method or in queueDelete is sort of what I'm testing/benchmarking as well. What that means is, if I use a Collection to start with, I'll have to make a secondary HashMap anyway which is an extra step that must be justifiable with performance gains. If I just use a Map and store the boolean x-times, I don't have to make another map, I can just call `deletionQueue#putAll`

New benchmarks indicate that using a LinkedList (and passing destroy at the end) is generally faster but that is not always the case, the TreeMap is second. HashMap and ArrayList are worse regardless of the scenario (most likely due to the amount of array/table resizing causing the GC to run more often). 
I'll update the PR after more benchmarking. 
I will also mention that JMH benchmarking indicates that even at 10000 elements, we are talking about an operation that is in the ballpark of 0.8ms-2.4ms (TreeMap being the best, ArrayList being the worst).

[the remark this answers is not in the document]
TheBusyBiscuit: Alright, I see

TheBusyBiscuit (approved):
Looks good from what I can tell

md5sha256:
Ok, so I've finalized the benchmark and it can be found here: https://github.com/md5sha256/BlockStorageTest. 

So now onto the results... 
image
So the first chart is the absolute time taken to what the future `BlockStorage#clearBlockInfo` would look like for every strategy. Error bars are present but because of how small the values are, you can only really discern the error bars for the x = 10k. The error for 10 elements is something like 5.7%. 

image
The second chart compares the how fast/slow each new strategy is compared to the old way. Positive values indicate it is slower and negative values indicate it was faster (negative is better).
```java
for (Location location : storage.keySet()) {
            if (location.getBlockX() >> 4 == chunkX && location.getBlockZ() >> 4 == chunkZ) {
               BlockStorage.clearBlockInfo(location);
            }
        }
```

The results suggest that it is slower to use a collection unless we know we will need to clear around 1000 entries. My analysis of why this is the case has to do with how the test is set up. It is very likely that due to the number of blocks and BlockStorage instances the test spawns, the garbage cleaner will need to run more often than if we had just run the test once on each VM. Using data structures further added pressure to the GC. Thus, it's very possible the GC is running during the test which leads me to think that the results are worse than they actually would be in the game. In fact, an in-game test from the previous comment indicated that using anything except for the for-loop would actually be faster. However, due to the complexities of testing in-game using the SF BlockStorage class, I have not used those results. 

Generally speaking, however, the results do match my expectations. For a small number of entries, it makes sense that it is faster to just use a for loop to queue the deletions in the ticker task and that for larger values, it makes more sense to store the results in a more optimal data structure before queueing the deletions. If my expectation held true, then one would expect that for 10000 elements, all the data structures should beat using a for-loop. However, I strongly believe the effect of the GC is more prevalent at larger element sizes which are influencing the tests in an unexpected manner. However, it is also unlikely that many block entries will be queued for deletion in one go anyway, so I would argue those results can be ignored in further discussion. 

The results indicate that using a HashMap is the most optimal at 100 elements and a TreeMap for 1000 elements; however, all HashMap, ArrayList, and LinkedList were faster than the for loop. The results also indicate that for a larger spread of values, HashMap's performance is more consistent. The HashMap performed significantly better than the TreeMap in the order of 10 and 100 elements and was only somewhat worse at 1000 -- but was importantly still better than the for-loop. 

Hence, based on the above benchmarking, I'd argue that the final implementation should use a HashMap. 

Finally, I'll also note that the time taken for these operations should really is small. The longest average time recorded for all operations was 1.17ms (HashMap at 10000 elements) which would be ~2.4% of a 50ms tick. That is to say, given the number of elements we are processing, the time spent is insignificant.

md5sha256:
Note: It is likely more performant to use `TickerTask#queueDelete(Map<Location, Boolean>)`. The reason is that an extra operation would need to be performed: converting the collection into a map. Therefore, if the caller had just stored the locations as a map in the first place, we could skip that step and only perform null checks. 

This PR adds an optimized `TickerTask#queueDelete(Collection<Location>, boolean)` for convenience purposes.

JustAHuman-xD (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
```diff
@@ -233,6 +235,29 @@ public void queueDelete(Location l, boolean destroy) {
...
+        Map<Location, Boolean> toDelete = new HashMap<>(locations.size(), 1.0F);
+        for (Location location : locations) {
+            Validate.notNull(location, "Locations must not contain null locations");
+            toDelete.put(location, destroy);
+        }
+        deletionQueue.putAll(toDelete);
+    }
+
+    @ParametersAreNonnullByDefault
+    public void queueDelete(Map<Location, Boolean> locations) {
```
WalshyDev, line 252: Why the boxed bool?

JustAHuman-xD: Is it possible to have maps with primitives?

JustAHuman-xD: image
Yeah like I thought you can't do that, or at least not with the standard `Map` and `boolean`

WalshyDev: ... I haven't wrote Java in so long :NotLikeThis:

It is but we'd need to use like fastutil

JustAHuman-xD: Gotcha

WalshyDev (approved):

PR #3802, opened 2023-03-18
fixed username not being found

J3fftw1:
## Description

Added st392 to ignored users.

## Proposed changes

Fixed st392's MC name

## Related Issues (if applicable)

fixed following error being thrown
[11:24:26] [Craft Scheduler Thread - 2/ERROR]: Exception while requesting skin: https://api.mojang.com/users/profiles/minecraft/st392
[11:24:26] [Craft Scheduler Thread - 0/WARN]: [Slimefun] Attempted to refresh skin cache, got this response: ExecutionException: java.io.FileNotFoundException: https://api.mojang.com/users/profiles/minecraft/st392

## Checklist

- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [x] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [x] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.16.* - 1.19.*).
- [x] I followed the existing code standards and didn't mess up the formatting.
- [ ] I did my best to add documentation to any public classes or methods I added.
- [ ] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

PR #3805, opened 2023-03-21
fix error on startup

J3fftw1:
## Description

Same issue as in #3802 other user
this user is how ever not findable on namemc
https://namemc.com/search?q=amarcais53

[15:21:46] [Craft Scheduler Thread - 4 - Slimefun/ERROR]: [dough: skins] Exception while requesting skin: https://api.mojang.com/users/profiles/minecraft/amarcais53
[15:21:46] [Craft Scheduler Thread - 2 - Slimefun/WARN]: [Slimefun] Attempted to refresh skin cache, got this response: ExecutionException: java.io.FileNotFoundException: https://api.mojang.com/users/profiles/minecraft/amarcais53

## Proposed changes

ignored amarcais53

## Checklist


TheBusyBiscuit:
I would prefer the actual issue to be fixed.
This PR is not a fix but rather just a lazy patch-up.

Constantly adding people to the blacklist is no long-term solution.
We should rather fix the API to handle the new response layout by Mojang's server.

PR #3806, opened 2023-03-22
[API] Add ShouldSave Parameter to BlockMenuPreset and BlockMenus

JustAHuman-xD:
## Description
Many Slimefun Blocks create menus to only convey information to the player, and they do not need to be saved to a file. This will help save time when saving on shutdown to help prevent the issues we see with BlockStorage taking too long and causing reverts, as well as the normal saving to file overhead no longer being a factor.

## Proposed changes
This PR adds a "shouldSave" parameter to BlockMenuPreset, thus to BlockMenu and UniversalBlockMenu.
Changes methods in BlockMenu and UniversalBlockMenu to only save if "shouldSave" is true

I don't believe there are any issues

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
@@ -22,17 +22,13 @@ public class BlockMenu extends DirtyChestMenu {
         return l.getWorld().getName() + ';' + l.getBlockX() + ';' + l.getBlockY() + ';' + l.getBlockZ();
     }
     
-    public BlockMenu(BlockMenuPreset preset, Location l, boolean display) {
+    public BlockMenu(BlockMenuPreset preset, Location l) {
         super(preset);
         this.location = l;
-        this.display = display;
+        this.display = preset.isDisplay();
         
         preset.clone(this);
         this.getContents();
-    }
-    
-    public BlockMenu(BlockMenuPreset preset, Location l) {
-        this(preset, l, false);
     }
 
     public BlockMenu(BlockMenuPreset preset, Location l, Config cfg) {
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
@@ -32,12 +32,7 @@ public class BlockMenu extends DirtyChestMenu {
     }
     
     public BlockMenu(BlockMenuPreset preset, Location l) {
-        super(preset);
-        this.location = l;
-        this.display = false;
-
-        preset.clone(this);
-        this.getContents();
+        this(preset, l, false);
     }
 
     public BlockMenu(BlockMenuPreset preset, Location l, Config cfg) {
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
@@ -14,9 +14,9 @@ import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 
 // This class will be deprecated, relocated and rewritten in a future version.
 public class BlockMenu extends DirtyChestMenu {
-
+    
+    private final boolean display;
     private Location location;
-    private boolean display;
 
     private static String serializeLocation(Location l) {
         return l.getWorld().getName() + ';' + l.getBlockX() + ';' + l.getBlockY() + ';' + l.getBlockZ();
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
@@ -15,17 +15,17 @@ import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 // This class will be deprecated, relocated and rewritten in a future version.
 public class BlockMenu extends DirtyChestMenu {
     
-    private final boolean display;
+    private final boolean shouldSave;
     private Location location;
 
     private static String serializeLocation(Location l) {
         return l.getWorld().getName() + ';' + l.getBlockX() + ';' + l.getBlockY() + ';' + l.getBlockZ();
     }
-    
+
     public BlockMenu(BlockMenuPreset preset, Location l) {
         super(preset);
         this.location = l;
-        this.display = preset.isDisplay();
+        this.shouldSave = preset.willSave();
         
         preset.clone(this);
         this.getContents();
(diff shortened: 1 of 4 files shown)


src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
```diff
@@ -90,6 +94,10 @@ public Block getBlock() {
     public Location getLocation() {
         return location;
     }
+    
+    public boolean isDisplay() {
+        return display;
```
J3fftw1, line 99: Can i suggest changing this do hasDisplay()

src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenu.java
```diff
@@ -14,24 +14,27 @@
...
-
+    
+    private final boolean display;
     private Location location;
 
     private static String serializeLocation(Location l) {
         return l.getWorld().getName() + ';' + l.getBlockX() + ';' + l.getBlockY() + ';' + l.getBlockZ();
     }
-
+
```
J3fftw1, line 24:
```suggestion

```

src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenuPreset.java
```diff
@@ -30,24 +30,30 @@ public abstract class BlockMenuPreset extends ChestMenu {
 
     // -1 means "automatically update according to the contents"
     private int size = -1;
-
+
```
J3fftw1, line 33:
```suggestion

```

src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenuPreset.java
```diff
@@ -30,24 +30,30 @@ public abstract class BlockMenuPreset extends ChestMenu {
...
     }
 
     protected BlockMenuPreset(@Nonnull String id, @Nonnull String title, boolean universal) {
+        this(id, title, universal, false);
+    }
+    
+    protected BlockMenuPreset(@Nonnull String id, @Nonnull String title, boolean universal, boolean display) {
         super(title);
-
+
```
J3fftw1, line 48:
```suggestion

```

src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenuPreset.java
```diff
@@ -30,24 +30,30 @@ public abstract class BlockMenuPreset extends ChestMenu {
...
+        this(id, title, universal, false);
+    }
+    
+    protected BlockMenuPreset(@Nonnull String id, @Nonnull String title, boolean universal, boolean display) {
         super(title);
-
+    
         Validate.notNull(id, "You need to specify an id!");
-
+
```
J3fftw1, line 50:
```suggestion

```

src/main/java/me/mrCookieSlime/Slimefun/api/inventory/BlockMenuPreset.java
```diff
@@ -30,24 +30,30 @@ public abstract class BlockMenuPreset extends ChestMenu {
...
         Validate.notNull(id, "You need to specify an id!");
-
+    
         this.id = id;
         this.inventoryTitle = title;
         this.universal = universal;
+        this.display = display;
         init();
-
+
```
J3fftw1, line 56:
```suggestion

```

J3fftw1 (changes requested):

Sefiraat:
Has this been tested on a block set to display? Because from what I am reading with the changes here, it will stop blocks marked as display from opening their inventory on Right Click, and if that's the intent then the resolution is to just not have a BlockMenu and/or Preset?

This will stop the `hasInventory()` check functioning correctly which would likely mean a player would not be thrown out of the inventory when the block is broken etc. leaving them with an unmanaged inventory they may be able to take items out of.

At the very least we need to be able to persist inventories whilst they are loaded in the world which means means they will be in memory and thus not save any, so having them in the Map would both resolve the issues above.

The goal to speed up saving is a good one, but this can be then condensed to a more refined parameter (shouldSave?) that just skips the blockmenu being saved by checking the blockmenu when saved is called and simply not saving?

JustAHuman-xD:
> Has this been tested on a block set to display? Because from what I am reading with the changes here, it will stop blocks marked as display from opening their inventory on Right Click, and if that's the intent then the resolution is to just not have a BlockMenu and/or Preset?
> 
> This will stop the `hasInventory()` check functioning correctly which would likely mean a player would not be thrown out of the inventory when the block is broken etc. leaving them with an unmanaged inventory they may be able to take items out of.
> 
> At the very least we need to be able to persist inventories whilst they are loaded in the world which means means they will be in memory and thus not save any, so having them in the Map would both resolve the issues above.
> 
> The goal to speed up saving is a good one, but this can be then condensed to a more refined parameter (shouldSave?) that just skips the blockmenu being saved by checking the blockmenu when saved is called and simply not saving?

I had not had time to do any testing yet but wanted to open the PR that way if there were any obvious issues such as the ones you brought up I would have been able to change accordingly as I now have.

The hasInventory method could easily be updated to be fixed however the closing the inventory on block break I don't believe is which would be the glaring problem with not putting them into the map, plus like you said it is already be in memory.

I have now switched the PR to just be a "shouldSave" parameter such as you suggested and will be doing the testing tonight 👍

TheBusyBiscuit (changes requested):
I really don't see the point of this addition.
The `BlockMenu` is just an extension of `ChestMenu` with the added functionality to save data based on location.
Why not just use a `ChestMenu` instead of making its entire functionality optional like many blocks in Slimefun do as well?

JustAHuman-xD:
> I really don't see the point of this addition. The `BlockMenu` is just an extension of `ChestMenu` with the added functionality to save data based on location. Why not just use a `ChestMenu` instead of making its entire functionality optional like many blocks in Slimefun do as well?

To put it simply, for convenience and QOL, it's easier to have a single format of doing things, when you are working with a machine that needs a saved inventory you can give it a saved inventory, if not then set saving to false. VS if you need an inventory that saves use this with this already very convenient formatting, if not use this: with not convenient formatting. It just makes it simpler for developers to develop. You said it has the added functionality of saving data based on location, but It's also just holding data, based on location. If you want to do chest menu's based on location you would have to setup your own system completely separate from Slimefun, unless you wanted to create a new ChestMenu every time they interacted with the block (such as you do in the current code for the gps control panel) which has its own added overhead and complication.

JustAHuman-xD:
> I really don't see the point of this addition. The `BlockMenu` is just an extension of `ChestMenu` with the added functionality to save data based on location. Why not just use a `ChestMenu` instead of making its entire functionality optional like many blocks in Slimefun do as well?

It is not making the entire block meu's functionality optional as you are still utilizing its per block functionality. The only difference is that it does not save to a file, everything else is still in affect.

JustAHuman-xD:
This'll be superceded by a new PR hopefully soon

PR #3822, opened 2023-04-28
fix: uncorrect exp spending on unlocking research

kingdom84521:
## Description

The amount of experience spending on unlocking researches is incorrect, since this process is done by Player.setLevel only.

## Proposed changes

1. Add some formula utils for calculating the proper value for unlocking
2. Refactor the original setLevel to new utils

## Checklist

- [v] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [v] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.16.* - 1.19.*).
- [v] I followed the existing code standards and didn't mess up the formatting.
- [v] I did my best to add documentation to any public classes or methods I added.
- [v] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
@@ -5,6 +5,7 @@ import java.util.function.Consumer;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.utils.ExperienceUtils;
 import org.bukkit.GameMode;
 import org.bukkit.entity.Player;
 import org.bukkit.inventory.ItemStack;
@@ -70,7 +71,11 @@ public interface SlimefunGuideImplementation {
         if (p.getGameMode() == GameMode.CREATIVE && Slimefun.getRegistry().isFreeCreativeResearchingEnabled()) {
             research.unlock(p, true, callback);
         } else {
-            p.setLevel(p.getLevel() - research.getCost());
+            float currentExp = ExperienceUtils.getPlayerCurrentExp(p.getLevel(), p.getExp());
+            float requiredExp = ExperienceUtils.convertLevelToFloatExp(research.getCost());
+            Number[] result = ExperienceUtils.convertExpToLevelAndProgress(currentExp - requiredExp);
+            p.setLevel((int) result[0]);
+            p.setExp((float) result[1]);
 
             boolean skipLearningAnimation = Slimefun.getRegistry().isLearningAnimationDisabled() || !SlimefunGuideSettings.hasLearningAnimationEnabled(p);
             research.unlock(p, skipLearningAnimation, callback);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
@@ -1,0 +1,71 @@
+package io.github.thebusybiscuit.slimefun4.utils;
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+
+/**
+ * Some helper methods for dealing with experience converting.
+ *
+ * @author kingdom84521
+ *
+ * @see io.github.thebusybiscuit.slimefun4.core.guide.SlimefunGuideImplementation
+ * @see <a href="https://minecraft.fandom.com/wiki/Experience"></>
+ *
+ */
+public final class ExperienceUtils {
+
+    /**
+     * Helper method to calculate the level to exact xp
+     *
+     * @param level
+     *            The level to convert ( in type of {@link Integer} )
+     *
+     * @return xp in type {@link Float}
+     */
+
+    public static @Nonnull float convertLevelToFloatExp(@Nonnull int level) {
+        if (level > 31) {
+            return (float) (4.5 * level * level - 162.5 * level + 2220);
+        } else if (level > 16) {
+            return (float) (2.5 * level * level - 40.5 * level + 360);
+        } else {
+            return level * level + 6 * level;
+        }
+    }
+
+    private static @Nonnull float getExpToNextLevel(@Nonnull int currentLevel) {
+        if (currentLevel > 30) {
+            return (float) 9 * currentLevel - 158;
+        } else if (currentLevel > 15) {
+            return (float) 5 * currentLevel - 38;
+        } else {
+            return (float) 2 * currentLevel + 7;
+        }
+    }
+
+    public static @Nonnull float getPlayerCurrentExp(@Nonnull int level, @Nonnull float progress) {
+        return convertLevelToFloatExp(level) + (getExpToNextLevel(level) * progress);
+    }
+
+    private static @Nonnull float convertExpToLevel(@Nonnull float exp) {
+        if (exp >= 1508) {
+            return (float) (325 + Math.sqrt(72 * exp - 54215)) / 18;
+        } else if (exp >= 353) {
+            return (float) (81 + Math.sqrt(40 * exp - 7839)) / 10;
+        } else {
+            return (float) Math.sqrt(exp + 9) - 3;
+        }
+    }
+
+    public static @Nonnull Number[] convertExpToLevelAndProgress(@Nonnull float exp) {
+        int level = (int) Math.floor(convertExpToLevel(exp));
+        float progress = convertExpToLevel(exp) - level;
+
+        Number[] levelAndProgress = new Number[2];
+
+        levelAndProgress[0] = level;
+        levelAndProgress[1] = progress;
+
+        return levelAndProgress;
+    }
+}
--- a/src/test/java/io/github/thebusybiscuit/slimefun4/utils/TestExperienceUtils.java
+++ b/src/test/java/io/github/thebusybiscuit/slimefun4/utils/TestExperienceUtils.java
@@ -1,0 +1,20 @@
+package io.github.thebusybiscuit.slimefun4.utils;
+
+import org.junit.jupiter.api.Assertions;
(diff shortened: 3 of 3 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
```diff
@@ -5,6 +5,7 @@
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.utils.ExperienceUtils;
```
ybw0014, line 8: import order

ybw0014 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
kingdom84521: fix

J3fftw1:
I feel like this should be a config option this has been this way for many years it would be weird to suddenly change this. Some addons would now also be easier to be unlocked for example I have uu matter set to 100 lvls taking in count that it’s actually levels

kingdom84521:
That's a point, I'll try to add a option for that.

Sefiraat:
The config option should be under the `research` heading. And I'm not a big fan of the name? `level-cost-as-exp`? Not sure that's much better xD

kingdom84521:
Actually I put it in research at first, but looks like other options in research seems to be like some options that can alter by player on their own? That get me feel a little bit weird, so I move it to option at last.

Also, I'm a pretty bad namer, so if anyone thought a better name, feel free to tell me, I can fix it right away.

SchnTgaiSpock:
`use-legacy-xp-spending`?

kingdom84521:
hmm, I will prefer "vanilla" than "legacy", cause vanilla minecraft is still using this kind of formula on enchanting.

JustAHuman-xD:
I agree with schnn I think it should be legacy, we are talking about slimefun, not minecraft

Sefiraat (changes requested):
Not legacy, that implies we are changing the default which we are not. 

Go with what I have put for the name for now but let’s hope something better comes up. And it still belongs in research category regardless.

src/test/java/io/github/thebusybiscuit/slimefun4/utils/TestExperienceUtils.java
```diff
@@ -0,0 +1,20 @@
+package io.github.thebusybiscuit.slimefun4.utils;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+public class TestExperienceUtils {
```
TheBusyBiscuit, line 7:
```suggestion
class TestExperienceUtils {
```

src/test/java/io/github/thebusybiscuit/slimefun4/utils/TestExperienceUtils.java
```diff
@@ -0,0 +1,20 @@
...
+    void testConvertLevelToXp() {
+        // case: level >= 32
+        Assertions.assertEquals(5345F, ExperienceUtils.convertLevelToFloatExp(50));
+
+        // case: level >= 17 && level <= 31
+        Assertions.assertEquals(910F, ExperienceUtils.convertLevelToFloatExp(25));
+
+        // case: level <= 16
+        Assertions.assertEquals(160F, ExperienceUtils.convertLevelToFloatExp(10));
+    }
```
TheBusyBiscuit, line 19: Make sure to also always test edge cases like -1, 0, MAX_VALUE

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
```diff
@@ -0,0 +1,71 @@
...
+        if (exp >= 1508) {
+            return (float) (325 + Math.sqrt(72 * exp - 54215)) / 18;
+        } else if (exp >= 353) {
+            return (float) (81 + Math.sqrt(40 * exp - 7839)) / 10;
+        } else {
+            return (float) Math.sqrt(exp + 9) - 3;
+        }
+    }
+
+    public static Number[] convertExpToLevelAndProgress(float exp) {
```
TheBusyBiscuit, line 60: I think a proper record/struct or small class would be much better than a `Number` array.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Some helper methods for dealing with experience converting.
+ *
+ * @author kingdom84521
+ *
+ * @see io.github.thebusybiscuit.slimefun4.core.guide.SlimefunGuideImplementation
+ * @see <a href="https://minecraft.fandom.com/wiki/Experience"></a>
+ *
+ */
+public final class ExperienceUtils {
```
TheBusyBiscuit, line 15: Please also add javadocs to all the public methods in this class, it's a utility class after all.

src/main/resources/config.yml
```diff
@@ -29,6 +29,7 @@ researches:
   free-in-creative-mode: true
   enable-fireworks: true
   disable-learning-animation: false
+  convert-level-cost-to-exp: false
```
TheBusyBiscuit, line 32: Please rename this.

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
```diff
@@ -70,7 +71,15 @@ default void unlockItem(Player p, SlimefunItem sfitem, Consumer<Player> callback
         if (p.getGameMode() == GameMode.CREATIVE && Slimefun.getRegistry().isFreeCreativeResearchingEnabled()) {
             research.unlock(p, true, callback);
         } else {
-            p.setLevel(p.getLevel() - research.getCost());
+            if (Slimefun.getCfg().getBoolean("researches.convert-level-cost-to-exp")) {
```
TheBusyBiscuit, line 74: We shouldn't read this from the config everytime a research is unlocked for performance reasons. Look into how it's handled with over config settings like the learning animation for example.

TheBusyBiscuit (changes requested):
Not sure what to think of this.

The XP behaviour has been consistently unchanged for almost TEN years. Changing it now will just result in a ton of bug tickets for weeks to come. Good that this is disabled by default, so if there really is a demand for this setting then I am willing to merge it

Sefiraat:
From my experience on a few servers now it is generally considered to be unintuitive to players coming into slimefun blind but once known it actually becomes a mechanic that rewards forward thinking and/or the use of mechanics offered in addons/3rd party plugins to make the most of xp. 

While I prefer the current system in the main, I'm all for the change so long as it's default-off like this.

src/test/java/io/github/thebusybiscuit/slimefun4/utils/TestExperienceUtils.java
kingdom84521: I...don't really know how to produce a test case for MAX_VALUE, but I add the other ones' case

src/test/java/io/github/thebusybiscuit/slimefun4/utils/TestExperienceUtils.java
kingdom84521: fix

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
kingdom84521: fix

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
kingdom84521: fix

src/main/resources/config.yml
kingdom84521: fix

src/main/java/io/github/thebusybiscuit/slimefun4/core/guide/SlimefunGuideImplementation.java
kingdom84521: fix

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public static class LevelAndProgress {
```
J3fftw1, line 16: We shouldnt use inner classes

src/main/java/io/github/thebusybiscuit/slimefun4/utils/ExperienceUtils.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    private static float getExpToNextLevel(int currentLevel) {
+        if (currentLevel > 30) {
+            return (float) 9 * currentLevel - 158;
+        } else if (currentLevel > 15) {
+            return (float) 5 * currentLevel - 38;
+        } else {
+            return (float) 2 * currentLevel + 7;
+        }
```
J3fftw1, line 54: not a fan of these methods. Isnt there a better way to do this?
I dont think the calculation is correct to be honest.

J3fftw1 (changes requested):
Sorry for the late review.

kingdom84521:
Sorry for the lack of focusing on this PR, my tasks were finally done and dusted, it's been a nightmare last year :(

PR #3826, opened 2023-05-07
newInstance hook for createPreset

test137E29B:
## Description

createPreset is a very useful function when writing addons, it allows us to not worry about re-writing the canOpen method that handles block permissions. The only downside for common use is that you can't bind menu click events easily since there's no support for newInstance without actually creating your own BlockMenuPreset instance directly. The problem with that is you must then overwrite canOpen yourself too - which is dangerous as an Addon developer. There are reasons you might want  to do that still (such as private machines only the placer can open), but for generic use it feels quite dangerous since responding to inventory clicks is quite a common use when you add custom functionality like collecting a non-item resource manually from machines (such as XP, or maybe money)

There is probably a much better way of doing this, it's more a "proof of concept" so I didn't worry too much about formatting being messed up or fully testing this.

## Proposed changes

An additional override for createPreset that allows passing a function to newInstance directly, thus negating the downsides for most common uses of createPreset.

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
@@ -20,8 +20,7 @@ import me.mrCookieSlime.Slimefun.api.item_transport.ItemTransportFlow;
 
 /**
  * 
- * @deprecated This interface is not designed to be used by addons. The entire
- *             inventory system will be replaced
+ * @deprecated This interface is not designed to be used by addons. The entire inventory system will be replaced
  *             eventually.
  *
  */
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
@@ -51,9 +51,7 @@ public interface InventoryBlock {
         createPreset(item, title, null, setup);
     }
 
-    default void createPreset(SlimefunItem item, String title,
-            @Nullable BiFunction<@Nonnull BlockMenu, @Nonnull Block, Void> onNewInstance,
-            Consumer<BlockMenuPreset> setup) {
+    default void createPreset(SlimefunItem item, String title, @Nullable BiFunction<@Nonnull BlockMenu, @Nonnull Block, Void> onNewInstance, Consumer<BlockMenuPreset> setup) {
         new BlockMenuPreset(item.getId(), title) {
 
             @Override
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
@@ -75,8 +75,7 @@ public interface InventoryBlock {
                 if (p.hasPermission("slimefun.inventory.bypass")) {
                     return true;
                 } else {
-                    return item.canUse(p, false) && Slimefun.getProtectionManager().hasPermission(p, b.getLocation(),
-                            Interaction.INTERACT_BLOCK);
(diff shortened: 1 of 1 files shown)


src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
```diff
@@ -43,6 +48,12 @@ default void createPreset(SlimefunItem item, Consumer<BlockMenuPreset> setup) {
     }
 
     default void createPreset(SlimefunItem item, String title, Consumer<BlockMenuPreset> setup) {
+        createPreset(item, title, null, setup);
+    }
+
+    default void createPreset(SlimefunItem item, String title,
+            @Nullable BiFunction<@Nonnull BlockMenu, @Nonnull Block, Void> onNewInstance,
+            Consumer<BlockMenuPreset> setup) {
```
J3fftw1, line 56:
```suggestion
    default void createPreset(SlimefunItem item, String title, @Nullable BiFunction<@Nonnull BlockMenu, @Nonnull Block, Void> onNewInstance, Consumer<BlockMenuPreset> setup) {
```

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
```diff
@@ -64,7 +75,15 @@ public boolean canOpen(Block b, Player p) {
                 if (p.hasPermission("slimefun.inventory.bypass")) {
                     return true;
                 } else {
-                    return item.canUse(p, false) && Slimefun.getProtectionManager().hasPermission(p, b.getLocation(), Interaction.INTERACT_BLOCK);
+                    return item.canUse(p, false) && Slimefun.getProtectionManager().hasPermission(p, b.getLocation(),
+                            Interaction.INTERACT_BLOCK);
```
J3fftw1, line 79:
```suggestion
                    return item.canUse(p, false) && Slimefun.getProtectionManager().hasPermission(p, b.getLocation(), Interaction.INTERACT_BLOCK);
```

Sefiraat:
The only issue with this is it is deprecated api and addons shouldn’t ever be using it in favour of creating your own presets and/or having your own factory method similar to this. That being said it’s been deprecated but still used for long enough that I wont close it off-the-cuff - though I’m going to assume cookie will not be wanting this class updated in any manner.

test137E29B:
> That being said it’s been deprecated but still used for long enough that I wont close it off-the-cuff - though I’m going to assum

The problem with creating your own is that you are forced to Overwrite a permission check, which means if Slimefun changes it in future to f ix a bug (that specific method, not one that gets called inside of it), then every single plugin also has to re-write their `canOpen` method for every preset too - even if they're just having default usage of it. I think writing your own Preset or abstraction is a good idea, except for forcing every developer to duplicate the Permissions logic which is quite dangerous imo.

Unfortunately I feel my Java knowledge isn't fresh enough around that specific area to really make those kind of changes, this stuff is much easier in TypeScript 😅

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/interfaces/InventoryBlock.java
```diff
@@ -1,22 +1,27 @@
...
-
+import me.mrCookieSlime.Slimefun.api.inventory.BlockMenu;
 import me.mrCookieSlime.Slimefun.api.inventory.BlockMenuPreset;
 import me.mrCookieSlime.Slimefun.api.item_transport.ItemTransportFlow;
 
 /**
  * 
- * @deprecated This interface is not designed to be used by addons. The entire inventory system will be replaced
+ * @deprecated This interface is not designed to be used by addons. The entire
+ *             inventory system will be replaced
```
test137E29B, line 24:
```suggestion
 * @deprecated This interface is not designed to be used by addons. The entire inventory system will be replaced
```

Sefiraat:
@test137E29B You still planning on working on this? The last update failed the build and will need resolving. Otherwise I can close as stale.

PR #3827, opened 2023-05-08
Fix disabled items preventing research completion

### LINKED ISSUE #3782: sf stats counting disabled items in the list of unlocked items, making us unable to achieve 100%
### ❗ Checklist

### 📍 Description

sf stats has a place where it shows total researches, and its counting the disabled ones as  a part of it

### 📑 Reproduction Steps

disable a item in sf and then unlock everything else, then check sf stats but you wont have 100% researches

### 💡 Expected Behavior

sf stats should not include disabled items into calculating % of  research done by player

### 📷 Screenshots / Videos

image
everything has been unlocked but still due to disabled items player is unable to reach 100% 

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

Sf version 1043 

### 🧭 Other plugins

this bug is not related to a plugin.

test137E29B:
## Description

When a SlimefunItem bound to a Research is disabled, the Research is not disabled. This means that the Research becomes locked permanently and the user can never achieve 100% Research unlock completion. If a SlimefunItem is globally disabled, it should not be bindable to Research since it's never usable. If a Research has no bound SlimefunItems (because they're all globall disabled), the research should act as not enabled also. Disabled researches should be filtered out too from the Progress stat.

## Proposed changes

Prevent disabled items being added to Research, filter out Researches with no Items bound when requested. This could also be triggered some other place that's more efficient than every time research is requested. Please suggest another place to perform this action if a better place exists.

Resolves #3782 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
@@ -649,6 +649,11 @@ public class SlimefunItem implements Placeable {
      *            The new {@link Research} for this {@link SlimefunItem}, or null
      */
     public void setResearch(@Nullable Research research) {
+        if (this.isDisabled()) {
+            // Disabled items should not be bindable to research
+            return;
+        }
+
         if (this.research != null) {
             this.research.getAffectedItems().remove(this);
         }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
@@ -337,8 +337,7 @@ public class PlayerProfile {
     public void sendStats(@Nonnull CommandSender sender) {
         Set<Research> unlockedResearches = getResearches();
         int levels = unlockedResearches.stream().mapToInt(Research::getCost).sum();
-        int allResearches = Slimefun.getRegistry().getResearches().stream()
-            .filter(research -> research.isEnabled()).collect(Collectors.toList()).size();
+        int allResearches = Slimefun.getRegistry().getResearches().size();
 
         float progress = Math.round(((unlockedResearches.size() * 100.0F) / allResearches) * 100.0F) / 100.0F;
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
@@ -205,12 +205,12 @@ public class PlayerProfile {
     }
 
     /**
-     * This method returns whether this {@link Player} has unlocked all {@link Research Researches}.
+     * This method returns whether this {@link Player} has unlocked all enabled {@link Research Researches}.
      * 
-     * @return Whether they unlocked every {@link Research}
+     * @return Whether they unlocked every enabled {@link Research}
      */
     public boolean hasUnlockedEverything() {
-        for (Research research : Slimefun.getRegistry().getResearches()) {
+        for (Research research : Slimefun.getRegistry().getEnabledResearches()) {
             // If there is a single Research not unlocked: They haven't unlocked everything.
             if (!hasUnlocked(research)) {
                 return false;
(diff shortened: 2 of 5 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/researches/Research.java
```diff
@@ -84,12 +84,13 @@ public Research(@Nonnull NamespacedKey key, int id, @Nonnull String defaultName,
     /**
      * This method returns whether this {@link Research} is enabled.
      * {@code false} can mean that this particular {@link Research} was disabled or that
-     * researches altogether have been disabled.
+     * researches altogether have been disabled, or that this {@link Research} as no bound
```
test137E29B, line 87:
```suggestion
     * researches altogether have been disabled, or that this {@link Research} has no bound
```

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
```diff
@@ -12,6 +12,7 @@
 import java.util.UUID;
 import java.util.function.Consumer;
 import java.util.logging.Level;
```
J3fftw1, line 14: Unused import

J3fftw1 (changes requested):

JustAHuman-xD:
For an example of this fix see: https://github.com/JustAHuman-xD/Slimefun4/commit/180731553df56eba65f3126ccc8d644bed283502

JustAHuman-xD:
For an example of this fix see: https://github.com/JustAHuman-xD/Slimefun4/commit/180731553df56eba65f3126ccc8d644bed283502

test137E29B:
> For an example of this fix see: https://github.com/JustAHuman-xD/Slimefun4/commit/180731553df56eba65f3126ccc8d644bed283502

But if you bind 2x researches to the same item, it unbinds the item from one of the researches (the first), thus leaving a research with NO items. This research is then NOT unlockable and again causes the same problem where you can't 100% research.

Researches with no items should be treated as not enabled because they are incomplete and cannot work correctly. This was conciously changed globally to prevent anywhere research is used from taking into account any research that is not set up correctly.

JustAHuman-xD:
Idk it still feels like an iffy change but others can give their opinion on it. I think that way because people could make dummy/marker researches with no items in it intentionally, or disabled items intentionally tho a reason for that I can't think of off the top of my head. And the change you made to getResearches is definitely an iffy change. I would suggest instead adding a method called getEnabledResearches. As getResearches implies all researches not just enabled ones.

JustAHuman-xD:
TLDR changing the behavior of old and crucial API is always scary

src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
```diff
@@ -200,7 +201,9 @@ public List<SlimefunItem> getEnabledSlimefunItems() {
      */
     @Nonnull
     public List<Research> getResearches() {
```
Sefiraat, line 203: When requesting a Set of all researches, one would expect to get the whole list, regardless of the researches current state.

If you want a list of enabled researches only, then a new method to return this should be made.

test137E29B: Added a new method for this, should have done it this way to begin with - also fixed other usages and added specific placeholder for total researches enabled as `researches_total_researches_enabled`

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
```diff
@@ -649,6 +649,11 @@ private void checkForDeprecations(@Nullable Class<?> c) {
      *            The new {@link Research} for this {@link SlimefunItem}, or null
      */
     public void setResearch(@Nullable Research research) {
+        if (this.isDisabled()) {
```
test137E29B, line 652: Should we still allow disabled items to be bound to researches?

src/main/java/io/github/thebusybiscuit/slimefun4/api/researches/Research.java
```diff
@@ -84,12 +84,13 @@ public Research(@Nonnull NamespacedKey key, int id, @Nonnull String defaultName,
...
      * {@code false} can mean that this particular {@link Research} was disabled or that
-     * researches altogether have been disabled.
+     * researches altogether have been disabled, or that this {@link Research} has no bound
+     * {@link SlimefunItem} and is therefore not-unlockable.
      * 
      * @return Whether this {@link Research} is enabled or not
      */
     public boolean isEnabled() {
-        return Slimefun.getRegistry().isResearchingEnabled() && enabled;
+        return Slimefun.getRegistry().isResearchingEnabled() && enabled && items.size() > 0;
```
test137E29B, line 93:
```suggestion
        return Slimefun.getRegistry().isResearchingEnabled() && enabled;
```

test137E29B:
**NB** adds new Placeholder `researches_total_researches_enabled`

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
test137E29B, line 652: If this check is not here, we'd have to check every affected item when getting enabled researches.

I don't think this is a good solution anymore though tbh

JustAHuman-xD:
Superceded by #3917

PR #3832, opened 2023-05-20
Add tradable attribute to SlimefunItem

ybw0014:
## Description

This allows addons to set whether the SlimefunItem is allowed to be traded with villagers and wandering traders.

## Proposed changes

- Add `tradable` attribute to Slimefun, with getter and setter.
- `VanillaItem` and `SyntheticEmerald` have `tradable` set to `true` by default.
- Change the method to determine whether the item is tradable in `VillagerTradingListener`, now it only reads `tradable`.

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
@@ -110,6 +110,7 @@ public class SlimefunItem implements Placeable {
     protected boolean disenchantable = true;
     protected boolean hidden = false;
     protected boolean useableInWorkbench = false;
+    protected boolean tradable = false;
 
     private Optional<String> wikiURL = Optional.empty();
 
@@ -736,6 +737,35 @@ public class SlimefunItem implements Placeable {
      */
     public @Nonnull SlimefunItem setUseableInWorkbench(boolean useable) {
         this.useableInWorkbench = useable;
+
+        return this;
+    }
+
+    /**
+     * This method returns whether this {@link SlimefunItem} is allowed to
+     * be traded with villagers or wandering traders.
+     *
+     * Items of type {@link VanillaItem} may be used in workbenches for example.
+     *
+     * @see #setTradable(boolean)
+     *
+     * @return Whether this {@link SlimefunItem} can be traded.
+     */
+    public boolean isTradable() {
+        return tradable;
+    }
+
+    /**
+     * This sets whether this {@link SlimefunItem} is allowed to be
+     * traded with villagers or wandering traders.
+     *
+     * @param tradable
+     *            Whether this {@link SlimefunItem} should be tradable
+     *
+     * @return This instance of {@link SlimefunItem}
+     */
+    public @Nonnull SlimefunItem setTradable(boolean tradable) {
+        this.tradable = tradable;
 
         return this;
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
@@ -48,7 +48,7 @@ public class VillagerTradingListener implements Listener {
             }
 
             if (e.getResult() == Result.DENY) {
-                Slimefun.getLocalization().sendMessage((Player) e.getWhoClicked(), "villagers.no-trading", true);
+                Slimefun.getLocalization().sendMessage(e.getWhoClicked(), "villagers.no-trading", true);
             }
         }
     }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
@@ -54,6 +54,6 @@ public class VillagerTradingListener implements Listener {
     }
 
     private boolean isUnallowed(@Nullable SlimefunItem item) {
-        return item != null && item.isTradable() && !item.isDisabled();
+        return item != null && !item.isTradable() && !item.isDisabled();
     }
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
@@ -54,6 +54,6 @@ public class VillagerTradingListener implements Listener {
     }
 
     private boolean isUnallowed(@Nullable SlimefunItem item) {
-        return item != null && !(item instanceof VanillaItem) && !(item instanceof SyntheticEmerald) && !item.isDisabled();
+        return item != null && item.isTradable() && !item.isDisabled();
     }
 }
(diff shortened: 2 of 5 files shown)


WalshyDev:
Can we get a test? :)

I like this.

ybw0014:
Testing, I got some problems. Turning it to draft for now.

ybw0014:
btw do you prefer `tradable` or `tradeable`?

ybw0014:
Also do we need to change the message shown? It now says "You cannot trade Slimefun items with Villagers!"
What about changing it to "You cannot trade this Slimefun item with Villagers or Wandering Traders!"

Sefiraat:
> Also do we need to change the message shown? It now says "You cannot trade Slimefun items with Villagers!"
> What about changing it to "You cannot trade this Slimefun item with Villagers or Wandering Traders!"

Yes change it otherwise it gives the impression that it’s all items :)

Sefiraat:
> btw do you prefer `tradable` or `tradeable`?

Either is fine. I’d stick with what you have now. I prefer it with the e personally but I think that’s a British vs American thing.

Sefiraat:
But of a side question - Assuming this is still to aid your work on the trading addon, and given it’s not going to sort out the root of the issue, has consideration been taken for the changes that would be needed to allow slimefun items to be traded regardless of this setting for a custom trade as defined?

ybw0014:
> But of a side question - Assuming this is still to aid your work on the trading addon, and given it’s not going to sort out the root of the issue, has consideration been taken for the changes that would be needed to allow slimefun items to be traded regardless of this setting for a custom trade as defined?

I haven't thought of this, maybe I will open a new pr for the trading item on the list.

Sefiraat:
> > But of a side question - Assuming this is still to aid your work on the trading addon, and given it’s not going to sort out the root of the issue, has consideration been taken for the changes that would be needed to allow slimefun items to be traded regardless of this setting for a custom trade as defined?
> 
> I haven't thought of this, maybe I will open a new pr for the trading item on the list.

The only reason I mention it is because that change will make any item subsequently tradable but only for a slimefun > slimefun trade meaning the term traceable becomes too ambiguous?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
```diff
@@ -48,12 +48,12 @@ public void onPreTrade(InventoryClickEvent e) {
             }
 
             if (e.getResult() == Result.DENY) {
-                Slimefun.getLocalization().sendMessage((Player) e.getWhoClicked(), "villagers.no-trading", true);
+                Slimefun.getLocalization().sendMessage(e.getWhoClicked(), "villagers.no-trading", true);
```
TheBusyBiscuit, line 51: I don't think we can do this here due to backwards compatibility.

src/main/resources/languages/en/messages.yml
```diff
@@ -346,7 +346,7 @@ smithing_table:
   not-working: '&4You cannot use a Slimefun item as a smithing material!'
 
 villagers:
-  no-trading: '&4You cannot trade Slimefun items with Villagers!'
+  no-trading: '&4You cannot trade this Slimefun item with Villagers or Wandering Traders!'
```
TheBusyBiscuit, line 349: You should define a new key since this change will neither transfer to other languages nor existing servers.

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
ybw0014: I looked through the javadoc from 1.16 to 1.19, the `getWhoClicked()` always returns a `HumanEntity`, which is a subinterface of `CommandSender`, so casting here isn't needed.

ybw0014: Do you remember why you add this casting?

src/main/resources/languages/en/messages.yml
```diff
@@ -347,6 +347,7 @@ smithing_table:
 
 villagers:
   no-trading: '&4You cannot trade Slimefun items with Villagers!'
+  no-trading-new: '&4You cannot trade this Slimefun item with Villagers or Wandering Traders!'
```
JustAHuman-xD, line 350: Maybe name this no-trading-individual ? just an idea

ybw0014: i really need a better name here...

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
```diff
@@ -740,6 +741,35 @@ public boolean isUseableInWorkbench() {
         return this;
     }
 
+    /**
+     * This method returns whether this {@link SlimefunItem} is allowed to
+     * be traded with villagers or wandering traders.
+     *
+     * Items of type {@link VanillaItem} may be used in workbenches for example.
```
Sfiguz7, line 748: Remnant of copy paste?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/VillagerTradingListener.java
Sfiguz7: Will resolve this as stale, backwards compat is dead and 1.16 on you said you checked so it's fine by me

src/main/resources/languages/en/messages.yml
Sfiguz7: Maybe no-trading-singular, or what Justin suggested: new sounds weird but anything else will work since there is an explanation literally next to it

Alessio-Colombo (changes requested):

src/main/resources/languages/en/messages.yml
ybw0014: The reason for adding a new message is that changes to the old message will not be deployed to existing servers or Crowdin, so I added a new message.
The old message will not be used anymore, not sure if `singular` is a proper name here.

Sefiraat: I see the reasoning, not a fan but given the alternative is rewriting the whole message system, I can follow the logic. 

While we’re here, it doesn’t really need to mention Villager or Wandering traders does it, given the message is given during the context of trading the player could just be told `You cannot trade using this Slimefun Item` or something to that effect? That was when mojang adds a… SoitheringMerchant in the future we don’t need to worry about further ‘new’ versions of this same message?

Sfiguz7: > The reason for adding a new message is that changes to the old message will not be deployed to existing servers or Crowdin, so I added a new message.
> 
> The old message will not be used anymore, not sure if `singular` is a proper name here.

I think I explained myself badly: I understand we won't be using the other one anymore but it's going to stay there for now (probably for a long time actually) so we need a way to distinguish the one referring to Slimefun Items as a whole and the one referring to a single one - hence why I was thinking either singular or individual would do.

I don't think perfect naming is needed, just one that is good enough: the error message it represents sits right next to it so the exact meaning can be deduced by context, otherwise we would need "no-trading-is-allowed-for-this-slimefun-item" or something which looks even worse in my opinion, way too many details.

Essentially, if this is the only thing keeping the PR open once the other requested change is addressed, I don't think we should spend too much time on it when it makes almost no difference overall

Issue #3831, opened 2023-05-18
Signs attached to custom blocks don't drop


ghost:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

When placing a sign on an output chest or edged/reinforced furnace. if you subsequently break the block the sign will not drop.

### 📑 Reproduction Steps

1. Place down an output chest
2. Place a sign on it
3. Break the output chest

### 💡 Expected Behavior

I expect the sign to drop.

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

DEV 1050

WalshyDev:
Did this break in 1.19.x or was this always an issue @Boomer-1 ?

Boomer-1:
i'm pretty sure it's been mentioned in the past, but this was the first time i think it got reported here

WalshyDev:
Interesting, thanks!

J3fftw1:
This is most likely the same as with vines buttons paintings etc.
I have seen this being reported on discord.
We even speculated a fix but it’s really vaguely in my mind.

I think we were talking about sending a block update on block break

PR #3834, opened 2023-06-01
Fix for blocks attached to custom blocks not dropping item

Lolboer9999999:
fix for issue #3831
fixed by setting the block to air and reenabling regular drops

## Description

This is a fix for issue #3831
## Proposed changes

i replaced 
`e.setDropItems(false);`
with
```
e.setCancelled(true);
e.getBlock().setType(Material.AIR);
e.setDropItems(true);
```

#3831
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -175,7 +175,10 @@ public class BlockListener implements Listener {
             // Fixes #2560
             if (e.isDropItems()) {
                 // Disable normal block drops
-                e.setDropItems(false);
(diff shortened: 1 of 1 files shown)


TheBusyBiscuit (commented):
Please give your pull request a more reasonable name.

J3fftw1:
When setting the material to air couldn’t you also cause a block update by setting the second optional parameter to true. That would fix this issue i think as well 
https://github.com/Slimefun/Slimefun4/issues/3182

Sefiraat:
Heya, on checking this will break several things after this method as well as causing a clear dupe for many items when breaking also. This isn't the correct approach to be able to fix the issue and I do not think there is a way for you to modify it to be suitable without completely undoing the current commits.

You're welcome to open a new PR with a different approach or take a look at the other existing PR for sensitive blocks to see if you can extend the fix you're after to that.

PR #3857, opened 2023-06-17
Fix #3260

### LINKED ISSUE #3260: Generators power multiple energy network with unnerfed rate

JustAHuman-xD:
## Description
I'm making this PR to fix the issues with energy duplicaiton.

## Proposed changes
Inside the EnergyNet class I override the addLocationToNetwork method and check if there is an EnergyNetwork already at it's location. If there isn't then I just call the super.addLocationToNetwork. This fixes the issue as if a generator is already connected to one network. It will find a EnergyNetwork and not add the generator to any new networks.

I also updated the annotations of some methods in EnergyNetwork to fit the code style guidelines.

Resolves #3260 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
@@ -143,6 +143,13 @@ public class EnergyNet extends Network implements HologramOwner {
         }
     }
 
+    @Override
+    protected void addLocationToNetwork(@Nonnull Location l) {
+        if (EnergyNet.getNetworkFromLocation(l) == null) {
+            super.addLocationToNetwork(l);
+        }
+    }
+
     public void tick(@Nonnull Block b) {
         AtomicLong timestamp = new AtomicLong(Slimefun.getProfiler().newEntry());
 
@@ -298,8 +305,7 @@ public class EnergyNet extends Network implements HologramOwner {
         }
     }
 
-    @Nullable
-    private static EnergyNetComponent getComponent(@Nonnull Location l) {
+    private static @Nullable EnergyNetComponent getComponent(@Nonnull Location l) {
         SlimefunItem item = BlockStorage.check(l);
 
         if (item instanceof EnergyNetComponent component) {
@@ -318,8 +324,7 @@ public class EnergyNet extends Network implements HologramOwner {
      *
      * @return The {@link EnergyNet} at that {@link Location}, or {@code null}
      */
-    @Nullable
-    public static EnergyNet getNetworkFromLocation(@Nonnull Location l) {
+    public static @Nullable EnergyNet getNetworkFromLocation(@Nonnull Location l) {
         return Slimefun.getNetworkManager().getNetworkFromLocation(l, EnergyNet.class).orElse(null);
     }
 
@@ -332,8 +337,7 @@ public class EnergyNet extends Network implements HologramOwner {
      * 
      * @return The {@link EnergyNet} at that {@link Location}, or a new one
      */
-    @Nonnull
-    public static EnergyNet getNetworkFromLocationOrCreate(@Nonnull Location l) {
+    public static @Nonnull EnergyNet getNetworkFromLocationOrCreate(@Nonnull Location l) {
         Optional<EnergyNet> energyNetwork = Slimefun.getNetworkManager().getNetworkFromLocation(l, EnergyNet.class);
 
         if (energyNetwork.isPresent()) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
```diff
@@ -143,6 +143,13 @@ public void onClassificationChange(Location l, NetworkComponent from, NetworkCom
         }
     }
 
+    @Override
+    protected void addLocationToNetwork(@Nonnull Location l) {
```
J3fftw1, line 147: Why do we override it and not just fix the actual method in the networks class?

JustAHuman-xD: I put it here because it doesn't make sense to have an energy network specific bug to be fixed in an abstract network class. An abstract class shouldn't call to its child class.

J3fftw1: Nothing else in Slimefun uses the network class except energy net tho. So I think it’s fine to add this check in there. I can be wrong tho

JustAHuman-xD: Cargo net uses the network class I thought? Also like you said if an addon extended it then that would be another usage.

JustAHuman-xD: Yep cargo net extends abstract item network which extends network

J3fftw1: Yes but if an addon uses this implementation it would also have the same issue

WalshyDev: Are there any addons extending Network and not EnergyNet? For an energy fix, it makes sense for it to be here. `Network` is a much more general-use class, not specific to energy.

JustAHuman-xD: > Are there any addons extending Network and not EnergyNet for energy? For an energy fix, it makes sense for it to be here. `Network` is a much more general-use class, not specific to energy.

Personally I don't know of any

Sfiguz7 (approved):
Seeing our conversation here and on discord (thread: https://discord.com/channels/565557184348422174/1121014648213352448 "Superimposed networks") I will approve and intend to merge

EDIT: maybe after a test with the new build

WalshyDev (approved):

Sefiraat:
Network locations are populated as/when connectors are added. 
If a connector is added and it's collection of valid location is denied the locations due to this change, those location will not be re-populated when the conditions of the denial are changed, either via a re-placement or a change in the 2nd network.

From testing this results in generators seemingly not producing, consumers not consuming or connecters extending the network to other connectors. In a small scale base this would be fairly easy for a player to debug, but for larger bases I can see this being fairly nightmarish, along the lines of what players experience now with finding what block has turned vanilla during a BlockStorage attack.

There either needs to be a trigger of somekind to re-populate or another solution to the core issue at hand?

Sefiraat (changes requested):
Just going to mark as requested changes while we discuss in Discord first quickly!

Sefiraat:
Also (offtopic), +1 for the preview builds. I did end up having to clone for JVM debug, but it got me there, gg @WalshyDev @J3fftw1

JustAHuman-xD:
Oh interesting, when I tested i I didn't run into any of that but that's probably due to the order I went through. Yeah then this'll be a little more complex

Sefiraat:
I’m happy to pop up a video if my rambling nonsense doesn’t cut it, but it’s also fairly clear from the code also. Tbh I like your solution here it’s just the code around it needs some changes to compensate. 

I also think this may be handled differently by the manner in which we handle energy, but that opens another barrel of worms. I’ve had some back and forth with Alessio, I’ll pop something more significant in the discord for discussion when I’m free also, assuming no other thoughts come to light

JustAHuman-xD:
Going to close this PR to await a proper fix that does not have any of the relevant problems 👍

PR #3862, opened 2023-06-21
Fix incorrect return in BlockListener

Sefiraat:
## Description
A recent merge resulted in an incorrect return if the block being broken was NOT a Slimefun item resulting in tool handlers no longer being called.

## Proposed changes
Change the logic to fire the event ONLY if the Slimefun block exists and then returning within that if statement only.

## Checklist


JustAHuman-xD (approved):
LGTM good catch

Sefiraat:
> LGTM good catch

Would have been a better catch yesterday ;)

WalshyDev (approved):

JustAHuman-xD:
Ofc it's the pull request labels check that failed 🤦‍♂️

PR #3863, opened 2023-06-21
Preview builds

J3fftw1:
## Description

This PR will add a way for reviewers/testers or whoever wants to test the PR out.

## Proposed changes
On push, the usual build workflow is ran, this will store the build and then the new workflow being introduced will upload to R2 (using [this Worker](https://github.com/Slimefun/preview-builds)) and comment on the PR with the preview link

## Checklist


WalshyDev:
Ok, this is ready :)

Sadly we can't test this until it's merged because the `workflow_run` requires it to be ran on the default branch
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run
> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

Hopefully it works but no big deal if not, we can still iterate.

JustAHuman-xD (approved):
Let's try it 👀

WalshyDev (approved):

Sfiguz7 (approved):

PR #3870, opened 2023-06-24
Download artifact by ID, it isn't available in the workflow_run

WalshyDev:
## Description
Apparently, download-artifact just checks the current workflow for it and not the run.

## Proposed changes
In our run-script, let's download the artifact as well. We know this is picking up the PR correctly: https://github.com/Slimefun/Slimefun4/actions/runs/5364131229/jobs/9732129940
So... this should work

N/A

## Checklist

### DIFF
--- a/.github/workflows/preview-builds.yml
+++ b/.github/workflows/preview-builds.yml
@@ -17,9 +17,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


JustAHuman-xD (approved):

Sefiraat (approved):
Imma just trust you here

PR #3875, opened 2023-06-28
Disable slimefun in world that contains illegal charcters.

### LINKED ISSUE #3874: Slimefun_Not_Working_As_Intended
### ❗ Checklist

### 📍 Description

When a member logged on and messed with the Slimefun components, one output chest was working while the other was not and when he tried to use an energy regulator it did not show him as owner rather it shown CS-CoreLib as owner and when he tried to destroy it and place it back it just shown it as CS-CoreLib's head rather than an actual component including the "Connecting..." hologram as well as the Simple Elevator when trying to be used it doesn't work and when destroyed and put into inventory it says Block of Quartz. After these problems rose, I saw in the console a message that states "[20:15:08 ERROR]: [Slimefun] An Exception was caught while ticking the Block Tickers Task for Slimefun vDEV - 1068 (git cab97b1a)
java.lang.IllegalStateException: World "Teknocraft S3.5" seems to have been deleted. Do not call unsafe methods directly!" Even though in the world settings folder of Slimefun has the world on it. The software used is a mixture of different softwares combined: Spigot, Paper, Purpur and Bukkit.

### 📑 Reproduction Steps

1. Updated  to the latest version
2. Uploaded it to my server's file manager
3. Restarted the server
4. Logged on and the issues still occured

### 💡 Expected Behavior

Was expecting for everything to work as normal after these issues, but alas, they still remain.

### 📷 Screenshots / Videos

SimpleElevatorIssue
WorldSettingsSlimefun
CoreLib'sHead
AnExceptionWasCaught-Slimefun
Owner_CSCoreLib

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Other (please specify in your description)

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

Slimefun4 - DEV 1068 (git cab97)

### 🧭 Other plugins

SimpleUtilities

J3fftw1:
## Description

This PR disabled Slimefun when there is an illegal charcter in 1 of the worlds name.

## Proposed changes

Disable Slimefun when there is an illegal char found

Resolves #3874 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
@@ -287,7 +287,6 @@ public final class Slimefun extends JavaPlugin implements SlimefunAddon {
         for (World world : Bukkit.getServer().getWorlds()) {
             if (pattern.matcher(world.getName()).find()) {
                 illegalWorld.add(world.getName());
-            return;
             }
         }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
@@ -1,11 +1,9 @@
 package io.github.thebusybiscuit.slimefun4.implementation;
 
 import java.io.File;
-import java.lang.reflect.Array;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -585,7 +583,7 @@ public final class Slimefun extends JavaPlugin implements SlimefunAddon {
     /**
      * This method creates all necessary directories (and sub directories) for Slimefun.
      */
-    private void  createDirectories() {
+    private void createDirectories() {
         String[] storageFolders = { "Players", "blocks", "stored-blocks", "stored-inventories", "stored-chunks", "universal-inventories", "waypoints", "block-backups" };
         String[] pluginFolders = { "scripts", "error-reports", "cache/github", "world-settings" };
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
@@ -162,7 +162,7 @@ public final class Slimefun extends JavaPlugin implements SlimefunAddon {
     /**
      * Keep track of if the server has a world file with an illegal character.
      */
-    private Set<String> illegalWorld = new HashSet<>();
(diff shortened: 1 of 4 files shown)


WalshyDev:
Not sure disabling is the right approach here. Why don't we just not load storage?

Also for the regex I'd probably just do like `[^a-z0-9_-]` so we have an allowlist rather than a blocklist

J3fftw1:
> Not sure disabling is the right approach here. Why don't we just not load storage?
> 
> Also for the regex I'd probably just do like `[^a-z0-9_-]` so we have an allowlist rather than a blocklist

Agreed with the allow list will change that later today.

However for the disabling if we do just not load storage. There will still be a storage folder created and I don’t think we prevent block placing in the worlds with not loaded storage

J3fftw1:
After a little more thought I don’t think an allow list is a good idea. We also support different languages which might support weird characters that aren’t in the English alphabet.
Chinese
Norwegian
German
Etc

Sefiraat:
I do agree that an approach in which we just stop slimefun storing data for an ‘illegal’ world would be better here. 
Assuming that we correctly identify that a world shouldn’t/cant have blockstorage and then do not create a storage object for said world, it should be easy enough in our listeners to stop block placement if there is a valid slimefun item in hand but with no blockstorage object available?

WalshyDev (commented):
Not a big fan of this collection approach

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
```diff
@@ -273,6 +280,17 @@ private void onPluginStart() {
...
+        Pattern pattern = Pattern.compile("[\\.\\\\\\/\":><|?*]");
+
+        // Check if world contains an illegal character
+        for (World world : Bukkit.getServer().getWorlds()) {
+            if (pattern.matcher(world.getName()).find()) {
+                illegalWorld.add(world.getName());
+            return;
+            }
+        }
+
```
JustAHuman-xD, line 293: Is that `return` indented properly?

WalshyDev: why are we even returning? That would stop this after the first match

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
```diff
@@ -157,6 +159,11 @@ public final class Slimefun extends JavaPlugin implements SlimefunAddon {
      */
     private boolean isNewlyInstalled = false;
 
+    /**
+     * Keep track of if the server has a world file with an illegal character.
+     */
+    private Set<String> illegalWorld = new HashSet<>();
```
WalshyDev, line 165: how about naming it like disabledWorlds?

WalshyDev: this also needs to be final

PR #3888, opened 2023-07-02
Fix breaking change in LongFallBoots

WalshyDev:
## Description
The sound system made a breaking change to LongFallBoots causing FoxyMachines to no longer start.

## Proposed changes
Re-add the old constructor and deprecate.

https://github.com/GallowsDove/FoxyMachines/issues/71

## Checklist


JustAHuman-xD (approved):
LGTM

GallowsDove:
Thanks for this <3

Should be fixed in the latest build of FoxyMachines (thanks Human), but I still think this is a good change, at least for a few versions.

Sfiguz7 (approved):

PR #3894, opened 2023-07-06
Cleanup and add version to preview builds

WalshyDev:
Changes the version in preview builds from `4.9-UNOFFICIAL` to `Preview Build #<pr-num>-<hash>`

Just a nice way to see what preview build we're running, check it's latest, check it's the right pr, etc.

Also does some cleanup, we no longer do preview stuff in the "Java CI" workflow and try to work around it firing on both push + pr. There's now just a PR workflow

### DIFF
--- a/.github/workflows/maven-compiler.yml
+++ b/.github/workflows/maven-compiler.yml
@@ -44,9 +44,3 @@ jobs:
 
     - name: Build with Maven
       run: mvn package --file pom.xml
-
-    - name: Upload the artifact
-      uses: actions/upload-artifact@v3
-      with:
-        name: slimefun-${{ github.event.number }}
-        path: 'target/Slimefun v4.9-UNOFFICIAL.jar'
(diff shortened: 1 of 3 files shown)


J3fftw1:
bit out of data but can we add the commit to the preview message as signature so we know if it has changed

Sfiguz7 (approved):

J3fftw1:
@WalshyDev can you fire a preview build so we know how it looks?

WalshyDev:
Nope, it runs on `main` (which is why it can access secrets and so forth). We'd need to merge it in to see

J3fftw1:
ugh

WalshyDev:
yeah trying to make this actually run as a test is annoying, in _theory_ it'll work now but I'm sure another error will happen

WalshyDev:
yeah i give up, we'll just test once merged

Sefiraat (approved):
Again, just trusting you know your stuff here xD <3

Sfiguz7 (approved):
Blob goes beep boop

PR #3898, opened 2023-07-09
Changed the registry to use world UUIDs

J3fftw1:
## Description
I have recently run into an issue where a world, when created quickly enough after deletion of a world with the same name, will not store data. I have tested this extensively and narrowed the issue down to [this](https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java#L102) ticker code and [this](https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java#L116) check for the world's existence. Since the ticker code only runs every half second, it does not remove the world from the registry until up to a Slimefun tick later and therefore causes issues in the constructor.

An easy way to recreate this issue is to delete a world and immediately create a new one with the same name.

Full credits go to @gecko10000 for some reason (my git skills most likely) his PR got closed while i tried to update it.

## Proposed changes
This PR changes the registry to use UUIDs instead of the world names themselves. This _does not_ affect the storage location in files and therefore these are not breaking changes.

This may not be the ideal way to do it; it may be better to remove the world from the registry [here](https://github.com/Slimefun/Slimefun4/blob/master/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/WorldListener.java#L28) instead.

No issue opened

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
@@ -85,7 +85,7 @@ public final class SlimefunRegistry {
     private final KeyMap<GEOResource> geoResources = new KeyMap<>();
 
     private final Map<UUID, PlayerProfile> profiles = new ConcurrentHashMap<>();
-    private final Map<String, BlockStorage> worlds = new ConcurrentHashMap<>();
+    private final Map<UUID, BlockStorage> worlds = new ConcurrentHashMap<>();
     private final Map<String, BlockInfoConfig> chunks = new HashMap<>();
     private final Map<SlimefunGuideMode, SlimefunGuideImplementation> guides = new EnumMap<>(SlimefunGuideMode.class);
     private final Map<EntityType, Set<ItemStack>> mobDrops = new EnumMap<>(EntityType.class);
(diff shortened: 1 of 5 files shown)


Sfiguz7 (changes requested):
I actually like this being a change in the underlying type as this also makes the registry more robust towards world names with dots or weird characters, which have been annoying in the past. It is simply a good idea to use unique IDs when they exists rather than compare strings.

Marking this requested changes to make sure we get it tested before it is merged

variananora:
Just tested it on 1.20.1 and it works fine.

Quick question tho, is disabling world with `.` character is still needed if we move this to UUID?

J3fftw1:
> Just tested it on 1.20.1 and it works fine.
> 
> Quick question tho, is disabling world with `.` character is still needed if we move this to UUID?

It's bad practice anyways to use dots in file names.
Is it needed to disable with this change honestly no clue, but i think its fine to stay like this.

src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
```diff
@@ -350,7 +350,7 @@ public Set<ItemHandler> getGlobalItemHandlers(@Nonnull Class<? extends ItemHandl
     }
 
     @Nonnull
-    public Map<String, BlockStorage> getWorlds() {
+    public Map<UUID, BlockStorage> getWorlds() {
```
WalshyDev, line 353: People rely on this method - https://github.com/search?q=BlockStorage+.getWorlds%28%29&type=code
This will break due to signature change.

WalshyDev (changes requested):

PR #3902, opened 2023-07-10
Add grace periods to RadiationTask

### LINKED ISSUE #3901: Radiation system instantly killing when player respawns with keep inventory.
### ❗ Checklist

### 📍 Description

This was discussed on https://discord.com/channels/565557184348422174/569249847258120222/1127853858962554880

What happened was a player picked up radiation items with over 100 value of radiation, after picking it up player will dies.
Because of keep inventory is on player is unable to drop, remove, or even wear a protective armor to combat this radiation.
If you are able to wear a protective armor, it took a while to register as being protected.

P.S. I haven't been able to test this a little more detail, I will update this tonight or tomorrow morning.

### 📑 Reproduction Steps

1. Enable keep inventory
2. Get some radiation items, preferably the one that kills you instantly
3. Just watch your self in a loop of death 

### 💡 Expected Behavior

Jeff and Sefi suggested there should be a 'cooldown' or 'grace period' or X seconds after player respawns, this allow the player to wear a protective armor or throw the item before going to die again.

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

```
Paper git-Paper-68 (MC: 1.20.1)
Slimefun DEV - 1081 (git aebab92e)
Metrics-Module #29
Java 17

No Addons installed
```

### 🧭 Other plugins

_No response_

Sefiraat:
## Description
Currently when a server has KeepInventory on (or a player dies on their respawn location) they can get caught in a tricky death-loop.

## Proposed changes
We add a 15 second grace period after dying before radiation effects trigger again. This gives more than enough time to remove the offending item(s) and/or put on the relevant armor. It is also short enough to discourage using this grace period to 'zerg' their way to their intended end result.

Tested quickly on 1.19.4 and worked well.

Fixes #3901 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
@@ -1,5 +1,10 @@
 package io.github.thebusybiscuit.slimefun4.implementation.tasks.armor;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 import org.bukkit.GameMode;
@@ -28,13 +33,20 @@ import net.md_5.bungee.api.chat.ComponentBuilder;
  */
 public class RadiationTask extends AbstractArmorTask {
 
+    public static final int GRACE_PERIOD_DURATION = 15;
+    private static final Map<UUID, Long> ACTIVE_GRACE_PERIODS = new HashMap<>();
+
     private final RadiationSymptom[] symptoms = RadiationSymptom.values();
 
     @Override
     @ParametersAreNonnullByDefault
     protected void onPlayerTick(Player p, PlayerProfile profile) {
-        int exposureTotal = 0;
+        if (withinGracePeriod(p)) {
+            // Player is within their grace period and shouldn't have radiation effects applied.
+            return;
+        }
 
+        int exposureTotal = 0;
         if (!profile.hasFullProtectionAgainst(ProtectionType.RADIATION) && p.getGameMode() != GameMode.CREATIVE && p.getGameMode() != GameMode.SPECTATOR) {
             for (ItemStack item : p.getInventory()) {
                 if (item == null || item.getType().isAir()) {
@@ -73,5 +85,40 @@ public class RadiationTask extends AbstractArmorTask {
                 p.spigot().sendMessage(ChatMessageType.ACTION_BAR, components);
             }
         }
+    }
+
+    /**
+     * Checks if the {@link Player} is within their grace period. A grace period is granted after death
+     * to give enough time to remove the radioactive items before being killed once more, which
+     * with KeepInventory on would result in a very hard-to-escape loop.
+     * @see io.github.thebusybiscuit.slimefun4.implementation.listeners.RadioactivityListener
+     *
+     * @param player
+     *              The {@link Player} to check against.
+     *
+     * @return Returns true if the {@link Player} is within their grace period.
+     */
+    private boolean withinGracePeriod(@Nonnull Player player) {
+        long gracePeriodEnd = ACTIVE_GRACE_PERIODS.getOrDefault(player.getUniqueId(), 0L);
+        if (gracePeriodEnd >= System.currentTimeMillis()) {
+            // Player is within their grace period
+            return true;
+        }
+
+        if (gracePeriodEnd > 0) {
+            // A grace period was present but has since lapsed, remove the entry.
+            ACTIVE_GRACE_PERIODS.remove(player.getUniqueId());
+        }
+
+        return false;
+    }
+
+    /**
+     * Adds the given {@link Player}'s grace period to the collection.
+     *
+     * @param player The player to add the grace period to.
+     */
+    public static void addGracePeriod(@Nonnull Player player) {
+        ACTIVE_GRACE_PERIODS.put(player.getUniqueId(), System.currentTimeMillis() + (GRACE_PERIOD_DURATION * 1000L));
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
```diff
@@ -28,13 +33,20 @@
  */
 public class RadiationTask extends AbstractArmorTask {
 
+    public static final int GRACE_PERIOD_DURATION = 15;
```
J3fftw1, line 36: lets config option this

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
```diff
@@ -74,4 +86,39 @@ protected void onPlayerTick(Player p, PlayerProfile profile) {
             }
         }
     }
+
+    /**
+     * Checks if the {@link Player} is within their grace period. A grace period is granted after death
+     * to give enough time to remove the radioactive items before being killed once more, which
+     * with KeepInventory on would result in a very hard-to-escape loop.
+     * @see io.github.thebusybiscuit.slimefun4.implementation.listeners.RadioactivityListener
```
J3fftw1, line 94: make this a @ link

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
```diff
@@ -74,4 +86,39 @@ protected void onPlayerTick(Player p, PlayerProfile profile) {
...
+     * @return Returns true if the {@link Player} is within their grace period.
+     */
+    private boolean withinGracePeriod(@Nonnull Player player) {
+        long gracePeriodEnd = ACTIVE_GRACE_PERIODS.getOrDefault(player.getUniqueId(), 0L);
+        if (gracePeriodEnd >= System.currentTimeMillis()) {
+            // Player is within their grace period
+            return true;
+        }
+
+        if (gracePeriodEnd > 0) {
```
J3fftw1, line 108: if its 0 its lapsed as well

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
Sefiraat: If it's 0 then it's not in the map so no need to remove

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
Sefiraat: I did think on this, and I'm not sure I see the direct benefit of adding it as a config option. Given the usage here it seems to fit more as a direct constant, not unlike the other behaviours surrounding radioactivity as a whole?

J3fftw1: people generally prefer to have a lot of values configurable.
Thats why i suggested it

JustAHuman-xD:
What if we moved the add grace period to clearExposure? That way things like vitamins and medicine also grants a grace period. This would also allow only giving the grace period if they are actually irradiated. I think that could be cleaner and more immersive but what do you guys think?

J3fftw1:
> What if we moved the add grace period to clearExposure? That way things like vitamins and medicine also grants a grace period. This would also allow only giving the grace period if they are actually irradiated. I think that could be cleaner and more immersive but what do you guys think?

that would be cool but technically out of scope

Sefiraat:
I guess the question is, do we additionally want those items that remove radiation to do so for an extended amount of time, as that would be a change in behaviour, and then open this up to requiring config, and on a per-item as well as global level. I like it as an idea essentially, but this PR is to fix an active reported issue so I think that’s perhaps a down-the-road conversation?

JustAHuman-xD:
> Edit: We also cant only give a grace period when they're actually irradiated, the point of this fix is to catch it before a player is irradiated upon respawning. Though I get your comments outside of this specific point.

If it was moved to clearExposure then it would be done before the radiation was actually removed was the idea. So the grace period would be added and then the radiation removed, keeping the current functionality but only when they actually were irradiated.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
WalshyDev: is there an existing mc/bukkit config value we can use? i know there's a general grace period in mc.

Sefiraat: Added config

variananora:
Tested this on commit: https://github.com/Slimefun/Slimefun4/commit/a91acfbabd4d7a5009b3449f4ab406ba469a4a27

The grace periods works like expected, the config also works properly too.

```

File: src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java [hunk 4]
```diff
@@ -74,4 +87,39 @@ protected void onPlayerTick(Player p, PlayerProfile profile) {
...
```
J3fftw1, line 109: Im a little confused for this

We are firstly checking
if the graceperiod time end is higher or is equal to current time then we return yes it is
but then if its just higher we remove it?

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/armor/RadiationTask.java
Sefiraat: First check checks that the graceperiod is still active, if it is the code returns. If the code doesn't return then we check if the value is above zero, meaning there was an active graceperiod which has lapsed and can be removed from the map. This could be changed to give the boxed object and do a null check, though it's all the same net result?

Sfiguz7 (approved):

J3fftw1 (approved):

PR #3905, opened 2023-07-11
Improve performance of clearAllBlockInfoAtChunk, don't get immutable map

### LINKED ISSUE #3432: Performance impact when clearing blockdata from a whole chunk

WalshyDev:
## Description
We're currently getting our ImmutableMap clone in clearAllBlockInfoAtChunk, instead we can _unsafely_ get the storage and save the overhead of duplicating the map (which for large servers is _massive_).

Right now it's taking basically all of the compute time:
image

## Proposed changes
Unsafely get the raw storage instead of our safe ImmutableMap.

Resolves #3432

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java
@@ -621,8 +621,8 @@ public class BlockStorage {
             return;
         }
         Map<Location, Boolean> toClear = new HashMap<>();
-        Map<Location, Config> storage = blockStorage.getRawStorage();
-        for (Location location : storage.keySet()) {
+        // Unsafe: get raw storage for this world
+        for (Location location : blockStorage.storage.keySet()) {
             if (location.getBlockX() >> 4 == chunkX && location.getBlockZ() >> 4 == chunkZ) {
                 toClear.put(location, destroy);
             }


WalshyDev:
Timing courtesy of @Skizzles - https://spark.lucko.me/5ZX5HHHvHi

`clearAllBlockInfoAtChunk` went from `52.67% 44440ms (self: 3404ms - 4.03%)` ->  `32.51% 11580ms (self: 112ms - 0.31%)`

44s -> 11s

We're now hitting Javaland code taking the rest of the majority:
image

22.03% (7.8s) from `ConcurrentHashMap.get()`

md5sha256:
> cc: @md5sha256 - not sure if you had any specific reason to use getRawStorage instead of accessing it directly. I believe we should be safe to use this though, we're only doing reads, not attempting any writes and aren't passing the ref down anywhere.

Either an oversight or I was erring on the side of caution. Either way, the performance benefit outweighs the perceived safety of operating on `#getRawStorage`

The only comment I have-ish, (and I've never done this, there was a reason why) but is it good practice/possible to just store the immutable map in BlockStorage since it's basically just a wrapper? This might become obsolete in the future, however.

WalshyDev:
> The only comment I have-ish, (and I've never done this, there was a reason why) but is it good practice/possible to just store (cache) the immutable map in the BlockStorage instance since it's basically just a wrapper? This might become obsolete in the future, however.

_probably_ but we'd really want to have partial updates done when needed to it (which kinda breaks the immutable part as it isn't fully) and afaik you can't do that with any Java-native implementations today.

There's also just the issue of added memory :(

md5sha256:
Yeah, fair enough. In that case LGTM

md5sha256 (approved):

JustAHuman-xD (approved):
LGTM

J3fftw1 (approved):

PR #3911, opened 2023-07-18
Deny placement when deletion is queued

Sefiraat:
## Description
When breaking/placing blocks too quickly you can place a block and have the BlockStorage data subsequently removed resulting in loss of items. This has also caused fringe cases of blocks with multiple BlockMenus, ghost blocks and other oddities.

## Proposed changes
If a block is placed and the location is still queued for deletion, deny the block placement with a message to the player to let them know why it was denied.

This is a bit of a patch job, but a true resolution would be part of the BlockStorage rewrite and I feel is important enough to warrant a fix prior to that to ease server owners burdens.

Tested on 1.19.4 and no issues, worked as intended. But given the scope of potential effects community testing here would be wonderful :)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -90,20 +90,34 @@ public class BlockListener implements Listener {
             Slimefun.getProtectionManager().logAction(e.getPlayer(), e.getBlock(), Interaction.PLACE_BLOCK);
         }
         if (sfItem != null && !(sfItem instanceof NotPlaceable)) {
-            if (!sfItem.canUse(e.getPlayer(), true)) {
+            Player player = e.getPlayer();
+
+            if (!sfItem.canUse(player, true)) {
                 e.setCancelled(true);
             } else {
-                SlimefunBlockPlaceEvent placeEvent = new SlimefunBlockPlaceEvent(e.getPlayer(), item, e.getBlock(), sfItem);
+                Block block = e.getBlockPlaced();
+
+                /*
+                 * Resolves an issue when placing a block in a location currently in the deletion queue
+                 * TODO This can be safely removed if/when the deletion no longer has a delay associated with it.
+                 */
+                if (Slimefun.getTickerTask().isDeletedSoon(block.getLocation())) {
+                    Slimefun.getLocalization().sendMessage(player, "messages.await-deletion");
+                    e.setCancelled(true);
+                    return;
+                }
+
+                SlimefunBlockPlaceEvent placeEvent = new SlimefunBlockPlaceEvent(player, item, block, sfItem);
                 Bukkit.getPluginManager().callEvent(placeEvent);
 
                 if (placeEvent.isCancelled()) {
                     e.setCancelled(true);
                 } else {
-                    if (Slimefun.getBlockDataService().isTileEntity(e.getBlock().getType())) {
-                        Slimefun.getBlockDataService().setBlockData(e.getBlock(), sfItem.getId());
+                    if (Slimefun.getBlockDataService().isTileEntity(block.getType())) {
+                        Slimefun.getBlockDataService().setBlockData(block, sfItem.getId());
                     }
 
-                    BlockStorage.addBlockInfo(e.getBlock(), "id", sfItem.getId(), true);
+                    BlockStorage.addBlockInfo(block, "id", sfItem.getId(), true);
                     sfItem.callItemHandler(BlockPlaceHandler.class, handler -> handler.onPlayerPlace(e));
                 }
             }
--- a/src/main/resources/languages/en/messages.yml
+++ b/src/main/resources/languages/en/messages.yml
@@ -172,6 +172,7 @@ messages:
   bee-suit-slow-fall: '&eYour Bee Wings will help you to get back to the ground safe and slow'
   deprecated-item: '&4This item has been deprecated and will be removed from Slimefun soon.'
   researching-is-disabled: '&cResearching has been disabled on this server. Everything is unlocked by default!'
+  await-deletion: '&cYou cannot place a Slimefun block so soon after breaking one. Try again shortly.'
 
   multi-tool:
     mode-change: '&b%device% mode changed to: &9%mode%'
--- a/src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunBlockBreakEvent.java
+++ b/src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunBlockBreakEvent.java
@@ -58,7 +58,7 @@ class TestSlimefunBlockBreakEvent {
         Player player = new PlayerMock(server, "SomePlayer");
 
         World world = server.addSimpleWorld("my_world");
-        Block block = new BlockMock(Material.GREEN_TERRACOTTA, new Location(world, 1, 1, 1));
+        Block block = new BlockMock(Material.GREEN_TERRACOTTA, new Location(world, TestUtilities.randomInt(), 100, TestUtilities.randomInt()));
 
         Slimefun.getRegistry().getWorlds().put("my_world", new BlockStorage(world));
         BlockStorage.addBlockInfo(block, "id", "FOOD_COMPOSTER");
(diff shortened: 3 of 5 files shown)


J3fftw1 (commented):
Code looks good not approving till further testing.
I really want to see some things tested
Breaking blocks on 1.16 and 1.20 
Placing blocks on 1.16 and 1.20
Combination of fast placing and fast breaking since that is what this should fix

JustAHuman-xD:
I agree with Jeff but I'm wondering if there's a better message option? Idk if I was a random player I think I'd get confused by the current message.

Sefiraat:
The core elements are cannot be placed yet and try again shortly. However putting that would trigger questions as to why. The reason in the middle is boringly abstract but it is the reason and removing that part I think w

src/main/resources/languages/en/messages.yml
```diff
@@ -172,6 +172,7 @@ messages:
   bee-suit-slow-fall: '&eYour Bee Wings will help you to get back to the ground safe and slow'
   deprecated-item: '&4This item has been deprecated and will be removed from Slimefun soon.'
   researching-is-disabled: '&cResearching has been disabled on this server. Everything is unlocked by default!'
+  await-deletion: '&cThis block cannot be placed yet as a previous removal has not yet been completed. Try again shortly.'
```
Sefiraat, line 175:
```suggestion
  await-deletion: '&cYou cannot place a Slimefun block so soon after breaking one. Try again shortly.'
```

WalshyDev (changes requested):
Just blocking this while I understand why we even would need to do this (feels very wrong)

JustAHuman-xD:
> Just blocking this while I understand why we even would need to do this (feels very wrong)

Currently if we don't do this when someone quickly breaks and replaces Slimefun blocks it messes up blockstorage, it can cause the newly placed block to vanillize. Make chest menu items drop that shouldn't drop, can mess up a lot of stuff.

WalshyDev:
Yes, Jeff told me the same thing but that shouldn't be a problem to begin with.

JustAHuman-xD:
> Yes, Jeff told me the same thing but that shouldn't be a problem to begin with. 

I mean if it's marked for deletion, it has to wait a tick for blockstorage to clear, which neither me sefi or Jeff know why afaik. So if someone places a block before that, which adds block storage with 0 delay. That new blockstorage will be cleared. Or some blockstorage stuff merges which I have no idea why that happens.

JustAHuman-xD:
_I believe anyway if I'm mis remembering any of this that's possible but iirc this is why, a part of it anyways_

Boomer-1:
i think originally when this was first reported it may have been someone trying to create either a dupe or some other situation, cause in order to do this, you either have to have it in your offhand and place after breaking or be insanely fast at switching your active hand.

WalshyDev:
> I mean if it's marked for deletion, it has to wait a sf tick for blockstorage to clear, which neither me sefi or Jeff know why afaik. 

yeah this shouldn't be a thing. That would just cause problems. So I want to understand if that is what is happening, understand why and understand if we do need that (i highly doubt it)

We never used to do that so the fact we do now is a problem. I want to fix the root of the problem not monkey patch on top.

Sefiraat:
Seriously? 

The reason I have not tackled the tick delay is because I asked you specifically in voice chat and you told me definitively that the tick delay in the queue deletion could not be changed. You didn’t give a shadow of a doubt in your response and said it could only be cleared as part of the blockstorage rewrite. Once again work is undertaken to provide relief while we await/work on the larger problems and you only question it AFTER a PR is raised, AFTER time is invested. Cmon Walshy you know how precious time is. 

Moreover this is a big problem on servers and has been for sometime and this is now causing problems with addon development as bad signals are being sent to addon so unless you have the answer to your questions above and you can confidently say the delay can be removed without causing issues, then we should absolutely patch the current issue while you work on the ‘true fix’ and feel free to remove this patch when completed. It is absolutely absurd to me that you’re happy to leave this in when we can provide this fix now when you do not know when/how your understanding of the fuller fix will come.

Sefiraat:
> > I mean if it's marked for deletion, it has to wait a sf tick for blockstorage to clear, which neither me sefi or Jeff know why afaik.
> 
> yeah this shouldn't be a thing. That would just cause problems. So I want to understand if that is what is happening, understand why and understand if we do need that (i highly doubt it)
> 
> We never used to do that so the fact we do now is a problem. I want to fix the root of the problem not monkey patch on top.

As a side note, I've checked the git logs for both little spans of code and it's awaited the ticker to run the deletion all the way back to 2016, unless I have missed something.

J3fftw1:
Here is where the 1 tick delay happens.
https://github.com/Slimefun/Slimefun4/blob/dc22e6658990dd7912dabbbfb79a0a24050eed74/src/main/java/me/mrCookieSlime/Slimefun/api/BlockStorage.java#L611

Sefiraat:
Tested on 1.17.1, 1.18.2, 1.19.4 and 1.20 with no differences in behaviour, though to be expected given there is nothing version specific within this code.

J3fftw1 (approved):

JustAHuman-xD (approved):
I think this is still a really necessary merge before the blockstorage rewrite is finished

Boomer-1:
in my testing this seems to be patched. i got the error message as intended and could not duplicate the original error after a dozen + attempts using various sf blocks.

test137E29B:
Would like to add this is very much needed - on servers with lots of people using slimefun, pending actions I've seen up to 5 million. This causes a lot of slimefun machines, or addon slimefun blocks like Sefi's network to turn into vanilla blocks. This is really annoying honestly as it can even happen over a forced server restart where not all pending actions are completed, or desync in some other way.

The amount of time invested into fixing these issues by server staff is insane, and especially on servers where hoppers are not craftable, it makes using Sefi's networks really frustrating since you can't just craft another grabber or pusher and replace it yourself. This temporary fix would solve the issues that players have, while being much less annoying to wait a little knowing it's not going to break the moment it's placed or at a random time in future.

JustAHuman-xD:
As someone who is constantly dealing with the ramifications of this I completely agree, this would help us sooooo much T>T

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -89,20 +89,33 @@ public void onBlockPlace(BlockPlaceEvent e) {
...
-                SlimefunBlockPlaceEvent placeEvent = new SlimefunBlockPlaceEvent(e.getPlayer(), item, e.getBlock(), sfItem);
+                Block block = e.getBlockPlaced();
+
+                /*
+                 * Resolves an issue when placing a block in a location currently in the deletion queue
+                 * TODO This can be safely removed if/when the deletion no longer has a delay associated with it.
+                 */
+                if (Slimefun.getTickerTask().isDeletedSoon(block.getLocation())) {
+                    Slimefun.getLocalization().sendMessage(player, "messages.await-deletion");
+                    e.setCancelled(true);
```
WalshyDev, line 105: We need to get out of here. We don't want to call SlimefunBlockPlaceEvent, we're currently still even writing to block storage and block data.

```suggestion
                    e.setCancelled(true);
                    return;
```

WalshyDev (changes requested):

WalshyDev:
I understand what is going on now and am fine with this going ahead.

@J3fftw1 fixed up the bad behaviour and we've added some new tests
I'd like it re-tested before I'm comfortable being ok to merge

Thanks @Sefiraat

J3fftw1 (approved):

Boomer-1:
on the second preview build, it is working as intended without the warning message, which feels much cleaner. i placed multiple blocks with the offhand the moment the main hand broke a block and all seemed to be working without any issue. I tested a dozen core slimefun machines and 5-6 from addons and not once did the item break

JustAHuman-xD:
> on the second preview build, it is working as intended without the warning message, which feels much cleaner. i placed multiple blocks with the offhand the moment the main hand broke a block and all seemed to be working without any issue. I tested a dozen core slimefun machines and 5-6 from addons and not once did the item break

there should still be a warning message, if you didnt ever get any it means you never placed when deletion was queued

Boomer-1:
well, i can't place it any faster than it is. any suggestions? or when is it being queued? is there at least a tick delay?

JustAHuman-xD:
easiest way would be to test it on some server with alot of lag

WalshyDev:
It's 1 Slimefun tick so you can make it super slow
`config.yml` -> `URID.custom-ticker-delay`

Boomer-1:
got a live production server owner who would be willing to add it to test?

Boomer-1:
> It's 1 Slimefun tick so you can make it super slow `config.yml` -> `URID.custom-ticker-delay` change to like 100

i'm reloading the server with it set to 100 to see what happens

Boomer-1:
over a dozen attempts and still no message

Boomer-1:
ok finally got it after a few more

Boomer-1:
after getting the message, all sf machines continued to work as expected

JustAHuman-xD (approved):

WalshyDev (approved):
Alright, @Boomer-1 managed to confirm this behaviour is indeed broken in the latest build and cannot break here (ty Boomer).

I'm happy for this now

PR #3919, opened 2023-07-24
skip already substituted parts when formatting message

iTwins:
## Description

``Debug::formatMessage`` doesn't work when ``vars[idx]`` cointains a ``{``
``msg.indexOf('{', i)`` will find the ``{`` that was just substituted in

## Proposed changes

replacing ``idx++;`` with ``i += ("" + vars[idx++]).length();`` should resolve this, because ``i`` will be moved to after anything that was just substituted in.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/debug/Debug.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/debug/Debug.java
@@ -106,7 +106,7 @@ public final class Debug {
         while ((i = msg.indexOf('{', i)) != -1 && msg.charAt(i + 1) == '}') {
             // Substring up to the opening brace `{`, add the variable for this and add the rest of the message
             msg = msg.substring(0, i) + vars[idx] + msg.substring(i + 2);
-            idx++;
+            i += ("" + vars[idx++]).length();
         }
 
         return msg;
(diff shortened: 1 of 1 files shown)


J3fftw1:
Im not sure what this PR actually fixes.
Do you have an example of the issue before this?

iTwins:
> Im not sure what this PR actually fixes. Do you have an example of the issue before this?

``formatMessage("some_debug1 = {} some_debug2 = {}", "substitution1{some_info}", "substitution2")``
now returns ``"some_debug1 = substitution1{some_info} some_debug2 = {}"``
while the expected result would be ``"some_debug1 = substitution1{some_info} some_debug2 = substitution2"``

The reason for this is that ``i = msg.indexOf('{', i)`` would find the ``{`` in ``substitution1{some_info}`` and then ``msg.charAt(i + 1) == '}'`` would return ``false`` so the loop stops. So if we add the length of the substituted part to ``i`` then ``msg.indexOf('{', i)`` will start searching after that.

J3fftw1 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/core/debug/Debug.java
```diff
@@ -106,7 +106,7 @@ public static void log(@Nonnull String test, @Nonnull String msg, @Nonnull Objec
         while ((i = msg.indexOf('{', i)) != -1 && msg.charAt(i + 1) == '}') {
             // Substring up to the opening brace `{`, add the variable for this and add the rest of the message
             msg = msg.substring(0, i) + vars[idx] + msg.substring(i + 2);
-            idx++;
+            i += ("" + vars[idx++]).length();
```
iTwins, line 109:
```suggestion
            i += String.valueOf(vars[idx++]).length();
```

J3fftw1 (approved):

Sfiguz7 (approved):

PR #3929, opened 2023-07-30
exploded sf blocks now drop their sf item

### LINKED ISSUE #3601: Enabling `call-explosion-event' will not drop Slimefun blocks when breaking it using explosive tools.
### ❗ Checklist

### 📍 Description

Enabling `call-explosive-event` on explosive tools caused only the targeted Slimefun blocks to be dropped and drop nothing on the exploded part. This only happens to Slimefun blocks, vanilla blocks will drop just fine. Also this seems to clear block storage data on that coordinate as well.

### 📑 Reproduction Steps

1. Enable `call-explosive-event` on explosive pickaxe
2. Try to break Slimefun blocks
3. See it only drops the one you are hitting with it

### 💡 Expected Behavior

It should drop all the Slimefun blocks

### 📷 Screenshots / Videos

https://youtu.be/OG7TwUCqqzg

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

```
Paper git-Paper-42 (MC: 1.19)
Slimefun DEV - 1015 (git 2c4f886f)
Metrics-Module #28
Java 17

Installed Addons: (2)
  CMILib v1.2.1.0
  CMI v9.2.0.3
```

### 🧭 Other plugins

```
Plugins (7): CMI, CMILib, Multiverse-Core, Slimefun, Vault, VoidGen, WorldEdit
```

iTwins:
## Description

Slimefun blocks didn't drop when exploded because the ``drops`` argument in ``BlockBreakHandler::onExplode`` did nothing.

## Proposed changes

~~Pass the ``SlimefunItem`` to the ``handleEplosion`` method in ``ExplosionsListener.java`` and drop the item there.~~

~~Other changes:~~
~~Removed unused import.~~

~~I didn't change the method signature for ``BlockBreakHandler::onExplode`` since it's a public method and that seemed like a bad idea to change.~~

EDIT: Changed approach completely. Now added drops logic to ``SimpleBlockBreakHandler::onExplode``.

Resolves #3601
Also resolves Slimefun blocks not dropping from explosions, but somehow there was no issue open
 
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/handlers/SimpleBlockBreakHandler.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/handlers/SimpleBlockBreakHandler.java
@@ -10,8 +10,11 @@ import org.bukkit.event.block.BlockBreakEvent;
 import org.bukkit.inventory.ItemStack;
 
 import io.github.thebusybiscuit.slimefun4.api.events.AndroidMineEvent;
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
 import io.github.thebusybiscuit.slimefun4.core.handlers.BlockBreakHandler;
 import io.github.thebusybiscuit.slimefun4.implementation.items.androids.MinerAndroid;
+
+import me.mrCookieSlime.Slimefun.api.BlockStorage;
 
 /**
  * This simple implementation of {@link BlockBreakHandler} will execute the same code
@@ -54,6 +57,9 @@ public abstract class SimpleBlockBreakHandler extends BlockBreakHandler {
     @Override
     public void onExplode(Block b, List<ItemStack> drops) {
         onBlockBreak(b);
+        SlimefunItem sfItem = BlockStorage.check(b);
+        if (sfItem != null)
+            drops.addAll(sfItem.getDrops());
     }
 
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/handlers/SimpleBlockBreakHandler.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/handlers/SimpleBlockBreakHandler.java
@@ -58,8 +58,9 @@ public abstract class SimpleBlockBreakHandler extends BlockBreakHandler {
     public void onExplode(Block b, List<ItemStack> drops) {
         onBlockBreak(b);
         SlimefunItem sfItem = BlockStorage.check(b);
-        if (sfItem != null)
+        if (sfItem != null) {
             drops.addAll(sfItem.getDrops());
+        }
     }
 
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ExplosionsListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ExplosionsListener.java
@@ -9,6 +9,7 @@ import javax.annotation.ParametersAreNonnullByDefault;
 
 import org.bukkit.Material;
 import org.bukkit.block.Block;
+import org.bukkit.entity.Entity;
 import org.bukkit.entity.EntityType;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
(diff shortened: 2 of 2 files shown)


J3fftw1 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/ExplosionsListener.java
```diff
@@ -83,23 +82,23 @@ private void removeResistantBlock(@Nonnull Block block) {
...
+    private void handleExplosion(BlockBreakHandler handler, Block block, SlimefunItem sfItem) {
         if (handler.isExplosionAllowed(block)) {
             BlockStorage.clearBlockInfo(block);
             block.setType(Material.AIR);
 
             List<ItemStack> drops = new ArrayList<>();
             handler.onExplode(block, drops);
 
-            for (ItemStack drop : drops) {
+            for (ItemStack drop : sfItem.getDrops()) {
```
JustAHuman-xD, line 101: This will break stuff right? If any current implementation adds items to that list to be dropped this just no longer drops them.

iTwins: In core Slimefun the ``drops`` argument of ``BlockBreakHandler::onExplode`` is never used, that's why I did it this way. But you're right that this doesn't account for custom handlers in addons. Perhaps a better way to implement this is in ``SimpleBlockBreakHandler::onExplode``? Reverting these changes and adding the following code there.
```java
        SlimefunItem sfItem = BlockStorage.check(b);
        if (sfItem != null)
            drops.addAll(sfItem.getDrops());
```
That would also work (already tested) and shouldn't interfere with current and future implementations.

J3fftw1 (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/handlers/SimpleBlockBreakHandler.java
```diff
@@ -54,6 +57,9 @@ public void onAndroidBreak(AndroidMineEvent e) {
     @Override
     public void onExplode(Block b, List<ItemStack> drops) {
         onBlockBreak(b);
+        SlimefunItem sfItem = BlockStorage.check(b);
+        if (sfItem != null)
+            drops.addAll(sfItem.getDrops());
```
JustAHuman-xD, line 62:
```suggestion
        if (sfItem != null) {
            drops.addAll(sfItem.getDrops());
        }
```

JustAHuman-xD (changes requested):
Code Style guidelines thing, otherwise looks good to me, can we get testing on this, feels like a big change, even if it's small code wise

JustAHuman-xD (approved):
LGTM

Boomer-1:
i built a wall of 5x5 slimefun items. I hit the center with the upgraded explosive pickaxe from fluffy machines and only the center block dropped. however using the core slimefun explosive pickaxe (3x3) everything dropped as expected and stayed slimefun items. creeper explosions and tnt did not drop any slimefun blocks, nor were the blocks damaged. that doesn't seem like that should be the case. if a vanilla block explodes, so should the slimefun block

iTwins:
I just tested this again, and for me the blocks were broken and dropped for the explosive pickaxe, tnt and creepers.
The Fluffy explosive pickaxe code says it shouldn't remove Slimefun blocks, so I guess that's working as intended.
```java
SlimefunItem sfItem = BlockStorage.check(b);

// Don't break SF blocks
if (sfItem != null) {
    return;
}
```

Boomer-1:
in doing additional testing, some slimefun items broke, where others do not. core slimefun that broke were regulators and cargo and core machines. addon blocks that i tested didnt break. so far i can't get any addon blocks to drop.

iTwins:
The way it's implemented now allows devs to choose what behaviour they want. This pr changes ``SimpleBlockBreakHandler`` which is used by core sf. If we wanted the blocks to drop regardless of how ``BlockBreakHandler::onExplode`` is implemented we could drop the blocks in the ``ExplosionsListener`` instead.

Boomer-1:
whichever way the team wants to go. just providing testing results. thanks itwins

iTwins:
And thank you a lot for testing 😁 !!

J3fftw1:
Closing this since it already works in the latest version.

J3fftw1:
it doesnt seem to work for me.
However i think this pr handles it already
https://github.com/Slimefun/Slimefun4/pull/4062
alessio just needs to adress those comments if thats merged we need to rebase this pr and see if its even still needed and/or if it fixes it

PR #3930, opened 2023-07-30
rename Magnesium to Magnesium Dust

iTwins:
## Description

Magnesium

## Proposed changes

Magnesium Dust

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
@@ -520,7 +520,7 @@ public final class SlimefunItems {
     public static final SlimefunItemStack ALUMINUM_DUST = new SlimefunItemStack("ALUMINUM_DUST", Material.SUGAR, "&6Aluminum Dust");
     public static final SlimefunItemStack LEAD_DUST = new SlimefunItemStack("LEAD_DUST", Material.GUNPOWDER, "&6Lead Dust");
     public static final SlimefunItemStack ZINC_DUST = new SlimefunItemStack("ZINC_DUST", Material.SUGAR, "&6Zinc Dust");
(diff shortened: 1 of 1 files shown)


J3fftw1:
This shouldnt be a breaking change anymore since we recently removed the backwards compatibility check but can someone please verify this?

variananora:
Not sure if this is a breaking changes tho, will test it later. 

But personally changing this now will just piss a shit ton of people that already configured their cargo, automation, current storage (especially fluffy barrel/infinity storage), etc, etc.

J3fftw1:
> Not sure if this is a breaking changes tho, will test it later.
> 
> But personally changing this now will just piss a shit ton of people that already configured their cargo, automation, current storage (especially fluffy barrel/infinity storage), etc, etc.

i would say if its not compatible with that its breaking enough to not accept the change.

iTwins:
Shall I just close the PR?

variananora:
Just tested it, changing the name will probably not broke existing cargo/automation setup, infinity and fluffy barrel storage, but it eventually will? I haven't tested this thoroughly.

This is because the item is being renamed back to "Magnesium" if there is already item on it and if there is not it will stay the new name.

Personally, I think this change doesn't really solve anything and probably just create confusion to existing servers.

Sefiraat:
While I agree this change is… ‘correct’ I think it causes more harm than good as it will disable stacking for an item player probably have a butt load of and just cause more grief for server owners/admins. I appreciate the time here however, please don’t get me wrong :). 
As ids match and the items are not marked distinctive they will ‘merge’ with their older counterparts until a new ItemStack is generated at which point the player will have to contend with two different types that vanilla Minecraft won’t, correctly, stack.

iTwins:
> Personally, I think this change doesn't really solve anything and probably just create confusion to existing servers.

I agree, it will probably cause more problems if this is changed and there isn't anything 'broken' in the current situation.

PR #3937, opened 2023-08-06
feat: allow addons to add wiki pages to items

ybw0014:
## Description

Nowadays there are more and more Slimefun addons, but only Slimefun items can have the wiki item in the recipe display. Some addons already have wiki, but players may not be able to know the wiki's existence, unless the developer add the wiki link via FlexItemGroup or using other ways. I think the wiki page item should be available to all the items that have wiki for them.

Tested with https://github.com/Sefiraat/Networks/pull/152.

## Proposed changes

- Add `getWikiURL()` in `SlimefunAddon` to provide a base URL for addon wiki.
- Add `getWikiPage(String)` in `SlimefunItem` to allow items to have wiki pages.
- Deprecate `addOfficialWikipage` in favor of `getWikiPage`.
- Display a different wiki icon for all addons.
- Move the code that setups wiki from `wiki.json` to `WikiUtils`.

N/A

## Checklist


JustAHuman-xD (commented):
I like the idea but limiting the addon to be under one url + a placeholder imo isn't ideal. If there was a way to add a full link to an item on top of this I think it would be a good addition

J3fftw1:
> I like the idea but limiting the addon to be under one url + a placeholder imo isn't ideal. If there was a way to add a full link to an item on top of this I think it would be a good addition

Yea i agree with this, maybe we can have the wiki link be defaulted to the repo as how you have it now but allow it to be overridable with custom links

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
```diff
@@ -895,6 +895,18 @@ public final void addWikiPage(@Nonnull String page) {
...
 
+    /**
+     * This method assign the given wiki URL to this Item.
+     *
+     * @param url
+     *            The associated wiki page URL.
+     */
+    public final void addCustomWikiPage(@Nonnull String url) {
+        Validate.notNull(url, "Wiki URL cannot be null.");
+        Validate.isTrue(getState() != ItemState.UNREGISTERED, "Wiki page can only be added after item has been registered.");
```
J3fftw1, line 906: please use preconditions

src/main/java/io/github/thebusybiscuit/slimefun4/utils/WikiUtils.java
```diff
@@ -0,0 +1,76 @@
...
+     * The formatter will make changes to the wiki page name before it is added to the item.
+     *
+     * @param plugin
+     *          The plugin to load the wiki pages for.
+     * @param formatter
+     *          The formatter to apply to the wiki page name.
+     */
+    public static void setupWiki(@Nonnull Plugin plugin, @Nonnull UnaryOperator<String> formatter) {
+        Validate.notNull(plugin, "The plugin cannot be null");
+        Validate.isTrue(plugin instanceof SlimefunAddon, "The plugin must be a SlimefunAddon");
```
J3fftw1, line 52: preconditions

J3fftw1 (changes requested):

ybw0014:
For anyone would like to test this build using the preview jar, here is the modified Networks that utilized the methods added in this PR.
Download link: 
If you want to compile it by yourself, the PR link is here: https://github.com/Sefiraat/Networks/pull/152

WalshyDev (changes requested):
I'm not fully opposed to giving a way to provide wiki links for addons.

However...

Arbitrary links doesn't feel good. Domains will be bought, not renewed and snatched by others. I think we need to figure out the right approach to this rather than just throw this out there.
We do not want to be the ones responsible for someone getting malware on their system or seeing adult material due to a snatched domain.

ybw0014:
> I'm not fully opposed to giving a way to provide wiki links for addons.
> 
> However...
> 
> Arbitrary links doesn't feel good. Domains will be bought, not renewed and snatched by others. I think we need to figure out the right approach to this rather than just throw this out there. We do not want to be the ones responsible for someone getting malware on their system or seeing adult material due to a snatched domain.

Maybe add a warning message before sending the link in the chat?

J3fftw1:
Warnings don’t mean much and are almost always ignored

ybw0014:
> Warnings don’t mean much and are almost always ignored

It is the player's choice to ignore the warning.

J3fftw1:
i dont think this is the right approach we will still be kept accountable if someone gets a virus or something else.
I do really like this feature, but i do agree with Walshy completely

StarWishsama:
Maintaining a list of trusted wiki domain names could be a compromise, since no official addon wiki provided.

J3fftw1:
Whoops miss clicked

WalshyDev:
I'd like @TheBusyBiscuit's views too

GitHub only definitely helps limit the scope. It doesn't solve the issue but it does help + I like the additional messaging.

J3fftw1:
I remember this being hard denied in the past.

ybw0014:
Yeah I submitted a PR before: https://github.com/Slimefun/Slimefun4/pull/3397

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/SlimefunItem.java
```diff
@@ -70,6 +74,14 @@
  */
 public class SlimefunItem implements Placeable {
 
+    /**
+     * This is the set of allowed wiki domains.
+     */
+    private static final Set<String> ALLOWED_WIKI_DOMAIN = Set.of(
+        "github.com",
+        "github.io"
+    );
```
TheBusyBiscuit, line 83: That's just really nonsensical in my opinion, first off: Anything can be put onto a `github.io` page, so that supposed "security" isn't much.

Second of all, any file can be uploaded to github and linked to this way too.

So I really don't see a point in this filter at all.

TheBusyBiscuit (dismissed):
Still very much against these changes in their current form.
I also find these filters a bit nonsensical.

Especially since addon developers can just do `p.sendMessage(MALICIOUS_LINK)` anytime they want to anyway and/or combine this with a custom FlexItemGroup. So this "line of defense" isn't stopping anyone who wants to abuse it.

I think the clear distinction between official and third-party wikis is a good step in the right direction. But in its current form, I am still opposing this.

ybw0014:
I will add a "wiki control panel" for server owners to control what wiki link the players can access.

Sefiraat:
> I will add a "wiki control panel" for server owners to control what wiki link the players can access.

Thanks for being a good sport while this is all being discussed Yb. Appreciated <3

TheBusyBiscuit:
> I will add a "wiki control panel" for server owners to control what wiki link the players can access.

I feel like if we opt for a control panel it should be a general control panel for various settings concerning addons, not just wiki-specific. Perhaps it may be a good idea to put this PR on hold for now and create such a panel first and then come back to this?

ybw0014:
> > I will add a "wiki control panel" for server owners to control what wiki link the players can access.
> 
> I feel like if we opt for a control panel it should be a general control panel for various settings concerning addons, not just wiki-specific. Perhaps it may be a good idea to put this PR on hold for now and create such a panel first and then come back to this?

Anything u can come up with in the panel besides addon wiki control?

J3fftw1:
Would still love this feature.
its now getting stale and definitely since cookie is gone and isnt looking into this anymore.

Minecraft has its own warning system already inplace for external links,
If that fires im fine with this being in place.

ybw0014:
> Minecraft has its own warning system already inplace for external links, If that fires im fine with this being in place.

Slimefun displays a message that contains the link when you click on the wiki button, so that will be fired.

ybw0014:
As I didnt get any response for panel thing, I just merged commits and resolved merge conflicts.
As Jeff mentioned, Minecraft has its warning system, as we just send the link to users and users will be warned when they click on a link in the chat.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
```diff
@@ -404,7 +404,7 @@ private void onPluginStart() {
     }
 
     @Override
-    public JavaPlugin getJavaPlugin() {
+    public @Nonnull JavaPlugin getJavaPlugin() {
```
J3fftw1, line 407:
```suggestion
    public JavaPlugin getJavaPlugin() {
```

J3fftw1: lets not make the diff any bigger then needed

J3fftw1 (changes requested):

J3fftw1:
Have you tested this with an addon that uses this?
Would love to see the button working.

ybw0014:
> Have you tested this with an addon that uses this? Would love to see the button working.

I tested this earlier with Networks (https://github.com/Sefiraat/Networks/pull/152), now I'm gonna test this again.

ybw0014:
Just tested and it still works. However, the item name seems to be a bit long, should we move some parts into item lore?


image
image

ybw0014:
I changed the wiki button name to 'View this item on the %addon% wiki`, where `%addon%` will be replaced by addon name.
And add a line to show if it is official Slimefun wiki, or third party wiki.

ybw0014:
image
image

JustAHuman-xD (changes requested):
How do we stop another plugin changing the wiki pages of another? 

If an addon or just random plugin wanted to they could loop through the items and change the wiki page right? 

And if they did that to the core slimefun items it would still appear as "official" even if it isn't 

Or you could make a slimefun item have an "official" wiki, if you register the item with the slimefun instance (which you shouldn't do but you could to trick this system)

ybw0014:
good point, maybe once the item has a wiki page, the wiki page link cant be changed further.

ybw0014:
Its been a while.
I added a state check on `addCustomWiki` so the wiki page cannot be added once it is already set.
Also, I use pattern match to check if the wiki URL is from official Slimefun wiki.

I wonder if I should remove `addOfficialWiki` since I don't think any addon would use this.

ybw0014:
Tested again with the modified Networks.

PR #3940, opened 2023-08-08
Fixes duplication glitch with explosive tools in combination with itemsadder

### LINKED ISSUE #3836: Duplication glitch with Explosive Pickaxe
### ❗ Checklist

### 📍 Description

When using the Explosive Pickaxe to break a custom block it will drop the item and block does not break.
This give unlimited items of this block.

### 📑 Reproduction Steps

1. Break custom block with explosive pickaxe
2. item drops but block stays
3. break custom block again
4. repeat

### 💡 Expected Behavior

Break the block and item drops block disappears

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

Paper git-Paper-471 (MCL1.19.4)
Slimefun DEV - 1050 git 22a7f
Metrics Module #28
Java 17

### 🧭 Other plugins

ItemsAdder plugin

J3fftw1:
## Description

This PR resolves a duplication glitch with the Items adder plugin and Slimefun.

## Proposed changes

Disable blockbreak when its a custom block from items adder.

Resolves #3836 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosiveTool.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/tools/ExplosiveTool.java
@@ -6,7 +6,6 @@ import java.util.List;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import dev.lone.itemsadder.api.CustomBlock;
 import org.bukkit.Bukkit;
 import org.bukkit.Effect;
 import org.bukkit.Material;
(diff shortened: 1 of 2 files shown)


J3fftw1:
keeping this on draft for now.
Looking into a way to allow itemsadder blocks to be broken as well.
Comment is standing out in https://github.com/PluginBugs/Issues-ItemsAdder/issues/2777

J3fftw1:
Sadly there doesnt seem to be away for us to actually drop the itemsadder block. This will be a sufficient fix for now. I do want to look into this more but that will be a follow up PR.

TheBusyBiscuit (changes requested):
This will cause a ClassNotFoundException on every server that does not use ItemsAdder :NotLikeThis:

JustAHuman-xD (dismissed):
LGTM (as much as it can, im not familiar with itemsadder api so 🤷, but the code itself is fine)

iTwins:
Don't know if it's possible, but if it is wouldn't the 'right' place to handle this be in the integrations package?

J3fftw1:
> Don't know if it's possible, but if it is wouldn't the 'right' place to handle this be in the integrations package?

good catch

Sfiguz7 (approved):
Same as above, trusting the API to be correct I don't see any issues

JustAHuman-xD (approved):
LGTM same as sfiguz

PR #3943, opened 2023-08-10
Evenly divide generated energy over all connected energy networks

### LINKED ISSUE #3260: Generators power multiple energy network with unnerfed rate

iTwins:
## Description

When a generator was connected to multiple networks it supplied its full generated output to all connected networks.

## Proposed changes

Evenly divide generated energy over all connected energy networks. output / networks.size * networks.size = output, so the generator will now produce the correct amount of energy when connected to multiple networks.

Resolves #3260

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
@@ -244,7 +244,7 @@ public class EnergyNet extends Network implements HologramOwner {
 
             try {
                 Config data = BlockStorage.getLocationInfo(loc);
-                int energy = provider.getGeneratedOutput(loc, data);
+                int energy = provider.getGeneratedOutput(loc, data) / Slimefun.getNetworkManager().getNetworksFromLocation(loc, EnergyNet.class).size();
 
                 if (provider.isChargeable()) {
                     energy = NumberUtils.flowSafeAddition(energy, provider.getCharge(loc, data));


src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
```diff
@@ -288,6 +293,10 @@ private int tickAllCapacitors() {
         return supply;
     }
 
+    private boolean isSaturated() {
```
JustAHuman-xD, line 296: Any reason this is private? Could be useful for Addons. Also you should probably add a javadoc.

iTwins: No reason really. It's public now.

JustAHuman-xD (dismissed):
LGTM but want to get some more testing on this before we send it 👍

iTwins:
I would say that would be intended. If there are more networks connected to a generator that the amount of the Joules it generates, each network would get <1 Joule which is floored to 0. You could also ceil the result of the division. The difference would be 1 Joule per network (max 6) which isn't very significant.

JustAHuman-xD:
I don't think the drop off is a problem. If we were to min it at 1 then that could lead to low level duplication if a generator only generated like 1-5 joules and gets split 6 ways. 

Plus if we want to give reasoning we can just say it's energy resistance being more an issue when using multiple regulators

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/energy/EnergyNet.java
```diff
@@ -244,7 +249,7 @@ private int tickAllGenerators(@Nonnull LongConsumer timings) {
 
             try {
                 Config data = BlockStorage.getLocationInfo(loc);
-                int energy = provider.getGeneratedOutput(loc, data);
+                int energy = provider.getGeneratedOutput(loc, data) / Math.max(1, (int) Slimefun.getNetworkManager().getNetworksFromLocation(loc, EnergyNet.class).stream().filter(Predicate.not(EnergyNet::isSaturated)).count());
```
TheBusyBiscuit, line 252: 1. Don't use Streams
2. Keep in mind this code will run A LOT so performance will suffer from this addition.

iTwins: I refactored the stream to a for loop. Is it better now?

JustAHuman-xD (approved):
LGTM

JustAHuman-xD:
Anyone up for testing this 👀

Boomer-1:
this still needs work. in placing 4 regulators and a lava generator, while all 4 went up nearly even, it was well above the rate that the lava generator produced. each went up 8 j/t, while the gen makes 10 total per tick, which created an excess of 22 per tick. additionally it sped up how fast the generator used up the lava. it's supposed to last 40 seconds, it lasted 10. then the amount of power in each network was way above what it should have been. this needs a lot more work

JustAHuman-xD:
> this still needs work. in placing 4 regulators and a lava generator, while all 4 went up nearly even, it was well above the rate that the lava generator produced. each went up 8 j/t, while the gen makes 10 total per tick, which created an excess of 22 per tick. additionally it sped up how fast the generator used up the lava. it's supposed to last 40 seconds, it lasted 10. then the amount of power in each network was way above what it should have been. this needs a lot more work

Sooooooo not sure how you got those results. This PR can **not** affect how long the lava lasts. And I've tested this previously without issue. Are you sure you used the right _tier_of lava generator?

iTwins:
I found out why. The pr didn't change how long the lava lasts, this was already the behavior before this pr. As for the excess power, that is because I only tested this for machines with no buffer 😅. The generated power is divided evenly ``floor(10/4)`` then stored in the buffer 4 times meaning the buffer increases by 8 every tick. But the whole buffer is still accessible by all networks.
The easiest fix imo is keeping ``EnergyNetProvider``s that implement ``MachineProcessHolder<FuelOperation>`` the same as they are now (4x the full output but fuel goes 4x faster) and ``EnergyNetProvider``s that do not will have their generated output evenly distributed. And if generators have a buffer also evenly distribute that.
The best fix imo would be changing the fuel consumption rate back to expected amount and evenly distributing it aswell, however this would be harder to make.

JustAHuman-xD:
> this still needs work. in placing 4 regulators and a lava generator, while all 4 went up nearly even, it was well above the rate that the lava generator produced. each went up 8 j/t, while the gen makes 10 total per tick, which created an excess of 22 per tick. additionally it sped up how fast the generator used up the lava. it's supposed to last 40 seconds, it lasted 10. then the amount of power in each network was way above what it should have been. this needs a lot more work

The reason the lava went 4x as quick was because each energy net ticked it down. That happens regardless of this pr like iTwins said. Honestly I think that should be the behavior

Boomer-1:
> > this still needs work. in placing 4 regulators and a lava generator, while all 4 went up nearly even, it was well above the rate that the lava generator produced. each went up 8 j/t, while the gen makes 10 total per tick, which created an excess of 22 per tick. additionally it sped up how fast the generator used up the lava. it's supposed to last 40 seconds, it lasted 10. then the amount of power in each network was way above what it should have been. this needs a lot more work
> 
> The reason the lava went 4x as quick was because each energy net ticked it down. That happens regardless of this pr like iTwins said. Honestly I think that should be the behavior

that would be fine if the power spread evenly as it should, and the total produced matched what the fuel would produce if just on one network

JustAHuman-xD:
iTwins if you can vc in a little bit to talk about some of this / make the changes necessary I'd be down.

PR #3944, opened 2023-08-11
change BREAK_BLOCK to PLACE_BLOCK in BlockPlaceEvent

iTwins:
## Description

The ``PlaceBlockEvent`` was logging block breaks

## Proposed changes

change ``Interaction.BREAK_BLOCK`` to ``Interaction.PLACE_BLOCK``

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -86,7 +86,7 @@ public class BlockListener implements Listener {
         
         // TODO: Protection manager is null in testing environment.
         if (!Slimefun.instance().isUnitTest()) {
-            Slimefun.getProtectionManager().logAction(e.getPlayer(), e.getBlock(), Interaction.BREAK_BLOCK);
+            Slimefun.getProtectionManager().logAction(e.getPlayer(), e.getBlock(), Interaction.PLACE_BLOCK);
         }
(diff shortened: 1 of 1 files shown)


JustAHuman-xD (approved):
I can't believe this hasn't been noticed for so long xd

JustAHuman-xD:
It would also be good if we can get some testing to make sure this change is the proper functionality

J3fftw1:
I did notice this before but thought it was intentional and didn’t give it any thoughts

Sefiraat:
It’s likely gone unnoticed because the permissions for breaking and placing normally both go hand-in-hand and for those few servers with a specific distinction it’s less likely someone would trigger the behaviour in the incorrect manner like this.

Sefiraat (approved):
Pending removal of the resting label, looking good

TheBusyBiscuit (approved):
good catch 👍🏻

TheBusyBiscuit:
> It would also be good if we can get some testing to make sure this change is the proper functionality

I don't see the need for testing this as it is mainly just passing the data onto other plugins.
Besides, in dough everything is try/catch wrapped anyway and it will log any issues with its protection modules should there be a problem.

However, I noticed that our dough Logblock module does not implement block placement yet:
https://github.com/baked-libs/dough/blob/main/dough-protection/src/main/java/io/github/bakedlibs/dough/protection/loggers/LogBlockLogger.java#L28-L35
So that should be changed, otherwise only CoreProtect users will notice any difference after this change.

This PR however can be merged in my opinion if that's okay with you.

PR #3953, opened 2023-08-20
fix geo miner voiding resources

iTwins:
## Description

When breaking a geo miner the resource it was currently mining would get voided.

## Proposed changes

Write logic for ``MachineOperation``s to allow for behavior to be implemented when a ``MachineOperation`` is interrupted without finishing. Use this to return the resource to the chunk when a ``MiningOperation`` is interrupted.
I also fixed annotations and Preconditions in the affected classes.

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/machines/MachineOperation.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/machines/MachineOperation.java
@@ -57,4 +57,10 @@ public interface MachineOperation {
         return getRemainingTicks() <= 0;
     }
 
+    /**
+     * This method is called when a {@link MachineOperation} is interrupted before finishing.
+     * Implement to specify behaviour that should happen in this case.
+     */
+    default void cancel() {}
+
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/geo/GEOMiner.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/geo/GEOMiner.java
@@ -7,7 +7,8 @@ import java.util.OptionalInt;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
-import org.apache.commons.lang.Validate;
+import com.google.common.base.Preconditions;
+
 import org.bukkit.ChatColor;
 import org.bukkit.Material;
 import org.bukkit.block.Block;
@@ -77,7 +78,7 @@ public class GEOMiner extends SlimefunItem implements RecipeDisplayItem, EnergyN
     }
 
     @Override
-    public MachineProcessor<MiningOperation> getMachineProcessor() {
+    public @Nonnull MachineProcessor<MiningOperation> getMachineProcessor() {
         return processor;
     }
 
@@ -121,8 +122,8 @@ public class GEOMiner extends SlimefunItem implements RecipeDisplayItem, EnergyN
      * 
      * @return This method will return the current instance of {@link GEOMiner}, so that can be chained.
      */
-    public final GEOMiner setCapacity(int capacity) {
-        Validate.isTrue(capacity > 0, "The capacity must be greater than zero!");
+    public final @Nonnull GEOMiner setCapacity(int capacity) {
+        Preconditions.checkArgument(capacity > 0, "The capacity must be greater than zero!");
 
         if (getState() == ItemState.UNREGISTERED) {
             this.energyCapacity = capacity;
@@ -140,8 +141,8 @@ public class GEOMiner extends SlimefunItem implements RecipeDisplayItem, EnergyN
      * 
      * @return This method will return the current instance of {@link GEOMiner}, so that can be chained.
      */
-    public final GEOMiner setProcessingSpeed(int speed) {
-        Validate.isTrue(speed > 0, "The speed must be greater than zero!");
+    public final @Nonnull GEOMiner setProcessingSpeed(int speed) {
+        Preconditions.checkArgument(speed > 0, "The speed must be greater than zero!");
 
         this.processingSpeed = speed;
         return this;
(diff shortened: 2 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/MiningOperation.java
```diff
@@ -22,20 +28,24 @@ public class MiningOperation implements MachineOperation {
...
 
+    private final GEOResource resource;
+    private final Block block;
     private final int totalTicks;
     private int currentTicks = 0;
 
-    public MiningOperation(@Nonnull ItemStack result, int totalTicks) {
-        Validate.notNull(result, "The result cannot be null");
-        Validate.isTrue(totalTicks >= 0, "The amount of total ticks must be a positive integer or zero, received: " + totalTicks);
+    public MiningOperation(@Nonnull GEOResource resource, Block block, int totalTicks) {
```
J3fftw1, line 36: this is a public api, this will break other addons using this

J3fftw1 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/MiningOperation.java
iTwins: Can I make a ``class GEOMiningOperation extends MiningOperation``, then make the geo miner use that?

src/main/java/io/github/thebusybiscuit/slimefun4/core/machines/MachineOperation.java
```diff
@@ -57,4 +57,10 @@ default boolean isFinished() {
         return getRemainingTicks() <= 0;
     }
 
+    /**
+     * This method is called when a {@link MachineOperation} is interrupted before finishing.
+     * Implement to specify behaviour that should happen in this case.
+     */
+    default void cancel() {}
```
WalshyDev, line 64: Can we give this a better name? onBreak, onInterrupt, onCancel, I don't know. It's an event triggered function though, it doesn't cancel operation itself.

iTwins: renamed to ``onCancel``

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/geo/GEOMiner.java
```diff
@@ -217,21 +216,18 @@ public void onBlockBreak(@Nonnull Block b) {
         };
     }
 
-    @Nonnull
     @Override
-    public int[] getInputSlots() {
+    public @Nonnull int[] getInputSlots() {
```
JustAHuman-xD, line 220: Null isn't passable on primitive but ig maybe it is for primitive arrays? Not sure but either way this is.... weird

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/geo/GEOMiner.java
```diff
@@ -217,21 +216,18 @@ public void onBlockBreak(@Nonnull Block b) {
...
         return new int[0];
     }
 
-    @Nonnull
     @Override
-    public int[] getOutputSlots() {
+    public @Nonnull int[] getOutputSlots() {
```
JustAHuman-xD, line 225: Null isn't passable on primitive but ig maybe it is for primitive arrays? Not sure but either way this is.... weird

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/operations/GEOMiningOperation.java
```diff
@@ -0,0 +1,46 @@
...
+import io.github.thebusybiscuit.slimefun4.implementation.items.geo.GEOMiner;
+
+/**
+ * This {@link MachineOperation} represents a {@link GEOMiner}
+ * mining a {@link GEOResource}.
+ *
+ * @author iTwins
+ *
+ * @see GEOMiner
+ *
```
JustAHuman-xD, line 22:
```suggestion
```

JustAHuman-xD (changes requested):
I wonder if this on cancel could potentially be used to fix the recipes in machines getting voided on server stop/restart

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/geo/GEOMiner.java
iTwins: the compiler lets you return null for an int array if you want.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/geo/GEOMiner.java
iTwins: the compiler lets you return null for an int array if you want.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/geo/GEOMiner.java
```diff
@@ -274,7 +270,7 @@ public boolean onClick(Player p, int slot, ItemStack cursor, ClickAction action)
 
                 @Override
                 public boolean onClick(InventoryClickEvent e, Player p, int slot, ItemStack cursor, ClickAction action) {
-                    return cursor == null || cursor.getType() == null || cursor.getType() == Material.AIR;
```
Sefiraat, line 277: I know that this is likely marked as Nonnull/NotNull yet it is possible for this to return null. It's rare and super annoying due to the warnings this raises in the IDE, but please revert this line.

iTwins: Ah I see, you are correct. Reverted.

JustAHuman-xD (dismissed):
LGTM

iTwins:
> I wonder if this on cancel could potentially be used to fix the recipes in machines getting voided on server stop/restart

Probably yes, if made correctly. If we add the ``BlockPosition`` from ``MachineProcessor::endOperation`` as argument in ``onCancel``, cause ``MachineOperation``s don't have their location in the object (and they shouldn't need to).

JustAHuman-xD (approved):
LGTM

J3fftw1 (dismissed):

J3fftw1 (approved):

PR #3963, opened 2023-09-03
Add network signal insulators

iTwins:
## Description

A feature for energy and cargo networks to stop nodes from connecting.

## Proposed changes

Add 2 items, 1 for energy and 1 for cargo, that you can place in between network nodes to stop them form connecting.
To accomplish this the following is necessary:
- Add ``INSULATOR`` to the network and energy component enums to classify them.
- Stop discovering neighbors in the current direction if an insulator is found.
- If an insulator is placed inside the network, rebuild it.
- if an insulator is broken inside the network update all its neighbors that are connected to the network.

Implements approved suggestion [#1652](https://discord.com/channels/565557184348422174/627048923122499584/933374758890926130)

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/api/network/Network.java
```diff
@@ -181,6 +182,8 @@ public boolean connectsTo(@Nonnull Location l) {
             return NetworkComponent.CONNECTOR;
         } else if (terminusNodes.contains(l)) {
             return NetworkComponent.TERMINUS;
+        } else if (insulatorNodes.contains(l)) {
+            return NetworkComponent.INSULATOR;
```
J3fftw1, line 186: would a switch case be better here?

J3fftw1 (changes requested):
quick review

J3fftw1:
I like this feature, only worried about performance on this.
Can we maybe get some testing on performance on it.

Also we need some general testing on this. This is a major change for 2 essential things in Slimefun.
We should block this till it has been tested a bunch

Sefiraat:
Agreed, needs some good testing :)

J3fftw1:
Can we also get a toggle on this maybe within a gui.
I think the reason few years ago was that people wanted this because the matter fabracitor uses so much energy and more machines do these days.
A toggle would solve this problem.

iTwins:
> Can we also get a toggle on this maybe within a gui. I think the reason few years ago was that people wanted this because the matter fabracitor uses so much energy and more machines do these days. A toggle would solve this problem.

What about right clicking the insulator to enable/disable it?

Also yes, please test this a lot. I have tried my best to break this already, but I would feel a lot safer if someone else tested this as well.

J3fftw1:
Yes that’s fine by me as well

Boomer-1:
both cargo and energy insulators seem to be working as intended. my suggestion would be alter the appearance as the only difference is a white dot on the top.

iTwins:
Have you tested this with a large network or small network?
Also if anyone if up for designing better textures, please send the file or the base64 encoded string :DD.

Boomer-1:
i didnt have a large network setup to test on. it was a pretty basic one

iTwins:
Alright, good to know. Maybe someone with a big network setup can also test this a little?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/HeadTexture.java
```diff
@@ -69,11 +69,13 @@ public enum HeadTexture {
     ELECTRIC_PRESS("8d5cf92bc79ec19f4106441affff1406a1367010dcafb197dd94cfca1a6de0fc"),
     ENERGY_REGULATOR("d78f2b7e5e75639ea7fb796c35d364c4df28b4243e66b76277aadcd6261337"),
     ENERGY_CONNECTOR("1085e098756b995b00241644089c55a8f9acde35b9a37785d5e057a923613b"),
+    ENERGY_INSULATOR("7f44bee34a0ead8af93260181747e573456abb54385bf234102ef14c4f2216d6"),
     NETHER_ICE("3ce2dad9baf7eaba7e80d4d0f9fac0aab01a76b12fb71c3d2af2a16fdd4c7383"),
     ENRICHED_NETHER_ICE("7c818aa13aabc7294838d21caac057e97bd8c89641a0c0f8a55442ff4e27"),
     NETHER_ICE_COOLANT_CELL("8d3cd412555f897016213e5d6c7431b448b9e5644e1b19ec51b5316f35840e0"),
     CARGO_MANAGER("e510bc85362a130a6ff9d91ff11d6fa46d7d1912a3431f751558ef3c4d9c2"),
     CARGO_CONNECTOR_NODE("07b7ef6fd7864865c31c1dc87bed24ab5973579f5c6638fecb8dedeb443ff0"),
+    CARGO_INSULATOR("ac8962e46ee8ba5517f3f524175391fca35af4ca3c01995a8da2c8898573c50c"),
```
JustAHuman-xD, line 78:
```suggestion
    CARGO_INSULATOR("c1620b325a36677d556d6b7588ae2d340fdc76bc99d77db4268422748e72c093"),
```

image
Made a new texture for the cargo insulator

src/main/java/io/github/thebusybiscuit/slimefun4/api/network/Network.java
```diff
@@ -241,6 +252,32 @@ private void discoverNeighbors(@Nonnull Location l) {
         discoverNeighbors(l, 0.0, 0.0, -1.0);
     }
 
+    private void updateNeighbors(@Nonnull Location l, double xDiff, double yDiff, double zDiff) {
+        for (int i = 1; i <= getRange(); i++) {
+            Location newLocation = l.clone().add(i * xDiff, i * yDiff, i * zDiff);
```
Sfiguz7, line 257: Slightly confused by this. What is this method supposed to do?

As it is written I don't see it ever being useful to pass anything but one unitary value (with sign) and two zeros, like you do below, therefore the method only updatesNeighbors in one direction.
But then the question arises on whether this format even makes sense, it would be a lot more readable (though not as compact) if it accepted a BlockFace at that point.

Was the idea just to avoid having to "decode" what versor each BlockFace refers to?

iTwins: This method checks if any Locations need to be added to the network when an insulator is broken. 

It indeed is intended to only check the 6 directions in which a network can connect. The reason I went with this signature is because the method is similar to discoverNeighbors and that used the same way of passing the arguments. If you'd rather have the method take a BlockFace would be happy to change it. Then perhaps discoverNeighbors should also be changed to take a BlockFace for consistency sake.

Sfiguz7: Oh my god I had not even noticed the other one. If you could make both accept a blockface that would be awesome

PR #3969, opened 2023-09-09
Remove Lore Check from Slimefun Guide

JustAHuman-xD:
## Description
When using the Slimefun Guide it does isItemSimilar and it respects lore, this was most likely due to the backwards compatibility and is since no longer needed. By removing the lore check we allow for any runes (core or addon) to work on the guide without breaking it, e.g. Soulbound (Core), Fireproof (Addon)

## Proposed changes
- No longer check lore in `SlimefunGuideListener#tryOpenGuide` 

N/A (I think)

## Checklist

- [] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values


TheBusyBiscuit (approved):
LGTM

PR #3976, opened 2023-09-19
fix breaking sf block with not unlocked item duping contents

iTwins:
## Description

I you broke a Slimefun block with a not unlocked item the contents of the inventory of the block would still drop. This is because the block handler would still get called no matter the results of the tool handler.

## Proposed changes

Move the tool handler code outside of the ``if (!e.isCancelled())`` block. That way if the tool handler cancels the event, the block handler won't get called. This should have no side effects since ``ignoreCancelled`` is true and there is nothing that cancels the event above the changed code.

https://discord.com/channels/565557184348422174/565569196218384385/1153792378327535738

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -141,15 +141,15 @@ public class BlockListener implements Listener {
             }
         }
 
+        int fortune = getBonusDropsWithFortune(item, e.getBlock());
+        List<ItemStack> drops = new ArrayList<>();
+
+        if (!item.getType().isAir()) {
+            callToolHandler(e, item, fortune, drops);
+        }
+
         if (!e.isCancelled()) {
             checkForSensitiveBlockAbove(e.getPlayer(), e.getBlock(), item);
-
-            int fortune = getBonusDropsWithFortune(item, e.getBlock());
-            List<ItemStack> drops = new ArrayList<>();
-
-            if (!item.getType().isAir()) {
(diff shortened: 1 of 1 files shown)


JustAHuman-xD (approved):
LGTM I think? I can't think of anything that might break with this but I'm not 100% sure

J3fftw1 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -141,15 +141,15 @@ public void onBlockBreak(BlockBreakEvent e) {
             }
         }
 
-        if (!e.isCancelled()) {
-            checkForSensitiveBlockAbove(e.getPlayer(), e.getBlock(), item);
+        int fortune = getBonusDropsWithFortune(item, e.getBlock());
```
WalshyDev, line 144: Can we move fortune down to just above callToolHandler?
There's no reason for it to be this high.

JustAHuman-xD: I mean it does have to be right? It's used in the next method call, it looks like it needs to be dropped into that if block but otherwise it's in the right place, at least that's how it looks anyways

iTwins: Done.

JustAHuman-xD (approved):
LGTM

J3fftw1 (approved):

WalshyDev (approved):

PR #3988, opened 2023-10-05
Add config option for radiation

iTwins:
## Description

There was some discussion about people wanting to disable radiation.

## Proposed changes

Add a config option that controls whether the radiation task will be scheduled.

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
@@ -367,7 +367,9 @@ public final class Slimefun extends JavaPlugin implements SlimefunAddon {
         // Armor Update Task
         if (config.getBoolean("options.enable-armor-effects")) {
             new SlimefunArmorTask().schedule(this, config.getInt("options.armor-update-interval") * 20L);
-            new RadiationTask().schedule(this, config.getInt("options.radiation-update-interval") * 20L);
+            if (config.getBoolean("options.enable-radiation")) {
+                new RadiationTask().schedule(this, config.getInt("options.radiation-update-interval") * 20L);
+            }
             new RainbowArmorTask().schedule(this, config.getInt("options.rainbow-armor-update-interval") * 20L);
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
```diff
@@ -367,7 +367,9 @@ private void onPluginStart() {
         // Armor Update Task
         if (config.getBoolean("options.enable-armor-effects")) {
             new SlimefunArmorTask().schedule(this, config.getInt("options.armor-update-interval") * 20L);
-            new RadiationTask().schedule(this, config.getInt("options.radiation-update-interval") * 20L);
+            if (config.getBoolean("options.enable-radiation")) {
+                new RadiationTask().schedule(this, config.getInt("options.radiation-update-interval") * 20L);
+            }
```
TheBusyBiscuit, line 372: Should this be moved outside of the enable-armor-effects condition?

iTwins: The reason this was inside the enable-armor-effects condition is probably that the hazmat suit wouldn't work if armor effects were disabled, making it impossible to handle radioactive items if the radiation task wouldn't also be disabled.

J3fftw1 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
TheBusyBiscuit: Therefore `options.enable-armor-effects` = false and `options.enable-radiation` = true is an invalid state.
I would strongly suggest to add a console warning in this case and warn that radiation will not work unless armor effects are enabled.

iTwins: Added a warning for the mentioned state

JustAHuman-xD (approved):
lgtm

J3fftw1 (approved):

WalshyDev (approved):

PR #3998, opened 2023-10-14
Optional player for SlimefunItemSpawnEvent

TheSilentPro:
## Description
This adds Optional player field for SlimefunItemSpawnEvent.
Some of the spawn reasons are tied to a player and there is not way to get it currently.

## Proposed changes
This adds a new constructor to the class that accepts a player field. This field is null by default.

n/a

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemSpawnEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemSpawnEvent.java
@@ -1,10 +1,14 @@
 package io.github.thebusybiscuit.slimefun4.api.events;
 
+import java.util.Optional;
+
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 import org.apache.commons.lang.Validate;
 import org.bukkit.Location;
+import org.bukkit.entity.Player;
 import org.bukkit.event.Cancellable;
 import org.bukkit.event.Event;
 import org.bukkit.event.HandlerList;
@@ -29,6 +33,7 @@ public class SlimefunItemSpawnEvent extends Event implements Cancellable {
     private ItemStack itemStack;
     private boolean cancelled;
     private final ItemSpawnReason itemSpawnReason;
+    private final Player player;
 
     @ParametersAreNonnullByDefault
     public SlimefunItemSpawnEvent(Location location, ItemStack itemStack, ItemSpawnReason itemSpawnReason) {
@@ -36,6 +41,25 @@ public class SlimefunItemSpawnEvent extends Event implements Cancellable {
         this.itemStack = itemStack;
         this.itemSpawnReason = itemSpawnReason;
         this.cancelled = false;
+        this.player = null;
+    }
+
+    @ParametersAreNonnullByDefault
+    public SlimefunItemSpawnEvent(@Nullable Player player, Location location, ItemStack itemStack, ItemSpawnReason itemSpawnReason) {
+        this.location = location;
+        this.itemStack = itemStack;
+        this.itemSpawnReason = itemSpawnReason;
+        this.cancelled = false;
+        this.player = player;
+    }
+
+    /**
+     * Optionally returns the {@link Player} responsible for this spawn reason.
+     *
+     * @return The player responsible if applicable.
+     */
+    public Optional<Player> getPlayer() {
+        return Optional.ofNullable(player);
     }
 
     /**
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -610,6 +610,23 @@ public final class SlimefunUtils {
         }
     }
 
+    public static @Nullable Item spawnItem(Player player, Location loc, ItemStack item, ItemSpawnReason reason, boolean addRandomOffset) {
+        SlimefunItemSpawnEvent event = new SlimefunItemSpawnEvent(player, loc, item, reason);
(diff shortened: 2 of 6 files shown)


TheSilentPro:
I forgot to make a new branch and comited to master by accident. Sorry

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemSpawnEvent.java
```diff
@@ -29,13 +33,33 @@ public class SlimefunItemSpawnEvent extends Event implements Cancellable {
...
+        this.cancelled = false;
+        this.player = player;
+    }
+
+    /**
+     * Optionally returns the {@link Player} responsible for this spawn reason.
+     *
+     * @return The player responsible if applicable.
+     */
+    public Optional<Player> getPlayer() {
```
TheBusyBiscuit, line 61:
```suggestion
    public @Nonnull Optional<Player> getPlayer() {
```

TheBusyBiscuit (commented):
Looks good but `player` might be a bit too ambiguous. One could think that that only this player can pick it up or whatever. Perhaps "causedBy" might be a better name here.

But it is also fine to stay as is, just a suggestion. 👍🏻

TheSilentPro:
> Looks good but `player` might be a bit too ambiguous. One could think that that only this player can pick it up or whatever. Perhaps "causedBy" might be a better name here.
> 
> But it is also fine to stay as is, just a suggestion. 👍🏻

I think player would fit fine, either way theres a javadoc on it.

J3fftw1 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemSpawnEvent.java
```diff
@@ -29,13 +33,33 @@ public class SlimefunItemSpawnEvent extends Event implements Cancellable {
...
     private final ItemSpawnReason itemSpawnReason;
+    private final Player player;
 
     @ParametersAreNonnullByDefault
     public SlimefunItemSpawnEvent(Location location, ItemStack itemStack, ItemSpawnReason itemSpawnReason) {
         this.location = location;
         this.itemStack = itemStack;
         this.itemSpawnReason = itemSpawnReason;
         this.cancelled = false;
+        this.player = null;
```
JustAHuman-xD, line 44: You can replace this with a `this()` call

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -610,6 +610,23 @@ public static boolean canPlayerUseItem(@Nonnull Player p, @Nullable ItemStack it
...
+
+            if (addRandomOffset) {
+                return world.dropItemNaturally(event.getLocation(), event.getItemStack());
+            } else {
+                return world.dropItem(event.getLocation(), event.getItemStack());
+            }
+        } else {
+            return null;
+        }
+    }
```
JustAHuman-xD, line 628: This is a lot of repeated code for no reason. You should instead add the player parameter to the original method and then add an overflow.

JustAHuman-xD (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
TheSilentPro: That was my first thought BUT: That would change the signature making it a breaking change.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemSpawnEvent.java
TheSilentPro: No, that would require me adding a Player field to the other constructor (see your other comment on why that can't be)

TheBusyBiscuit, line 44: You can simplify this by calling the other constructor using `this(...)`

TheBusyBiscuit (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
TheBusyBiscuit: @JustAHuman-xD is right, way too much repeated code.
You can simplify this by changing the original method to just call this one with player = null

TheSilentPro:
Resolved

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -589,12 +589,14 @@ public static boolean canPlayerUseItem(@Nonnull Player p, @Nullable ItemStack it
      *            The {@link ItemSpawnReason} why the item is being dropped
      * @param addRandomOffset
      *            Whether a random offset should be added (see {@link World#dropItemNaturally(Location, ItemStack)})
+     * @param player
+     *            The player that caused this {@link org.bukkit.event.entity.ItemSpawnEvent}
```
JustAHuman-xD, line 593: Qualifier should be removed

JustAHuman-xD (changes requested):

JustAHuman-xD (approved):
LGTM

JustAHuman-xD (approved):
LGTM, only thing that you *could* but don't need to change is in some places you use e.getPlayer() which is called multiple times within that code block so you could define a variable for it but it's not a big deal

Sfiguz7 (approved):
Justin's point makes sense but I agree it's fine as is. Just acknowledging before I merge

PR #4002, opened 2023-10-20
Research unlock with vault economy provider

misdocumeno:
## Description
Adds the option to use money instead of experience to unlock research, without major changes in the codebase or config files.

## Proposed changes
I added two new keys to config.yml inside researches. `enable-vault-economy` and `economy-price-multiplier`. If `enable-vault-economy` is true, money will be used instead of experience, and to avoid having to define the prices again for each investigation, the price is calculated by doing xp levels cost times `economy-price-multiplier`, that way the ratio between prices is the same as the previous exp cost system.
I created a VaultIntegration class with a few methods, in case anyone wants to mess with the economy in the future.
The lore that shows the cost of unlocking a research shows the price formatted correctly as a currency.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/researches/Research.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/researches/Research.java
@@ -249,7 +249,9 @@ public class Research implements Keyed {
         }
 
         boolean creativeResearch = p.getGameMode() == GameMode.CREATIVE && Slimefun.getRegistry().isFreeCreativeResearchingEnabled();
-        return creativeResearch || p.getLevel() >= cost;
+        boolean economyResearch = Slimefun.getIntegrations().isVaultInstalled() && Slimefun.getIntegrations().getVaultIntegration().hasBalanceForResearch(p, cost);
+        boolean expResearch = !Slimefun.getRegistry().isVaultEconomyEnabled() && p.getLevel() >= cost;
+        return creativeResearch || economyResearch || expResearch;
     }
 
     /**
--- a/src/main/resources/config.yml
+++ b/src/main/resources/config.yml
@@ -31,6 +31,8 @@ researches:
   free-in-creative-mode: true
   enable-fireworks: true
   disable-learning-animation: false
+  enable-vault-economy: false
+  economy-price-multiplier: 1000.0
 
 URID:
   info-delay: 3000
--- a/pom.xml
+++ b/pom.xml
@@ -513,6 +513,12 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.github.MilkBowl</groupId>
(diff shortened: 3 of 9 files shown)


ybw0014:
Duplicate of https://github.com/Slimefun/Slimefun4/pull/3942

misdocumeno:
> Duplicate of #3942

not the same thing, this implementation is much simpler, and it only adds two new optional settings, the other one creates an unlock provider and what not, i guess it depends on the way you want to add this feature, this is quick and simple, the other one is more scalable probably

src/main/resources/config.yml
```diff
@@ -31,6 +31,8 @@ researches:
   free-in-creative-mode: true
   enable-fireworks: true
   disable-learning-animation: false
+  enable-vault-economy: false
+  economy-price-multiplier: 1000.0
```
TheBusyBiscuit, line 35: I think this should be grouped and simplified
```suggestion
  vault-integration
    enabled: false
    cost-multiplier: 1000.0
```

src/main/java/io/github/thebusybiscuit/slimefun4/api/researches/Research.java
```diff
@@ -249,7 +249,9 @@ public boolean canUnlock(@Nonnull Player p) {
         }
 
         boolean creativeResearch = p.getGameMode() == GameMode.CREATIVE && Slimefun.getRegistry().isFreeCreativeResearchingEnabled();
-        return creativeResearch || p.getLevel() >= cost;
+        boolean economyResearch = Slimefun.getIntegrations().isVaultInstalled() && Slimefun.getIntegrations().getVaultIntegration().hasBalanceForResearch(p, cost);
+        boolean expResearch = !Slimefun.getRegistry().isVaultEconomyEnabled() && p.getLevel() >= cost;
+        return creativeResearch || economyResearch || expResearch;
```
TheBusyBiscuit, line 254: This is a method with high traffic, so think about if we really need to calculate all of this everytime...
If you split this into smaller methods then the `||` operator would ensure that the sub-methods are not called unnecessarily.

TheBusyBiscuit (changes requested):

TheBusyBiscuit:
> > Duplicate of #3942
> 
> not the same thing, this implementation is much simpler, and it only adds two new optional settings, the other one creates an unlock provider and what not, i guess it depends on the way you want to add this feature, this is quick and simple, the other one is more scalable probably

Agreed.

Also we don't really mark PRs as duplicates unless absolutely obvious. The same feature can be implement in numerous ways and we may choose one implementation over the other for reasons.

src/main/resources/config.yml
misdocumeno: do you mean you want `vault-integration` to be inside `researches`? or you want it at the top level?

PR #4018, opened 2023-11-07
Add Horizontal Offset support to Holograms

test137E29B:
## Description

Hologram projectors are really nice things to use, but they lack horizontal offset support meaning any design without a central block annoyingly has an offset hologram. This PR adds support for them to be up to 32 blocks away horizontally. This offers more options to hide the projector in the wall or ceiling, while also allowing them to be centered nicely along any axis.

## Proposed changes

- Add X offset to hologram projectors
- Add Z offset to hologram projectors
- Clean up code a little with the new additions

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
@@ -153,11 +153,7 @@ public class HologramProjector extends SlimefunItem implements HologramOwner {
             double offsetY = getYOffset(projector);
             double offsetZ = getZOffset(projector);
 
-            if (offsetX < -MAX_HORIZONTAL_OFFSET) {
-                offsetX = -MAX_HORIZONTAL_OFFSET;
-            } else if (offsetX > MAX_HORIZONTAL_OFFSET) {
-                offsetX = MAX_HORIZONTAL_OFFSET;
-            }
+            offsetX = Math.min(Math.max(offsetX, -MAX_HORIZONTAL_OFFSET), MAX_HORIZONTAL_OFFSET);
             
             ArmorStand hologram = getArmorStand(projector, true);
             Location l = new Location(projector.getWorld(), projector.getX() + offsetX, projector.getY() + offsetY, projector.getZ() + offsetZ);
@@ -189,11 +185,7 @@ public class HologramProjector extends SlimefunItem implements HologramOwner {
             double offsetY = getYOffset(projector);
             double offsetZ = NumberUtils.reparseDouble(getZOffset(projector) + (action.isRightClicked() ? -0.1F : 0.1F));
 
-            if (offsetZ < -MAX_HORIZONTAL_OFFSET) {
-                offsetZ = -MAX_HORIZONTAL_OFFSET;
-            } else if (offsetZ > MAX_HORIZONTAL_OFFSET) {
-                offsetZ = MAX_HORIZONTAL_OFFSET;
-            }
+            offsetZ = Math.min(Math.max(offsetZ, -MAX_HORIZONTAL_OFFSET), MAX_HORIZONTAL_OFFSET);
 
             ArmorStand hologram = getArmorStand(projector, true);
             Location l = new Location(projector.getWorld(), projector.getX() + offsetX, projector.getY() + offsetY, projector.getZ() + offsetZ);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -119,14 +147,59 @@ private void openEditor(@Nonnull Player p, @Nonnull Block projector) {
...
+        menu.addMenuClickHandler(1, (pl, slot, item, action) -> {
+            double offsetX = NumberUtils.reparseDouble(getXOffset(projector) + (action.isRightClicked() ? -0.1F : 0.1F));
+            double offsetY = getYOffset(projector);
+            double offsetZ = getZOffset(projector);
+
+            if (offsetX < -MAX_HORIZONTAL_OFFSET) {
+                offsetX = -MAX_HORIZONTAL_OFFSET;
+            } else if (offsetX > MAX_HORIZONTAL_OFFSET) {
+                offsetX = MAX_HORIZONTAL_OFFSET;
+            }
```
JustAHuman-xD, line 160: Math.max and Math.min should be used here, there might also be a method to do both

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -119,14 +147,59 @@ private void openEditor(@Nonnull Player p, @Nonnull Block projector) {
...
+        menu.addMenuClickHandler(1, (pl, slot, item, action) -> {
+            double offsetX = getXOffset(projector);
+            double offsetY = getYOffset(projector);
+            double offsetZ = NumberUtils.reparseDouble(getZOffset(projector) + (action.isRightClicked() ? -0.1F : 0.1F));
+
+            if (offsetZ < -MAX_HORIZONTAL_OFFSET) {
+                offsetZ = -MAX_HORIZONTAL_OFFSET;
+            } else if (offsetZ > MAX_HORIZONTAL_OFFSET) {
+                offsetZ = MAX_HORIZONTAL_OFFSET;
+            }
```
JustAHuman-xD, line 196: Math.max and Math.min should be used here, there might also be a method to do both

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -47,6 +48,13 @@
 public class HologramProjector extends SlimefunItem implements HologramOwner {
 
     private static final String OFFSET_PARAMETER = "offset";
+    private static final String OFFSET_PARAMETER_X = "offset-x";
+    private static final String OFFSET_PARAMETER_Z = "offset-z";
+    private static final double MAX_HORIZONTAL_OFFSET = 32.0D; // Up to 2 chunks away in any direction horizontally
```
JustAHuman-xD, line 53: I feel like this might be too big, how this works with chunkloading already worries me, allowing it to go this far only worsens that

JustAHuman-xD: If getting the armorstand causes the chunk to load, which I think it does (could be totally wrong tho), this would now act as a chunk loader

JustAHuman-xD (changes requested):
Not sure about this one

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
```diff
@@ -119,14 +147,59 @@ private void openEditor(@Nonnull Player p, @Nonnull Block projector) {
...
+            BlockStorage.addBlockInfo(projector, OFFSET_PARAMETER_X, String.valueOf(offsetX));
+            openEditor(pl, projector);
+            return false;
+        });
+
+        menu.addItem(1, new CustomItemStack(Material.CLOCK, "&7Offset Y: &e" + getYOffset(projector) + 1.0D, "", "&fLeft Click: &7+0.1", "&fRight Click: &7-0.1"));
         menu.addMenuClickHandler(1, (pl, slot, item, action) -> {
-            double offset = NumberUtils.reparseDouble(Double.valueOf(BlockStorage.getLocationInfo(projector.getLocation(), OFFSET_PARAMETER)) + (action.isRightClicked() ? -0.1F : 0.1F));
+            double offsetX = getXOffset(projector);
+            double offsetY = NumberUtils.reparseDouble(getYOffset(projector) + (action.isRightClicked() ? -0.1F : 0.1F));
```
JustAHuman-xD, line 174: Is there any limit to the yOffset like how you have the x and z limited? If not I think one should be added for both consistency but also just it makes sense

test137E29B: Currently in SF there's no limit, and realistically anything over 10 blocks away is annoyingly hard to see to line up anyway. I wouldn't want to add a limit on something currently unlimited either just incase someone clicks the button to move it up 0.1 and suddenly it's 10 blocks higher and can't go back down

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/HologramProjector.java
JustAHuman-xD: Alr in that case it should probably be fine 👍

PR #4021, opened 2023-11-10
Add Yaw to GPS Waypoint Location

test137E29B:
## Description
Currently GPS Waypoints use the Block's Location (and thus Yaw). This means the player always is south facing when spawning as the default is 0.

## Proposed changes
Preserve the Yaw of the player when creating GPS Marker Waypoints.

N?A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/gps/GPSMarkerTool.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/gps/GPSMarkerTool.java
@@ -36,7 +36,11 @@ public class GPSMarkerTool extends SimpleSlimefunItem<ItemUseHandler> implements
 
             if (e.getClickedBlock().isPresent()) {
                 Block b = e.getClickedBlock().get().getRelative(e.getClickedFace());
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/gps/GPSMarkerTool.java
```diff
@@ -36,7 +36,11 @@ public ItemUseHandler getItemHandler() {
 
             if (e.getClickedBlock().isPresent()) {
                 Block b = e.getClickedBlock().get().getRelative(e.getClickedFace());
-                Slimefun.getGPSNetwork().createWaypoint(e.getPlayer(), b.getLocation());
+                Location l = b.getLocation();
+                Location locationWithYaw = new Location(
+                    l.getWorld(), l.getX(), l.getY(), l.getZ(), e.getPlayer().getLocation().getYaw(), 0.0f
+                );
```
JustAHuman-xD, line 42: Rather than making a new location aren't you able to call setYaw on l? Block locations don't change the block so I think that's all that's required here

JustAHuman-xD (approved):
lgtm

PR #4048, opened 2023-12-09
Hotfix for the Sensistive Block Check

JustAHuman-xD:
## Description
A bug slipped through the craps on the sensitive blocks PR

## Proposed changes
gets and then re sets the block data of the block for sensitive blocks

## Checklist


JustAHuman-xD:
I wouldn't say I've *fully* tested these changes but I made sure that it fixed the bug and that the bug was legit

Sfiguz7 (approved):

J3fftw1 (approved):

PR #3920, opened 2023-07-25
Adds a TalismanActivatedEvent which fires when a Player uses a Talisman.

Sniperkaos:
## Description
I found it odd that there wasn't an event to listen for when Talismans are activated, and thought it would be
helpful for one to be added.

## Proposed changes
Added TalismanActivationEvent, which extends PlayerEvent.
It is fired when the static function activateTalisman is called.

## Checklist
- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [x] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.16.* - 1.20.*).
- [ ] I followed the existing code standards and didn't mess up the formatting.
- [x] I did my best to add documentation to any public classes or methods I added.
- [x] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

PR #4054, opened 2023-12-12
Actually check if the negative tests do their job

Sfiguz7:
## Description

While adding tests to #3920 I pushed some tests which would never fail, this fixes them

## Proposed changes

Set a flag if the event failure is correctly caught. If not the test has actually failed, so we throw a new exception of our own.

## Checklist

### DIFF
--- a/src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestTalismanActivateEvent.java
+++ b/src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestTalismanActivateEvent.java
@@ -80,32 +80,18 @@ class TestTalismanActivateEvent {
         server.getPluginManager().clearEvents();
 
         // Assert the normal talisman does not activate in the ender chest
-        boolean caught = false;
-        try {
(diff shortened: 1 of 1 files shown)


JustAHuman-xD:
This works but ima check when I'm on my computer if there is a cleaner way to do this

JustAHuman-xD:
(I may or may not have completely forgot about this, I'd say just go for it since nothing jumps out as better)

JustAHuman-xD (dismissed):
LGTM

WalshyDev:
Actually a much easier way to do this:
```java
Assertions.assertThrows(
    AssertionError.class,
    () -> server.getPluginManager().assertEventFired(TalismanActivateEvent.class, e -> true)
);
```

JustAHuman-xD:
Walshy chad moment

Sfiguz7:
Oh wow, cool trick

JustAHuman-xD (approved):
LGTM (idk if I'm allowed to approve a pr that I committed to like this....)

WalshyDev (approved):

Issue #4058, opened 2023-12-14
Custom player heards are broken on 1.20.2.


MagmaBro123:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

Title. All player heads from Slimefun and all addons are broken.

Pufferfish fork is being used.

### 📑 Reproduction Steps

Use the plugin and any addons on a 1.20.2 server.

### 💡 Expected Behavior

Everything to work correctly.

### 📷 Screenshots / Videos

![2023-12-13_19 12 42](https://github.com/Slimefun/Slimefun4/assets/51277766/69338f11-7d29-4ec4-af41-bd435ba637b6)

### 📜 Server Log

N/A.

### 📂 `/error-reports/` folder

N/A.

### 💻 Server Software

Other (please specify in your description)

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

![2023-12-13_19 08 28](https://github.com/Slimefun/Slimefun4/assets/51277766/49496b95-bf89-4068-b571-b1fd797aeee8)
![2023-12-13_19 08 30](https://github.com/Slimefun/Slimefun4/assets/51277766/2f774c18-3545-4c0f-beb9-db25dd65bf01)

### 🧭 Other plugins

None.

ybw0014:
you didnt search existing issues... why you tick that

variananora:
Slimefun already fixed this and moved the builds page to the new https://blob.build/
Version 1104 or lower won't automatically update to the new builds page, you need to update your Slimefun manually.

MagmaBro123:
> Slimefun already fixed this and moved the builds page to the new https://blob.build/ Version 1104 or lower won't automatically update to the new builds page, you need to update your Slimefun manually.

I wasn't made aware of this new page. No announcement, ping, etc. was ever made, but thanks.

ybw0014:
There was an @everyone announcement on discord server. https://discord.com/channels/565557184348422174/565569425210867715/1181501726516318218

Issue #4052, opened 2023-12-11
 head items have the player head texture


TerrariaMinecraftFan44:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

All the heads are just player heads without texture (I'm on a premium account and my server is online)

### 📑 Reproduction Steps

Use a player head mechanic 

### 💡 Expected Behavior

Player heads with the texture 

### 💻 Server Software

Spigot

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

Slimefun - DEV 1104
![prueba slimefun](https://github.com/Slimefun/Slimefun4/assets/152024790/1560e408-6a07-4540-a522-1bc671ce798d)

### 🧭 Other plugins

Not in another plugin but it happens in the ExoticGarden addon

TerrariaMinecraftFan44:
A user posted something similar on the exotic garden addon and no one answered him

Sfiguz7:
There was an issue due to a Spigot change, this has since been fixed. You can download the latest build here: https://blob.build/

TerrariaMinecraftFan44:
Thanks, it worked.

Issue #4004, opened 2023-10-22
Slimefun items shows as player heads (texture bug)


RamsTwii:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

The first time i played with slimefun the textures worked perfectky, the following day when i restarted the server all items that uses player heads showed as steve, alex, etc
(items that i crafted the first day when it was working still shows the right textures)
How can i solve this ?

### 📑 Reproduction Steps

1. Start your server
2. Panic because the backpack you just crafted is now a human head

### 💡 Expected Behavior

I expected the textures to work i guess ?

### 💻 Server Software

Spigot

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

![Minecraft 1 20 2 - Multiplayer (3rd-party Server) 22_10_2023 22_35_40](https://github.com/Slimefun/Slimefun4/assets/148721520/420e4bae-a6e5-4b24-a9f8-b810c78796d9)

RamsTwii:
Also the items are still working they did not turn into cs-core heads the bug is just the textures

Boomer-1:
confirmed from server log mojang servers were unable to be reached

PR #4059, opened 2023-12-14
chore: update issue template to include download source

ybw0014:
## Description

We have some bug reports with the latest version from the old builds site. We may need to add one field to the checklist to mention the new builds system as well.
Although I think some people will just tick everything on the checklist without actually checking all of them, this is still something.

## Proposed changes

Add a field in the checklist about the download source.

#4058, #4052, #4004

## Checklist


J3fftw1:
I wonder if there is a way to make it that if you tick a box that it makes the report fail

ybw0014:
can use actions to check issues as always.

J3fftw1 (dismissed):

J3fftw1 (approved):

JustAHuman-xD (approved):
LGTM

TheBusyBiscuit (approved):

PR #4062, opened 2023-12-16
Fix #4051

### LINKED ISSUE #4051: Break a block with an explosive block with a Slimefun block under it wont break the block mined
### ❗ Checklist

### 📍 Description

When you mine a block with an explosive pickaxe when there is a Slimefun block below the mined block.
The mined block is broken but not dropped

### 📑 Reproduction Steps

Place a 1x3 tower
Cobble
Dirt
Enhanced furnace

break Dirt with the explosive pickaxe

### 💡 Expected Behavior

The dirt should be dropped in this case

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Spigot

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

Slimefun 1115

### 🧭 Other plugins

_No response_

Sfiguz7:
## Description

When using an ExplosiveTool to break vanilla blocks in the vicinity of Slimefun Blocks, drops from the vanilla block would disappear. This adds back those drops.

## Proposed changes

When drops are dropped, check if the main block broken was vanilla. If so, drop what it should.

#4051

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -15,6 +15,7 @@ import org.bukkit.Material;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
+import org.bukkit.block.Container;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.enchantments.Enchantment;
 import org.bukkit.entity.Player;
@@ -174,7 +175,7 @@ public class BlockListener implements Listener {
 
             callBlockHandler(e, item, drops, sfItem);
 
-            dropItems(e, drops);
+            dropItems(e, item, drops);
 
             // Checks for vanilla sensitive blocks everywhere
             checkForSensitiveBlocks(e.getBlock(), 0, e.isDropItems());
@@ -224,7 +225,7 @@ public class BlockListener implements Listener {
     }
 
     @ParametersAreNonnullByDefault
-    private void dropItems(BlockBreakEvent e, List<ItemStack> drops) {
+    private void dropItems(BlockBreakEvent e, ItemStack item, List<ItemStack> drops) {
         if (!drops.isEmpty()) {
             // TODO: properly support loading inventories within unit tests
             if (!Slimefun.instance().isUnitTest()) {
@@ -237,10 +238,16 @@ public class BlockListener implements Listener {
                 // Disable normal block drops
                 e.setDropItems(false);
 
+                // Fixes #4051
+                Block b = e.getBlock();
+                if (BlockStorage.check(b) == null) {
+                    b.breakNaturally(item);
+                }
+
                 for (ItemStack drop : drops) {
                     // Prevent null or air from being dropped
                     if (drop != null && drop.getType() != Material.AIR) {
-                        e.getBlock().getWorld().dropItemNaturally(e.getBlock().getLocation(), drop);
+                        b.getWorld().dropItemNaturally(b.getLocation(), drop);
                     }
                 }
             }


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -15,6 +15,7 @@
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
+import org.bukkit.block.Container;
```
WalshyDev, line 18: Unused import

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -237,10 +238,16 @@ private void dropItems(BlockBreakEvent e, List<ItemStack> drops) {
...
+                Block b = e.getBlock();
+                if (BlockStorage.check(b) == null) {
+                    b.breakNaturally(item);
+                }
+
                 for (ItemStack drop : drops) {
                     // Prevent null or air from being dropped
                     if (drop != null && drop.getType() != Material.AIR) {
-                        e.getBlock().getWorld().dropItemNaturally(e.getBlock().getLocation(), drop);
+                        b.getWorld().dropItemNaturally(b.getLocation(), drop);
```
WalshyDev, line 250: Could we ever hit the above case and this one therefore having dupe drops?

JustAHuman-xD: We couldn't find one no

WalshyDev: Would we ever want to hit this? If not, can we just return above?

JustAHuman-xD: OH sorry I misunderstood, we can hit both cases, but there are no duped drops

JustAHuman-xD: If the above case is hit it means they aren't added to drops already, we couldn't find a case where they were

WalshyDev: Cool, let's leave a comment just being like "don't add to drops what will be dropped by this block" but with better Englandish

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -237,10 +238,16 @@ private void dropItems(BlockBreakEvent e, List<ItemStack> drops) {
                 // Disable normal block drops
                 e.setDropItems(false);
 
+                // Fixes #4051
+                Block b = e.getBlock();
+                if (BlockStorage.check(b) == null) {
```
WalshyDev, line 243: Do the callers of this function have the Slimefun block (or know it's null)?
Would be nice to save doing this check and just pass through as a parameter

WalshyDev (approved):

J3fftw1 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
Sfiguz7: Unsure what comment and where you want, I don't understand what confused you or is ambiguous

J3fftw1 (commented):
besides the unused import this LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -237,24 +238,30 @@ private void callBlockHandler(BlockBreakEvent e, ItemStack item, List<ItemStack>
...
             }
 
             // Fixes #2560
             if (e.isDropItems()) {
                 // Disable normal block drops
                 e.setDropItems(false);
 
+                // Fixes #4051
+                if (sfBlock == null) {
+                    block.breakNaturally(item);
```
mcchampions, line 256: call Block#breakNaturally will drop items, it will caused many bugs(double drops,drops that should disappear fall normally)

mcchampions (changes requested):

PR #4065, opened 2023-12-19
Storage rewrite - Phase 1

WalshyDev:
## Description
Ok... bear with me on this!

This PR is the start of our new storage layer. Famously known as the "BlockStorage rewrite" or "SQL for player data".
This PR goes over the goals, and how we get there and starts implementation of phase 1.

Please read through the whole ADR as that describes the goal that I'm going for here.
Please review and comment on the ADR as well as the code being done here.

This PR is mainly about kicking things off, sharing the direction/goal and starting the process. It is not about implementing the API, the additional backends or anything else. The API seen in this PR is a bare minimum and will change to be more inline with what we will probably go for going forward.

## Proposed changes
* Save and load the PlayerProfile with the new storage layer
  * Move saving/loading outside of the PlayerProfile class as much as possible
* Move away from passing around PlayerProfile into things
* Start to implement a new storage API (this is **not** final and will drastically change)
* Moved data out of PlayerProfile to make it easier to handle -- see open questions
* Remove dirty from PlayerProfile -- see open questions

N/A

## Checklist

## Open questions

* Where should PlayerProfile data live?
  * Within this PR, it was moved to a PlayerData class as it was much easier to manage. PlayerProfile does a _lot_. Does this make sense? Do we want it back in PlayerProfile?
* Do we care about tracking dirty state?
  * Generally, does this make sense if we save/load efficiently and always off-main-thread? The PlayerProfile was always marked dirty on the first save anyway but would not be re-saved after (unless someone did something like open a backpack).
* Backpacks are currently lazy-loaded but this moves it away from lazy-loading. Is this ok?
  * I don't think backpacks make up much memory in general so I think this is fine. It's more of a lift and complication to lazy-load just backpacks. Something that can be done but probably a question for down the road.
* Does the direction of the ADR make sense? Agree with the steps?

## Testing

This PR focuses on the PlayerProfile and associated data. Therefore, we'd like testing on:
* Do researches work still? If you research something and restart, is it still researched? Can you keep researching things? Is anything broken there?
* Do backpacks work still? If you create a backpack with items and restart, does it still look good? Is anything broken there?
* Do waypoints work still? If you create a waypoint and restart, does it still look good? Is anything broken there? Player specific and shared (not sure how shared work -- I suspect those are broken)

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
@@ -1,11 +1,13 @@
 package io.github.thebusybiscuit.slimefun4.api.gps;
 
 import java.util.Objects;
+import java.util.UUID;
 
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
 import org.apache.commons.lang.Validate;
+import org.bukkit.Bukkit;
 import org.bukkit.Location;
 import org.bukkit.World.Environment;
 import org.bukkit.entity.Player;
@@ -30,7 +32,7 @@ import io.github.thebusybiscuit.slimefun4.implementation.items.teleporter.Telepo
  */
 public class Waypoint {
 
-    private final PlayerProfile profile;
+    private final UUID uuid;
     private final String id;
     private final String name;
     private final Location location;
@@ -48,26 +50,40 @@ public class Waypoint {
      *            The name of this {@link Waypoint}
      */
     @ParametersAreNonnullByDefault
-    public Waypoint(PlayerProfile profile, String id, Location loc, String name) {
-        Validate.notNull(profile, "Profile must never be null!");
+    public Waypoint(UUID uuid, String id, Location loc, String name) {
+        Validate.notNull(uuid, "UUID must never be null!");
         Validate.notNull(id, "id must never be null!");
         Validate.notNull(loc, "Location must never be null!");
         Validate.notNull(name, "Name must never be null!");
 
-        this.profile = profile;
+        this.uuid = uuid;
         this.id = id;
         this.location = loc;
         this.name = name;
     }
 
     /**
-     * This returns the owner of the {@link Waypoint}.
+     * This returns the owner {@link UUID} of the {@link Waypoint}.
      * 
-     * @return The corresponding {@link PlayerProfile}
+     * @return The corresponding owner {@link UUID}
      */
     @Nonnull
+    public UUID getUuid() {
+        return this.uuid;
+    }
+
+    /**
+     * This returns the owner of the {@link Waypoint}.
+     *
+     * @return The corresponding {@link PlayerProfile}
+     *
+     * @deprecated Use {@link #getUuid()} instead
+     */
+    @Nonnull
+    @Deprecated
     public PlayerProfile getOwner() {
-        return profile;
+        // This is jank and should never actually return null
+        return PlayerProfile.find(Bukkit.getOfflinePlayer(uuid)).orElse(null);
     }
 
     /**
@@ -126,7 +142,7 @@ public class Waypoint {
      */
     @Override
     public int hashCode() {
-        return Objects.hash(profile.getUUID(), id, name, location);
+        return Objects.hash(this.uuid, this.id, this.name, this.location);
     }
 
     /**
@@ -139,7 +155,9 @@ public class Waypoint {
         }
 
         Waypoint waypoint = (Waypoint) obj;
-        return profile.getUUID().equals(waypoint.getOwner().getUUID()) && id.equals(waypoint.getId()) && location.equals(waypoint.getLocation()) && name.equals(waypoint.getName());
+        return this.uuid.equals(waypoint.getUuid())
+            && id.equals(waypoint.getId())
+            && location.equals(waypoint.getLocation())
+            && name.equals(waypoint.getName());
     }
-
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
@@ -63,9 +63,9 @@ public class Waypoint {
     }
 
     /**
-     * This returns the owner {@link UUID} of the {@link Waypoint}.
+     * This returns the owner's {@link UUID} of the {@link Waypoint}.
      * 
-     * @return The corresponding owner {@link UUID}
+     * @return The corresponding owner's {@link UUID}
      */
     @Nonnull
     public UUID getOwnerId() {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
@@ -39,6 +39,26 @@ public class Waypoint {
 
     /**
      * This constructs a new {@link Waypoint} object.
+     *
+     * @param profile
+     *            The owning {@link PlayerProfile}
+     * @param id
+     *            The unique id for this {@link Waypoint}
+     * @param loc
+     *            The {@link Location} of the {@link Waypoint}
+     * @param name
+     *            The name of this {@link Waypoint}
+     *
+     * @deprecated Use {@link #Waypoint(UUID, String, Location, String)} instead
+     */
+    @Deprecated
+    @ParametersAreNonnullByDefault
+    public Waypoint(PlayerProfile profile, String id, Location loc, String name) {
+        this(profile.getUUID(), id, loc, name);
+    }
+
+    /**
+     * This constructs a new {@link Waypoint} object.
      * 
      * @param ownerId
      *            The owning {@link Player}'s {@link UUID}
@@ -77,7 +97,7 @@ public class Waypoint {
      *
      * @return The corresponding {@link PlayerProfile}
      *
-     * @deprecated Use {@link #getUuid()} instead
+     * @deprecated Use {@link #getOwnerId()} instead
      */
     @Nonnull
     @Deprecated
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
@@ -32,7 +32,7 @@ import io.github.thebusybiscuit.slimefun4.implementation.items.teleporter.Telepo
  */
 public class Waypoint {
 
-    private final UUID uuid;
+    private final UUID ownerId;
     private final String id;
     private final String name;
     private final Location location;
@@ -40,8 +40,8 @@ public class Waypoint {
     /**
      * This constructs a new {@link Waypoint} object.
      * 
-     * @param profile
-     *            The owning {@link PlayerProfile}
+     * @param ownerId
+     *            The owning {@link Player}'s {@link UUID}
      * @param id
      *            The unique id for this {@link Waypoint}
      * @param loc
@@ -50,13 +50,13 @@ public class Waypoint {
      *            The name of this {@link Waypoint}
      */
     @ParametersAreNonnullByDefault
-    public Waypoint(UUID uuid, String id, Location loc, String name) {
-        Validate.notNull(uuid, "UUID must never be null!");
+    public Waypoint(UUID ownerId, String id, Location loc, String name) {
+        Validate.notNull(ownerId, "owner ID must never be null!");
         Validate.notNull(id, "id must never be null!");
         Validate.notNull(loc, "Location must never be null!");
         Validate.notNull(name, "Name must never be null!");
 
-        this.uuid = uuid;
+        this.ownerId = ownerId;
         this.id = id;
         this.location = loc;
         this.name = name;
@@ -68,8 +68,8 @@ public class Waypoint {
      * @return The corresponding owner {@link UUID}
      */
     @Nonnull
-    public UUID getUuid() {
-        return this.uuid;
+    public UUID getOwnerId() {
+        return this.ownerId;
     }
 
     /**
@@ -83,7 +83,7 @@ public class Waypoint {
     @Deprecated
     public PlayerProfile getOwner() {
         // This is jank and should never actually return null
-        return PlayerProfile.find(Bukkit.getOfflinePlayer(uuid)).orElse(null);
+        return PlayerProfile.find(Bukkit.getOfflinePlayer(ownerId)).orElse(null);
     }
 
     /**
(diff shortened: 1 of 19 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/legacy/LegacyStorage.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.UUID;
+import java.util.logging.Level;
+
+@Beta
+public class LegacyStorage implements Storage {
+
+    @Override
+    public PlayerData loadPlayerData(@Nonnull UUID uuid) {
+        Config playerFile = new Config("data-storage/Slimefun/Players/" + uuid + ".yml");
+        // Not too sure why this is it's own file
```
Sfiguz7, line 33:
```suggestion
        // Not too sure why this is its own file
```

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
```diff
@@ -157,12 +130,6 @@ public boolean isDirty() {
...
      */
     public void save() {
-        for (PlayerBackpack backpack : backpacks.values()) {
-            backpack.save();
-        }
-
-        waypointsFile.save();
-        configFile.save();
         dirty = false;
     }
```
Sfiguz7, line 134: Shouldn't we call Storage#savePlayerData here before marking dirty as false?

src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/legacy/LegacyStorage.java
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        return new PlayerData(researches, backpacks, waypoints);
+    }
+
+    @Override
+    public void savePlayerData(@Nonnull UUID uuid, @Nonnull PlayerData data) {
+        Config playerFile = new Config("data-storage/Slimefun/Players/" + uuid + ".yml");
+        // Not too sure why this is it's own file
+        Config waypointsFile = new Config("data-storage/Slimefun/waypoints/" + uuid + ".yml");
```
Sfiguz7, line 85: Should we extract these and make them class variables? They are used in both load and save, and the first one is still referenced in PlayerProfile.java just for a getter that I see, and who knows if they are elsewhere that I have yet to see or can't remember. Sounds more robust to just have them be in one place, or is there a reason to keep them elsewhere too?

src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/legacy/LegacyStorage.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+
+        return new PlayerData(researches, backpacks, waypoints);
+    }
+
+    @Override
+    public void savePlayerData(@Nonnull UUID uuid, @Nonnull PlayerData data) {
+        Config playerFile = new Config("data-storage/Slimefun/Players/" + uuid + ".yml");
+        // Not too sure why this is it's own file
```
Sfiguz7, line 84:
```suggestion
        // Not too sure why this is its own file
```

Sfiguz7 (commented):
Haven't reviewed all, will continue in the next days. Still left some comments on what I went over

src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/legacy/LegacyStorage.java
WalshyDev: Bit awkward since they rely on a UUID. I could put the string as a global and format but it's only used twice in here so wasn't the biggest fan of that.
This API isn't great and not really expandable so I plan to work on figuring out how that should look post this PR. Once that is done, we may be able to globalise in some way. Not sure how yet.

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -97,9 +112,14 @@ public int getId() {
      * 
      * @return The owning {@link PlayerProfile}
      */
+    @Deprecated
     @Nonnull
     public PlayerProfile getOwner() {
-        return profile;
+        return profile != null ? profile : PlayerProfile.find(Bukkit.getOfflinePlayer(ownerId)).orElse(null);
```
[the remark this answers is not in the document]
WalshyDev: same but not sure how else to keep compatibility here haha

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
WalshyDev: 🤦 yes, ty

src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
```diff
@@ -48,26 +50,40 @@ public class Waypoint {
...
-        this.profile = profile;
+        this.ownerId = ownerId;
         this.id = id;
         this.location = loc;
         this.name = name;
     }
 
     /**
-     * This returns the owner of the {@link Waypoint}.
+     * This returns the owner {@link UUID} of the {@link Waypoint}.
```
J3fftw1, line 66:
```suggestion
     * This returns the owners {@link UUID} of the {@link Waypoint}.
```

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -49,6 +61,7 @@ public class PlayerBackpack {
      * @param id
      *            The id of this Backpack
      */
+    @Deprecated
```
J3fftw1, line 64: What do we need to use now?

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -97,9 +112,14 @@ public int getId() {
      * 
      * @return The owning {@link PlayerProfile}
      */
+    @Deprecated
```
J3fftw1, line 115: What do we need to use now?

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -34,13 +37,22 @@ public class PlayerBackpack {
 
     private static final String CONFIG_PREFIX = "backpacks.";
 
-    private final PlayerProfile profile;
+    private final UUID ownerId;
     private final int id;
-    private final Config cfg;
 
+    @Deprecated
```
J3fftw1, line 43: What do we need to use now?

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -34,13 +37,22 @@ public class PlayerBackpack {
...
+    private PlayerProfile profile;
+    @Deprecated
```
J3fftw1, line 45: What do we need to use now?

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -188,6 +207,7 @@ public void setSize(int size) {
     /**
      * This method will save the contents of this backpack to a {@link File}.
      */
+    @Deprecated
```
J3fftw1, line 210: What do we need to use now?

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -198,8 +218,43 @@ public void save() {
      * This method marks the backpack dirty, it will then be queued for an autosave
      * using {@link PlayerBackpack#save()}
      */
+    @Deprecated
```
J3fftw1, line 221: What do we need to use now?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
```diff
@@ -258,6 +264,9 @@ private void onUnitTestStart() {
         registry.load(this, config);
         loadTags();
         soundService.reload(false);
+        // TODO: What do we do if tests want to use another storage backend (e.g. testing new feature on legacy + sql)?
+        // Do we have a way to override this?
```
J3fftw1, line 268: I think we should have a way to override this. but not for right now maybe a follow up??

src/main/java/io/github/thebusybiscuit/slimefun4/storage/data/PlayerData.java
```diff
@@ -0,0 +1,96 @@
...
+
+        for (Waypoint wp : waypoints) {
+            if (wp.getId().equals(waypoint.getId())) {
+                throw new IllegalArgumentException("A Waypoint with that id already exists for this Player");
+            }
+        }
+
+        // Limited to 21 due to limited UI space and no pagination
+        if (waypoints.size() >= 21) {
+            return; // not sure why this doesn't throw but the one above does...
```
J3fftw1, line 86: Since we are touching this anyways can we throw?

J3fftw1 (changes requested):

J3fftw1:
Placing items
restarting the server
items are missing

J3fftw1:
>* Where should PlayerProfile data live?
>	* Within this PR, it was moved to a PlayerData class as it was much easier to manage. PlayerProfile does a _lot_. Does this make sense? Do we want it back in PlayerProfile?

If its easier to manage we should keep it out of PlayerProfile. No reason to make our own lifes harder

>* Do we care about tracking dirty state?
>	* Generally, does this make sense if we save/load efficiently and always off-main-thread? The PlayerProfile was always marked dirty on the first save anyway but would not be re-saved after (unless someone did something like open a backpack).

If its handled correctly no if its not yes.
So no since we are gonna handle it correctly :p

>* Backpacks are currently lazy-loaded but this moves it away from lazy-loading. Is this ok?
>	* I don't think backpacks make up much memory in general so I think this is fine. It's more of a lift and complication to lazy-load just backpacks. Something that can be done but probably a question for down the road.

We might need some examples of servers with mass use of backpacks to answer this.
If the backpacks are big in size we should move away from lazy-loading.


> * Does the direction of the ADR make sense? Agree with the steps?

On phase 3 we mark the new backend as stable for playerprofile. then the quesiton should we enable it by default for the servers is yes this shouldnt be a maybe otherwise we arent going into phase 3.
Same actually for all other phases.
We cant go to the next phase when we dont know its stable.
The maybe should be removed.

I agree otherwise with the steps.
I think we should lay some grounds on testing. like what do we need to be tested. maybe write up a doc on what we want to be tested.


General question.
Once we are within phases do we want to "change freeze" other PR's cuz of this or do we just keep continuing to work since we dont know the pace of the work being done since this isnt just you doing the job.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
WalshyDev: agreed

src/main/java/io/github/thebusybiscuit/slimefun4/storage/data/PlayerData.java
WalshyDev: I'd rather not throw, I'd rather both return a boolean value however that is more involved than I'd like. Since this keeps existing behaviour, I'll punt on this.

WalshyDev:
> Once we are within phases do we want to "change freeze" other PR's cuz of this or do we just keep continuing to work since we dont know the pace of the work being done since this isnt just you doing the job.

Imo, this is a question we'll ask ourselves on a case by case basis. I think for this one, not needed. BS phase, maybe since it's so massive. This one should be easy to spot the bugs coming from this.

J3fftw1 (approved):
LGTM i think we should move to phase 2

Seggan:
Agree

Sfiguz7 (approved):
Some painful things but discussed in DMs and they're fine. LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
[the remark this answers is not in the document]
iTwins: Maybe ``orElseThrow`` to keep the nonnull contract and give some info in an error if this somehow fails?

J3fftw1:
Walked through everything again and it seems to work.

src/main/java/io/github/thebusybiscuit/slimefun4/storage/data/PlayerData.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import javax.annotation.Nonnull;
+
+import org.apache.commons.lang.Validate;
+
+/**
+ * The data which backs {@link io.github.thebusybiscuit.slimefun4.api.player.PlayerProfile}
```
JustAHuman-xD, line 19: is there a reason we have the full package name for PlayerProfile here?

JustAHuman-xD (commented):
Not sure if what I mentioned should be "gating" changes so just doing comment

src/main/java/io/github/thebusybiscuit/slimefun4/storage/data/PlayerData.java
WalshyDev: It's only used in one comment here, no real reason to import for that.

JustAHuman-xD: Fair enough

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerBackpack.java
```diff
@@ -199,7 +227,40 @@ public void save() {
      * using {@link PlayerBackpack#save()}
      */
     public void markDirty() {
```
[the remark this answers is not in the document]
WalshyDev: It's still mostly used (hence not removed) and may be in the future again. I also don't really expect this to be used as a public API

src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/Waypoint.java
```diff
@@ -48,26 +50,40 @@ public class Waypoint {
      *            The name of this {@link Waypoint}
      */
     @ParametersAreNonnullByDefault
-    public Waypoint(PlayerProfile profile, String id, Location loc, String name) {
```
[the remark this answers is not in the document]
WalshyDev: Yes, thought I had added that, ty

JustAHuman-xD (approved):
LGTM, if I missed anything it would be on the tests but I think it's good 😅

J3fftw1 (approved):

Sfiguz7 (approved):

PR #4067, opened 2023-12-21
Add new analytics service

WalshyDev:
## Description
As discussed in #developers, we want some better analytic insights into Slimefun (especially for the new storage layer work in progress).

This creates a new service and adds it to our current timings.

This can be debugged with `/sf debug slimefun_analytics`. Debugging will just show the JSON being sent and if the send was ignored.
Users can disable in the `config.yml` at `metrics.analytics`.

## Proposed changes
Creates a new analytics service

## Checklist

## Still to do

- [x] Whip up the analytic service and put it on GitHub
- [x] Test once the service is stood up


src/main/java/io/github/thebusybiscuit/slimefun4/core/services/AnalyticsService.java
```diff
@@ -0,0 +1,106 @@
...
+            double totalTimings = Slimefun.getProfiler().getAndResetAverageNanosecondTimings();
+            double avgPerMachine = Slimefun.getProfiler().getAverageTimingsPerMachine();
+
+            if (totalTimings == 0 || avgPerMachine == 0) {
+                Debug.log(TestCase.ANALYTICS, "Ignoring analytics data for server_timings as no data was found"
+                    + " - total: " + totalTimings + ", avg: " + avgPerMachine);
+                // Ignore if no data
+                return;
+            }
+
```
J3fftw1, line 61: can we add build id.
So we know what timings go with which build

WalshyDev:
Got a base site built out for this, just seeing if we can grab the slimefun.dev domain and then I can close this out

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/ThreadService.java
```diff
@@ -0,0 +1,97 @@
...
+        });
+    }
+
+    /**
+     * Invoke a new thread from the cached thread pool with the given name.
+     * This is a much better alternative to using
+     * {@link BukkitScheduler#runTaskAsynchronously(org.bukkit.plugin.Plugin, Runnable)}
+     * as this will show not only the plugin but a useful name.
+     * By default, Bukkit will use "Craft Scheduler Thread - <x> - <plugin>" which nice to show the plugin but
+     * it's impossible to track exactly what thread this is.
```
J3fftw1, line 41:
```suggestion
     * By default, Bukkit will use "Craft Scheduler Thread - <x> - <plugin>" which is nice to show the plugin but
     * it's impossible to track exactly what thread that is.
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/ThreadService.java
```diff
@@ -0,0 +1,97 @@
...
+        });
+    }
+
+    /**
+     * Invoke a new scheduled thread from the cached thread pool with the given name.
+     * This is a much better alternative to using
+     * {@link BukkitScheduler#runTaskTimerAsynchronously(org.bukkit.plugin.Plugin, Runnable, long, long)}
+     * as this will show not only the plugin but a useful name.
+     * By default, Bukkit will use "Craft Scheduler Thread - <x> - <plugin>" which nice to show the plugin but
+     * it's impossible to track exactly what thread this is.
```
J3fftw1, line 62:
```suggestion
     * By default, Bukkit will use "Craft Scheduler Thread - <x> - <plugin>" which is nice to show the plugin but
     * it's impossible to track exactly what thread that is.
```

J3fftw1 (changes requested):

J3fftw1 (approved):
LGTM,  I really like this addition.
We can now see if our PRs actually improve code.

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/AnalyticsService.java
```diff
@@ -0,0 +1,153 @@
...
+                "AnalyticsService - Timings",
+                sendTimingsAnalytics(),
+                1,
+                1,
+                TimeUnit.MINUTES
+            );
+        }
+    }
+
+    // Timings we'll send the average every minute
```
Sfiguz7, line 62: Confused about what this means but may be cause I am tired

Sfiguz7 (changes requested):
Typo and question because am ded

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/AnalyticsService.java
J3fftw1: That should be will iirc

WalshyDev: I wrote this while tired too 😅

WalshyDev:
```suggestion
    // We'll send some timing data every minute,
    // Today, we collect; the tick interval, the avg timing per tick and avg timing per machine
```

J3fftw1 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/AnalyticsService.java
```diff
@@ -0,0 +1,158 @@
...
+                sendTimingsAnalytics(),
+                1,
+                1,
+                TimeUnit.MINUTES
+            );
+        }
+    }
+
+    // We'll send some timing data every minute,
+    // Today, we collect; the tick interval, the avg timing per tick and avg timing per machine
```
Sfiguz7, line 63:
```suggestion
    // We'll send some timing data every minute.
    // To date, we collect the tick interval, the avg timing per tick and avg timing per machine
```

Sfiguz7 (approved):

J3fftw1 (approved):

PR #4068, opened 2023-12-21
Remove Unirest and replace with Java's native HttpClient

WalshyDev:
## Description
When we first implemented Unirest, the goal was always to remove it once we moved to Java 11 minimum. It's been a while since this happened so let's do this.

This reduces the jar size from 4.7 MB (4782834 bytes) to 2.4 MB (2480486 bytes) so we remove ~2.3 MB.

## Proposed changes
Removes Unirest with Java's native HttpClient (introduced in Java 11)

N/A

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -205,10 +205,6 @@
                             <shadedPattern>io.github.thebusybiscuit.slimefun4.libraries.paperlib</shadedPattern>
                         </relocation>
                         <relocation>
-                            <pattern>kong.unirest</pattern>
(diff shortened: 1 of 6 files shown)


J3fftw1 (changes requested):
please create unit tests

WalshyDev:
This code cannot be unit tested, it's all private or ran as part of a task. Making it public doesn't make sense here

J3fftw1 (dismissed):

J3fftw1 (approved):

Sfiguz7 (approved):

PR #4073, opened 2023-12-28
Multi tools won't reset their modes after server restart

womzil:
## Description

Multi tools reset their mode after server restart. I found it really annoying and so did my players, so I decided to fix that.
This pull request also adds a display of current multi tool mode in its lore.
Feel free to improve my code. I tried to achieve my goal without changing too much stuff, so it *shouldn't* break anything.

Captura de pantalla 2023-12-28 231315

## Proposed changes

- Multi tools keep their current mode after server restart
- Current multi tool mode is displayed in its lore (works with already existing multi tools)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
@@ -8,7 +8,6 @@ import java.util.OptionalInt;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Consumer;
-import java.util.stream.IntStream;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -35,6 +34,8 @@ import io.github.thebusybiscuit.slimefun4.api.items.HashedArmorpiece;
 import io.github.thebusybiscuit.slimefun4.api.researches.Research;
 import io.github.thebusybiscuit.slimefun4.core.attributes.ProtectionType;
 import io.github.thebusybiscuit.slimefun4.core.attributes.ProtectiveArmor;
+import io.github.thebusybiscuit.slimefun4.core.debug.Debug;
+import io.github.thebusybiscuit.slimefun4.core.debug.TestCase;
 import io.github.thebusybiscuit.slimefun4.core.guide.GuideHistory;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 import io.github.thebusybiscuit.slimefun4.implementation.items.armor.SlimefunArmorPiece;
@@ -237,6 +238,7 @@ public class PlayerProfile {
      * The profile can then be removed from RAM.
      */
     public final void markForDeletion() {
+        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Marking {} ({}) profile for deletion", name, ownerId);
         markedForDeletion = true;
     }
 
@@ -244,6 +246,7 @@ public class PlayerProfile {
      * Call this method if this Profile has unsaved changes.
      */
     public final void markDirty() {
+        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Marking {} ({}) profile as dirty", name, ownerId);
         dirty = true;
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/SlimefunTabCompleter.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/SlimefunTabCompleter.java
@@ -16,6 +16,7 @@ import org.bukkit.command.TabCompleter;
 
 import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
 import io.github.thebusybiscuit.slimefun4.api.researches.Research;
+import io.github.thebusybiscuit.slimefun4.core.debug.TestCase;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 
 class SlimefunTabCompleter implements TabCompleter {
@@ -33,6 +34,13 @@ class SlimefunTabCompleter implements TabCompleter {
     public List<String> onTabComplete(CommandSender sender, Command cmd, String label, String[] args) {
         if (args.length == 1) {
             return createReturnList(command.getSubCommandNames(), args[0]);
+        } else if (args.length == 2) {
+            if (args[0].equalsIgnoreCase("debug")) {
+                return createReturnList(TestCase.VALUES_LIST, args[1]);
+            } else {
+                // Returning null will make it fallback to the default arguments (all online players)
+                return null;
+            }
         } else if (args.length == 3) {
             if (args[0].equalsIgnoreCase("give")) {
                 return createReturnList(getSlimefunItems(), args[2]);
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/debug/TestCase.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/debug/TestCase.java
@@ -1,5 +1,7 @@
 package io.github.thebusybiscuit.slimefun4.core.debug;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.Locale;
 
 import javax.annotation.Nonnull;
@@ -17,7 +19,15 @@ public enum TestCase {
      * being checked and why it is comparing IDs or meta.
      * This is helpful for us to check into why input nodes are taking a while for servers.
      */
-    CARGO_INPUT_TESTING;
+    CARGO_INPUT_TESTING,
+
+    /**
+     * Debug information regarding player profile loading, saving and handling.
+     * This is an area we're currently changing quite a bit and this will help ensure we're doing it safely
+     */
+    PLAYER_PROFILE_DATA;
+
+    public static final List<String> VALUES_LIST = Arrays.stream(values()).map(TestCase::toString).toList();
 
     TestCase() {}
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/AutoSavingService.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/AutoSavingService.java
@@ -13,6 +13,8 @@ import org.bukkit.block.Block;
 import org.bukkit.entity.Player;
 
 import io.github.thebusybiscuit.slimefun4.api.player.PlayerProfile;
+import io.github.thebusybiscuit.slimefun4.core.debug.Debug;
+import io.github.thebusybiscuit.slimefun4.core.debug.TestCase;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 
 import me.mrCookieSlime.Slimefun.api.BlockStorage;
@@ -39,9 +41,8 @@ public class AutoSavingService {
     public void start(@Nonnull Slimefun plugin, int interval) {
         this.interval = interval;
 
-        plugin.getServer().getScheduler().runTaskTimer(plugin, this::saveAllPlayers, 2000L, interval * 60L * 20L);
+        plugin.getServer().getScheduler().runTaskTimerAsynchronously(plugin, this::saveAllPlayers, 2000L, interval * 60L * 20L);
         plugin.getServer().getScheduler().runTaskTimerAsynchronously(plugin, this::saveAllBlocks, 2000L, interval * 60L * 20L);
-
     }
 
     /**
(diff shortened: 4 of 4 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -86,12 +94,33 @@ protected ItemUseHandler getItemUseHandler() {
                     }
                 }
             } else {
+                final String LORE_PREFIX = ChatColors.color("&8\u21E8 &7Mode: ");
+                final Pattern REGEX = Pattern.compile(ChatColors.color("(&c&o)?" + LORE_PREFIX) + "(.+)");
```
WalshyDev, line 98: Global these too pls, right now we're compiling this regex on every tool use. We only need to compile it once

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -86,12 +94,33 @@ protected ItemUseHandler getItemUseHandler() {
...
 
                 SlimefunItem selectedItem = modes.get(index).getItem();
                 String itemName = selectedItem != null ? selectedItem.getItemName() : "Unknown";
                 Slimefun.getLocalization().sendMessage(p, "messages.multi-tool.mode-change", true, msg -> msg.replace("%device%", "Multi Tool").replace("%mode%", ChatColor.stripColor(itemName)));
-                selectedMode.put(p.getUniqueId(), index);
+                
+                ItemMeta meta = item.getItemMeta();
+                List<String> lore = meta.hasLore() ? meta.getLore() : new ArrayList<>();
+                
+                meta.getPersistentDataContainer().set(key, PersistentDataType.INTEGER, index);
```
WalshyDev, line 108: Let's not rely on indexes, it's bad we do it currently let's not expand that. Let's instead just set the MultiToolMode here

womzil: I tried to replace indexes with item IDs, hopefully this is what you meant.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -73,25 +78,49 @@ protected ItemUseHandler getItemUseHandler() {
...
 
                 SlimefunItem selectedItem = modes.get(index).getItem();
                 String itemName = selectedItem != null ? selectedItem.getItemName() : "Unknown";
                 Slimefun.getLocalization().sendMessage(p, "messages.multi-tool.mode-change", true, msg -> msg.replace("%device%", "Multi Tool").replace("%mode%", ChatColor.stripColor(itemName)));
-                selectedMode.put(p.getUniqueId(), index);
+
+                ItemMeta meta = item.getItemMeta();
+                List<String> lore = meta.hasLore() ? meta.getLore() : new ArrayList<>();
+
+                meta.getPersistentDataContainer().set(key, PersistentDataType.STRING, modes.get(index).getItemId());
```
JustAHuman-xD, line 108: Again you can use PersistentDataAPI

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiToolMode.java
```diff
@@ -8,19 +8,22 @@
...
+    @Nonnull
+    String getItemId() { return itemId; }
+    
     @Nullable
     SlimefunItem getItem() {
-        return SlimefunItem.getById(item.getValue());
+        return SlimefunItem.getById(itemId);
     }
 
     boolean isEnabled() {
```
JustAHuman-xD, line 29: Why are we changing this? I think this would be a breaking change config wise.

JustAHuman-xD: If the reason is to make sure it can't be changed and break existing multi tools you should just make it so that when it looks for a tool if it doesn't find one it resets for the first one

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -73,25 +78,49 @@ protected ItemUseHandler getItemUseHandler() {
...
+            
+            int index = 0;
+            if (container.has(key, PersistentDataType.STRING)) {
+                for (int i = 0; i < modes.size(); i++) {
+                    if (modes.get(i).getItemId() == container.get(key, PersistentDataType.STRING)) {
+                        index = i;
+                        break;
                     }
                 }
+            }
```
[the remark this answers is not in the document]
JustAHuman-xD: Could probably use a map

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -73,25 +78,49 @@ protected ItemUseHandler getItemUseHandler() {
...
+                        lore.set(i, LORE_PREFIX + ChatColor.stripColor(itemName));
+                        meta.setLore(lore);
+                        item.setItemMeta(meta);
+                        return;
+                    }
+                }
+
+                lore.add(2, LORE_PREFIX + ChatColor.stripColor(itemName));
+                meta.setLore(lore);
+                item.setItemMeta(meta);
```
JustAHuman-xD, line 123: I feel like this can be abstracted or just done in a better way potentially, not entirely sure

JustAHuman-xD (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiToolMode.java
womzil: For some reason I thought it has nothing to do with the config and I didn't get any errors during testing because of it, so I didn't notice. I'll revert unnecessary changes here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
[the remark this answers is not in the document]
womzil: I thought about using a HashMap here, but it would be more difficult to get the next element later, so I decided to stick with a List.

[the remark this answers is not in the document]
JustAHuman-xD: You can do a linked hash map, does index's iirc, that or it's just sorted, one of the two

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
womzil: I thought using a Persistent Data Container here would be enough, also I saw PDCs being used this way in many places in Slimefun. Can you give me more guidance on how it should be implemented?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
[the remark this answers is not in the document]
womzil: I just tried to use it and it seems it's still not possible to get rid of that for loop, at least I don't know how to do that. I tried creating a Linked Hash Map with a String as the key (ID of specific multi tool mode) and with an Integer as the key (so very similar syntax to that of a List), and I always ended up using a for loop somewhere. Probably it's possible to implement it in a better way with a Linked Hash Map, but I simply don't know how.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -73,25 +78,49 @@ protected ItemUseHandler getItemUseHandler() {
...
-                if (removeItemCharge(item, COST)) {
-                    SlimefunItem sfItem = modes.get(index).getItem();
-
-                    if (sfItem != null) {
-                        sfItem.callItemHandler(ItemUseHandler.class, handler -> handler.onRightClick(e));
+            
+            int index = 0;
+            if (container.has(key, PersistentDataType.STRING)) {
+                for (int i = 0; i < modes.size(); i++) {
+                    if (modes.get(i).getItemId().equals(container.get(key, PersistentDataType.STRING))) {
```
WalshyDev, line 87: I thought this MultiToolMode would be an enum _sigh_, ok so it's an object which we use very awkwardly right now...

What we can do but I don't like is to instead have a getId() getter, this is gonna be the index so we can just do `modes.get(getId())`

We should have an actual ID here, I won't put that onto you now though

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
WalshyDev: PersistentDataAPI is a helper, you can use it like `PersistentDataAPI.setString(meta, key, modes.get(index).getId())`. It's just a nicer API

womzil: Oh okay, I'll use it then. Thanks.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -73,25 +77,50 @@ protected ItemUseHandler getItemUseHandler() {
...
-            if (!p.isSneaking()) {
-                if (removeItemCharge(item, COST)) {
-                    SlimefunItem sfItem = modes.get(index).getItem();
+            SlimefunItem sfItem = modes.get(index).getItem();
 
-                    if (sfItem != null) {
-                        sfItem.callItemHandler(ItemUseHandler.class, handler -> handler.onRightClick(e));
-                    }
+            if (!p.isSneaking()) {
+                if (removeItemCharge(item, COST) && sfItem != null) {
```
WalshyDev, line 96:
```suggestion
                if (sfItem != null && removeItemCharge(item, COST)) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
```diff
@@ -73,25 +77,50 @@ protected ItemUseHandler getItemUseHandler() {
...
-            int index = selectedMode.getOrDefault(p.getUniqueId(), 0);
+            int index = 0;
+            if (PersistentDataAPI.hasString(meta, key)) {
+                for (int i = 0; i < modes.size(); i++) {
+                    if (modes.get(i).getItemId().equals(PersistentDataAPI.getString(meta, key, modes.get(index).getItemId()))) {
+                        index = i;
+                        break;
+                    }
+                }
+            }
```
Phoenix-Starlight, line 91: Note: Don't forget the imports.
For loops are possibly substitutable via `forEach`, `Stream`s. Although in this case the only way to replace it is via a Stream.
```suggestion
            int index = !PersistentDataAPI.hasString(meta, key) ? 0 :
                IntStream.range(0, modes.size()).
                filter(i -> modes.get(i).getItemId().equals(
                    PersistentDataAPI.getString(meta, key, modes.get(0).getItemId())
                )).findFirst()
                .orElse(0);
```

Phoenix-Starlight (changes requested):
The classes are definitely in need of a refactor, however its out-of-scope IMO. (oops I submitted two reviews)

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
womzil: I thought so too, but I'm out of ideas for now.

womzil: Nevermind, it was simpler than I thought. Fixed in https://github.com/Slimefun/Slimefun4/pull/4073/commits/c220609fd2e3b8fe25e99f33c918b9c009c02b92.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
J3fftw1: I have seen you update the branch many times but didnt adress this commnet

WalshyDev:
Just to elevate Jeff's comment, this is currently blocked on - https://github.com/Slimefun/Slimefun4/pull/4073#discussion_r1438454999

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/gadgets/MultiTool.java
womzil: Sorry, seems like I forgot about this one. From what I remember, SlimefunItem already has a `getId()` getter and Multitool class inherits it. I tried to find a way to create such a getter under a different name for it, and I have no idea what it would actually do. Replace the _index_? If so, the code above uses _meta_ variable, which is provided by an event - I would need to pass it as a variable to that getter (so basically I would have to do `getIndex(meta)` or something like that) and it feels redundant. Unless I'm missing something here, I think it would be better to leave it as it is for now and refactor it in the future.

If it changes anything, it uses a filter() method now instead of a for loop, like that:
`int index = !PersistentDataAPI.hasString(meta, key) ? 0 :
                IntStream.range(0, modes.size()).
                filter(i -> modes.get(i).getItemId().equals(
                    PersistentDataAPI.getString(meta, key, modes.get(0).getItemId())
                )).findFirst()
                .orElse(0);
`

Sorry if I missed something crucial or obvious here, I don't have much experience with Java, let alone Slimefun.

womzil:
> Just to elevate Jeff's comment, this is currently blocked on - [#4073 (comment)](https://github.com/Slimefun/Slimefun4/pull/4073#discussion_r1438454999)

I answered, just in case you missed it like I did 😉 - https://github.com/Slimefun/Slimefun4/pull/4073#discussion_r1460606053

WalshyDev:
master branches are such a pain to work off 😅 

Opened #4122 with my small improvement on top of this change

PR #4075, opened 2024-01-02
added custom texture support for everything but heads

Kevin-writes-code:
## Description

Minecraft supports custom Textures for items since 1.14 and I thought it would be a great addition to allow TexturePack creators to add custom textures to ingots, items and so on.
## Proposed changes

I added added customModelData to each items Metadata during registration,
excluding heads since that'd break their textures.
I also added a setter for the counter so add-ons can claim their own range for compatibility.
Furthermore, and here I am a little uncertain, I had to change a line in customTextureService to not reset the customModelData.
The provided Tests all passed and I am currently testing the changes on a private server.
The variable customModelData could be used to get item-ids in guide.
## Checklist

(For the ones I tested the customModelData increased as expected, the Items seem to behave as expected)
(CustomModelData seems to have been added in 1.14)


ybw0014:
hmm we already have `item-models.yml`

PR #4078, opened 2024-01-04
Recipe Rewrite

SchnTgaiSpock:
## Description
Hey guys, this is my attempt at the recipe rewrite.

I have written up an [ADR](https://github.com/Slimefun/Slimefun4/pull/4078/files#diff-b225a6f01a4a9496f99a188c0b09c59baf6d37c96c7185cc174bc73a2f528e35) that goes over how the new API currently works, but in short:
- Recipe system sucks
- Can't translate recipes in the dispenser (like crafting Copper Wire on the top row)
- Items can only have one* recipe (mostly)
- Can't have true shapeless recipes

Any feedback on the API or discussion around the new recipe system in general would be greatly appreciated!

## Proposed changes
- Changes how multiblocks and machines store and search recipes
- Allows recipe inputs with multiple options (e.g. Vanilla/Slimefun Copper)
- Allows true shaped/shapeless recipes (does not replace the old versions)

N/A

## Checklist


SchnTgaiSpock:
It's still a draft for now, since i haven't moved the workstations over to the new system yet, but if yall have any opinions on the API/rewrite, please tell me :D

WalshyDev:
Thank you for the ADR <3

Will do a initial review this weekend

JustAHuman-xD:
Do you plan on/is it possible to add the serializing to file to be editable component to this rewrite? It was a big part of previously proposed ideas for the rewrite, allowing for the recipes to be edited via editing the file not unlike recipes in datapacks and mods.

JustAHuman-xD:
This is honestly more so a question to Walshy / Other devs, in some parts of the rewrite you have labeled most local scope variables final, I do the same thing myself but while it isn’t specifically agains the code style guidelines provided for Slimefun it doesn’t follow the code style of the rest of the code. So my question to other devs: Should we remove all of that or is it okay? The code style guidelines are already in dire need of updating as has been discussed in #developers so I’m not entirely sure.

WalshyDev:
> in some parts of the rewrite you have labeled most local scope variables final, I do the same thing myself but while it isn’t specifically agains the code style guidelines provided for Slimefun it doesn’t follow the code style of the rest of the code. So my question to other devs: Should we remove all of that or is it okay? The code style guidelines are already in dire need of updating as has been discussed in #developers so I’m not entirely sure.

so, this is something I do as well. It's imo good practice to always mark things as final (just like it is in JS and others)
It can also help with optimisation but is very much a micro-optimisation. 

However, this is considered against the style guide today and PRs have had to do changes due to usage in the past. 

So, today yes.
However, as you said, we really should update the guide and this is something we can bring up as an open question. 

---

Also, not forgot about this pr. It's top of my list, I've just been sick for the past week. I'll get to it when I can :)

SchnTgaiSpock:
> Do you plan on/is it possible to add the serializing to file to be editable component to this rewrite? It was a big part of previously proposed ideas for the rewrite, allowing for the recipes to be edited via editing the file not unlike recipes in datapacks and mods.

i think its possible, ill give it a shot

J3fftw1:
Now that we are on the latest of MockBukkit i would like all the new behaviour and also the old behaviour to be unit tested.
So we just make sure everything works.
This is a huge change and there is no way we wont forget something unit tests should help with that.

SchnTgaiSpock:
the new API only pr is here #4093

PR #4088, opened 2024-01-13
Fixes #4087 and some cleanup

### LINKED ISSUE #4087: You can take items out of Menus after Breaking a Block
### ❗ Checklist

### 📍 Description

There has been a long standing bug and known dupe using InfinityExpansion where you have two people, one person breaks a storage container and the other takes items out of it while they do so. There is a small period of time where you can take out items after the item has been dropped, effectively duping the Items. 

I suspect the cause is the one tick delay before blockstorage is cleared after a block is broken. So after its broken but before blockstorage is cleared there is a period of time to take out items. Theoretically this should be possible with any menus but its easy for IE because of the buttons to easily take out items. The laggier the server the easier it is to achieve.

When I get out of school ill attach more information but other than description and steps I can’t give much right now.

### 📑 Reproduction Steps

1. Find someone to help you
2. Place a Storage Container filled with Items
3. Have one person break the storage container and the other open it and spam the button to take out items
4. See if you were successful, if not goto step 2

### 💡 Expected Behavior

After breaking the block the menu should be “broken” too. It should not be able to be opened, interacted with, or remain open, etc.

### 📷 Screenshots / Videos

(When I get out of school I’ll add a video if I can)

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Spigot (Is not software dependent)

### 🎮 Minecraft Version

1.20.x (But is not version dependent)

### ⭐ Slimefun version

Dev 1123 (But it works on every build afaik)

### 🧭 Other plugins

_No response_

WalshyDev:
## Description
Per #4087 if you can get into the inventory of a broken block you can easily dupe items (take them out of the inv before cleanup - 500ms by default and from block drops).
We want to fix that and there's also just some general cleanup

## Proposed changes
There are 2 parts to this fix, the first is disabling opening the inventory if the block is broken. The second is closing the inventory on block create (rather than an SF tick later). Created a pretty long test to validate this behaviour and fully verified it through debugger.

I also did a bit of cleanup, and commented out `checkForSensitiveBlocks` rather than the code in the function so we don't have random unused code. Moved @Nonnull annotations in BiomeMap as vscode was freaking out over it being in the middle of generics.

Fixes #4087 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -178,7 +178,7 @@ public class BlockListener implements Listener {
             dropItems(e, drops);
 
             // Checks for vanilla sensitive blocks everywhere
-            checkForSensitiveBlocks(e.getBlock(), 0, e.isDropItems());
+            // checkForSensitiveBlocks(e.getBlock(), 0, e.isDropItems());
         }
     }
 
@@ -306,8 +306,7 @@ public class BlockListener implements Listener {
     // Disabled for now due to #4069 - Servers crashing due to this check
     // There is additionally a second bug with `getMaxChainedNeighborUpdates` not existing in 1.17
     @ParametersAreNonnullByDefault
-    private void checkForSensitiveBlocks(Block block, Integer count, boolean isDropItems) {
-        /*
+    private void checkForSensitiveBlocks(Block block, int count, boolean isDropItems) {
         if (count >= Bukkit.getServer().getMaxChainedNeighborUpdates()) {
             return;
         }
@@ -329,7 +328,6 @@ public class BlockListener implements Listener {
         // Set the BlockData back: this makes it so containers and spawners drop correctly. This is a hacky fix.
         block.setBlockData(state.getBlockData(), false);
         state.update(true, false);
-        */
     }
 
     /**
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -18,6 +18,7 @@ import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.enchantments.Enchantment;
+import org.bukkit.entity.HumanEntity;
 import org.bukkit.entity.Player;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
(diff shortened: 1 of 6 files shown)


WalshyDev:
cc: @JustAHuman-xD

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -220,6 +221,14 @@ private void callBlockHandler(BlockBreakEvent e, ItemStack item, List<ItemStack>
...
             drops.addAll(sfItem.getDrops());
+            // Partial fix for #4087 - We don't want the inventory to be usable post break, close it for anyone still inside
+            // The main fix is in SlimefunItemInteractListener preventing opening to begin with
+            // Close the inventory for all viewers of this block
+            // TODO(future): Remove this check when MockBukkit supports viewers
+            if (!Slimefun.instance().isUnitTest()) {
+                BlockStorage.getInventory(e.getBlock()).toInventory().getViewers().forEach(HumanEntity::closeInventory);
+            }
+            // Remove the block data
             BlockStorage.clearBlockInfo(e.getBlock());
```
JustAHuman-xD, line 232: I will say this scares me 😅 sense this just kinda goes around the existing one tick delay that we don't know the reason behind existing

WalshyDev: huh? This is existing code, this is how we remove the data on break (which triggers the cleanup a tick later). This is normal?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/biomes/BiomeMap.java
```diff
@@ -140,25 +140,28 @@ public String toString() {
...
-    public static <T> @Nonnull BiomeMap<T> fromJson(NamespacedKey key, String json, BiomeDataConverter<T> valueConverter, boolean isLenient) throws BiomeMapException {
+    public static <T> BiomeMap<T> fromJson(NamespacedKey key, String json, BiomeDataConverter<T> valueConverter, boolean isLenient) throws BiomeMapException {
         // All parameters are validated by the Parser.
         BiomeMapParser<T> parser = new BiomeMapParser<>(key, valueConverter);
         parser.setLenient(isLenient);
         parser.read(json);
         return parser.buildBiomeMap();
     }
 
+    @Nonnull
```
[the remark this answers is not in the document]
WalshyDev: Yes but it's also messy as it was before along with just throwing syntax errors in other editors.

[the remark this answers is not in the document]
J3fftw1: Everyone hates the inline stuff anyways.
We might need to rethink this.

J3fftw1 (approved):
I would like this to be tested but this is an annoying thing to be tested.
But there is unit tests so LGTM

Sfiguz7 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
JustAHuman-xD: Ah I misunderstood, mb, review on my phone at like midnight 💀

PR #4093, opened 2024-01-15
Recipe Rewrite (API Only)

SchnTgaiSpock:
## Description
This PR supersedes #4078

### Description from #4078
Hey guys, this is my attempt at the recipe rewrite.

I have written up an [ADR](https://github.com/SchnTgaiSpock/Slimefun4/blob/api/recipe-rewrite-3/docs/adr/0002-recipe-rewrite.md) that goes over how the new API currently works, but in short:
- Recipe system sucks
- Can't translate recipes in the dispenser (like crafting Copper Wire on the top row)
- Items can only have one* recipe (mostly)
- Can't have true shapeless recipes

Any feedback on the API or discussion around the new recipe system in general would be greatly appreciated!

## Proposed changes
- Adds a new recipe API which:
  - Allows recipe inputs with multiple options (e.g. Vanilla/Slimefun Copper)
  - Allows true shaped/shapeless recipes (does not replace the old versions)
  - Allows machines and workstations to search recipes more efficiently
  
N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
@@ -13,7 +13,6 @@ import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
-import javax.annotation.ParametersAreNonnullByDefault;
 
 import io.github.thebusybiscuit.slimefun4.storage.Storage;
 import io.github.thebusybiscuit.slimefun4.storage.backend.legacy.LegacyStorage;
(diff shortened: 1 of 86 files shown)


J3fftw1 (changes requested):
Can we get unit tests on this?

SchnTgaiSpock:
yup thats my prio rn

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/Slimefun.java
```diff
@@ -795,7 +803,7 @@ private static void validateInstance() {
 
     /**
      * This method returns out {@link MinecraftRecipeService} for Slimefun.
-     * This service is responsible for finding/identifying {@link Recipe Recipes}
+     * This service is responsible for finding/identifying {@link org.bukkit.inventory.Recipe Recipes}
```
J3fftw1, line 806: Please import this

SchnTgaiSpock:
unit tests should be done, if i missed something lmk

JustAHuman-xD:
Might be worth waiting for the item stack component changes before doing much more with this

PR #4099, opened 2024-01-17
:sparkles: feat(api): Introduce SlimefunItemRegistryFinalizedEvent

ProfElements:
## Description

This Event is provided so that plugins that have integrations with each other aren't stuck in a state in which some SlimefunItems may not be available to one plugin or the other.

I will use this event to provide a cleaner integration between Dynatech and other plugins. Currently InfinityXpansion, ExoticGarden, and Gastronomicon.

## Proposed changes

Make an event to be able to listen to Slimefun's Registry finalizing all static items. 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemRegistryFinalizedEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemRegistryFinalizedEvent.java
@@ -11,7 +11,8 @@ import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 
 /**
  * This {@link Event} is fired after {@link Slimefun} finishes loading the
- * {@link SlimefunItem} registry. 
+ * {@link SlimefunItem} registry. We recommend listening to this event if you
+ * want to register recipes using items from other addons.
  *
  * @author ProfElements
  */
@@ -31,5 +32,4 @@ public class SlimefunItemRegistryFinalizedEvent extends Event {
     public HandlerList getHandlers() {
         return getHandlerList(); 
     }
-
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemRegistryFinalizedEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemRegistryFinalizedEvent.java
@@ -11,15 +11,15 @@ import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
 
 /**
  * This {@link Event} is fired after {@link Slimefun} finishes loading the
- * {@link SlimefunItem} registry before auto loading
+ * {@link SlimefunItem} registry. 
  *
  * @author ProfElements
  */
-public class SlimefunRegistryFinalizedEvent extends Event {
+public class SlimefunItemRegistryFinalizedEvent extends Event {
     
     private static final HandlerList handlers = new HandlerList();
 
-    public SlimefunRegistryFinalizedEvent() {}
+    public SlimefunItemRegistryFinalizedEvent() {}
 
     @Nonnull
     public static HandlerList getHandlerList() {
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunRegistryFinalizedEvent.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunRegistryFinalizedEvent.java
@@ -1,0 +1,35 @@
+package io.github.thebusybiscuit.slimefun4.api.events;
+
+import javax.annotation.Nonnull;
+
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+
+
+/**
+ * This {@link Event} is fired after {@link Slimefun} finishes loading the
+ * {@link SlimefunItem} registry before auto loading
+ *
+ * @author ProfElements
+ */
+public class SlimefunRegistryFinalizedEvent extends Event {
+    
+    private static final HandlerList handlers = new HandlerList();
+
+    public SlimefunRegistryFinalizedEvent() {}
+
+    @Nonnull
+    public static HandlerList getHandlerList() {
+        return handlers;
+    }
+
+    @Nonnull
+    @Override
+    public HandlerList getHandlers() {
+        return getHandlerList(); 
+    }
+
+}
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java
@@ -24,6 +24,7 @@ import org.bukkit.inventory.ItemStack;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
+import io.github.thebusybiscuit.slimefun4.api.events.SlimefunItemRegistryFinalizedEvent;
 import io.github.thebusybiscuit.slimefun4.api.events.SlimefunRegistryFinalizedEvent;
 import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
 import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
@@ -78,6 +79,8 @@ public final class PostSetup {
             }
         }
 
+        Bukkit.getPluginManager().callEvent(new SlimefunItemRegistryFinalizedEvent());
+        
         loadOreGrinderRecipes();
         loadSmelteryRecipes();
 
@@ -110,7 +113,7 @@ public final class PostSetup {
 
         Slimefun.getItemCfg().save();
         Slimefun.getResearchCfg().save();
-        Bukkit.getPluginManager().callEvent(new SlimefunRegistryFinalizedEvent());
+
         Slimefun.getRegistry().setAutoLoadingMode(true);
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java
@@ -112,7 +112,6 @@ public final class PostSetup {
 
         Slimefun.getItemCfg().save();
         Slimefun.getResearchCfg().save();
-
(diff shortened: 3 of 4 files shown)


J3fftw1:
Can we get unit tests?

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,35 @@
...
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+
+
+/**
+ * This {@link Event} is fired after {@link Slimefun} finishes loading the
+ * {@link SlimefunItem} registry before auto loading
```
WalshyDev, line 14: "before auto loading"
before auto loading what?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java
```diff
@@ -109,6 +110,7 @@ public static void loadItems() {
 
         Slimefun.getItemCfg().save();
         Slimefun.getResearchCfg().save();
+        Bukkit.getPluginManager().callEvent(new SlimefunRegistryFinalizedEvent());
```
WalshyDev, line 113: Seems like a bit of a strange place to put this call, any reason it's not higher up? Either right under item load or right under the recipe load.

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author ProfElements
+ */
+public class SlimefunRegistryFinalizedEvent extends Event {
```
WalshyDev, line 18:
```suggestion
public class SlimefunItemRegistryFinalizedEvent extends Event {
```

ProfElements:
Fixed up according to suggestions

JustAHuman-xD:
I'm not sure about the use case of this tbch, for example: if one addon waits for this event to then register its items to integrate with another addon, but then *another* addon waits for this event to integrate with that addon, then you would still have the same problem.

Currently if you soft depend on any addon you can utilize its items without any problem, in my experience anyways. I guess it depends on the addon but I've never experienced that not working.

The use case you have also provided encourages registering items, after load items is called, which feels wrong. 

Idk this could just be me but I don't think this is the right solution.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/setup/PostSetup.java
```diff
@@ -109,6 +112,7 @@ public static void loadItems() {
 
         Slimefun.getItemCfg().save();
         Slimefun.getResearchCfg().save();
+
```
JustAHuman-xD, line 115:
```suggestion
```

src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+    public static void load() {
+        server = MockBukkit.mock();
+        plugin = MockBukkit.load(Slimefun.class);
+    }
+
+    @AfterAll
+    public static void unload() {
+        MockBukkit.unmock();
+    }
+
```
JustAHuman-xD, line 30:
```suggestion
```

src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Test
+    @DisplayName("Test that SlimefunRegistryFinalizedEvent is fired")
+    void testEventIsFired() {
+
```
JustAHuman-xD, line 35:
```suggestion
```

src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,45 @@
+package io.github.thebusybiscuit.slimefun4.api.events;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
JustAHuman-xD, line 3: Static imports are against the code style guidelines ([See here](https://github.com/Slimefun/Slimefun4/blob/master/CONTRIBUTING.md#1-imports))

JustAHuman-xD (changes requested):

ProfElements:
> I'm not sure about the use case of this tbch, for example: if one addon waits for this event to then register its items to integrate with another addon, but then _another_ addon waits for this event to integrate with that addon, then you would still have the same problem.
> 
> Currently if you soft depend on any addon you can utilize its items without any problem, in my experience anyways. I guess it depends on the addon but I've never experienced that not working.
> 
> The use case you have also provided encourages registering items, after load items is called, which feels wrong.
> 
> Idk this could just be me but I don't think this is the right solution.

This probably isn't the right solution, but I'm currently in the situation when I want to integrate items into another project. Which I can't load after because its `softdepend` on DynaTech. I don't thing we should add items to registry after this point, I'm using it to add all my recipes using another projects items IE my SeedPlucker adding Gastro Seeds to its recipes

JustAHuman-xD:
> This probably isn't the right solution, but I'm currently in the situation when I want to integrate items into another project. Which I can't load after because its `softdepend` on DynaTech. I don't thing we should add items to registry after this point, I'm using it to add all my recipes using another projects items IE my SeedPlucker adding Gastro Seeds to its recipes


Ah so you're in a circular dependency problem. Which is similar to the issue I showed with this. One solution I can think of is you PR to gastro and make gastro add it to the seed plucker but yeah I get the problem.

I think what you may be able to do is something kinda like slimefun, slimefun adds a lot of recipes **to machines** on post register. So if all you want to do is add machine recipes, then this event is actually a good solution. In which case I misunderstood what this was meant for. I thought you meant registering items dependent on addons items 😅

ProfElements:
Yeah Im sorry i know my explanation was kinda confusing

JustAHuman-xD (dismissed):
LGTM, apologizes for the confusion 😅

src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,42 @@
...
+    public static void load() {
+        server = MockBukkit.mock();
+        plugin = MockBukkit.load(Slimefun.class);
+    }
+
+    @AfterAll
+    public static void unload() {
+        MockBukkit.unmock();
+    }
+    @Test
```
JustAHuman-xD, line 30:
```suggestion

    @Test
```

JustAHuman-xD (commented):
(You double removed the empty line here, not a big deal I'll re add it lol)

JustAHuman-xD (approved):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,35 @@
...
+    public static HandlerList getHandlerList() {
+        return handlers;
+    }
+
+    @Nonnull
+    @Override
+    public HandlerList getHandlers() {
+        return getHandlerList(); 
+    }
+
```
WalshyDev, line 34:
```suggestion
```

src/main/java/io/github/thebusybiscuit/slimefun4/api/events/SlimefunItemRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,35 @@
...
+import org.bukkit.event.Event;
+import org.bukkit.event.HandlerList;
+
+import io.github.thebusybiscuit.slimefun4.api.items.SlimefunItem;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+
+
+/**
+ * This {@link Event} is fired after {@link Slimefun} finishes loading the
+ * {@link SlimefunItem} registry.
```
WalshyDev, line 14:
```suggestion
 * {@link SlimefunItem} registry. We recommend listening to this event if you
 * want to register recipes using items from other addons.
```

J3fftw1 (dismissed):

src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @AfterAll
+    public static void unload() {
+        MockBukkit.unmock();
+    }
+
+    @Test
+    @DisplayName("Test that SlimefunRegistryFinalizedEvent is fired")
+    void testEventIsFired() {
+        //Make sure post setup does not throw
```
WalshyDev, line 34:
```suggestion
        // Make sure post setup does not throw
```

src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,43 @@
...
+        Assertions.assertDoesNotThrow(() -> PostSetup.loadItems());
+        
+        //Make sure post setup sent the event
```
WalshyDev, line 37:
```suggestion
        // Make sure post setup sent the event
```

src/test/java/io/github/thebusybiscuit/slimefun4/api/events/TestSlimefunRegistryFinalizedEvent.java
```diff
@@ -0,0 +1,43 @@
...
+        server.getPluginManager().assertEventFired(SlimefunItemRegistryFinalizedEvent.class, ignored -> true);
+ 
+        server.getPluginManager().clearEvents();
+    }
+
```
WalshyDev, line 42:
```suggestion
```

WalshyDev (approved):

J3fftw1 (approved):

PR #4106, opened 2024-01-19
Replace contributor section in guide with a link

J3fftw1:
## Description

This PR replaces the contributor section in the guide with a link.
The contributor section is causing a bunch of warning because contributors change their name. It might not link up with their MC name anymore. This will cause a bunch of warnings. We either replace it by a robust system or change it like this.
The section is barely looked at since there is no benefit to it.
Making a more robust system would just be a waste of time imo.

## Proposed changes

Delete contributor section and replace it with a url to https://github.com/Slimefun/Slimefun4/graphs/contributors

There is no open issue.

## Checklist


src/main/resources/languages/en/messages.yml
```diff
@@ -136,7 +136,7 @@ guide:
     description:
     - '&7Slimefun is an open-source project'
     - '&7and maintained by a large community of people.'
-    - '&7Over &e%contributors% &7people have worked on'
+    - '&7Many people have worked on'
```
JustAHuman-xD, line 139: Iirc this won't update for servers automatically so instead it'll have a not filled placeholder

J3fftw1: for english it will automaticall update.
for other languages it need to be manually changed through crowdin.

JustAHuman-xD: Isn't that only for new keys? It won't override it afaik

JustAHuman-xD: Otherwise the point of configs would be kinda moot

PR #4119, opened 2024-02-04
Move PlayerProfile saving off the main thread

### LINKED ISSUE #4118: move auto saves from main thread to off thread
### ❗ Checklist

### 📍 Description

See the discord thread below

https://discord.com/channels/565557184348422174/565569196218384385/1203819747687268403

and walshy's notes

https://discord.com/channels/565557184348422174/565569196218384385/1203820974470860893

### 📑 Reproduction Steps

review the spark report linked above

### 💡 Expected Behavior

moving save to off the main thread to reduce lag when saving

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

none

none

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

1133

### 🧭 Other plugins

_No response_

WalshyDev:
## Description
Moves the PlayerProfile saving off the main thread

## Proposed changes
Move PlayerProfile off main thread, add debugs and improve tab completion for debug

Moved the PlayerProfile saving off the main thread, we generally load this off-thread but now we also save off-thread. I thought we were already doing this but apparently not, especially with our current YAML stuff this should definitely be done

Also done a small change to ensure that we don't remove the PlayerProfile from memory if the player is still online. I don't think we ever had a reported issue from this but it's kinda weird behaviour

Finally, added some debug logs to the saving logic, this can be enabled with `/sf debug slimefun_player_profile_data`. Also added auto-complete to /sf debug because it's nice, this only works for Slimefun test cases rather than addons but that's fine. Mostly internal anyway

---

**Please consider any cases where this will cause issues and bring them up!**

I think this should be generally safe since we've been saving on-thread but loading off-thread anyway for ages but we should exercise caution. I think with the extra enforcement of the player definitely still being offline there's no issue here but give it a think before approving :)

Resolves #4118 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
@@ -246,7 +246,7 @@ public class PlayerProfile {
      * Call this method if this Profile has unsaved changes.
      */
     public final void markDirty() {
-        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Marking {} ({}) profile as diry", name, ownerId);
+        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Marking {} ({}) profile as dirty", name, ownerId);
         dirty = true;
     }
 
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
@@ -8,7 +8,6 @@ import java.util.OptionalInt;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Consumer;
-import java.util.stream.IntStream;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


WalshyDev:
I tested a basic profile with just researches and with research + backpacks (both Slimefun items and vanilla). Each one could save without issue

J3fftw1 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/api/player/PlayerProfile.java
```diff
@@ -237,13 +238,15 @@ public void removeWaypoint(@Nonnull Waypoint waypoint) {
...
     public final void markForDeletion() {
+        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Marking {} ({}) profile for deletion", name, ownerId);
         markedForDeletion = true;
     }
 
     /**
      * Call this method if this Profile has unsaved changes.
      */
     public final void markDirty() {
+        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Marking {} ({}) profile as diry", name, ownerId);
```
Sfiguz7, line 249:
```suggestion
        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Marking {} ({}) profile as dirty", name, ownerId);
```

Sfiguz7 (approved):
Can't think of a case in which this would fail if the original worked so LGTM

J3fftw1 (approved):

PR #4120, opened 2024-02-05
replace isitemsimilar

J3fftw1:
## Description

This PR replaces isItemSImilar with a new method for now called compareItem()

## Proposed changes

Replace isItemSImilar with a better check

## Checklist

## ToDo
- [x] add tests for all lines of the new code
- [x] Add comments
- [ ] performance test


src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -424,6 +427,46 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+        if (itemDataOne.isPresent() && itemDataOne.equals(itemDataTwo)) {
+            Optional<DistinctiveItem> distinctiveItemOne = getDistinctiveItem(itemDataOne.get());
+            Optional<DistinctiveItem> distinctiveItemTwo = getDistinctiveItem(itemDataTwo.get());
+            if (distinctiveItemOne.isPresent() && distinctiveItemTwo.isPresent()) {
+                return distinctiveItemOne.get().canStack(itemMetaOne, itemMetaTwo);
+            }
+            return true;
+        }
+
+        return itemOne.equals(itemTwo);
```
JustAHuman-xD, line 467: Doesn't this leave us with performance problems again?

WalshyDev: The idea is that the last resort is Spigot's slow one (see below though). 
We check IDs and our own stuff before this.

Only vanilla items with itemmeta will go down here and right now, the main performance loss is from getDisplayName. Considering we won't do that on almost all items hitting this... I think this is good. There's no way to not compare the itemstacks but still... know they're equal

WalshyDev:
Boomer:
> I even added teleporting, basic dust farming, tested a few nodes going thru cargo, and everything seems to work as expected. backpacks are fine. no issues on ancient altar. all i tested with though was core slimefun. no addons or outside plugins. Added multiple power generators, geo mining, teleporting with the gps pad, and everything seems to be running as expected. 

> the environment was 1.20.4, paper version 405. probably need to update that but, eh for now

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -424,6 +427,81 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+        boolean itemTwoHasMeta = itemTwo.hasItemMeta();
+
+        // If one has meta and the other does not, they are not equal
+        if ((!itemOneHasMeta && itemTwoHasMeta) || (itemOneHasMeta && !itemTwoHasMeta)) {
+            return false;
+        }
+
+        // If neither have meta, compare the types and return
+        if (!itemOneHasMeta && !itemTwoHasMeta) {
+            return itemOne.getType() == itemTwo.getType();
```
ybw0014, line 458: why do a type check again here?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -424,6 +427,81 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+        }
+
+        ItemMeta itemMetaOne = itemOne.getItemMeta();
+        ItemMeta itemMetaTwo = itemTwo.getItemMeta();
+        // If both metas are null, compare the types and return
+        if (itemMetaOne == null && itemMetaTwo == null) {
+            return itemOne.getType() == itemTwo.getType();
```
ybw0014, line 465: and here?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -424,6 +427,81 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+            }
+
+            // Otherwise return true since the IDs are equal
+            return true;
+        }
+
+        // -- Vanilla items --
+        // This should only be vanilla items now
+        // Compare types and metas
+        return itemOne.getType() == itemTwo.getType()
```
ybw0014, line 501: and here?

ybw0014 (changes requested):
the type check is the third check,

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -424,6 +427,81 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+
+        // If one meta is null and the other is not, they are not equal
+        } else if (itemMetaOne == null || itemMetaTwo == null) {
+            return false;
+        }
+
+        Optional<String> itemDataOne = Slimefun.getItemDataService().getItemData(itemMetaOne);
+        Optional<String> itemDataTwo = Slimefun.getItemDataService().getItemData(itemMetaTwo);
+        // If one item has a Slimefun ID and the other does not, they are not equal
+        if ((itemDataOne.isPresent() && !itemDataTwo.isPresent()) || (!itemDataOne.isPresent() && itemDataTwo.isPresent())) {
```
TheBusyBiscuit, line 475:
```suggestion
        if (itemDataOne.isPresent() != itemDataTwo.isPresent()) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
```diff
@@ -200,7 +200,7 @@ public boolean test(@Nonnull ItemStack item) {
              * and thus only perform .getItemMeta() once
              */
             for (ItemStackWrapper stack : items) {
-                if (SlimefunUtils.isItemSimilar(subject, stack, checkLore, false)) {
+                if (SlimefunUtils.compareItem(subject, stack)) {
```
TheBusyBiscuit, line 203: The "checkLore" option is ignored here

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/accelerators/AnimalGrowthAccelerator.java
```diff
@@ -40,7 +40,7 @@ protected void tick(Block b) {
 
         for (Entity n : b.getWorld().getNearbyEntities(b.getLocation(), RADIUS, RADIUS, RADIUS, this::isReadyToGrow)) {
             for (int slot : getInputSlots()) {
-                if (SlimefunUtils.isItemSimilar(inv.getItemInSlot(slot), organicFood, false, false)) {
+                if (SlimefunUtils.compareItem(inv.getItemInSlot(slot), organicFood)) {
```
TheBusyBiscuit, line 43: This compares with an unobtainable blank organic food item.
So this would break this machine.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/accelerators/CropGrowthAccelerator.java
```diff
@@ -58,7 +58,7 @@ private boolean grow(Block machine, BlockMenu inv, Block crop) {
 
         if (ageable.getAge() < ageable.getMaximumAge()) {
             for (int slot : getInputSlots()) {
-                if (SlimefunUtils.isItemSimilar(inv.getItemInSlot(slot), organicFertilizer, false, false)) {
+                if (SlimefunUtils.compareItem(inv.getItemInSlot(slot), organicFertilizer)) {
```
TheBusyBiscuit, line 61: This compares with an unobtainable blank organic fertilizer item.
So this would break this machine.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/accelerators/TreeGrowthAccelerator.java
```diff
@@ -120,7 +120,7 @@ private boolean updateSaplingData(Block machine, Block block, BlockMenu inv, Sap
     }
 
     protected boolean isFertilizer(@Nullable ItemStack item) {
-        return SlimefunUtils.isItemSimilar(item, organicFertilizer, false, false);
+        return SlimefunUtils.compareItem(item, organicFertilizer);
```
TheBusyBiscuit, line 123: This compares with an unobtainable blank organic fertilizer item.
So this would break this machine.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/entities/AutoBreeder.java
```diff
@@ -111,7 +111,7 @@ protected void tick(Block b) {
 
         for (Entity n : b.getWorld().getNearbyEntities(b.getLocation(), 4.0, 2.0, 4.0, this::canBreed)) {
             for (int slot : getInputSlots()) {
-                if (SlimefunUtils.isItemSimilar(inv.getItemInSlot(slot), organicFood, false)) {
+                if (SlimefunUtils.compareItem(inv.getItemInSlot(slot), organicFood)) {
```
TheBusyBiscuit, line 114: This compares with an unobtainable blank organic food item.
So this would break this machine.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/multiblocks/OreWasher.java
```diff
@@ -131,7 +131,7 @@ public void onInteract(Player p, Block b) {
                         }
 
                         return;
-                    } else if (SlimefunUtils.isItemSimilar(input, new ItemStack(Material.SAND, 2), false)) {
+                    } else if (SlimefunUtils.compareItem(input, new ItemStack(Material.SAND))) {
```
TheBusyBiscuit, line 134: The required amount is ignored here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -24,6 +24,7 @@
 import org.bukkit.event.EventPriority;
 import org.bukkit.event.Listener;
 import org.bukkit.event.block.BlockBreakEvent;
+import org.bukkit.event.block.BlockDropItemEvent;
```
TheBusyBiscuit, line 27: unused import?

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -424,6 +427,81 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+        // If types aren't equal items aren't equal
+        if (itemOne.getType() != itemTwo.getType()) {
+            return false;
+        }
+
+        boolean itemOneHasMeta = itemOne.hasItemMeta();
+        boolean itemTwoHasMeta = itemTwo.hasItemMeta();
+
+        // If one has meta and the other does not, they are not equal
+        if ((!itemOneHasMeta && itemTwoHasMeta) || (itemOneHasMeta && !itemTwoHasMeta)) {
```
TheBusyBiscuit, line 452:
```suggestion
        if (itemOneHasMeta != itemTwoHasMeta) {
```

src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
J3fftw1: We already check lore in the method. We don’t specifically need to check for it with a boolean

TheBusyBiscuit: The option exists in the gui yet it has no effect on the check.

ybw0014:
Type check is the first, so items with changed item type (e.g. explosive pickaxe is diamond pickaxe, and using the improvement forge from FoxyMachines to upgrade it to netherite) will not match. 
maybe we can have some sort of whitelist for tools and weapons. and other type mismatches will return immediately.

WalshyDev:
An allowlist just leads to unmaintable again

I'd say just do this when we know it isn't an Slimefun item but that will slow down a bunch...

PR #4124, opened 2024-02-10
Fixes #4123 - Coal Generator will no longer be locked after researching

### LINKED ISSUE #4123: When researching coal generator, if you leave and rejoin after save, coal generator is un-researched
### ❗ Checklist

- [x] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

when researched coal generator and leave and rejoin, coal generator un-researched

### 📑 Reproduction Steps

1. see coal generator researched
2. leave and rejoin after few minutes
3. you can see coal generator un-researched

### 💡 Expected Behavior

boris expected coal generator research progress loaded properly after rejoining server

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

dev-1133

### 🧭 Other plugins

_No response_

WalshyDev:
## Description
Due to a logic bug in the Legacy storage backend if there was a duplicate ID it would mark it as researched for the first, then see it researched already and remove it on the second. This was happening for the Coal Generator and Bio Reactor here. Both shared he same research ID 173.

We're just doing this fix for now until we can move away from the legacy backend (work in progress).

## Proposed changes
If we already see the value as researched in the file just double-check that the player does not have it researched.

This change also needed us to pin Paper due to a Paper PR that caused MockBukkit to fail. Not a fan of this but it should only be temporary until MockBukkit updates.
I hate that we always use a "1.x" version which versions are pushed to... Everyone really should always pin to just specific versions to avoid these kinda problems. But that's another topic for another day.

Fixes #4123 

## Checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -406,7 +406,20 @@
                     <groupId>org.jetbrains</groupId>
                     <artifactId>annotations</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>io.papermc.paper</groupId>
+                    <artifactId>paper-api</artifactId>
+                </exclusion>
             </exclusions>
+        </dependency>
+        <!-- Override MockBukkit's Paper to a pinned slightly older version -->
+        <!-- This is because MockBukkit currently fails after this PR: -->
+        <!-- https://github.com/PaperMC/Paper/pull/9629 -->
+        <dependency>
+            <groupId>io.papermc.paper</groupId>
+            <artifactId>paper-api</artifactId>
+            <version>1.20.4-R0.1-20240205.114523-90</version>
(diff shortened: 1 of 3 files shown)


Sfiguz7 (approved):
Sigh

J3fftw1 (commented):
Not a fan of doing this imo we should just change the id. We can still have this patch for future errors but this patch shouldn’t be needed

WalshyDev:
It's a hard no on changing the ID. This has been this way for years, having every single player lose the research, no.

J3fftw1 (approved):

PR #4044, opened 2023-12-07
Strip colors from search term

JustAHuman-xD:
## Description
To allow any servers that have colored chat to still use the slimefun search system

## Proposed changes
- In `SurvivalSlimefunGuide#openSearch` when `searchTerm` is defined I wrapped it with `ChatColor#stripColor`
- In `TestGuideOpening` I changed `testOpenSearch` to `testOpenSearchHistory` to better reflect the test
- In `TestGuideOpening` I created a new `testOpenSearch` method that tests with a normal, and then colored version of the same search. It opens the search for both and asserts that the item registered that should be found by the search is in the upper inventory (where the search would be displayed)

## Related Issues (if applicable)
Resolves #4043 

## Checklist

- [ ] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [ ] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.16.* - 1.20.*).
- [x] I followed the existing code standards and didn't mess up the formatting.
- [ ] I did my best to add documentation to any public classes or methods I added.
- [ ] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [x] I added sufficient Unit Tests to cover my code.

PR #4125, opened 2024-02-11
Fixes guide search when using colored chat

WalshyDev:
## Description
Fixes guide search when using colored chat

## Proposed changes
With https://github.com/baked-libs/dough/pull/250 handle the ChatInput as colored text rather than user friendly (&) text.

Also pinned Spigot to resolve build issues.

N/A

## Checklist

## TODO
- [x] merge dough PR and update pom

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -345,11 +345,10 @@
         </dependency>
 
         <!-- Shaded packages -->
-        <!-- TODO: Revert changes back to maven when dough 1.3 released -->
         <dependency>
             <groupId>com.github.baked-libs.dough</groupId>
             <artifactId>dough-api</artifactId>
-            <version>1108163a49</version>
+            <version>0130f8d9ce</version>
             <scope>compile</scope>
         </dependency>
         <dependency>


JustAHuman-xD:
The color search was fixed here was it not? https://github.com/Slimefun/Slimefun4/pull/4044

J3fftw1:
> The color search was fixed here was it not? #4044

this didnt fix hex color codes / gradient color codes

WalshyDev:
> The color search was fixed here was it not? #4044

It was not. We aren't getting through real colored text so stripping wasn't actually doing anything. We're instead getting user friendly colored text (& - e.g. &6test)

So, we're changing to pass through real colored text and strip.

JustAHuman-xD:
> > The color search was fixed here was it not? #4044
> 
> 
> 
> It was not. We aren't getting through real colored text so stripping wasn't actually doing anything. We're instead getting user friendly colored text (& - e.g. &6test)
> 
> 
> 
> So, we're changing to pass through real colored text and strip.

Okay that makes sense on the dough end, but what I didn't understand is why are you adding an additional strip color to the method parameter, it will call it already in the method

WalshyDev:
Simple answer is I didn't see it getting stripped anywhere

JustAHuman-xD:
> Simple answer is I didn't see it getting stripped anywhere

Fair enough, it gets stripped in the survival slimefun guides implementation

WalshyDev:
Yee I see it now, we call the interface method but in the same class is the impl func where we strip

This PR is basically just gonna be update dough then

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/guide/SurvivalSlimefunGuide.java
```diff
@@ -602,7 +602,7 @@ public void createHeader(Player p, PlayerProfile profile, ChestMenu menu) {
             pl.closeInventory();
 
             Slimefun.getLocalization().sendMessage(pl, "guide.search.message");
-            ChatInput.waitForPlayer(Slimefun.instance(), pl, msg -> SlimefunGuide.openSearch(profile, msg, getMode(), isSurvivalMode()));
+            ChatInput.waitForPlayer(Slimefun.instance(), pl,msg -> SlimefunGuide.openSearch(profile, ChatColor.stripColor(msg), getMode(), isSurvivalMode()));
```
Sfiguz7, line 605:
```suggestion
            ChatInput.waitForPlayer(Slimefun.instance(), pl, msg -> SlimefunGuide.openSearch(profile, ChatColor.stripColor(msg), getMode(), isSurvivalMode()));
```

Sfiguz7 (dismissed):

J3fftw1 (approved):

JustAHuman-xD (approved):
👀

Sfiguz7 (approved):

Issue #3170, opened 2021-07-11
🚀 Roadmap / Upcoming changes and features


TheBusyBiscuit:
## 📔 Introduction
Over the last few months and years you all have often requested for us to be more open in regards to upcoming changes.
So we decided its time to finally release a proper roadmap.
Note here that this roadmap will never be complete and final, we may add or change some goals from time to time and we will refrain from setting a release date for things until absolutely certain.

This simply serves as a rough guide for you to see what changes you can be excited about.
Many of those changes will likely be of technical nature though, so they may not be as exciting or intesting at first.

Also, we aren't actively looking for pull requests towards these goals unless explicitly requested.
If you wish to make a pull request, choose any open issue or approved suggestion and you are more than welcome :heart:
These changes however are mainly targeted towards the core team.

## ⭐ Upcoming changes
Some of these milestones will have a date assigned to them.
But the majority of milestones are still to far away to even have a date ready here.
Those are marked as *TBD* (to be decided).

| Symbol | Type of change |
| :---: | :------- |
| 🔧 | Technical change / API |
| 🎈 | New Features |
| 💡 | Performance improvements |

### 🎈 (Q3 2021) Changes to the radiation system
Thanks to @Semisol you will soon be able to enjoy a new radiation system which works by calculating a so-called "exposure level" to a radioactive substance. You won't experience radiation symptoms until a certain exposure level has been reached.
This will also make the different radiation levels much more distinct from each other and the features comes with a few good-looking UI changes too! You can track the progress of that here: https://github.com/Slimefun/Slimefun4/pull/3131

### 🔧 (September 3rd - 6th 2021) API relocation
The first upcoming change is a major relocation of some of our classes to unify some packages, get rid of old classpaths and to follow the general conventions. You can read more about this here: https://github.com/Slimefun/Slimefun4/pull/3139

### 🔧 (Q3/Q4 2021) Inventory system rewrite
Next up is a rewrite for our inventory-system. The API is very rusty and aged. We don't have an ETA yet, perhaps closely after or along with the aforementioned API relocation.
The new system will be much more flexible and easier to use for developers.

### 🎈(Q4 2021) New sound-system
Soon we will allow server owners to fully configure any sound in this plugin.
This will also work with custom sounds from resource packs. Additionally you can change the pitch and volume of any sound or even mute it.
You can see the progress here: https://github.com/Slimefun/Slimefun4/pull/2844

[...]

PR #4126, opened 2024-02-11
[WIP] Binary storage for player data

WalshyDev:
## Description
Starting phase 2 of the new storage system

Once again there's a lot to unpack here but it's a much smaller change than phase 1!

### What changed
* There's a new config property - `storage.player-data`
  * This accepts two values, `legacy` and `binary`
  * `binary` is marked as experimental
* If `binary` (this PR) is selected in the config for player data we will show a large warning banner in the console:
experimental warning
* Created a new BinaryStorage to handle writing the files as binary (more on this later)
* This BinaryStorage is changing the way researches are saved, they're now using their NamespacedKey's rather than incremental IDs. This has been a change we should have made forever ago and this is step 1 to removing these number IDs.

### Storage choice

Ok so, as you can see from the PR, the storage is done with NBT. The question I'm sure you're asking is why NBT over xyz?
Well, there's a few reasons;
1. Familiarity
  a. Lots of server owners and developers are familiar with NBT. This makes adoption internally and externally _much_ easier (especially with how much MC has leaned into it over the years)
3. Solid storage
  a. NBT is about as small as it can get. Some people think it's inflated, the format isn't. Usage however is. That isn't something any format can really prevent.
4. Existing tooling
  a. Lots of existing tooling exists for server owners to read these files (a bit more on this later as well)

> [!NOTE]
> This is a simple implementation which would have to have the whole file rewritten each time. I'll explore ways to avoid this but without limiting file size, there isn't a great way to do it.

### Usage example

Legacy file with all researches:

Legacy file with all researches, a backpack with numerous items and a waypoint:

Binary file with all researches (no compression):

Binary file with all researches (gzip compression):

### Tooling

- TODO

## Proposed changes
See above

https://github.com/Slimefun/Slimefun4/issues/3170

## Checklist

## For the reviewers

- TODO

## Testing

- TODO

## TODO

- [ ] Sign off from reviewers we want to go this way
- [ ] Finish up NBT side on dough
  - [ ] Write any more methods that'd be helpful
  - [ ] Document all functions (snooze)
  - [ ] Add more tests covering more complex cases + compression
- [ ] TODO

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/binary/BinaryStorage.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/binary/BinaryStorage.java
@@ -1,0 +1,151 @@
+package io.github.thebusybiscuit.slimefun4.storage.backend.binary;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+
+import io.github.thebusybiscuit.slimefun4.core.debug.Debug;
+import io.github.thebusybiscuit.slimefun4.core.debug.TestCase;
+import org.bukkit.Location;
+import org.bukkit.NamespacedKey;
+
+import io.github.bakedlibs.dough.nbt.TagType;
+import io.github.bakedlibs.dough.nbt.streams.CompressionType;
+import io.github.bakedlibs.dough.nbt.streams.TagInputStream;
+import io.github.bakedlibs.dough.nbt.streams.TagOutputStream;
+import io.github.bakedlibs.dough.nbt.tags.CompoundTag;
+import io.github.bakedlibs.dough.nbt.tags.ListTag;
+import io.github.bakedlibs.dough.nbt.tags.StringTag;
+import io.github.bakedlibs.dough.nbt.tags.Tag;
+import io.github.thebusybiscuit.slimefun4.api.gps.Waypoint;
+import io.github.thebusybiscuit.slimefun4.api.player.PlayerBackpack;
+import io.github.thebusybiscuit.slimefun4.api.researches.Research;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+import io.github.thebusybiscuit.slimefun4.storage.Storage;
+import io.github.thebusybiscuit.slimefun4.storage.backend.binary.serializers.LocationSerializer;
+import io.github.thebusybiscuit.slimefun4.storage.data.PlayerData;
+
+public class BinaryStorage implements Storage {
+
+    private static final NamespacedKey PLAYER_DATA = new NamespacedKey(Slimefun.instance(), "player_data");
+    private static final NamespacedKey RESEARCHES_KEY = new NamespacedKey(Slimefun.instance(), "researches");
+    private static final NamespacedKey BACKPACKS_KEY = new NamespacedKey(Slimefun.instance(), "backpacks");
+    private static final NamespacedKey WAYPOINTS_KEY = new NamespacedKey(Slimefun.instance(), "waypoints");
+
+    // TODO: Move to lz4 or zstd
+    private static final CompressionType compression = CompressionType.GZIP;
+
+    /*
+     * Structure as JSON:
+     * {
+     *   "slimefun:researches": ["slimefun:some_research", "slimefun:some_other_research"],
+     *   "slimefun:backpacks": [
+     *     {
+     *       "id": x,
+     *       "item": y
+     *     }
+     *   ],
+     *   "slimefun:waypoints": {
+     *     "<ID>": {
+     *       "name": "",
+     *       "location": {}
+     *     }
+     *   }
+     * }
+     */
+
+    @Override
+    public PlayerData loadPlayerData(UUID uuid) {
+        Debug.log(TestCase.PLAYER_PROFILE_DATA, "Loading player data from binary storage for {}", uuid);
+        File file = new File("data-storage/Slimefun/Players/" + uuid + ".dat");
+
+        if (!file.exists()) {
+            return new PlayerData(Set.of(), Map.of(), Set.of());
+        }
+
+        CompoundTag root;
+        try {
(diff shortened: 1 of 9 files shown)


J3fftw1:
> * Are we good with NBT? Any other format people _really_ want?
     * If another format, why? What advantage does it give?

I think NBT is the right way, we already discussed loads about this and about other storage options.
but the gist of our discussion was NBT is good since most of the MC community is known with this.
Even non admins/devs know about NBT this is widely used even by players using singeplayer.
> * Do we want to save player data elsewhere?
    * We've been saving to `data-storage/Slimefun/Players` and `data-storage/Slimefun/waypoints` for a long time. If we want to change that, this is the time
    * One change at least I'd like to make is moving waypoints into the player files. I need to figure out exactly how the shared waypoints work though before I do this, they seem to not do anything different but then confused how they're shared.
  
My vote would be to move this to `~/plugins/Slimefun/data-storage`.
We dont get reports of this often but we do occasionally that people got rid of the data by moving servers.
Saving in `~/plugins/Slimefun/data-storage` would make more sense to me.


**My questions**

How do we get people to use this, we couldn't answer this in our DMs either.
It's an opt-in option people barely read configs anyways in this space so I doubt people will opt-in.
I would love to see stats on this as well how many opt-in how many opt-out.
Do we maybe want to throw it straight in dev its development anyways but other point is we also tell servers to run dev and advertise RC for plugin use only. soo /shrug


Is there a big difference between gzip, zlib and lz4. 
Can we move later on to lz4 if not lets just use what ever is supported now.
on the other hand people wont move to different standards if no one supports it.
Same discussion we had on to use postgres even though no one supports it.
Should we care about existing tooling, yes but no idc much. People need to adapt anyways but idk i will leave that to the others.


**Note:**

We should run these tests again on a bigger scale bigger server more player data.

WalshyDev:
> How do we get people to use this, we couldn't answer this in our DMs either.
It's an opt-in option people barely read configs anyways in this space so I doubt people will opt-in.
I would love to see stats on this as well how many opt-in how many opt-out.
Do we maybe want to throw it straight in dev its development anyways but other point is we also tell servers to run dev and advertise RC for plugin use only. soo /shrug

The age old question of how to make people use the thing. 
Not sure there's really an easy way sadly. We can enable by default for new servers but are we likely to find bugs... or just push people away? New users aren't usually the ones to report stuff. 
However, this is how we rolled out item IDs back in the day. Compat mode was disabled by default for new servers and it slowly moved over to be everyone.

> We should run these tests again on a bigger scale bigger server more player data.

Agreed, I need the biggest player file we can find and a bunch of blockstorage later too. 


We'll definitely log to bStats too... though I can't make a new panel so gonna have to throw it somewhere else

> Is there a big difference between gzip, zlib and lz4.
Can we move later on to lz4 if not lets just use what ever is supported now.

We can do whatever we want :p 
As for difference, mostly comes down to compression ratio and speed. Lz4 doesn't compress as good as others but it is very fast. 
Zstd has the best compression ratio but it's slower. 
I'd say these days we value speed over ratio. Disk space is much easier to get than time.

ProfElements:
I think NBT is a good format. 

I think that binary storage in general is a good idea. 

I think moving away from `data-storage` might be a misstep if taken do to `data-storage` being easily accessible instead of a multi-folder dive into plugins or something similar. 

Moving way-points into players is probably going to be a big win unless it constantly duplicates data. 

 I think we should support any compression format Minecraft supports. Zlib/Gzip is a fine alternative until its better for us to  break it for a faster or more efficient compression like lz4

Sfiguz7:
I agree with most of what has been said above.

NBT sounds like a good format to use due to the accessibility - researches taking a bit more space is honestly not a problem.

As long as we compress I don't mind which compression method is used, it's going to be quite the jump compared to what we have had for years no matter what, and if anything speed is probably something that is more important when we have tens of thousands of blocks to take care of, so the fact it is already available just makes it even more welcome of a choice.

I agree with Jeff that moving the data folder inside Slimefun's folder would make sense, no other plugin is ever going to use that folder ever and as a user I had always found it weird how it is not a subfolder: /plugins/Slimefun is the first place I would check when looking for Slimefun related anything.

Overall, yay for me

src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/binary/CommonKeys.java
```diff
@@ -0,0 +1,17 @@
...
+public class CommonKeys {
+
+    public static final NamespacedKey ID = new NamespacedKey(Slimefun.instance(), "id");
+    public static final NamespacedKey NAME = new NamespacedKey(Slimefun.instance(), "name");
+    // Location stuff
+    public static final NamespacedKey LOCATION = new NamespacedKey(Slimefun.instance(), "location");
+    public static final NamespacedKey WORLD = new NamespacedKey(Slimefun.instance(), "world");
+    public static final NamespacedKey POSITION = new NamespacedKey(Slimefun.instance(), "position");
+    public static final NamespacedKey PITCH = new NamespacedKey(Slimefun.instance(), "pitch");
+    public static final NamespacedKey YAW = new NamespacedKey(Slimefun.instance(), "yaw");
```
Seggan, line 16: Why aren't these in `LocationSerializer`?

src/main/java/io/github/thebusybiscuit/slimefun4/storage/backend/binary/BinaryStorage.java
```diff
@@ -0,0 +1,151 @@
...
+        // Save waypoints
+        CompoundTag waypointMap = new CompoundTag();
+        for (Waypoint waypoint : data.getWaypoints()) {
+            CompoundTag waypointTag = new CompoundTag();
+
+            waypointTag.putString(CommonKeys.ID, waypoint.getId());
+            waypointTag.putString(CommonKeys.NAME, waypoint.getName());
+            waypointTag.putCompound(CommonKeys.LOCATION, LocationSerializer.INSTANCE.serialize(waypoint.getLocation()));
+
+            waypointMap.putCompound(new NamespacedKey(Slimefun.instance(), waypoint.getId()), waypointTag);
```
Seggan, line 139: Why don't we use a `BinarySerializer` for this?

Seggan (changes requested):

Seggan:
Has the NBT code for dough been PR'd yet? I don't see it so I assume not.

PR #4127, opened 2024-02-14
Fixes exhaustion when loading large profiles

### LINKED ISSUE #4011: RAM overload with backpacks
### ❗ Checklist

### 📍 Description

One player on my server uses backpacks heavily, his Slimefun data file weights 10.1MB, when he joins, there is about 50% chance that the server will crash with out of memory error, this error happens when the server has 16 GB of RAM allocated and he is the only player on the server.

### 📑 Reproduction Steps

Have a player with huge amount of backpacks full of items with 10.1MB slimefun player data file in total, join the server and observe about 50% of times it crashing or ram usage going high

### 💡 Expected Behavior

I expected RAM usage of backpacks to not be so high, maybe load them only on use instead of all at once on join?

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

During out of memory crash, I saw mention of: io.github.thebusybiscuit.slimefun4.api.player.PlayerProfile and io.github.thebusybiscuit.slimefun4.libraries.dough.config.Config and org.bukkit.configuration.file.YamlConfiguration.loadConfiguration 

No recent errors here

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

Slimefun DEV - 1104 (git c977454e)

### 🧭 Other plugins

_No response_

### LINKED ISSUE #4116: possibly out of memory or process/resource limits reached
### ❗ Checklist

### 📍 Description

server crashed

error: https://mclo.gs/3rRGEww

I reported it to the ajLeaderboards plugin, and it said that the problem is not with it but with slimefun

" your server is running out of threads, and it appears that Slimefun is using the most threads (162). ajlb is using 70 "

### 📑 Reproduction Steps

I don't know how it happened but the server crashed

### 💡 Expected Behavior

that doesn't happen

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

https://prnt.sc/PnU70Fb6nTaT

### 🧭 Other plugins

_No response_

WalshyDev:
## Description
Fixes exhaustion when loading large PlayerProfiles due to multiple load calls being made.
If we're already loading the profile, we don't want to spin up a whole new thread and load the profile again/more. This can very easily cause CPU, memory and thread exhaustion if the profile is large

We saw AbstractArmorTask constantly try to get the PlayerProfile (every 10 mc ticks) even though the profile was still being loaded (due to it's very large size). This compounded the issue and ended up killing servers off entirely.

## Proposed changes
Add tracking for when we're loading the PlayerProfile from file into memory. If the profile is loading, we will simply return and not try to load it again. This does mean that callbacks for PlayerProfile#get are currently tossed away, this isn't ideal but IMO is acceptable here. It will just mean if someone interacts with a block/item quickly and they have a large profile, instead of the interaction happening after load it will just not happen and they will need to re-interact. This is only for first load so not a big deal but worth calling out.

Fixes #4011
Fixes #4116

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/Threads.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/Threads.java
@@ -1,0 +1,23 @@
+package io.github.thebusybiscuit.slimefun4;
+
+import javax.annotation.ParametersAreNonnullByDefault;
+
+import org.bukkit.plugin.java.JavaPlugin;
+
+public class Threads {
+
+    @ParametersAreNonnullByDefault
+    public static void newThread(JavaPlugin plugin, String name, Runnable runnable) {
+        // TODO: Change to thread pool
+        new Thread(runnable, plugin.getName() + " - " + name).start();
+    }
+
+    public static String getCaller() {
+        // First item will be getting the call stack
+        // Second item will be this call
+        // Third item will be the func we care about being called
+        // And finally will be the caller
+        StackTraceElement element = Thread.currentThread().getStackTrace()[3];
+        return element.getClassName() + "." + element.getMethodName() + ":" + element.getLineNumber();
+    }
+}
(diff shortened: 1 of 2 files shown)


J3fftw1 (approved):
since its just the initial request and only on the first few ticks of the player being joined i would say its fine
LGTM

Sfiguz7 (approved):

PR #4096, opened 2024-01-16
Add update warning to /sf versions

WalshyDev:
## Description
Adds a "Update available" to /sf versions if someone isn't on the latest Slimefun version. Hovering over will tell the user they should update and not to report bugs on this version.

Needs https://github.com/baked-libs/dough/pull/245 merged

## Proposed changes
Check if we're on the latest version in the versions command and if not, send a update available message.

image
image

## Related Issues (if applicable)
N/A

## Checklist
- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [ ] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.16.* - 1.20.*).
- [x] I followed the existing code standards and didn't mess up the formatting.
- [ ] I did my best to add documentation to any public classes or methods I added.
- [ ] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

## TODO

- [x] Update dough

PR #4129, opened 2024-02-14
chore: add rc warnings to `sf versions` command

ybw0014:
## Description

People who use RC versions still report bugs to Discord/GitHub, adding warning messages may possibily stop them from reporting. (just possible, people without eyes will still report)
Also fixed a issue introduced in https://github.com/Slimefun/Slimefun4/pull/4096, where all the text after updater available have hover event.

This will only apply to future RC versions.

Test run (I changed the version manually to RC - 38, and this is only screenshot for first run):
image

## Proposed changes

Add warning messages to `/sf versions`.

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/VersionsCommand.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/VersionsCommand.java
@@ -84,7 +84,7 @@ class VersionsCommand extends SubCommand {
                     .append(")").color(ChatColor.GRAY);
             }
 
-            builder.append("\n");
+            builder.append("\n").event((HoverEvent) null);
             // @formatter:on
 
             if (Slimefun.getMetricsService().getVersion() != null) {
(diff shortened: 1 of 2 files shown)


ybw0014:
Where else we can add warning messages? maybe the version in sf guide?

JustAHuman-xD:
I think we should re-word this, “we don’t provide support for stable versions” won’t make sense to 99% of people if they don’t know RC is meant for plugin developers.

src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/VersionsCommand.java
```diff
@@ -97,6 +98,17 @@ public void onExecute(@Nonnull CommandSender sender, @Nonnull String[] args) {
 
             addJavaVersion(builder);
 
+            if (Slimefun.getUpdater().getBranch() == SlimefunBranch.STABLE) {
+                builder.append("\nYou are using a RC (stable) version, " +
+                        "\nDO NOT report any bugs to Discord/GitHub.\n")
+                    .color(ChatColor.RED)
+                    .event(new HoverEvent(HoverEvent.Action.SHOW_TEXT, new Text(
+                        "We do not provide support for stable versions, use with caution!\n" +
```
J3fftw1, line 106: Why `use with caution!` this is a stable build there should be no worries.
I get what you mean but i dont think this is the way to say it

J3fftw1 (changes requested):

J3fftw1:
> Where else we can add warning messages? maybe the version in sf guide?

on the guide settings for players maybe.

Im in general not a huge fan of giving a warning.
reports for RC should always be treated with a grain of salt since it can be fixed in dev versions as we see  99% of the time.
But we have gotten valid reports on stable versions.
Im not gonna put a blocker on it since i know many like this

ybw0014:
if you don't want such messages I will turn this into a bug fix PR.

PR #4135, opened 2024-02-19
Optimize SlimefunUtils#isItemSimilar / Optimize SlimefunUtils#equalsItemMeta

IAISI:
Optimize SlimefunUtils#isItemSimilar
Optimize SlimefunUtils#equalsItemMeta

## Description

We can skip some of slow ItemMeta checks that relay on getDisplayName() and getLore() which both deserialize string from JSON internally in Spigot

## Proposed changes

We check SF Item IDs first if those are available and then fallback to ItemMeta check.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -408,6 +408,11 @@ public final class SlimefunUtils {
                         return optionalDistinctive.get().canStack(possibleSfItemMeta, itemMeta);
                     }
                     return true;
+                } else if (id != null && possibleItemId != null) {
+                    /*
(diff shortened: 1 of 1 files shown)


J3fftw1:
We have another PR changing this already as said many times on discord.
https://github.com/Slimefun/Slimefun4/pull/4120

IAISI:
The PR doesn't remove those function from SlimeFunUtils class tho? If so it's possible they would still be used? By other addons and stuff that is not related to Cargo?

WalshyDev:
We cannot just remove public API without a deprecation period. The PR deprecates them as is our policy. That PR changes all Slimefun usages of the functions including Cargo. 

They will be removed after the next RC when it's merged.

IAISI:
Sounds good, I guess I can just close this?

J3fftw1:
Closed in favour of #4120

PR #4143, opened 2024-02-24
fix: set hover event to null for texts after update warning

ybw0014:
## Description

When your mouse hovers over java/metrics version, it will show the update warning too.
Set the hover event to null to let hover event only show for the "update available" text.

## Proposed changes

Set the hover event to `null` for texts after update warning.

## Checklist


WalshyDev:
Thank you for moving this from the other PR, we'll get back to that when we figure out the future of RC

WalshyDev (approved):

JustAHuman-xD (approved):
LGTM

PR #4158, opened 2024-03-03
feat: add item group key to item group items

ybw0014:
## Description

ItemGroup items can be arbitrary. A Slimefun item, a simple renamed vanilla item, or a complex item can be an ItemGroup item.
It would be good if they could have their `NamespacedKey` in their PDC so other plugins can read this as an ItemGroup item.

## Proposed changes

ItemGroup items now have `slimefun:item_group` in their PDC.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
@@ -216,11 +216,7 @@ public class ItemGroup implements Keyed {
      */
     public @Nonnull ItemStack getItem(@Nonnull Player p) {
         return new CustomItemStack(item, meta -> {
-            String name = Slimefun.getLocalization().getItemGroupName(p, getKey());
-
-            if (name == null) {
-                name = meta.getDisplayName();
-            }
+            String name = getDisplayName(p);
 
             if (this instanceof SeasonalItemGroup) {
                 meta.setDisplayName(ChatColor.GOLD + name);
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
```diff
@@ -214,11 +216,7 @@ public void remove(@Nonnull SlimefunItem item) {
      */
     public @Nonnull ItemStack getItem(@Nonnull Player p) {
         return new CustomItemStack(item, meta -> {
-            String name = Slimefun.getLocalization().getItemGroupName(p, getKey());
-
-            if (name == null) {
-                name = item.getItemMeta().getDisplayName();
-            }
```
WalshyDev, line 221: how come this was changed?

WalshyDev (commented):
What's the use case for the PDC here?

src/main/java/io/github/thebusybiscuit/slimefun4/api/items/ItemGroup.java
ybw0014: `getDisplayName` has same logic with this piece of code, so I simplify it. if u don't like it I can revert this part.

ybw0014:
> What's the use case for the PDC here?

For my addon SlimefunTranslation. Slimefun's localization only covers Slimefun's ItemGroups, it is a bit annoying to add localization for addons' ItemGroups. (unless I missed something)

PR #4163, opened 2024-03-22
Fix: solve teleportation time overflow problem.

Violet-Nonbloosom:
## Description

In TP time calculation, $\rm complexity \times complexity$ may result in int overflow when complexity is higher than $4\times 10^{4}$, not difficult to achieve with several popular addons.

Compared with original solution based on `Math` function, it is better to avoid overflow directly.

## Proposed changes

Use `long` to storage temporary speed variable, which hardly overflow in TP time calculation.

If $\rm speed > distance$, time cost must be 1 tick, so we can skip calculation.

Otherwise, $\rm \dfrac {distance}{speed} \geq 1$, while speed must be in range of `int`.

Therefore we can simply convert speed to int variable to calculate ultimate time cost.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/TeleportationManager.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/TeleportationManager.java
@@ -161,11 +161,15 @@ public final class TeleportationManager {
             return 100;
         }
 
-        int speed = 50_000 + complexity * complexity;
-        int unsafeTime = Math.min(4 * distanceSquared(source, destination) / speed, 40);
+        long speed = 50_000L + (long)complexity * complexity;
+        int distance = 4 * distanceSquared(source, destination), time = 1;
 
-        // Fixes #3573 - Using Math.max is a safer way to ensure values > 0 than relying on addition.
-        return Math.max(1, unsafeTime);
+        // If speed is greater than distance, ultimate time cost must be 1 tick.
+        // Otherwise, speed WON'T overflow the range of int.
+        if (speed <= distance)
+            time = Math.min(distance / (int)speed, 40);
+
+        return time;
     }
 
     @ParametersAreNonnullByDefault


src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/TeleportationManager.java
```diff
@@ -161,11 +161,15 @@ public int getTeleportationTime(int complexity, @Nonnull Location source, @Nonnu
             return 100;
         }
 
-        int speed = 50_000 + complexity * complexity;
-        int unsafeTime = Math.min(4 * distanceSquared(source, destination) / speed, 40);
+        long speed = 50_000L + (long)complexity * complexity;
+        int distance = 4 * distanceSquared(source, destination), time = 1;
```
Sfiguz7, line 165:
```suggestion
        int distance = 4 * distanceSquared(source, destination)
        int time = 1;
```
I don't think we ever declare two variables on the same row so for consistency this makes more sense

Sfiguz7 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/TeleportationManager.java
Sfiguz7: Actually let me just commit the change

Sfiguz7 (dismissed):

src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/TeleportationManager.java
```diff
@@ -161,11 +161,16 @@ public int getTeleportationTime(int complexity, @Nonnull Location source, @Nonnu
             return 100;
         }
 
-        int speed = 50_000 + complexity * complexity;
-        int unsafeTime = Math.min(4 * distanceSquared(source, destination) / speed, 40);
+        long speed = 50_000L + (long)complexity * complexity;
+        int distance = 4 * distanceSquared(source, destination)
```
Sfiguz7, line 165:
```suggestion
        int distance = 4 * distanceSquared(source, destination);
```

Sfiguz7 (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/TeleportationManager.java
```diff
@@ -161,11 +161,16 @@ public int getTeleportationTime(int complexity, @Nonnull Location source, @Nonnu
             return 100;
         }
 
-        int speed = 50_000 + complexity * complexity;
-        int unsafeTime = Math.min(4 * distanceSquared(source, destination) / speed, 40);
+        long speed = 50_000L + (long)complexity * complexity;
+        int distance = 4 * distanceSquared(source, destination);
+        int time = 1;
```
JustAHuman-xD, line 166:
```suggestion
```

src/main/java/io/github/thebusybiscuit/slimefun4/api/gps/TeleportationManager.java
```diff
@@ -161,11 +161,16 @@ public int getTeleportationTime(int complexity, @Nonnull Location source, @Nonnu
...
 
-        // Fixes #3573 - Using Math.max is a safer way to ensure values > 0 than relying on addition.
-        return Math.max(1, unsafeTime);
+        // If speed is greater than distance, ultimate time cost must be 1 tick.
+        // Otherwise, speed WON'T overflow the range of int.
+        if (speed <= distance)
+            time = Math.min(distance / (int)speed, 40);
+
+        return time;
```
JustAHuman-xD, line 173:
```suggestion
        // If speed is greater than distance, ultimate time cost must be 1 tick.
        // Otherwise, speed WON'T overflow the range of int.
        if (speed <= distance) {
            return Math.min(distance / (int) speed, 40);
        }

        return 1;
```

JustAHuman-xD (commented):
Just some code style stuff

J3fftw1:
Why are we moving it to a long? Aren’t we delaying the issue then, this isn’t solving the issue.

should limit it to a value instead of moving the problem in my opinion

Violet-Nonbloosom:
> Why are we moving it to a long? Aren’t we delaying the issue then, this isn’t solving the issue.
> 
> should limit it to a value instead of moving the problem in my opinion

Though it may be more robust, I have no idea what will be happening in future overflow bug.

If `long` was not enough for complexity², `int` might not be suitable for complexity as well, which meant a thorough rework would be needed.

Besides, I have another subject to work at present, so my solution ends here.

Violet-Nonbloosom:
@J3fftw1 As you wish.

Violet-Nonbloosom:
> Why are we moving it to a long? Aren’t we delaying the issue then, this isn’t solving the issue.
> 
> should limit it to a value instead of moving the problem in my opinion

@J3fftw1 As you wish.

PR #4169, opened 2024-04-10
Writtenbook soulbound fix

ShiftSad:
## Description

When cloning an Written Book in the crafting slot with a Writable Book, it keeps all the item data, like the SoulBound Status. With should not be intended since it "dupes it".

## Proposed changes

Simply setting the soul bound status false when a book is cloned

## Checklist


ShiftSad:
This is my first pull request, if there is anything I can improve I would be glad to do so. I have already noticed the labels, next time I'll make sure to use them.

Issue #4166, opened 2024-03-31
Machines find it difficult to find recipes for similar items


lijinhong11:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

When I was testing my addon, I found this problem. And, the bug has existed for more than 5 months.

If a recipe needs [64,32] (total 64+32 = 96 items), the machine(s) can't find the recipe, but I place the items like [32,64], it works.
However, it still can find the recipe, but if you place items like [64,64], machine(s) can't find the recipe again.

### 📑 Reproduction Steps

1.The recipe needs items that amount >64 (but not a multiple of 64)
2.place similar items like [64,64] in machines' input slots
3.Machines doesn't take items

### 💡 Expected Behavior

The machines will take 96 items and there's only 32 items in the input slots

### 📷 Screenshots / Videos

![653ee832166478c65b426b5625b1d258](https://github.com/Slimefun/Slimefun4/assets/62691956/07756f14-4a20-4b2f-a0eb-c9ac3d07ae9e)
![12be32879ac2d7f4293ef94d9ba62c0a](https://github.com/Slimefun/Slimefun4/assets/62691956/1b6be4a1-a012-493a-93f5-247c743ff2a9)

https://github.com/Slimefun/Slimefun4/assets/62691956/6297bd1c-bfa6-4b9a-b9ae-5a1b0f79265e

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

![d3c1b819574d70c7f6b1579e3f899559](https://github.com/Slimefun/Slimefun4/assets/62691956/5720bbe8-c624-46d6-ac5b-a94457d9dc96)

### 🧭 Other plugins

Slimefun and its addons

JustAHuman-xD:
We're going to need some more programming specifics, what class are you using for this, I'm assuming AContainer but you need to specify. Those kinds of things. Excerpts of code would also make this a much easier process.

lijinhong11:
> process

just AContainer and i don't modify findNextRecipe method

lijinhong11:
[...]

PR #4172, opened 2024-04-18
fix AContainer can't find recipes with similar items

lijinhong11:
## Description

#4166 

## Proposed changes

Skip scanned slots

#4166 

## Checklist

### DIFF
--- a/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
+++ b/src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
@@ -422,6 +422,8 @@ public abstract class AContainer extends SlimefunItem implements InventoryBlock,
             for (ItemStack input : recipe.getInput()) {
                 for (int slot : getInputSlots()) {
                     if (SlimefunUtils.isItemSimilar(inventory.get(slot), input, true)) {
+                        if (found.containsKey(slot)) continue; //skip scanned slots
+                        
                         found.put(slot, input.getAmount());
                         break;
                     }


JustAHuman-xD:
Have you tested this?

lijinhong11:
not :( but I will test later

lijinhong11:
Tested and it did.

src/main/java/me/mrCookieSlime/Slimefun/Objects/SlimefunItem/abstractItems/AContainer.java
```diff
@@ -422,6 +422,8 @@ protected MachineRecipe findNextRecipe(BlockMenu inv) {
             for (ItemStack input : recipe.getInput()) {
                 for (int slot : getInputSlots()) {
                     if (SlimefunUtils.isItemSimilar(inventory.get(slot), input, true)) {
+                        if (found.containsKey(slot)) continue; //skip scanned slots
```
JustAHuman-xD, line 425: You should invert this check and merge it with the surrounding of statement, have it be the first boolean expression

JustAHuman-xD (commented):
Just a code style (and kinda performance but just micro optimization) thing

lijinhong11:
So there's the only way is to stack items' amount

JWJUN233233:
im afraid that it wont work.
if there is a recipe a[Diamondx64] and a recipe b[Diamondx64, Iron_ingot] and there are 64 diamonds in the AContainer, it should match recipe a instead of b. i may give another solution later.

lijinhong11:
u r more excellent than me 🙁
________________________________
发件人: Ddggdd135 ***@***.***>
发送时间: 2024年4月27日 08:34
收件人: Slimefun/Slimefun4 ***@***.***>
抄送: lijinhong11 ***@***.***>; Author ***@***.***>
主题: Re: [Slimefun/Slimefun4] fix AContainer can't find recipes with similar items (PR #4172)


im afraid that it wont work.
if there is a recipe a[Diamondx64] and a recipe b[Diamondx64, Iron_ingot] and there are 64 diamonds in the AContainer, it should match recipe a instead of b. i may give another solution later.

JWJUN233233:
and i wonder why not use hashcode. if the NBT of item is different from that in recipe, it shouldnt match.

JustAHuman-xD:
> and i wonder why not use hashcode. if the NBT of item is different from that in recipe, it shouldnt match.

Because there are tons of cases where nbt wouldn't be identical but the ingredient is still valid

JWJUN233233:
> > and i wonder why not use hashcode. if the NBT of item is different from that in recipe, it shouldnt match.
> 
> Because there are tons of cases where nbt wouldn't be identical but the ingredient is still valid 

Could you please give an example?

JustAHuman-xD:
> Could you please give an example?

Sure! We already have problems with it in other areas so I can just give a couple of those: the recipe calls for a diamond hoe, you have a diamond hoe that was used and then repaired, nbt won't be identical.

2. You have a recipe that calls for fertilizer, the fertilizer has a custom model nbt because the server previously had a resource pack but now does not or the model number changed, nbt won't be identical

Etc etc

JWJUN233233:
thats so bad

JustAHuman-xD:
> thats so bad

?

JWJUN233233:
i crrated my pr #4177

PR #4173, opened 2024-04-18
Use a thread-safe Set impl for the TickerTask's tickingLocations

### LINKED ISSUE #3696: An Exception was caught while ticking the Block Tickers Task
### ❗ Checklist

### 📍 Description

Hello, there's a bug on our server that probably affects Slimefun in general, that's why I'm posting it here and not in some special room. The bug is that SF electricity generators are changing to vanilla items, but all items can change this way based on the code from the error I'm posting below. 

Please fix this as soon as possible, this is a very serious bug as players may start to abuse it and ask us to replace a vanilla block with an op generator, but the block was never a generator.

### 📑 Reproduction Steps

We don't know, it happens randomly.

### 💡 Expected Behavior

Bug fix.

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

https://pastebin.com/wbGr8thN

_No response_

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.19.x

### ⭐ Slimefun version

Slimefun vDEV - 1034 (git 7b9d769b)

### 🧭 Other plugins

_No response_

md5sha256:
## Description

Bugfix

## Proposed changes

* Replace the use of a `HashSet` with `ConcurrentHashMap#newKeySet` in the ticker task to prevent `CME`s from ocurring when we are ticking
* Updated relevant javadocs
* Replace getOrDefault creations of `HashSet` with `Collections#emptySet`

Fixes #3696 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/tasks/TickerTask.java
@@ -44,6 +44,7 @@ public class TickerTask implements Runnable {
 
     /**
      * This Map holds all currently actively ticking locations.
+     * The value of this map (Set entries) MUST be thread-safe and mutable.
      */
     private final Map<ChunkPosition, Set<Location>> tickingLocations = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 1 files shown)


md5sha256:
Haven't tested this on the server yet but I expect the chance of a major bug ocurring are unlikely.

WalshyDev (approved):

WalshyDev:
I looked for the HashSet a while ago and couldn't find it, ty!

TheBusyBiscuit (approved):

PR #4175, opened 2024-04-22
Convert to Gradle Kotlin DSL

md5sha256:
## Description

PoC of using Gradle as the buildscript

## Proposed changes

Remove all maven buildscripts
Use the gradle wrapper and the gradle Kotlin DSL for buildscripts

## Checklist


.github/workflows/gradle-compiler.yml
```diff
@@ -34,6 +34,16 @@ jobs:
...
         architecture: x64
+        cache: gradle
+
+    - name: Setup Gradle
+      uses: gradle/actions/setup-gradle@v3
+
+    - name: Validate Gradle Wrapper
+      uses: gradle/actions/wrapper-validation@v3
+
+    - name: Grant Wrapper Permission
```
Seggan, line 45: I suggest we do this locally once (and push) instead of doing it every time in every task we want to invoke gradle

build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+
+group = "com.github.slimefun"
+version = property("project.version").toString()
+description = "Slimefun"
+
+java.toolchain.languageVersion.set(JavaLanguageVersion.of(targetJavaVersion))
+
+java {
+    withSourcesJar()
+}
```
Seggan, line 70:
```suggestion
java {
    toolchain.languageVersion.set(JavaLanguageVersion.of(targetJavaVersion))
    withSourcesJar()
}
```

build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+            windowTitle = "Slimefun4 - Javadocs"
+            // doclet = "org.gradle.external.javadoc.StandardJavadocDocletOptions"
+            // FIXME offline links, groups, and additional options.
+        }
+
+    }
+    assemble {
+        dependsOn(shadowJar)
+
+    }
```
Seggan, line 97:
```suggestion
```
This bit is unneeded, `shadowJar` automatically assembles

build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+        // Relocations
+        val destination = "io.github.thebusybiscuit.slimefun4.libraries"
+        relocate("io.github.bakedlibs.dough", "$destination.dough")
+        relocate("io.papermc.lib", "$destination.paperlib")
+        relocate("org.apache.commons.lang", "$destination.commons.lang")
+        // Merge the META-INF files
+        mergeServiceFiles()
+    }
+    processResources {
+
```
Seggan, line 115:
```suggestion
```
Rouge newline

build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+        include("plugin.yml")
+        include("config.yml")
+        include("item-models.yml")
+        include("wiki.json")
+        include("languages/translators.json")
+        include("tags/*.json")
+        include("biome-maps/*.json")
+        include("languages/**/*.yml")
```
Seggan, line 123: Is this really needed? AFAIK all resources are automatically included

build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+        filesMatching("plugin.yml") {
```
Seggan, line 124: This can be replaced with the `plugin-yml` gradle plugin, tho idk how much others will want it

build.gradle.kts
```diff
@@ -0,0 +1,169 @@
...
+                    url = "https://github.com/Slimefun/Slimefun4/issues"
+                }
+                licenses {
+                    license {
+                        name = "GNU General Public License v3.0"
+                        url = "https://github.com/Slimefun/Slimefun4/blob/master/LICENSE"
+                        distribution = "repo"
+                    }
+                }
+
```
Seggan, line 165:
```suggestion
```
Another rouge newline

Seggan (changes requested):

build.gradle.kts
md5sha256: I'm pretty sure it doesnt. Without this calling `gradle build` will only build the unshaded jar. Nonetheless, we can remove this and just do `gradle shadowJar`

build.gradle.kts
md5sha256: Not sure, I'd have thought so too but I wonder why it was explicitly declared in the pom so I honored that explicit declaration. Can remove if there are no other reasons for it being explicitly declared.

build.gradle.kts
md5sha256: Yes, haven't gotten to that part yet so I left it as we used to with maven resource filtering.

build.gradle.kts
Seggan: Yep that's what I do with my gradle spigot plugins, no need to attach it to `build`

.github/workflows/gradle-compiler.yml
Seggan: See also https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git

build.gradle.kts
Seggan: Can confirm all stuff is automatically included without that code

PR #3880, opened 2023-06-29
Implement new sound system

J3fftw1:
## Description

Merged all commits from master
Updated some minor things that java 16 added
Please squash this pr because of all those commits
Changed more sounds over to the new system

## Proposed changes

Moved almost all sounds to the new system
Merged all commits from master
Updated some minor things that java 16 added

## Checklist

- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [ ] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [ ] I have made sure that the proposed changes do not break compatibility across the supported Minecraft versions (1.14.* - 1.19.*).
- [x] I followed the existing code standards and didn't mess up the formatting.
- [ ] I did my best to add documentation to any public classes or methods I added.
- [ ] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

## Note
Build will fail since mockbuckit doesnt support strings as sound as cookie stated in #2844.
This however does build completely fine

PR #2844, opened 2021-02-26
[CI Skip] Allow server owners to fully customize sounds

TheBusyBiscuit:
## Description

This pull request adds the `SoundService` and the `SoundEffect` enum which allow us to easily play sounds
but also allow server owners to easily configure and customize sounds.
Note here that I opted for the minecraft-ids of sounds rather than the bukkit constants because this will allow
server owners to use completely custom sounds in combination with resource packs.
If a sound does not exist, the player will simply not hear anything, no error will be thrown.

When values are misconfigured, a helpful warning will appear in the console.
This will also protect us against Mojang sound changes in the future, as we don't have to worry about the support of changed enum constants or such.

Here is an example of the new config file:
```yaml
PORTABLE_CRAFTER_OPEN_SOUND:
  sound: entity.wolf.howl
  volume: 1.0
  pitch: 0.123456789
SPLINT_CONSUME_SOUND:
  sound: entity.skeleton.hurt
  volume: 1.0
  pitch: 1.0
TRASH_CAN_OPEN_SOUND:
  sound: i.dont.exist.btw.spinach.sucks.and.pineapple.on.pizza.rules
  volume: 1.0
  pitch: 1.0
```

### ToDo (Why this is still a draft)
* [x] MockBukkit does not support String-sounds atm, so Unit Tests will fail, I will need to update MockBukkit for that first
* [ ] I would like to add Unit Tests for this sound service too
* [ ] Not all sounds have been converted yet

## Checklist

- [x] I have fully tested the proposed changes and promise that they will not break everything into chaos.
- [x] I have also tested the proposed changes in combination with various popular addons and can confirm my changes do not break them.
- [x] I followed the existing code standards and didn't mess up the formatting.
- [x] I did my best to add documentation to any public classes or methods I added.
- [x] I have added `Nonnull` and `Nullable` annotations to my methods to indicate their behaviour for null values
- [ ] I added sufficient Unit Tests to cover my code.

PR #4180, opened 2024-05-03
Added TestSoundService

### LINKED ISSUE #3881: Unit tests for new sound system
As a followup to #2844 & and #3880 we need unit tests to cover these changes.
This issue tracks that.

Intybyte:
## Description

 Unit tests for new sound system

## Proposed changes

Adding unitest

Resolves #3881, #2844, #3880  

## Checklist

### DIFF
--- a/src/test/java/io/github/thebusybiscuit/slimefun4/core/services/sounds/TestSoundService.java
+++ b/src/test/java/io/github/thebusybiscuit/slimefun4/core/services/sounds/TestSoundService.java
@@ -1,0 +1,34 @@
+package io.github.thebusybiscuit.slimefun4.core.services.sounds;
+
+import be.seeseemelk.mockbukkit.MockBukkit;
+import io.github.thebusybiscuit.slimefun4.implementation.Slimefun;
+import org.junit.jupiter.api.*;
+
+public class TestSoundService {
+
+    private static Slimefun plugin;
+
+    @BeforeAll
+    public static void load() {
+        MockBukkit.mock();
+        plugin = MockBukkit.load(Slimefun.class);
+    }
+
+    @AfterAll
+    public static void unload() {
+        MockBukkit.unmock();
+    }
+
+    @Test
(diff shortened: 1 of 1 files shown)


J3fftw1:
Thanks for your first PR thank you for reviewing it.

J3fftw1:
Does this cover all of the soundservice?

Intybyte:
> Does this cover all of the soundservice?

It should, I get all the element in the` SoundEffect` enum and check if it has a configuration using the `SoundService`, as it might be null even if it should not.

In fact the `SoundEffect` enum has a method that gets the `SoundService` instance in order to check it, which before returning the value checks if it is `null`, and prints a warning (the comment itself says that this should not happen in the first place), so by adding this test you are checking all of the elements of `SoundEffect` ahead in the testing.

It is complete unless the `reload` method requires some testing too.

src/test/java/io/github/thebusybiscuit/slimefun4/core/services/sounds/TestSoundService.java
```diff
@@ -0,0 +1,34 @@
...
+    @Test
+    @DisplayName("Testing each SoundConfiguration")
+    void testSounds() {
+        SoundService service = Slimefun.getSoundService();
+        SoundEffect[] effects = SoundEffect.values();
+
+        for (SoundEffect i : effects) {
+            SoundConfiguration element = service.getConfiguration(i);
+            Assertions.assertNotNull(element, "The sound effect must not be null!");
+        }
```
TheBusyBiscuit, line 31: Having a for loop in a unit test is not a good approach.
Please have a look at Parameterized Unit Tests:
https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests
Hint: If you are unable to map the parameters to the annotation parameter you may also want to look at the `@MethodSource` annotation

TheBusyBiscuit (changes requested):

PR #4186, opened 2024-05-05
Update to 1.20.5

WalshyDev:
## Description
Relatively easy update just a bunch of enum changes for which we can just provide a compatible API for.

## Proposed changes
Created a bunch of VersionedX classes, I don't really like this but it works for now. Ideally, we build a proper mapping system but for now, I think this is fine. These names have been wrong a long time, I don't see them changing much from here.

N/A

## Checklist

## Still to do

- [x] Testing finishes
- [x] Merge dough (https://github.com/Slimefun/dough/pull/271)
- [x] Update dough in pom.xml


src/main/java/io/github/thebusybiscuit/slimefun4/api/MinecraftVersion.java
```diff
@@ -133,8 +161,10 @@ public boolean isVirtual() {
      * 
      * @return Whether this {@link MinecraftVersion} matches the specified version id
      */
-    public boolean isMinecraftVersion(int minecraftVersion) {
-        return !isVirtual() && this.majorVersion == minecraftVersion;
+    public boolean isMinecraftVersion(int minecraftVersion, int patchVersion) {
```
JustAHuman-xD, line 164: We would deprecate the old method, otherwise it might break some addons

JustAHuman-xD (changes requested):
Also should this be something we maybe put in dough instead of SF itself?

WalshyDev:
> Also should this be something we maybe put in dough instead of SF itself?

MinecraftVersion or the versioned stuff? We have a MinecraftVersion in dough, just not with the actual MC vers. The versioned classes, maybe but I don't intend for them to be around forever or widely used.

JustAHuman-xD:
> > Also should this be something we maybe put in dough instead of SF itself?
> 
> 
> 
> MinecraftVersion or the versioned stuff? We have a MinecraftVersion in dough, just not with the actual MC vers. The versioned classes, maybe but I don't intend for them to be around forever or widely used.

I was meaning the versioned stuff but yeah if they're temp then it's probably fine

WalshyDev:
Tests are due to mix of 1.20.6 and 1.20.4, will wait for MockBukkit

This looks alright to me on 1.20.4, not tested 1.20.6 yet since Paper is broken

J3fftw1 (changes requested):

J3fftw1:
https://github.com/MockBukkit/MockBukkit/pull/1008

ybw0014:
> [MockBukkit/MockBukkit#1008](https://github.com/MockBukkit/MockBukkit/pull/1008)

PR is merged, time to have some progress?

JustAHuman-xD:
Java version issue by the look of the errors

J3fftw1:
Ah I saw an open PR for that on mockbukkit

JustAHuman-xD:
I think we need to bump Java up to 21 for tests since that's what mockbukkits Java is now... Could be wrong tho

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -486,7 +486,7 @@ private static boolean equalsItemMeta(@Nonnull ItemMeta itemMeta, @Nonnull ItemM
         }
 
         if (itemMeta instanceof PotionMeta && sfitemMeta instanceof PotionMeta) {
-            return ((PotionMeta) itemMeta).getBasePotionData().equals(((PotionMeta) sfitemMeta).getBasePotionData());
+            return ((PotionMeta) itemMeta).getBasePotionType().equals(((PotionMeta) sfitemMeta).getBasePotionType());
```
ybw0014, line 489: `getBasePotionType` is added in 1.20.2, this code will break backward compatibility.
also, `getBasePotionType` becomes nullable in 1.20.5, need to check `hasBasePotionType` before.

This will break all potion items.

ybw0014 (changes requested):

md5sha256:
Okay so I bumped tests for Java 21 and the mockbukkit version. I've also fixed the merge conflict so now all thats left for the tests to pass is dough... not entirely sure how to fix this. The ItemUtils class doesn't compile on 1.20.5+ due to the enchantments constants changes. So unless dough adopts some sort of Versioned enchantment, we'd have to make dough support 1.20.5+ exclusively for this to work.

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
md5sha256: There were multiple places where the new `PotionMeta#getBasePotionType` were used. I've refactored all of them to do a version check to ensure backwards compat.

J3fftw1:
the e2e tester also needs to be updated otherwise it will check against 1.21

ybw0014:
putting the link to MockBukkit's basePotionType PR here: https://github.com/MockBukkit/MockBukkit/pull/1059

md5sha256:
> putting the link to MockBukkit's basePotionType PR here: [MockBukkit/MockBukkit#1059](https://github.com/MockBukkit/MockBukkit/pull/1059)

the PR targets 1.21 though, we won't be able to use it until then...

ybw0014:
We should probably temporarily disable the tests that use the base potion type. When we work on 1.21, we re-enable and fix those tests.

J3fftw1:
We would have to disable the unit test for 1.20.5 and 1.20.6
kinda annoying

WalshyDev:
Testing is done: https://canary.discord.com/channels/565557184348422174/565570276038017044/1282538803940163584

md5sha256 (dismissed):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/api/MinecraftVersion.java
```diff
@@ -47,7 +47,13 @@ public enum MinecraftVersion {
      * This constant represents Minecraft (Java Edition) Version 1.20
      * ("The Trails &amp; Tales Update")
      */
-    MINECRAFT_1_20(20, "1.20.x"),
+    MINECRAFT_1_20(20, 0, 4,"1.20.x"),
```
ybw0014, line 50:
```suggestion
    MINECRAFT_1_20(20, 0, 4, "1.20.x"),
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -485,11 +485,24 @@ private static boolean equalsItemMeta(@Nonnull ItemMeta itemMeta, @Nonnull ItemM
...
-        if (itemMeta instanceof PotionMeta && sfitemMeta instanceof PotionMeta) {
-            return ((PotionMeta) itemMeta).getBasePotionData().equals(((PotionMeta) sfitemMeta).getBasePotionData());
+        if (!(itemMeta instanceof PotionMeta potionMeta) || !(sfitemMeta instanceof PotionMeta sfPotionMeta)) {
+            return true;
         }
-
-        return true;
+        MinecraftVersion current = Slimefun.getMinecraftVersion();
+
+        if (current.isBefore(20, 2)) {
```
ybw0014, line 493: using `isBefore` here will go to the 1.20.5+ check directly on 1.20-1.20.4.

ybw0014 (changes requested):

J3fftw1 (commented):
LGTM after the points from YB have been addressed

src/main/java/io/github/thebusybiscuit/slimefun4/api/MinecraftVersion.java
```diff
@@ -193,4 +271,14 @@ public boolean isBefore(@Nonnull MinecraftVersion version) {
...
     }
 
+    /**
+     * Checks whether this {@link MinecraftVersion} is older than the specified minecraft and patch versions
+     * @param minecraftVersion The minecraft version
+     * @param patchVersion The patch version
+     * @return True if this version is before, False if this version is virtual or otherwise.
+     */
+    public boolean isBefore(int minecraftVersion, int patchVersion) {
+        return !isVirtual() && (this.majorVersion < minecraftVersion && this.minorVersion < patchVersion);
```
JustAHuman-xD, line 281: If for 1.21 we have 1_21, wouldn't calling 1_20_5#isBefore(21, 0) fail because the minor version 5 is larger than -1? And if so would we need to adjust it to be `this.majorVersion < minecraftVersion || this.minorVersion < patchVersion` or am I missing something

ybw0014, line 281:
```suggestion
        return !isVirtual() && (this.majorVersion < minecraftVersion || (this.majorVersion == minecraftVersion && this.minorVersion < patchVersion));
```

ybw0014 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/utils/FireworkUtils.java
```diff
@@ -31,6 +33,8 @@ public final class FireworkUtils {
         Color.RED, Color.SILVER, Color.TEAL, Color.WHITE, Color.YELLOW
     };
     // @formatter:on
+    
+    private static final EntityType firework = VersionedEntityType.FIREWORK;
```
svr333, line 37: Not sure why you make it a static variable when its only used once

pom.xml
```diff
@@ -387,19 +399,13 @@
...
         </dependency>
-        <!-- Override MockBukkit's Paper to a pinned slightly older version -->
-        <!-- This is because MockBukkit currently fails after this PR: -->
-        <!-- https://github.com/PaperMC/Paper/pull/9629 -->
+        <!-- Override Spigot with Paper tests as a CommandMap ctor which MockBukkit uses only exists on Paper but not in Spigot -->
         <dependency>
             <groupId>io.papermc.paper</groupId>
             <artifactId>paper-api</artifactId>
-            <version>1.20.4-R0.1-20240205.114523-90</version>
+            <version>1.20.6-R0.1-SNAPSHOT</version>
```
svr333, line 408: -SNAPSHOT is still latest version?

svr333 (approved):
Everything looks good, should add that it's been a while since I've done mc related coding

JustAHuman-xD (approved):
LGTM as far as I can see as long as it's been tested too!

src/main/java/io/github/thebusybiscuit/slimefun4/utils/FireworkUtils.java
WalshyDev: not sure either but can be cleaned up later

PR #4190, opened 2024-05-13
Fix/cargo enchants

### LINKED ISSUE #3341: Cargo is not filtering Vanilla Enchanted Books (All treated as same with/without Lore filter)

Intybyte:
## Description

This PR aims to fix a confirmed bug regarding enchants not being filtered.

## Proposed changes

Adding a filter option to cargo nodes for vanilla enchants.

Resolves #3341 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/networks/cargo/ItemFilter.java
@@ -1,18 +1,15 @@
 package io.github.thebusybiscuit.slimefun4.core.networks.cargo;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.function.Predicate;
 
 import javax.annotation.Nonnull;
 
-import io.github.thebusybiscuit.slimefun4.implementation.SlimefunItems;
 import org.bukkit.Bukkit;
 import org.bukkit.Material;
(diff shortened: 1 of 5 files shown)


J3fftw1:
before i review this i would really like some benchmarks for this.
Anything cargo may slow down the server massively

Intybyte:
> before i review this i would really like some benchmarks for this. Anything cargo may slow down the server massively

Now it uses `SlimefunUtils#isItemSimilar` so the benchmark is the same as that method, however it uses `ItemFactory` to create ItemMetas of vanilla items, as for some reason `ItemStackWrapper` class doesn't save them. However as it is an heavy method, I cache the item meta in an hashmap so it is called just once per Material type

PR #4193, opened 2024-05-20
Fixes #4020

### LINKED ISSUE #4020: Books and player heads metadata copy/dupe
### ❗ Checklist

### 📍 Description

If you connect two chests via Cargo Nodes, you can copy metadata of signed books and player heads from one to another, resulting in, for example, duplication of "cheque" books, or other valuable books or heads.

### 📑 Reproduction Steps

1. Have two signed books (A and B).
2. Place these blocks (listed from left to right, but placement order does not matter, appaerently):
   - Chest,
   - Cargo Node (Input),
   - Cargo Manager
   - Cargo Node (Output),
   - Chest.
 3. Set the Cargo Node (Input) type to Blacklist.
 4. Insert the book to be copied (A) into the right (output) chest.
 5. Insert the other book (B) into the left (input) chest.
 6. In the right chest you will have the copied book (A) two times.

The same applies to player heads.

Also it works for higher quantities of items: for example, if you were to fill the right chest with the A books, and then proceed to insert some B books into the left chest, the item stacks in the right chest will all get eventually filled up.

On the other hand, if you would only have 1 A book in the right chest and you were to put 17 B books into the left chest, you would end up with 16 A books and 2 B books in the right chest. So the issue is clearly **linked to incomplete item stacks**.

### 💡 Expected Behavior

I would expect the B book to transfer to the right (output) chest, instead of getting another A book.

### 📷 Screenshots / Videos

See the bug reproduction process screencap: https://ctrlv.tv/ZLMh

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Paper

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

Tested both on:
```
Paper git-Paper-411 (MC: 1.17.1)
Slimefun RC - 35 (git 5e901972)
Metrics-Module #29
Java 17
Installed Addons: (2)
  CMILib v1.2.5.3
  CMI v9.2.1.1
```
and:
```
Paper git-Paper-411 (MC: 1.17.1)
Slimefun DEV - 1104 (git c977454e)
Metrics-Module #29
Java 17
Installed Addons: (2)
  CMILib v1.2.5.3
  CMI v9.2.1.1
```

### 🧭 Other plugins

_No response_

Dinnerwolph:

Resolves #4020 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -21,8 +21,10 @@ import org.bukkit.entity.Item;
 import org.bukkit.entity.Player;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.ItemStack;
+import org.bukkit.inventory.meta.BookMeta;
 import org.bukkit.inventory.meta.ItemMeta;
 import org.bukkit.inventory.meta.PotionMeta;
+import org.bukkit.inventory.meta.SkullMeta;
 import org.bukkit.metadata.FixedMetadataValue;
 import org.bukkit.persistence.PersistentDataContainer;
 import org.bukkit.persistence.PersistentDataType;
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -489,6 +491,23 @@ private static boolean equalsItemMeta(@Nonnull ItemMeta itemMeta, @Nonnull ItemM
...
+            if(itemBookMeta.getTitle() != sfItemBookMeta.getTitle())
+                return false;
+            if(itemBookMeta.getPageCount() != sfItemBookMeta.getPageCount())
+                return false;
+        }
+
+        if (itemMeta instanceof SkullMeta itemHeadMeta && sfitemMeta instanceof SkullMeta sfItemHeadMeta) {
+            if(itemHeadMeta.getOwningPlayer() != sfItemHeadMeta.getOwningPlayer())
+                return false;
+        }
```
JustAHuman-xD, line 508: Other than the codestyle this looks good, only thing I think is wrong is the if statements are missing brackets and some spaces. 

Other than that functionally its probably okay but you can inline the returns to be like
`return itemBookMeta.getAuthor().equals(sfItemBookMeta.getAuthor() && etc etc`
or
return `itemHeadMeta.getOwningPlayer().equals(sfItemHeadMeta.getOwningPlayer());`
I'm not 100% sure that the player check is a good idea since it's nullable iirc, so you might need to change that, I'd have to check to be completely sure

JustAHuman-xD (changes requested):

PR #4196, opened 2024-05-24
feat: silent give command

ybw0014:
## Description

Adds a `silent_give` command to give items to players without notifying them.

## Proposed changes

Move the logics to `AbstractGiveCommand`, and pass arguments to determine whether the receiver should get a notification.

Discord suggestion [2579](https://discord.com/channels/565557184348422174/693130800853418055/1243284910769438741)

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/AbstractGiveCommand.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/commands/subcommands/AbstractGiveCommand.java
@@ -1,0 +1,109 @@
+package io.github.thebusybiscuit.slimefun4.core.commands.subcommands;
+
+import java.util.Locale;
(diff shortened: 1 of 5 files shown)


JustAHuman-xD:
Is there a reason for the class being abstract here? if all you are changing is the constructer arguments then the class doesn't need to be abstract, just have a give command that's created twice with the corresponding silent and not silent arguments. unless I'm missing something anyways

ybw0014:
> Is there a reason for the class being abstract here? if all you are changing is the constructer arguments then the class doesn't need to be abstract, just have a give command that's created twice with the corresponding silent and not silent arguments. unless I'm missing something anyways

to keep the `SlimefunSubCommands` format consistent.
image

ybw0014:
> Hmmmm, like fair but also aaaaaa

I would like to see others' opinions.

Issue #3176, opened 2021-07-21
Holograms not disappearing again


Skizzles:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

Seems Cargo and other holos are remaining after breaking their linked item again. You can find the same information under #2927 

With a possible fix located #3003 which actually had fixed the issue until recently.

### 📑 Reproduction Steps

Place down a holo based item
Proceed to break the item
Sometimes the holo stays

This is completely random and reproduction steps are still hard to reproduce as last time.

### 💡 Expected Behavior

For holos to disappear.

### 📷 Screenshots / Videos

![image](https://user-images.githubusercontent.com/30563757/126508964-563b6b0f-cb53-419f-841e-8c8b43760a08.png)

### 📜 Server Log

None shown

### 📂 `/error-reports/` folder

None available

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.17.x

### ⭐ Slimefun version

![image](https://user-images.githubusercontent.com/30563757/126508997-949368e5-a4b4-4b3e-ba9e-1876cfee80da.png)

### 🧭 Other plugins

Not applicable to this issue.

FoxWithSox:
similar issue on my server but instead the holograms duplicate when the server restarts 
![image](https://user-images.githubusercontent.com/25159633/126881400-bbb39bfb-a0d9-4d98-81ea-4c38d0c8c1e2.png)
![image](https://user-images.githubusercontent.com/25159633/126881402-b77828ac-bbcb-4493-b960-e4aaa1d0d54b.png)

Syrent:
same issue

variananora:
> same issue

saying 'same issue' without information is useless, please add some information to help us to pin point the problem.

andris155:
The server restart cause this. Duplicate holograms.
https://www.youtube.com/watch?v=99payQA5brg

variananora:
As Panda pointed out in Discord you can just place and break Energy Regulator and have like 10-20% chance of holograms not disappearing. I'm still finding a better way to replicate this more consistently now.

EpicPlayerA10:
I found an issue why this is happening.
[...]

PR #4197, opened 2024-05-29
Fix not disappearing holograms

EpicPlayerA10:
## Description
Fixes not disappearing holograms caused by delayed hologram placement when called from async ( https://github.com/Slimefun/Slimefun4/issues/3176#issuecomment-1025863596 )

## Proposed changes
- Makes `HologramsService#updateHologram` to allows calls only from sync code. Calls from async code should not be allowed as this on its own causes undefined behaviour. For example the `HologramsService#removeHologram` function already does not allow async.
- Make some changes in `HologramOwner#updateHologram` to allow calls from async code. Checks `isDeletedSoon` to be sure that the hologram owner still exists.

- #3176

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
@@ -115,6 +115,14 @@ public class HologramsService {
     }
 
     /**
+     * @see HologramsService#getHologram(Location, boolean)
+     */
+    @Nonnull
+    private Hologram getHologramOrCreate(@Nonnull Location loc) {
+        return getHologram(loc, true);
+    }
+
+    /**
      * This returns the {@link Hologram} associated with the given {@link Location}.
      * If createIfNoneExists is set to true a new {@link ArmorStand} will be spawned
      * if no existing one could be found.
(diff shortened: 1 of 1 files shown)


JustAHuman-xD:
I’ll try and test and review this today, long standing bug but if it’s being called async that would make sense as to why it does that weird behavior, thanks for the PR!

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
```diff
@@ -245,30 +245,20 @@ private Hologram getAsHologram(@Nonnull BlockPosition position, @Nonnull Entity
      * 
      * @param loc
      *            The {@link Location}
-     * @param consumer
-     *            The callback to run
      */
-    private void updateHologram(@Nonnull Location loc, @Nonnull Consumer<Hologram> consumer) {
+    @Nullable
+    private Hologram updateHologram(@Nonnull Location loc) {
```
JustAHuman-xD, line 250: I definitely don't agree with the changes here, this turns update hologram essentially into a wrapper of getHologram and removes the functionality defined in the javadocs

JustAHuman-xD (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
EpicPlayerA10: Hmm, you are totally right. So I think we can safely remove method `updateHologram` and replace its usages with `getHologram(loc, true)` or create a shorthand for this like `getHologramOrCreate`. What do you think about it?

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
```diff
@@ -114,6 +114,14 @@ private void purge() {
...
     }
 
+    /**
+     * @see HologramsService#getHologram(Location, boolean)
+     */
+    @Nonnull
+    private Hologram getHologramOrCreate(@Nonnull Location loc) {
+        return getHologram(loc, true);
+    }
+
```
JustAHuman-xD, line 124:
```suggestion
```
I don't really see the point of making this method instead of just using `getHologram(loc, true)`. If it was used en masse _maybe_, but it's only used once and it's private so I don't think it's really necessary.

JustAHuman-xD (changes requested):
Only other thing that may or may not be a problem is the removal of the exception catching and therein logging

src/main/java/io/github/thebusybiscuit/slimefun4/core/services/holograms/HologramsService.java
EpicPlayerA10: Yeah, but the only difference in the `getHologram` and `getHologramOrCreate` methods is that the `getHologram` has `@Nullable` annotation and `getHologramOrCreate` has `@Nonnull` annotation. I just created this method for cleaner code and so that IntelliJ doesn't yell at me that i'm using a possibly null variable.

EpicPlayerA10:
> Only other thing that may or may not be a problem is the removal of the exception catching and therein logging

The error will be logged by java anyway because the thrown exception from `getHologram` will be propagated to the stacktrace. I think it is better to throw exception and stop execution of further code where `getHologram` was called, than only printing error in the console and allowing further code execution creating undefined behaviour.

Issue #4069, opened 2023-12-23
sensitive blocks server crashes


100petr:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

Server crashes with error about sensitive blocks (last thing about it I found this PR https://github.com/Slimefun/Slimefun4/pull/4050)

### 📑 Reproduction Steps

it specifically happened to me when I was destroying a custom tree from itemsadder
to someone else when harvesting wines in bulk https://discord.com/channels/565557184348422174/565569196218384385/1187986406484750406

### 💡 Expected Behavior

the server will not be overloaded and will not crash

### 📷 Screenshots / Videos

https://www.youtube.com/watch?v=pqGNxOiH5N0

### 📜 Server Log

https://pastebin.com/pjg05cMU

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

Slimefun version Dev - 1117

WalshyDev:
Related: https://github.com/Slimefun/Slimefun4/pull/3966 & https://github.com/Slimefun/Slimefun4/pull/4048 & https://github.com/Slimefun/Slimefun4/pull/4050

cc: @iTwins @JustAHuman-xD @Sfiguz7 

We've had a few reports of this now hanging servers, I haven't dug into this code much or the situations around it but want to highlight for y'all

JustAHuman-xD:
To fix this there are a couple things we could do, I'll talk to the others about it, I get home in a couple days so I can make a PR then

WalshyDev:
This code was disabled in build 1121, we'll figure out a better path forward (and probably config option) for next time.

Issue #4037, opened 2023-12-02
Explosive pickaxe duplicates elevator plate when breaking it (or in the explode range)


EienMiku:
### ❗ Checklist

- [X] I am using the official english version of Slimefun and did not modify the jar.
- [X] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [X] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [X] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

When I break a elevator plate or nearby block with an explosive pickaxe, the plate gets duplicated.

### 📑 Reproduction Steps

1. Break a elevator plate or nearby block with an explosive pickaxe.
2. The elevator plate got duplicated.

### 💡 Expected Behavior

The correct drops should be 1 of each, not 2.

### 📷 Screenshots / Videos

[21:52:55 ERROR]: [Slimefun] An Error occurred while copying a temporary File for Slimefun DEV - 1104 (git c977454e)
java.nio.file.AccessDeniedException: data-storage\Slimefun\stored-blocks\world\ELECTRIC_SMELTERY_2.sfb.tmp -> data-storage\Slimefun\stored-blocks\world\ELECTRIC_SMELTERY_2.sfb
        at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:89) ~[?:?]
        at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103) ~[?:?]
        at sun.nio.fs.WindowsFileCopy.move(WindowsFileCopy.java:317) ~[?:?]
        at sun.nio.fs.WindowsFileSystemProvider.move(WindowsFileSystemProvider.java:293) ~[?:?]
        at java.nio.file.Files.move(Files.java:1432) ~[?:?]
        at me.mrCookieSlime.Slimefun.api.BlockStorage.save(BlockStorage.java:336) ~[Slimefun4 - DEV 1104 (git c9774).jar:?]
        at io.github.thebusybiscuit.slimefun4.core.services.AutoSavingService.saveAllBlocks(AutoSavingService.java:95) ~[Slimefun4 - DEV 1104 (git c9774).jar:?]
[...]

PR #4200, opened 2024-05-30
Fix not dropping sensitive blocks once and for all

EpicPlayerA10:
## Description
Fixed not dropping sensitive blocks once and for all

## Proposed changes
This PR takes a different approach from all other solutions. It listens for `BlockPhysicsEvent` which fires before block update. Inside listener it checks if block can be placed in the location where event was fired. This is the best method to listen for collapsing sensitive blocks without a headache with `SENSITIVE_MATERIALS` tag.

The only downside of this solution is that it require at least version 1.18 because it uses `BlockData#isSupported` method.

- #3182
- #3831
- #3009
- #4069
- #4037

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/handlers/BlockBreakHandler.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/handlers/BlockBreakHandler.java
@@ -68,7 +68,7 @@ public abstract class BlockBreakHandler implements ItemHandler {
     /**
      * Called when sensitive block was collapsed by breaking the supporting block
      *
-     * @param block Our {@link Block}
+     * @param block Collapsed {@link Block}
      */
     @ParametersAreNonnullByDefault
     public void onCollapse(Block block) {


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockPhysicsListener.java
```diff
@@ -112,4 +115,34 @@ public void onBucketUse(PlayerBucketEmptyEvent e) {
...
+                    if (drop != null && !drop.getType().isAir()) {
+                        block.getWorld().dropItemNaturally(block.getLocation(), drop);
+                    }
+                }
+
+                // Fixes #2944 - Don't forget to clear the Block Data
+                BlockStorage.clearBlockInfo(block);
+            }
+        }
+    }
```
JustAHuman-xD, line 147: Not a huge fan of using this event at all due to the mass amounts of times its called and the lag that can be easily created by listening to it.

If we do go this route (which I'm not completely against just hesitant personally), we need to ensure that we do it as performantly as possible. Starting by re-arranging some of these filter ifs like doing the sf item checks before the supported check (as blockstorage checks are just checks to a map and are definitely more performant then the is supported method). 

We should also ensure we are only executing this code when the physics check calls for it, i.e. when the block is being broken because it's no longer supported, as there are definite cases we don't need to check at all.

JustAHuman-xD (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockPhysicsListener.java
EpicPlayerA10: > Not a huge fan of using this event at all due to the mass amounts of times its called and the lag that can be easily created by listening to it.

Well I also thought like you but I ran a spark profiler on my local server and to my surprise it takes very little tick time.

> If we do go this route (which I'm not completely against just hesitant personally), we need to ensure that we do it as performantly as possible. Starting by re-arranging some of these filter ifs like doing the sf item checks before the supported check (as blockstorage checks are just checks to a map and are definitely more performant then the is supported method).

Done

> We should also ensure we are only executing this code when the physics check calls for it, i.e. when the block is being broken because it's no longer supported, as there are definite cases we don't need to check at all.

I don't understand what you mean by writing this.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockPhysicsListener.java
```diff
@@ -112,4 +115,34 @@ public void onBucketUse(PlayerBucketEmptyEvent e) {
...
+                block.setType(Material.AIR, true);
+
+                for (ItemStack drop : sfItem.getDrops()) {
+                    if (drop != null && !drop.getType().isAir()) {
+                        block.getWorld().dropItemNaturally(block.getLocation(), drop);
+                    }
+                }
+
+                // Fixes #2944 - Don't forget to clear the Block Data
+                BlockStorage.clearBlockInfo(block);
```
JustAHuman-xD, line 144: We are also missing the call to block break handler, which could lead to massive amounts of unintended behavior

EpicPlayerA10: Ah, good catch. Done

PR #4214, opened 2024-06-28
Fix errors in SlimefunUtils

### LINKED ISSUE #4212: Wrongs in SlimefunUtils
### ❗ Checklist

### 📍 Description

From #4211 

In a word,there is a wrong check.And some incorrect use of variables(or wrong name?)(Found out while researching this issue)
<details>
<summary>Problem tracing</summary>

https://github.com/Slimefun/Slimefun4/blob/226af9be61d1bfc582cbe235909579b9f4f605cd/src/main/java/me/mrCookieSlime/Slimefun/api/inventory/DirtyChestMenu.java#L178-L183

https://github.com/Slimefun/Slimefun4/blob/226af9be61d1bfc582cbe235909579b9f4f605cd/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java#L306-L322

https://github.com/Slimefun/Slimefun4/blob/226af9be61d1bfc582cbe235909579b9f4f605cd/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java#L348-L387

Herein lies the problem——`equalsItemMeta`method

https://github.com/Slimefun/Slimefun4/blob/226af9be61d1bfc582cbe235909579b9f4f605cd/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java#L459-L501

**Here's a thing to say about the `checkCustomModelData` variable in this method called `bypassCustomModelCheck`, which actually ignores checking, so it should be reversed when it is passed in, but it is not reversed when it is passed in**

This code checks Lore & Name,The problem arises

https://github.com/Slimefun/Slimefun4/blob/226af9be61d1bfc582cbe235909579b9f4f605cd/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java#L364-L386

Check here to see if you have overlooked a situation where it is a slimefun item and the two items are not the same slimefun item

https://github.com/Slimefun/Slimefun4/blob/226af9be61d1bfc582cbe235909579b9f4f605cd/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java#L358-L378

The variable name is `sfItem`, which means that `sfItem` is a slimefun item by default. In the case of `id==null` (i.e. the item is not a Slimefun item), it is't necessary to continue to determine if it is necessary to return false if one is a Slimefun item and the other is not.

Also, since `sfitem.hasItemMeta()` has been judged
Why should the variable of this ItemMeta be named `possibleSfItemMeta`, and the same goes for `possibleItemId`

In a word,maybe change like:
```java
if (id != null && id.equals(possibleItemId)) { 
                     Debug.log(TestCase.CARGO_INPUT_TESTING, "  Item IDs matched!"); 
  
                     /* 
                      * PR #3417 
                      * 
                      * Some items can't rely on just IDs matching and will implement Distinctive Item 
                      * in which case we want to use the method provided to compare 
                      */ 
                     Optional<DistinctiveItem> optionalDistinctive = getDistinctiveItem(id); 
                     if (optionalDistinctive.isPresent()) { 
                         return optionalDistinctive.get().canStack(possibleSfItemMeta, itemMeta); 
                     } 
                     return true; 
                 }
                 return false;
```
</details>

### 📑 Reproduction Steps

~~Put different slimefun items with the same lore and name into the charging bench
And then they stacks.~~
After the method is called, an error result is returned.
There is a wrong checking.

### 💡 Expected Behavior

~~they shouldn't stack~~

The method should run correctly.

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Purpur

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

latest

### 🧭 Other plugins

_No response_

mcchampions:
## Description
Fix #4212 

This issue explains why to change.

**or move `equalsItemMeta` check to `if (id != null && id.equals(possibleItemId)) { `?**

## Proposed changes
- remove `equalsItemMeta` check
- rename var

#4212 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -38,7 +38,7 @@ import io.github.thebusybiscuit.slimefun4.utils.PatternUtils;
  *
  * @see StormStaff
  */
-public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> {
+public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> implements DistinctiveItem {
 
     private final NamespacedKey defaultUsageKey;
     private int maxUseCount = -1;
@@ -149,4 +149,13 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
         }
     }
 
+    @Override
+    public boolean canStack(ItemMeta itemMetaOne, ItemMeta itemMetaTwo) {
+        NamespacedKey key = getStorageKey();
+        PersistentDataContainer pdc1 = itemMetaOne.getPersistentDataContainer();
+        int usesLeft1 = pdc1.getOrDefault(key, PersistentDataType.INTEGER, getMaxUseCount());
+        PersistentDataContainer pdc2 = itemMetaTwo.getPersistentDataContainer();
+        int usesLeft2 = pdc2.getOrDefault(key, PersistentDataType.INTEGER, getMaxUseCount());
+        return usesLeft1 == usesLeft2;
+    }
 }
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
@@ -160,6 +160,6 @@ public abstract class LimitedUseItem extends SimpleSlimefunItem<ItemUseHandler> 
         NamespacedKey key = getStorageKey();
         int usesLeft1 = PersistentDataAPI.getInt(itemMetaOne, key);
         int usesLeft2 = PersistentDataAPI.getInt(itemMetaTwo, key);
-        return usesLeft1 != -1 && usesLeft1 == usesLeft2;
+        return usesLeft1 == usesLeft2;
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -149,4 +150,16 @@ private void updateItemLore(ItemStack item, ItemMeta meta, int usesLeft) {
         }
     }
 
+    @Override
+    public boolean canStack(ItemMeta itemMetaOne, ItemMeta itemMetaTwo) {
+        if (Slimefun.getItemDataService().getItemData(itemMetaOne) != Slimefun.getItemDataService().getItemData(itemMetaTwo)) {
```
JustAHuman-xD, line 155: This returns optionals right? You shouldn't use a != here

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
```diff
@@ -149,4 +150,16 @@ private void updateItemLore(ItemStack item, ItemMeta meta, int usesLeft) {
...
+            return false;
+        }
+        NamespacedKey key = getStorageKey();
+        PersistentDataContainer pdc1 = itemMetaOne.getPersistentDataContainer();
+        int usesLeft1 = pdc1.getOrDefault(key, PersistentDataType.INTEGER, getMaxUseCount());
+        PersistentDataContainer pdc2 = itemMetaTwo.getPersistentDataContainer();
+        int usesLeft2 = pdc2.getOrDefault(key, PersistentDataType.INTEGER, getMaxUseCount());
```
JustAHuman-xD, line 162: You can also use PersistantDataAPI instead of the raw pdc

JustAHuman-xD (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/LimitedUseItem.java
mcchampions: oh, i didn't notice it return an optional(beacuse i change it in github)

PR #4220, opened 2024-08-10
Use BlockPositions internally instead of Locations

JustAHuman-xD:
## Description
BlockStorage & related classes currently use Locations internally instead of BlockPositions. Locations are comprised of a Reference<World>, 3 doubles, and 2 floats, while BlockPositions are comprised of only a WeakReference<World>, and a long.
By switching to using BlockPositions we would save MASSIVELY on memory with the current implementation of BlockStorage. 

Once this PR is ready to go and approved I will make a matching PR that updates all of Slimefun's internals to use the new methods instead of the deprecated old ones.

## Proposed changes
- (Very simple overview, will write a full blown one later)
- Refactors BlockStorage, TickerTask, BlockMenu, BlockMenuPreset, SlimefunProfiler, ProfiledBlock, and ErrorReport to use BlockPositions instead of Locations
  - A couple methods needed to be renamed due to their signature & to avoid breaking changes
  - I still need to setup the javadocs for the new methods
  - To avoid breaking changes all old methods still exist and just overflow to the new methods.
    - I still need to update java docs and add deprecation warnings to all of these

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/api/ErrorReport.java
@@ -17,6 +17,7 @@ import java.util.stream.IntStream;
 import javax.annotation.Nonnull;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.bakedlibs.dough.blocks.BlockPosition;
 import org.bukkit.Bukkit;
 import org.bukkit.Location;
 import org.bukkit.plugin.Plugin;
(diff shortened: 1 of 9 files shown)


JustAHuman-xD:
*Potentially* controversial things I want opinions on:
- adding `BlockStorage#getBlockInfo` for the new versions of `BlockStorage#getLocationInfo` as most other methods call it BlockInfo
- `BlockStorage#getRawStorage` & `TickerTask#getLocations` return empty maps/sets now, we *could* make it remap the data to use locations and return that but that would kinda defeat the point of the memory conservation effort being made with this PR, open to discussion ofc tho
- Deprecating the old methods, I feel this is necessary since there would be no reason to *not* use the new methods & it would encourage using BlockPosition and not Location
- Possibly other things but these are the main ones I can think of rn

JustAHuman-xD:
Once this PR is ready/in a good state I'll make an additional PR that updates all of the slimefun implementations, tests, etc, that use the old methods to use the new ones

md5sha256:
> _Potentially_ controversial things I want opinions on:
> 
> * adding `BlockStorage#getBlockInfo` for the new versions of `BlockStorage#getLocationInfo` as most other methods call it BlockInfo
> * `BlockStorage#getRawStorage` & `TickerTask#getLocations` return empty maps/sets now, we _could_ make it remap the data to use locations and return that but that would kinda defeat the point of the memory conservation effort being made with this PR, open to discussion ofc tho
> * Deprecating the old methods, I feel this is necessary since there would be no reason to _not_ use the new methods & it would encourage using BlockPosition and not Location
> * Possibly other things but these are the main ones I can think of rn

I'm pretty sure `#getRawStorage` and `#getLocations` are used so they absolutely cannot return empty values. I think we should deprecate them for the new methods but still have them return a copy.

JustAHuman-xD:
> I'm pretty sure `#getRawStorage` and `#getLocations` are used so they absolutely cannot return empty values. I think we should deprecate them for the new methods but still have them return a (immutable?) copy. 

The only usages I could find for getrawstorage outside of base SF was headlimiter (which I can PR to  fix) and my own project that isn't released.

As for getLocations I couldn't find any usages but I could obviously have missed some.

The thing with copying it is that it'll use WAY more memory than the old method did since every time the method is called itll create a NEW location instance for every block position. While previously the immutable map would have (I assume) just been references.

Sfiguz7:
Closure of shame incoming

JustAHuman-xD:
*me and Idra did major bad napkin math, the memory save is negligible* #blame-idra

Issue #514, opened 2017-11-30
CargoNet.matchesFilter() creates a lot of lag


krusic22:
## Description
Describe your Issue AS DETAILED AS POSSIBLE.
CargoNet.matchesFilter() creates a lot of lag...

## Server Log
Upload your FULL SERVER LOG (/logs/latest.log) to http://pastebin.com/ and paste your link in here.
Without this File, we may not be able to help you at all.

Might not be directly connected!

`java.lang.IllegalStateException:
Asynchronous Async Chunk Load! Blocking thread until it returns 
at net.minecraft.server.v1_11_R1.MCUtil.ensureMain(MCUtil.java:51)
at org.bukkit.craftbukkit.v1_11_R1.chunkio.ChunkIOExecutor.syncChunkLoad(ChunkIOExecutor.java:18)
at net.minecraft.server.v1_11_R1.ChunkProviderServer.getChunkAt(ChunkProviderServer.java:164)
at net.minecraft.server.v1_11_R1.ChunkProviderServer.getChunkAt(ChunkProviderServer.java:147)
at net.minecraft.server.v1_11_R1.ChunkProviderServer.getChunkAt(ChunkProviderServer.java:143)
at org.bukkit.craftbukkit.v1_11_R1.CraftWorld.getChunkAt(CraftWorld.java:174)
at org.bukkit.craftbukkit.v1_11_R1.CraftWorld.getBlockAt(CraftWorld.java:118)
at org.bukkit.craftbukkit.v1_11_R1.CraftWorld.getBlockAt(CraftWorld.java:606)
at org.bukkit.Location.getBlock(Location.java:87)
at me.mrCookieSlime.Slimefun.Setup.SlimefunSetup$50.generateEnergy(SlimefunSetup.java:3181)
at me.mrCookieSlime.Slimefun.api.energy.EnergyNet.tick(EnergyNet.java:102)
at me.mrCookieSlime.Slimefun.Setup.SlimefunSetup$47.tick(SlimefunSetup.java:3112)
at me.mrCookieSlime.Slimefun.api.TickerTask.run(TickerTask.java:204)
at org.bukkit.craftbukkit.v1_11_R1.scheduler.CraftTask.run(CraftTask.java:58)
at org.bukkit.craftbukkit.v1_11_R1.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:52)
[...]

PR #4231, opened 2024-09-04
api/Open up range field

Intybyte:
## Description
Add new EXP collectors

EXP Collector II: Collects XP orbs in a 7x7x7 area.
EXP Collector III: Collects XP orbs in a 15x15x15 area.

## Proposed changes
Make EXP collector's constructor have a range and remove hardcoded capacity and consumption

## Approved suggestion
Discord approved suggestion number [#514](https://discord.com/channels/565557184348422174/627048923122499584/1265651763818659953)

## What is left to do
I need someone to tell me the recipes and or changes to the consumption and capacity of these new machines, so that I can register them into `SlimefunItemSetup`

## Checklist


Boomer-1:
is there any real application to this? other than wolves killing mobs, and zombified piglins, both farms of which are in tight narrow kill areas, what would be the purpose to go out 15 blocks?

Intybyte:
> It's worth noting a few things:
> 
>     * You've not actually made SlimefunItems for the two additional SlimefunItemStacks you've made
> 
>     * Increasing range when checking for entities can harm performance so the reason would need to be considered vs the new, additional, cost
> 
>     * The power requirement to range increase seems low to me, though I could be wrong here
> 
> 
> Overall I like the idea of having range within the object so at least addons can extend the class to do something similar here? Would like to hear further opinions before I do anything more.

The addon one was one of the main reason for me too, all the points you made are valid, reason why in my PR description towards the end I said that I am waiting for an adjustement for energy and a recipe, i don't want to register something with null recipe, if the idea is accepted and the recipe made and energy cost made I will of course add them

JustAHuman-xD:
Yeah the 15x15x15 seems incredibly unnessesary and taxing in performance to boot imo

Intybyte:
> Yeah the 15x15x15 seems incredibly unnecessary and taxing in performance to boot imo

There are many machines with only tier 2, we can just remove the tier 3 if deemed unnecessary, this way we add a new feature and allow addon makers to make their own xp collector easily if needed, still better than the way everything was hardcoded in that class

Sefiraat:
> > Yeah the 15x15x15 seems incredibly unnecessary and taxing in performance to boot imo
> 
> There are many machines with only tier 2, we can just remove the tier 3 if deemed unnecessary, this way we add a new feature and allow addon makers to make their own xp collector easily if needed, still better than the way everything was hardcoded in that class

I’d go so far as to say perhaps to leave the new tiers out completely and just have the api opened up like you have done. That’s would be a quick accept as far as I’m concerned.

Intybyte:
Updated now it only opens up the api

JustAHuman-xD (dismissed):
LGTM

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/entities/ExpCollector.java
```diff
@@ -44,35 +44,38 @@ public class ExpCollector extends SlimefunItem implements InventoryBlock, Energy
...
-    private static final int ENERGY_CONSUMPTION = 10;
     private static final String DATA_KEY = "stored-exp";
 
+    private final double range;
+    private int energyConsumedPerTick = -1;
+    private int energyCapacity = -1;
+
     @ParametersAreNonnullByDefault
-    public ExpCollector(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+    public ExpCollector(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe, double range) {
```
WalshyDev, line 54: if anyone is extending this today it's a breaking change

new constructor with range and a default passed in will need to be done here.

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/entities/ExpCollector.java
```diff
@@ -44,35 +44,38 @@ public class ExpCollector extends SlimefunItem implements InventoryBlock, Energy
...
         super(itemGroup, item, recipeType, recipe);
 
         createPreset(this, this::constructMenu);
 
+        this.range = range;
```
WalshyDev, line 59: any reason this is done below createPreset?
this. in constructor should always be at the top

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/entities/ExpCollector.java
```diff
@@ -44,35 +44,42 @@ public class ExpCollector extends SlimefunItem implements InventoryBlock, Energy
...
     private static final String DATA_KEY = "stored-exp";
 
+    private final double range;
+    private int energyConsumedPerTick = -1;
+    private int energyCapacity = -1;
+
     @ParametersAreNonnullByDefault
     public ExpCollector(ItemGroup itemGroup, SlimefunItemStack item, RecipeType recipeType, ItemStack[] recipe) {
+        this(itemGroup, item, recipeType, recipe, 4.0);
+    }
```
JustAHuman-xD, line 56: should we maybe deprecate this constructer as well?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/electric/machines/entities/ExpCollector.java
```diff
@@ -44,35 +44,43 @@ public class ExpCollector extends SlimefunItem implements InventoryBlock, Energy
...
     private final int[] border = { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26 };
 
-    private static final int ENERGY_CONSUMPTION = 10;
     private static final String DATA_KEY = "stored-exp";
 
+    private final double range;
+    private int energyConsumedPerTick = -1;
+    private int energyCapacity = -1;
+
+    @Deprecated
```
WalshyDev, line 53:
```suggestion
    @Deprecated(since = "RC-37", forRemoval = true)
```

WalshyDev (approved):

JustAHuman-xD (approved):

Issue #1592, opened 2020-02-27
Slimefun told me to post this issue through the console


meisjuan:
## Description (Required)
the plugin just told me to post the error report here

## Steps to reproduce the Issue (Required)
when someone interact with a certain slimefun block

## Expected behavior (Required)
players have not reported anything, the plugin just told me to post it here

## Server Log / Error Report
https://pastebin.com/hmG2iSHn

## Environment (Required)

 - Minecraft Version: PaperMC 1.14.4 (paper-243.jar) 
 - Slimefun Version: Slimefun4 - DEV 444 (git 8e3d4)
 - CS-CoreLib Version: CS-CoreLib+v1.6.1

Note: in the error report it says i'm running CS-CoreLib v1.6.2 and Slimefun vDEV - 449 (git 05de09c8) which is odd.. because my server is not running that version..

TheBusyBiscuit:
Duplicate of #1577

TheBusyBot:
> Your issue has been closed!

If your issue was closed **and not resolved** by one of our developers it may fall under one or more of these categories:

* [ ] You did not follow our template. Please follow the Issue template to help us identify your issue more quickly.
* [ ] You did not provide any information about your versions (We absolutely need your exact versions, \"latest\" is not helpful)
* [ ] You did not provide a proper description to the problem.
* [ ] Your issue is not a bug, it is intended behaviour.
* [ ] Your issue was already reported, it is a duplicate. Check the other open issues first before posting!
* [ ] You posted an error without using pastebin. Please always post errors via pastebin.
* [ ] Your issue has already been fixed in a later version of Slimefun or CS-CoreLib, you should update.
* [ ] You are using an outdated version of Slimefun / CS-CoreLib, again, you should update.
* [ ] You are using an unsupported version of Minecraft. We only provide support for the latest version of Minecraft, old versions are not supported anymore.
* [ ] You are using a \"stable\" version of Slimefun (prefixed with \"RC - \"), your issue may have been fixed in a development build, so we only accept bug reports from those.
* [ ] You are on the wrong issue tracker. We would like to remind you that this Issue Tracker is **only for Slimefun**, to report bugs on any addons, head to the corresponding issue tracker.

Make sure to check out our article on [How to report bugs](https://github.com/TheBusyBiscuit/Slimefun4/wiki/How-to-report-bugs) for even more information.

PR #4232, opened 2024-09-04
Feature/amethyst farm

Intybyte:
## Description
An amethyst virtual farm, that would work like the others virtual farms, could be useful considering some players play on worlds where the nearest Amethyst Geode can be really far as their world started in earlier versions.

## Proposed changes
Create an abastract class dedicated to generating item using energy only, the recipes will require 1 input and 1 output max and clicking on the input slot will toggle the item generated.

CrystalGrower which will grow amethyst will be the first block to use such class.

## TODO
Decide recipe
Decide power settings (capacity, speed, consumption)

## Approved suggestion
Discord approved suggestion number [#1592](https://discord.com/channels/565557184348422174/627048923122499584/1265647680911310941)

## Checklist


Boomer-1:
for what it's worth, there's currently 3 or 4 addons that provide ways to get amethyst. not against it, but want to weigh that in

Sfiguz7:
I have yet to look at it but if it's thought of as a virtual farm shouldn't it go in infinity with the other virtual farms?

Intybyte:
> I have yet to look at it but if it's thought of as a virtual farm shouldn't it go in infinity with the other virtual farms?

I have no idea if that was the intention of the creator of the suggestion, tbh i just fish ideas that i would like to code and make PR about them and if they are merged good, if they are not i had fun while doing it, the whole suggestion channel imo should be divided into a `#suggesitons` and `#addon-suggestions`

TheBusyBiscuit (commented):
A nice idea but I am not a fan of such a machine in vanilla Slimefun. It might work good for addons and apparently there are already addons providing such features.

So from my side I would reject this change as I do not see it as a good fit for the base project.

TheBusyBiscuit:
> > I have yet to look at it but if it's thought of as a virtual farm shouldn't it go in infinity with the other virtual farms?
> 
> I have no idea if that was the intention of the creator of the suggestion, tbh i just fish ideas that i would like to code and make PR about them and if they are merged good, if they are not i had fun while doing it, the whole suggestion channel imo should be divided into a `#suggesitons` and `#addon-suggestions`

It is a good point, we tried to heavily enforce things like "only suggest ideas for slimefun, not addons" in the past but it did not work out. Seperating channels is also something we have had not much luck with.
In general in many players' minds there is no real seperation between Slimefun and addons as there is from a developer's perspective. But I absolutely get your point! Glad you had fun doing it and hope to see you in a future PR o/

PR #4234, opened 2024-09-06
Remove block info if present when placing on air (more in desc)

JustAHuman-xD:
## Description
Currently SF supports removing ghost blocks by placing a block where the "ghost block" is, however it only supports ghost blocks of registered items. If an item is no longer registered but the data is still there, it does not get removed. This just changes the ghost block handling to account for that as well.

## Proposed changes
Make the clearing require block info, not an sf item

Issue with Dynatech's new update

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
@@ -70,18 +70,18 @@ public class BlockListener implements Listener {
         Block block = e.getBlock();
 
         // Fixes #2636 - This will solve the "ghost blocks" issue
-        if (e.getBlockReplacedState().getType().isAir()) {
+        if (e.getBlockReplacedState().getType().isAir() && BlockStorage.hasBlockInfo(block) && !Slimefun.getTickerTask().isDeletedSoon(block.getLocation())) {
             SlimefunItem sfItem = BlockStorage.check(block);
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BlockListener.java
```diff
@@ -70,18 +70,18 @@ public void onBlockPlaceExisting(BlockPlaceEvent e) {
         Block block = e.getBlock();
 
         // Fixes #2636 - This will solve the "ghost blocks" issue
-        if (e.getBlockReplacedState().getType().isAir()) {
+        if (e.getBlockReplacedState().getType().isAir() && BlockStorage.hasBlockInfo(block) && !Slimefun.getTickerTask().isDeletedSoon(block.getLocation())) {
             SlimefunItem sfItem = BlockStorage.check(block);
 
-            if (sfItem != null && !Slimefun.getTickerTask().isDeletedSoon(block.getLocation())) {
```
J3fftw1, line 76: Removing #isDeletedSoon will cause issues as well we added that so people can’t spam place blocks and dupe

J3fftw1 (approved):
LGTM but would love for it to be tested

Boomer-1:
tested and can confirm it removed the old block data, so new blocks placed in the same spot as the old ones work as intended and ghost block data is removed

Sfiguz7 (approved):
LGTM

PR #4241, opened 2024-09-12
Add powder snow block to the list of blocks that the Explosive Shovel can break

RenanMsV:
## Description

Hello, I was playing the minecraft 1.20.4 and noticed that the [Explosive Shovel](https://github.com/Slimefun/Slimefun4/wiki/Explosive%20Shovel)  can't break powder snow blocks. I believe they should be able to break this block. 

## Proposed changes

Add powder snow block to the list of blocks that the Explosive Shovel can break.

An issue is that this block came in minecraft 1.17, so most likely 1.16 would not be compatible.
This might not be an issue but this block drops nothing, since it can only be collected with buckets like a liquid.

## Checklist

### DIFF
--- a/src/main/resources/tags/explosive_shovel_blocks.json
+++ b/src/main/resources/tags/explosive_shovel_blocks.json
@@ -5,6 +5,7 @@
 		"#slimefun:dirt_variants",
 		"minecraft:snow",
 		"minecraft:snow_block",
+		"minecraft:powder_snow",
 		"minecraft:gravel",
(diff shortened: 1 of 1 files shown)


Boomer-1:
with 1.21, support will be dropped for 1.16.x and 1.17, so it might be worth waiting to merge until after the 1.21 release if it's approved.

ybw0014:
to make it compatible right now, u need to mark it as not required, like the `soul_soil` field.

PR #4242, opened 2024-09-15
Fixed bad formatting

m3z0id:
Fixed language

## Description
Fixed the bad formatting


ybw0014:
Localization should be updated on Crowdin

PR #4247, opened 2024-09-26
Feature/flint and steal unbreaking

Intybyte:
## Description
Make the Automatic Ignition chamber respect the level of the Unbreaking enchantment on the tool selected for durability loss (thus, an Unbreaking 3 Flint and Steel would last 4 times as long on average).

## Proposed changes
Refactor the code releated to damaging the flint and steal into a `damageFlintAndSteel` method that uses vanilla MC unbreaking calculation to determine when to damage the flint and steal

[Suggestion 2429](https://discord.com/channels/565557184348422174/627048923122499584/1273071970149339138
) #approved

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/IgnitionChamber.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/IgnitionChamber.java
@@ -4,6 +4,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.ParametersAreNonnullByDefault;
 
+import io.github.thebusybiscuit.slimefun4.utils.compatibility.VersionedEnchantment;
 import org.apache.commons.lang.Validate;
 import org.bukkit.Material;
 import org.bukkit.block.Block;
@@ -11,6 +12,7 @@ import org.bukkit.block.BlockFace;
 import org.bukkit.block.BlockState;
 import org.bukkit.block.Dropper;
 import org.bukkit.block.data.type.Dispenser;
+import org.bukkit.enchantments.Enchantment;
 import org.bukkit.entity.Player;
 import org.bukkit.inventory.Inventory;
 import org.bukkit.inventory.ItemStack;
@@ -28,6 +30,8 @@ import io.github.thebusybiscuit.slimefun4.implementation.items.multiblocks.Smelt
 import io.papermc.lib.PaperLib;
 
 import me.mrCookieSlime.Slimefun.api.BlockStorage;
+
+import java.util.concurrent.ThreadLocalRandom;
 
 /**
  * The {@link IgnitionChamber} is used to re-ignite a {@link Smeltery}.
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/IgnitionChamber.java
```diff
@@ -82,31 +85,43 @@ public static boolean useFlintAndSteel(Player p, Block smelteryBlock) {
...
+        Damageable damageable = (Damageable) meta;
+
+        if (meta.isUnbreakable()) {
+            return;
+        }
+
+        Enchantment unbreaking = VersionedEnchantment.UNBREAKING;
+        int lvl = flintAndSteel.getEnchantmentLevel(unbreaking);
+
+        if (UnbreakingAlgorithm.TOOLS.evaluate(lvl)) {
```
JustAHuman-xD, line 114:
```suggestion
        if (!UnbreakingAlgorithm.TOOLS.evaluate(lvl)) {
```
this needs to be inverted, evaluate returns if unbreaking proc'd, not if it didn't

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/IgnitionChamber.java
```diff
@@ -82,31 +85,43 @@ public static boolean useFlintAndSteel(Player p, Block smelteryBlock) {
...
+            damageable.setDamage(damageable.getDamage() + 1);
+        }
+
+        if (damageable.getDamage() >= flintAndSteel.getType().getMaxDurability()) {
```
JustAHuman-xD, line 118: not sure how we should handle components here, since items can now have their own max durability, we might need to have a util somewhere that returns getType#getMaxDura if version is below components otherwise returns damageable.getMaxDamage() (or whatever the method is can't remember)

JustAHuman-xD (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/items/blocks/IgnitionChamber.java
Intybyte: A concrete example would be helpful, also i think it is too soon to consider this problem as i doubt that feature is fully implemented

JustAHuman-xD: That feature is fully implemented and I have used it myself lol.

For example an addon adds "Flint and Damascus-Steel", it's normal flint and steel except has more durability using the new component system. Lets say it has 256 durability unlike the normal (64? I think).

If they put it in the ignition chamber once the durability is 192/256 or less it'll get removed even though it has more to spend

Intybyte: ok didn't know that, when was that added?

JustAHuman-xD: 1.20.5+ when item components were introduced, any item regardless of type can be given durability

PR #4252, opened 2024-10-24
[Api] Give & Pick Slimefun Item api

JustAHuman-xD:
## Description
Addons with complex SlimefunItem's that can have data different for the same slimefun item often have problems with the middle click & give item command, as they may lack some data that would be there when attained in survival, this PR adds 2 methods to help solve that problem.

## Proposed changes
- Added `SlimefunItem#getGiveItemResult(Player)`
  - Defaults to `getItem` and is called in `GiveCommand` in place of `getItem`
- Added `SlimefunItem#getPickBlockResult(Player, Block)`
  - Also defaults to `getItem` and is called in `MiddleClickListener` in place of `getItem`
    - Also adds an instanceof to `MiddleClickListener` as the event returns `HumanEntity` not `Player` and I didn't want it to be inconsistent, testing is needed to make sure it still works

N/A

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
@@ -42,7 +42,10 @@ public class MiddleClickListener implements Listener {
          * ClickType is not MIDDLE but CREATIVE (because this ClickType covers
          * multiple cases, we have to filter out more later on)
          */
-        if (e.getClick() == ClickType.CREATIVE && e.getSlotType() == SlotType.QUICKBAR && e.getWhoClicked() instanceof Player player) {
+        if (e.getClick() == ClickType.CREATIVE
+        	&& e.getSlotType() == SlotType.QUICKBAR
+        	&& e.getWhoClicked() instanceof Player player
+        ) {
             // get the block the player is looking at for later
             Block b = player.getTargetBlockExact(5);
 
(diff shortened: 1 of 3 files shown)


Seggan (approved):

Intybyte (approved):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/MiddleClickListener.java
```diff
@@ -41,8 +42,7 @@ public void onInventoryCreativeEvent(InventoryCreativeEvent e) {
          * ClickType is not MIDDLE but CREATIVE (because this ClickType covers
          * multiple cases, we have to filter out more later on)
          */
-        if (e.getClick() == ClickType.CREATIVE && e.getSlotType() == SlotType.QUICKBAR) {
-            HumanEntity player = e.getWhoClicked();
+        if (e.getClick() == ClickType.CREATIVE && e.getSlotType() == SlotType.QUICKBAR && e.getWhoClicked() instanceof Player player) {
```
WalshyDev, line 45:
```suggestion
        if (e.getClick() == ClickType.CREATIVE
        	&& e.getSlotType() == SlotType.QUICKBAR
        	&& e.getWhoClicked() instanceof Player player
        ) {
```

WalshyDev (dismissed):
I really like this, great job!

I'm not the _biggest_ fan of giveItem since /sf give is a command not just a special mc thingy but, I think it's fine. Generally, I try to make APIs that work first-party the same as they do third-party. So, any addon could do something similar rather than it being a special Slimefun hardcoded thing

For this situation, I think it's fine. I haven't seen others asking for a similar functionality and if someone wants to make one, we can easily move over.

TheBusyBiscuit (dismissed):
Also a fan of this, simple but a very useful addition to the API 👍🏻 
Thanks!

WalshyDev (approved):

Intybyte (approved):

PR #4256, opened 2024-11-04
Recipe Rewrite (API Only)

SchnTgaiSpock:
## Description

The long awaited recipe rewrite (supersedes #4093)

## Proposed changes

In short, adds true shaped/shapeless crafting, tagged inputs, and json (de)serialization (custom recipes)

Full ADR [here](https://github.com/SchnTgaiSpock/Slimefun4/blob/api/recipe-rewrite-4/docs/adr/0002-recipe-rewrite.md)

N/A

## Checklist

### DIFF
--- a/docs/adr/0002-recipe-rewrite.md
+++ b/docs/adr/0002-recipe-rewrite.md
@@ -100,7 +100,7 @@ This would remove the need to use ItemSettings to determine the gold pan weights
 
 ### RecipeService
 
-This is the public interface for the recipe system, there are methods here to add,
+This is the public interface for the recipe system, there are methods here to register,
 load, save, and search recipes. It also stores a map of `MatchProcedures` and
 `RecipeType` by key for conversions from a string
 
--- a/docs/adr/0002-recipe-rewrite.md
+++ b/docs/adr/0002-recipe-rewrite.md
@@ -1,7 +1,7 @@
 # 2. Recipe rewrite
 
 Date: 2024-11-03
-Last update: 2024-11-03
+Last update: 2024-11-08
 
 **DO NOT rely on any APIs introduced until we finish the work completely!**
 
@@ -32,9 +32,9 @@ Slimefun, focusing on
   Slimefun recipes
 - Performance: Should not blow up any servers
 
-The new system should also be completely backwards compatible with the old.
+The new recipe system should also be completely backwards compatible.
 
-## API Changes
+## API Additions
 
 ### 5 main recipe classes
 
@@ -61,7 +61,7 @@ An `RecipeOutputItem`s controls how an output is generated when the recipe is
 crafted. It can be a single item (see `RecipeOutputItemStack`, `RecipeOutputSlimefunItem`),
 or a group of items each with a certain weight of being output (see `RecipeOutputGroup`).
 
-#### Examples
+#### Examples (pseudocode)
 
 Here are the inputs and outputs of the recipe for a vanilla torch
 
@@ -104,9 +104,9 @@ This is the public interface for the recipe system, there are methods here to ad
 load, save, and search recipes. It also stores a map of `MatchProcedures` and
 `RecipeType` by key for conversions from a string
 
-## JSON Serialization
+### JSON Serialization
 
-All recipes should be able to be serialized to and deserialized
+All recipes are able to be serialized to and deserialized
 from JSON. The schemas are shown below.
 
 Here, `key` is the string representation of a namespaced key
@@ -124,7 +124,8 @@ Here, `key` is the string representation of a namespaced key
 }
 ```
 
-The recipe deserializer also needs a `__filename` field, which is inserted when the file is read, so it doesn't (and shouldn't) be in the schema
+The recipe deserializer technically needs a `__filename` field, but it is
+inserted when the file is read, so it isn't (and shouldn't) be in the schema
 
 `RecipeInput`
 
@@ -174,7 +175,7 @@ The recipe deserializer also needs a `__filename` field, which is inserted when 
 }
 ```
 
-*In addition to those schemata, items can be in short form:
+*In addition to those schemas, items can be in short form:
 
 - Single items: `<namespace>:<id>|<amount>`
 - Tags: `#<namespace>:<id>|<amount>`
@@ -183,22 +184,76 @@ The recipe deserializer also needs a `__filename` field, which is inserted when 
 
 The 5 main recipe classes are all polymorphic, and subclasses can be used in their
 stead, and should not affect the recipe system (as long as the right methods are
-override, see javadocs)
+overriden, see javadocs)
 
 ### Custom serialization/deserialization
 
 The default deserializers recognize subclasses with custom deserializers by
-the presence of a `class` field in the json, which is the key of a
+the presence of a `class` field in the json, which should be the key of a
 custom deserializer registered with Slimefun's `RecipeService`.
 For custom serializers, override the `serialize` method on the subclass,
 and ensure they also add the `class` field
+
+## Recipe Lifecycle
+
+### Stage 1a
+
+When Slimefun is enabled, all recipes in the resources folder will be
+moved to `plugins/Slimefun/recipes/` (unless a file with its name already exists).
+
+Addons should do the same. (We recommend saving to
+`plugins/Slimefun/recipes/<your-addon-name>/` but it's not required).
+
+To prevent unnecessary file operations, Slimefun/addons first send a list of
+filenames of recipes present in the resources folder to the recipe service,
+which then filters out all the files that already exist. Then each recipe can
+be read and copied over.
+
+### Stage 1b
+
+Also on enable, recipes defined in code should be registered. These two steps
+can be done in no particular order.
+
+### Stage 2
+
+On the first server tick, all recipes in the `plugins/Slimefun/recipes` folder
+are read and added to the `RecipeService`, removing all recipes with the
+same filename. This is why recipes should ideally be *defined* in JSON,
+to prevent unnecessary work.
+
+When loading JSON recipes, we also need to be able to tell the difference between
+a server owner changing a recipe, and a developer changing a recipe. To do this,
+we use a system called Recipe Overrides; it allows for updates to recipes from
+developers while also preserving custom recipes by server owners
+
+- Slimefun/addons should tell the recipe service it will apply a recipe
+  override on enable, **before** any JSON recipes are copied from the resources
+  folder
+- The recipe service checks all recipe overrides that have already run
+  (in the file `plugins/Slimefun/recipe-overrides`) and if it never received
+  that override before, it deletes the old files and all recipes inside them.
+  Then all recipes are loaded as before.
+
+### Stage 3
+
+While the server is running, recipes can be modified in code, saved to disk, or
+re-loaded from disk. New recipes can also be added, however not to any existing
+file (unless forced, which is not recommended)
+
+### Stage 4
+
+On server shutdown (or `/sf recipe save`), **all** recipes are saved to disk.
+This means any changes made while the server is running will be overwritten.
+Server owners should run `/sf recipe reload <file-name?>` to load new recipes
+dynamically from disk.
 
 ## Phases
 
 Each phase should be a separate PR
 
 - Phase 1 - Add the new API
-- Phase 2 - Migrate Slimefun toward the new API
+- Phase 2 - Migrate Slimefun items/multiblocks/machines toward the new API
+- Phase 3 - Update the Slimefun Guide to use the new API
 
 The entire process should be seamless for the end users, and
 backwards compatible with addons that haven't yet migrated
--- a/docs/adr/0002-recipe-rewrite.md
+++ b/docs/adr/0002-recipe-rewrite.md
@@ -204,11 +204,6 @@ moved to `plugins/Slimefun/recipes/` (unless a file with its name already exists
 Addons should do the same. (We recommend saving to
 `plugins/Slimefun/recipes/<your-addon-name>/` but it's not required).
 
-To prevent unnecessary file operations, Slimefun/addons first send a list of
-filenames of recipes present in the resources folder to the recipe service,
-which then filters out all the files that already exist. Then each recipe can
-be read and copied over.
-
 ### Stage 1b
 
 Also on enable, recipes defined in code should be registered. These two steps
--- a/docs/adr/0002-recipe-rewrite.md
+++ b/docs/adr/0002-recipe-rewrite.md
@@ -30,7 +30,7 @@ Slimefun, focusing on
   of recipes with this system.
 - Customizability: Server owners should be able to customize any and all
   Slimefun recipes
-- Performance: Should not blow up any servers
+- Performance: Should be on par or better than the current system.
 
 The new recipe system should also be completely backwards compatible.
 
--- a/docs/adr/0002-recipe-rewrite.md
+++ b/docs/adr/0002-recipe-rewrite.md
@@ -117,7 +117,7 @@ Here, `key` is the string representation of a namespaced key
 {
     "input"?: RecipeInput
     "output"?: RecipeOutput
-    "type": key | key[]
+    "type": NamespacedKey | NamespacedKey[]
     "energy"?: int
     "crafting-time"?: int
     "permission-node"?: string | string[]
@@ -135,7 +135,7 @@ inserted when the file is read, so it isn't (and shouldn't) be in the schema
     "key": {
         [key: string]: RecipeInputItem
     }
-    "match"?: key
+    "match"?: NamespacedKey
 }
 ```
 
@@ -151,11 +151,11 @@ inserted when the file is read, so it isn't (and shouldn't) be in the schema
 
 ```txt
 {
-    "id": key
+    "id": NamespacedKey
     "amount"?: int
     "durability"?: int
 } | {
-    "tag": key
+    "tag": NamespacedKey
     "amount"?: int
     "durability"?: int
 } | {
--- a/docs/adr/0002-recipe-rewrite.md
+++ b/docs/adr/0002-recipe-rewrite.md
@@ -1,7 +1,7 @@
 # 2. Recipe rewrite
 
 Date: 2024-11-03
-Last update: 2024-11-08
+Last update: 2025-03-11
 
 **DO NOT rely on any APIs introduced until we finish the work completely!**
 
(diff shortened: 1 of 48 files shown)


SchnTgaiSpock:
I'm going to throw together a showcase addon sometime next week to facilitate testing since I don't want to migrate any Slimefun stuff in this PR.

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
+# 2. Recipe rewrite
```
WalshyDev, line 1: initial review is just on the adr - thank you for this by the way, it makes large changes like this much better :)

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+The goal of this rewrite is to introduce an improved recipe system to
+Slimefun, focusing on
+
+- Ease of use: The API should be clean and the system intuitive for
+  developers to use
+- Extensibility: Addons should be able to create and use their own types
+  of recipes with this system.
+- Customizability: Server owners should be able to customize any and all
+  Slimefun recipes
+- Performance: Should not blow up any servers
```
WalshyDev, line 33:
```suggestion
- Performance: Should be on par or better than the current system.
```

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+All recipes are now `Recipe` objects. It is an association between
+inputs (see `RecipeInput`) and outputs (see `RecipeOutput`), along with other metadata
+for how the recipe should be crafted -- recipe type, energy cost, base crafting duration, etc.
+
+`RecipeInput`s are a list of `RecipeInputItem`s plus a `MatchProcedure` -- how the inputs of
+the recipe should be matched to items in a multiblock/machine when crafting. The base ones are:
+
+- Shaped/Shapeless: Exactly the same as vanilla
+- Subset: How the current smeltery, etc. craft
```
WalshyDev, line 49: can we expand and say what this actually is

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+### 5 main recipe classes
+
+All recipes are now `Recipe` objects. It is an association between
+inputs (see `RecipeInput`) and outputs (see `RecipeOutput`), along with other metadata
+for how the recipe should be crafted -- recipe type, energy cost, base crafting duration, etc.
+
+`RecipeInput`s are a list of `RecipeInputItem`s plus a `MatchProcedure` -- how the inputs of
+the recipe should be matched to items in a multiblock/machine when crafting. The base ones are:
+
+- Shaped/Shapeless: Exactly the same as vanilla
```
WalshyDev, line 48:
```suggestion
- Shaped/Shapeless: Exactly the same as vanilla. Shaped recipes need to be crafted exactly as defined. Shapeless need to be in the same order but can move (e.g. 3 sticks down the y axis could be on any x axis)
```

(ngl shaped + shapeless are both wrong in mc...)

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+- Subset: How the current smeltery, etc. craft
+- Shaped-flippable: The recipe can be flipped on the Y-axis
```
WalshyDev, line 50: give an example

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+- Shaped-rotatable: The recipe can be rotated (currently only 45deg, 3x3)
```
WalshyDev, line 51: let's give a simple example

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+`RecipeInputItem`s describe a single slot of a recipe and determines what
+items match it. There can be a single item that matches (see `RecipeInputSlimefunItem`,
+`RecipeInputItemStack`), or a list (tag) of items all of which can be used
+in that slot (see `RecipeInputGroup`, `RecipeInputTag`).
+
+`RecipeOutput`s are just a list of `RecipeOutputItem`s, all of which are crafted by the recipe.
+
+An `RecipeOutputItem`s controls how an output is generated when the recipe is
+crafted. It can be a single item (see `RecipeOutputItemStack`, `RecipeOutputSlimefunItem`),
+or a group of items each with a certain weight of being output (see `RecipeOutputGroup`).
```
WalshyDev, line 62: can i return just multiple items in any way? 
also i'd be clear that this weighting is still outputing a single item, just with different chances - i didn't get what this was saying until the example below

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+        35 RecipeOutputSlimefunItem(SIFTED_ORE)
+    )
+)
+```
+
+This would remove the need to use ItemSettings to determine the gold pan weights
+
+### RecipeService
+
+This is the public interface for the recipe system, there are methods here to add,
```
WalshyDev, line 103: i'd probably keep the whole registration naming/process rather than "add"

```suggestion
This is the public interface for the recipe system, there are methods here to register,
```

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+Here, `key` is the string representation of a namespaced key
+
+`Recipe`
+
+```txt
+{
+    "input"?: RecipeInput
+    "output"?: RecipeOutput
+    "type": key | key[]
```
WalshyDev, line 120: A recipe being multiple types is insteresting
```suggestion
    "type": NamespacedKey | NamespacedKey[]
```

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+    "energy"?: int
+    "crafting-time"?: int
```
WalshyDev, line 122: I'm not fully sure we should have these at least top level
My thinking is around addons and custom properties, I'd rather not pollute top level with all these custom properties (and we should treat anything sf as a custom property too - we should essentially be the main addon)

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+    "permission-node"?: string | string[]
```
WalshyDev, line 123: any reason to use - by the way here?
i feel like camelCase is much more common within json

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+The recipe deserializer technically needs a `__filename` field, but it is
```
WalshyDev, line 127: interesting 🤔

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+All recipes are able to be serialized to and deserialized
+from JSON. The schemas are shown below.
+
+Here, `key` is the string representation of a namespaced key
+
+`Recipe`
+
+```txt
+{
```
WalshyDev, line 117: i don't see any like "id" field

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+```
+
+The recipe deserializer technically needs a `__filename` field, but it is
+inserted when the file is read, so it isn't (and shouldn't) be in the schema
+
+`RecipeInput`
+
+```txt
+{
+    "items": string | string[]
```
WalshyDev, line 134: how do these string references work? if i want to point to a vanilla item or an sf item - what do i do there?
(may be worth giving examples for these)

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+    "key": {
+        [key: string]: RecipeInputItem
+    }
```
WalshyDev, line 137: I don't get how this is meant to look today

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+    "match"?: key
```
WalshyDev, line 138:
```suggestion
    "match"?: NamespacedKey
```

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+{
+    "items": RecipeOutputItem[]
+}
+```
+
+`RecipeInputItem`*
+
+```txt
+{
+    "id": key
```
WalshyDev, line 154:
```suggestion
    "id": NamespacedKey
```

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+    "amount"?: int
+    "durability"?: int
```
WalshyDev, line 156: why durability?

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+} | {
+    "tag": key
```
WalshyDev, line 158:
```suggestion
    "tag": NamespacedKey
```

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+`RecipeOutputItem`*
+
+```txt
+{
+    "id": key
+    "amount"?: int
+} | {
+    "group": RecipeInputItem[]
+    "weights"?: int[]
```
WalshyDev, line 174: guessing weights have to line up index wise? If so, that's not great dx. I could add a new item and forget a weight - what happens then?

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+*In addition to those schemas, items can be in short form:
+
+- Single items: `<namespace>:<id>|<amount>`
+- Tags: `#<namespace>:<id>|<amount>`
```
WalshyDev, line 181: tbh i'm not a big fan of this, i'd rather be more explicit. I can feel this getting unwieldy in the future

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+stead, and should not affect the recipe system (as long as the right methods are
+overriden, see javadocs)
+
+### Custom serialization/deserialization
+
+The default deserializers recognize subclasses with custom deserializers by
+the presence of a `class` field in the json, which should be the key of a
+custom deserializer registered with Slimefun's `RecipeService`.
+For custom serializers, override the `serialize` method on the subclass,
+and ensure they also add the `class` field
```
WalshyDev, line 195: nts: come back to this class field

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+## Recipe Lifecycle
+
+### Stage 1a
+
+When Slimefun is enabled, all recipes in the resources folder will be
+moved to `plugins/Slimefun/recipes/` (unless a file with its name already exists).
+
+Addons should do the same. (We recommend saving to
+`plugins/Slimefun/recipes/<your-addon-name>/` but it's not required).
```
WalshyDev, line 205: wonder if we just make this required - otherwise i think we just have people overriding eachother.

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+Also on enable, recipes defined in code should be registered. These two steps
+can be done in no particular order.
+
+### Stage 2
+
+On the first server tick, all recipes in the `plugins/Slimefun/recipes` folder
+are read and added to the `RecipeService`, removing all recipes with the
+same filename. This is why recipes should ideally be *defined* in JSON,
+to prevent unnecessary work.
```
WalshyDev, line 217: err interesting...

any reason to not just make plugins register their recipes? avoids the whole 1 tick thing.
Also ideally we just don't ever depend on a filename, that's not great.

`RecipeService.registerRecipes(Path.of(getPluginFolder(), "recipes", "recipes.json"));`
(this is reading from addon folder which i also think is fine)

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+When loading JSON recipes, we also need to be able to tell the difference between
```
WalshyDev, line 219: why do we need to do this? what's the use case?

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+- The recipe service checks all recipe overrides that have already run
+  (in the file `plugins/Slimefun/recipe-overrides`) and if it never received
+  that override before, it deletes the old files and all recipes inside them.
+  Then all recipes are loaded as before.
+
+### Stage 3
+
+While the server is running, recipes can be modified in code, saved to disk, or
+re-loaded from disk. New recipes can also be added, however not to any existing
+file (unless forced, which is not recommended)
```
WalshyDev, line 236: why not?

docs/adr/0002-recipe-rewrite.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Stage 4
+
+On server shutdown (or `/sf recipe save`), **all** recipes are saved to disk.
```
WalshyDev, line 240: Does this include ones where addons register them _without_ a file? If so, why?
I like owners being able to change but I'd also like to not _force_ that to be the case.

WalshyDev (commented):
Some questions and comments my side but very excited for this, thank you!

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: i think we have differing definitions of shaped/shapeless; for me:
- shaped means as long as the order and relative positions of the items are the same, they can be moved anywhere on the crafting grid and it will be crafted
- shapeless means as long as the all the specified items (and nothing extra) are there it will be crafted regardless of order

these are what mc does and i want to keep the naming consistent. i also don't see why they are 'wrong' per se, although ordered and orderless might have been a better name tbh

what you defined as shaped i think i would call smth like 'fixed', and as a player i dislike those kinds of recipes -- copper wire for example needs 3 copper ingots in the middle row, which means i can't shift click them in (to the top row), whereas if it were 'shaped' i could.

i probably wasn't clear enough that i wanted to move away from those recipes since they are pretty annoying as a player to use

ill add it in (albeit w/ different name), but i don't see a reason to use it over mc shaped

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: just to be clear, the short form is only used for items that are *just* an id (mc or sf) and amount

i think this would just be a lot cleaner when writing recipes, since most inputs won't have any other fields anyways

for example
```
{
    "id": "ELECTROMAGNET",
    "input": {
        "items": [
            "123",
            " 4 ",
            "   "
        ],
        "key": {
            "1": "slimefun:nickel_ingot",
            "2": "slimefun:magnet",
            "3": "slimefun:cobalt_ingot",
            "4": "slimefun:battery"
        },
        "match": "slimefun:shaped"
    },
    "output": {
        "items": [
            "slimefun:electromagnet"
        ]
    },
    "type": "slimefun:enhanced_crafting_table"
}
```
vs
```
{
    "id": "ELECTROMAGNET",
    "input": {
        "items": [
            "123",
            " 4 ",
            "   "
        ],
        "key": {
            "1": {
                "id": "slimefun:nickel_ingot"
            },
            "2": {
                "id": "slimefun:magnet"
            },
            "3": {
                "id": "slimefun:cobalt_ingot"
            },
            "4": {
                "id": "slimefun:battery"
            }
        },
        "match": "slimefun:shaped"
    },
    "output": {
        "items": [
            {
                "id": "slimefun:electromagnet"
            }
        ]
    },
    "type": "slimefun:enhanced_crafting_table"
}
```

i forgot to mention the `|<amount>` bit was optional if its just 1; ill edit that in.

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: yeah good point, not sure why i did it like this

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: i remember some mods that had recipes like this, where you could put an item and some kind of tool in, and when you crafted, it would use up some dura on the tool (i *think* it was farmers delight but i could be misremembering)

i thought that would be cool to add, but if you think its out of scope or should just be a custom recipe ill remove it

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: each key in `key` is a character that you used in the `items` field
```
"input": {
    "items": [
        "123",
        " 4 ",
        "   "
    ],
    "key": {
        "1": "slimefun:nickel_ingot",
        "2": "slimefun:magnet",
        "3": "slimefun:cobalt_ingot",
        "4": "slimefun:battery"
    },
    "match": "slimefun:shaped"
},
```

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: added an example in comment below, ill also put it in the adr

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: no particular reason, ill change it to camelCase

SchnTgaiSpock:
@WalshyDev I updated the ADR with most of your review

For custom server recipes, that design was pretty confusing, so I replaced it with one where developers' recipes are in `plugins/Slimefun/recipes` as usual, and custom server recipes are in `plugins/Slimefun/recipes/custom`, which takes precedent

Regarding your comments about recipe filenames -- can you elaborate on why we can't rely on filenames for recipes? It makes overriding them with custom recipes easier. Afaik resource packs also use paths/filenames to override certain resources

Lmk if I missed any parts

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: If an addon registers a new recipe in code it will have to provide a filename, otherwise we wouldn't know where to save it and server owners would have no way to override it. Also disallowing owners from overriding a recipe doesn't do much since they can fork the addon and change it there

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: i also forgot why 😅 the new system should handle this fine i think so i removed it

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: I wanted all recipes to be in a single place for the convenience of server owners; like with `Items.yml`

With how custom recipes now work, they will have to be loaded on the first tick too since they take precedence over all other addons' recipes

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: yeah coming back 5 months later it seems unnecessary, removed it

docs/adr/0002-recipe-rewrite.md
SchnTgaiSpock: i might have another solution: registering a deserializer to work addon-wide, but ig there is still the issue of if addon A wants to depend on addon B and use B's deserializers

PR #4261, opened 2024-11-12
revert: revert #4258 except those for Biome

ybw0014:
## Description

#4258 changed all EnumMap/EnumSet usages to HashMap/HashSet. EnumMap/EnumSet is optimized for enums, and changing them for those that are still enums may cause worse performance issues.

Keep those EnumMap/EnumSet until their keys are no longer enums.

## Proposed changes

Reverted #4258.
Changed the EnumMap/EnumSet for `Biome` as key.

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
@@ -47,7 +47,7 @@ import me.mrCookieSlime.Slimefun.api.inventory.UniversalBlockMenu;
 /**
  * This class houses a lot of instances of {@link Map} and {@link List} that hold
  * various mappings and collections related to {@link SlimefunItem}.
- *
+ * 
  * @author TheBusyBiscuit
  *
  */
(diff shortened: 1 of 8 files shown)


JustAHuman-xD:
Is the performance regression that bad? The point behind switching was that way it doesn't break once they are no longer enums, which could be any update now with all of the breaking changes.

ybw0014:
https://docs.oracle.com/javase/7/docs/api/java/util/EnumMap.html
> Enum maps are represented internally as arrays. This representation is extremely compact and efficient.
> Implementation note: All basic operations execute in constant time. They are likely (though not guaranteed) to be faster than their HashMap counterparts.

md5sha256:
I'm not in favour for this unless there is a clear and obvious performance obsetved performance regression. Just because its theoretically slower does not mean its too slow to accept practially. Would like to see some benchmarks.

JustAHuman-xD:
> I'm not in favour for this unless there is a clear and obvious performance observed performance regression. Just because its theoretically slower does not mean its too slow to accept practically. Would like to see some benchmarks.

Same here

PR #4266, opened 2024-11-19
Add permission check to SF Search in cheat mode

JustAHuman-xD:
## Description
In the unlikely event someone manages to get a cheat sheet they can cheat in items with the search feature, we don’t want that.

## Proposed changes
Change the logic for search items to match the logic for normal items when clicked. (Have a permission check for cheating, cheat in stack when shifting)

## Checklist


TheBusyBiscuit (approved):
LGTM

PR #4299, opened 2025-03-09
Walshy/mc 1.21 bundles

Intybyte:
Reopened version of #4278

## Description
Bundles in 1.21.3+ will create some problems with backpacks that need to be addressed

## Issues
- Players can put bundles into backpack (Minor problem, maybe worth considering a config option for this)
- Players can put backpacks into bundles (Major problem)

## Proposed changes
- [x] Make a tag similar to shulker boxes in order to stop people from putting bundles inside backpacks
- [x] Block people from bundling up backpacks (Major problem)

## Checklist


src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
```diff
@@ -277,7 +277,9 @@ public enum SlimefunTag implements Tag<Material> {
     /**
      * All tile entities.
      */
-    TILE_ENTITIES;
+    TILE_ENTITIES,
+
+    BUNDLES;
```
ybw0014, line 285: javadocs

src/main/resources/tags/bundles.json
```diff
@@ -0,0 +1,72 @@
+{
+  "values" : [
+    {
+      "id" : "minecraft:bundle",
+      "required" : false
+    },
+    {
+    "id" : "minecraft:white_bundle",
```
ybw0014: format?

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BundlingListener.java
ybw0014, line 1: there is already a BackpackListener for handling backpack-related events.

ybw0014 (changes requested):

src/main/java/io/github/thebusybiscuit/slimefun4/implementation/listeners/BundlingListener.java
Intybyte: As this one handles bundles only, I think it is better keeping these 2 separate, in case we need in the future to prevent some items from being bundled together

src/main/java/io/github/thebusybiscuit/slimefun4/utils/tags/SlimefunTag.java
Intybyte: meh, do i really need to? those "javadocs" aren't descriptive at all, the code is self explainatory

PR #4306, opened 2025-04-07
Replace copper ingot with copper from original Minecraft.

Outfit8TSB:
## Description

on 1.17, the copper was added,
however, slimefun didnt removed it yet. causing the copper ingot(sf) still remaining.
this commit replaces Copper Ingot(SF) in favor of Copper ingot from vanilla minecraft.

all users making a new addon
should replace all usage of SlimefunItems.COPPER_INGOT to new ItemStack(Material.COPPER_INGOT).
## Proposed changes

Remove SlimefunItems.COPPER_INGOT and replace with new ItemStack(Material.COPPER_INGOT)
## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/implementation/SlimefunItems.java
@@ -456,7 +456,6 @@ public final class SlimefunItems {
     public static final SlimefunItemStack RAINBOW_GLAZED_TERRACOTTA_HALLOWEEN = new SlimefunItemStack("RAINBOW_GLAZED_TERRACOTTA_HALLOWEEN", Material.ORANGE_GLAZED_TERRACOTTA, "&5Rainbow Glazed Terracotta &7(Halloween)", "", HALLOWEEN);
 
     /* Ingots */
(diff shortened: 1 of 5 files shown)


ybw0014:
I think this would never be considered before the recipe system rewrite (support interchangeable ingredients), but we can see what admins think.

Boomer-1:
agreed ybw. there's too much impact that wasn't considered as it will affect every addon and break them all, along with every players base, sorting systems, network systems, etc., much more care has to be put into this before it would be rolled out

Outfit8TSB:
D'oh, i didnt reailze that. 😵
But you should consider that Copper was Added in 1.17.
but the copper from slimefun was really like existed in 1.17~1.21. Why do you even want to keep it? They are `@Deprecated`. It Must be replaced with `new Itemstack(Material.COPPER_INGOT)`.
If Slimefun copper is there, whats the point for Copper in Minecraft?
For that, i think the old `SlimefunItems.COPPER_INGOT` Should be removed,
And Must be replaced to `new ItemStack(Material.COPPER_INGOT)`.

gphg:
Let's ship it on Slimefun 5!

Outfit8TSB:
> Let's ship it on Slimefun 5!

shush! no saying about slimefun 5. this is slimefun 4.

PR #4327, opened 2025-06-13
Fix DistinctiveItem#canStack

Intybyte:
kudos to ybw0014 for spotting it

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
@@ -342,12 +342,13 @@ public final class SlimefunUtils {
             return false;
         }
 
-        SlimefunItem sf_sfitem = SlimefunItem.getByItem(sfitem);
         if (!item.hasItemMeta()) {
             return !sfitem.hasItemMeta();
         }
 
         Debug.log(TestCase.CARGO_INPUT_TESTING, "SlimefunUtils#isItemSimilar - item.hasItemMeta()");
+
+        SlimefunItem sf_sfitem = SlimefunItem.getByItem(sfitem);
         ItemMeta itemMeta = item.getItemMeta();
 
         if (sf_sfitem != null) {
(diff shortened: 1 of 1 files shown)


src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -341,90 +341,78 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
         } else if (checkAmount && item.getAmount() < sfitem.getAmount()) {
             return false;
         }
+
         SlimefunItem sf_sfitem = SlimefunItem.getByItem(sfitem);
```
ybw0014: `SlimefunItem.getByItem` calls `getItemMeta()` so probably put this after has meta check

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -341,90 +341,78 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
+        if (sf_sfitem != null) {
+            String id = Slimefun.getItemDataService().getItemData(itemMeta).orElse(null);
+
+            if (id == null) {
+                ItemMeta meta = sfitem.getItemMeta();
+                return equalsItemMeta(itemMeta, meta, checkLore);
+            }
+
+            if (!checkDistinction) {
+                return id.equals((sf_sfitem.getId()));
```
ybw0014:
```suggestion
                return id.equals(sf_sfitem.getId());
```

src/main/java/io/github/thebusybiscuit/slimefun4/utils/SlimefunUtils.java
```diff
@@ -341,90 +341,78 @@ public static boolean isItemSimilar(@Nullable ItemStack item, @Nullable ItemStac
...
-                        }
-                    }
-                    return id.equals((sf_sfitem.getId()));
-                }
 
-                ItemMeta meta = sf_sfitem.getItem().getItemMeta();
-                return equalsItemMeta(itemMeta, meta, checkLore);
-            } else if (sfitem instanceof ItemStackWrapper && sfitem.hasItemMeta()) {
-                Debug.log(TestCase.CARGO_INPUT_TESTING, "  is wrapper");
+            return id.equals((sf_sfitem.getId()));
```
ybw0014:
```suggestion
            return id.equals(sf_sfitem.getId());
```

ybw0014 (changes requested):
only found these so far

PR #4336, opened 2025-08-28
fix: remove bartering drop order bias by shuffling candidates (#4326)

### LINKED ISSUE #4326: Piglin Barter Drops are Biased
### ❗ Checklist

- [x] I am using the official english version of Slimefun and did not modify the jar.
- [x] I downloaded the official version from the new build site [Blob Builds](https://blob.build/).
- [x] I am using an up to date "DEV" (not "RC") version of Slimefun.
- [x] I am aware that issues related to Slimefun addons need to be reported on their bug trackers and not here.
- [x] I searched for similar open issues and could not find an existing bug report on this.

### 📍 Description

The piglin barter drop recipe type uses a hashset and iterates over it when generating the custom drop, the hash set, after its done being modified (post finalized event) will have a deterministic order for the runtime of the server.

Because of this, the items at the beginning of the iteration order have a bias towards being dropped.

### 📑 Reproduction Steps

Look at the code

### 💡 Expected Behavior

Drops would be properly randomized always

### 📷 Screenshots / Videos

_No response_

### 📜 Server Log

_No response_

_No response_

### 💻 Server Software

Spigot

### 🎮 Minecraft Version

1.20.x

### ⭐ Slimefun version

Image

### 🧭 Other plugins

_No response_

1vers1on:
## Description
This PR fixes the issue where the piglin barter drop uses a `HashSet` for iteration, which results in a deterministic order after the set is finalized.

## Proposed changes
- Added a new method in `SlimefunRegistry` called `getRandomizedBarteringDrops()`, which copies the `barterDrops` into a `List` and shuffles it with `ThreadLocalRandom`.  
- Updated the piglin drop event listener to use the randomized list instead of iterating over the original `HashSet`. 

Resolves #4326 

## Checklist

### DIFF
--- a/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
+++ b/src/main/java/io/github/thebusybiscuit/slimefun4/core/SlimefunRegistry.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadLocalRandom;
 
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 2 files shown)


1vers1on:
i broke the commit name

